Issue #65, opened 2026-06-02
tech-debt(graph-loader): APPEARS_IN edge key 'hadith_number' drifts from ig#935 canonical 'hadith_number_in_book'


parametrization:
## Summary

Cross-repo APPEARS_IN edge-property naming drift between the two ingest paths:

- **ingest-platform** (canonical, per ig#935 AppearsIn Pydantic model): normalize worker emits `hadith_number_in_book`
- **data-acquisition** (legacy direct loader): `src/graph/load_edges.py` `_APPEARS_IN_QUERY` still emits the bare legacy key `hadith_number`

Same fact, two different edge property names depending on which ingest path wrote it. Node shapes converged after da#62 (the #59 fix); the edge-key naming did not.

## Impact

Graph queries that read the edge property must currently handle both names (or silently miss data from one path). This is a correctness landmine for any consumer that assumes one canonical key.

## Proposed fix

Align data-acquisition's `load_edges.py` edge property to `hadith_number_in_book` (the ig#935-ratified canonical), with a one-time migration note for existing edges written under the legacy key.

## Provenance

Surfaced by Jean-Claude Habimana (architecture lens) during da#62 review, 2026-06-02. Out of scope for #59 (which correctly only touched node writes).

Origin-Verification: `src/graph/load_edges.py` exists at `main`

parametrization:
Requestor: Fatima Okonkwo
Requestee: Kwesi Boateng
RequestOrReplied: Request

**Wave 1 Kickoff — Phase 4**

This issue is assigned to you for p4-wave-1.
- Peer reviewer: Jean-Claude Habimana
- Secondary reviewer: Oyunbileg Batbayar
- Branch from: `deployments/phase-4/wave-1`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

parametrization:
**Post-merge runtime item resolved — no migration needed (2026-06-03).**

PR #68 flagged a one-time prod Neo4j migration for any `APPEARS_IN` edges written under the legacy `hadith_number` key. Checked live against prod Neo4j (`docker exec noorinalabs-neo4j-1 cypher-shell`):

```
MATCH ()-[r:APPEARS_IN]->() RETURN count(r);                         -> 0   (total)
MATCH ()-[r:APPEARS_IN]->()
[...]

PR #68, opened 2026-06-02
fix(graph-loader): align APPEARS_IN edge key to canonical hadith_number_in_book (#65)

parametrization:
## Summary

Aligns the data-acquisition direct edge loader's APPEARS_IN edge property to the ig#935-ratified canonical name `hadith_number_in_book`, eliminating the cross-repo naming drift described in #65.

`src/graph/load_edges.py` `_APPEARS_IN_QUERY` was writing the edge property as the legacy bare `hadith_number`. The drift was loader-only:
- This repo's own `AppearsIn` Pydantic model (`src/models/edges.py`) already uses `hadith_number_in_book`.
- isnad-graph's canonical `src/models/edges.py` uses `hadith_number_in_book` (verified at origin HEAD — see below).
- The ingest-platform normalize path already emits `hadith_number_in_book`.

The direct loader was the lone divergent writer, so consumers reading the edge by the canonical name got nulls for edges produced via this path.

### Change scope (one Cypher property key)

```diff
 MERGE (h)-[:APPEARS_IN {
     book_number: row.book_number,
     chapter_number: row.chapter_number,
-    hadith_number: row.hadith_number
+    hadith_number_in_book: row.hadith_number
 }]->(c)
```

Only the Cypher **edge property key** changed. The bound value `row.hadith_number` is the staging-table (`hadiths_*.parquet`) column name and is intentionally unchanged — that's the source-data field, not the graph property.

## Canonical-name verification

`gh api repos/noorinalabs/noorinalabs-isnad-graph/contents/src/models/edges.py?ref=main` (decoded) →
```
hadith_number_in_book: int | None = None
```
Also confirmed in the org ontology (`ontology/domain.yaml` APPEARS_IN attributes: `[book_number, chapter_number, hadith_number_in_book]`).

## Data migration (runtime item — not in this PR)

Per #65, edges already loaded under the legacy `hadith_number` key in any live Neo4j environment will NOT be retroactively renamed by this code change. A one-time migration is needed where the old key already exists:

```cypher
MATCH ()-[r:APPEARS_IN]->()
WHERE r.hadith_number IS NOT NULL AND r.hadith_number_in_book IS NULL
SET r.hadith_number_in_book = r.hadith_number
REMOVE r.hadith_number
```

The issue does not ask this PR to write the migration, so it is flagged here as a post-merge runtime item for the environment owner. The loader is idempotent, so a fresh `make load` writes only the canonical key going forward.

## Test Plan

- New regression test `tests/test_graph/test_load_edges.py::TestLoadAppearsIn::test_appears_in_edge_uses_canonical_property_key` asserts `_APPEARS_IN_QUERY` uses `hadith_number_in_book: row.hadith_number` and not the legacy bare `hadith_number:` key.
- Local CI-equivalent (child repo config, line-length 100, ruff 0.15.11):
  - `ruff check src/ tests/` — passed
  - `ruff format --check src/ tests/` — 116 files already formatted
  - `mypy src/` — Success: no issues found in 62 source files
  - `pytest tests/ -m "not integration"` — 525 passed, 3 skipped, 13 deselected
- Integration tests (real Neo4j) run in CI's separate job.

Closes #65

Co-Authored-By: Kwesi Boateng <parametrization+Kwesi.Boateng@gmail.com>
Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/src/graph/load_edges.py
+++ b/src/graph/load_edges.py
@@ -279,7 +279,7 @@ MATCH (c:Collection {id: row.collection_id})
 MERGE (h)-[:APPEARS_IN {
     book_number: row.book_number,
     chapter_number: row.chapter_number,
-    hadith_number: row.hadith_number
+    hadith_number_in_book: row.hadith_number
 }]->(c)
 """
 
--- a/tests/test_graph/test_load_edges.py
+++ b/tests/test_graph/test_load_edges.py
@@ -6,7 +6,12 @@ from pathlib import Path
 
 import pytest
 
-from src.graph.load_edges import EdgeLoadResult, _build_chain_pairs, load_all_edges
+from src.graph.load_edges import (
+    _APPEARS_IN_QUERY,
+    EdgeLoadResult,
+    _build_chain_pairs,
+    load_all_edges,
+)
 from tests.test_graph.conftest import (
     MockNeo4jClient,
     write_hadiths,
@@ -197,6 +202,13 @@ class TestLoadAppearsIn:
         ai_result = results[2]
         assert ai_result.edge_type == "APPEARS_IN"
         assert ai_result.created == 1
+
+    def test_appears_in_edge_uses_canonical_property_key(self) -> None:
+        # The APPEARS_IN edge property MUST be the ig#935-canonical
+        # ``hadith_number_in_book`` (matches AppearsIn model + isnad-graph
+        # src/models/edges.py), NOT the legacy bare ``hadith_number`` (da#65).
+        assert "hadith_number_in_book: row.hadith_number" in _APPEARS_IN_QUERY
+        assert "hadith_number:" not in _APPEARS_IN_QUERY
 
     def test_missing_collection_name_skipped(
         self, mock_client: MockNeo4jClient, staging_dir: Path, curated_dir: Path


parametrization:
Requestee: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

## Data-architect review — PR #68 (closes da#65)

Verified against the artifact at head `919237012589c0d18023e077cb53aeb3af69b0e0` (gh pr diff + full clone at head + gh api contents). Five checks, all pass.

### 1. Canonical-name claim — VERIFIED at source of truth
isnad-graph `src/models/edges.py?ref=main` `class AppearsIn` declares `hadith_number_in_book: int | None = None` (line 57). Org `ontology/domain.yaml` APPEARS_IN attributes are `[book_number, chapter_number, hadith_number_in_book]`. This repo's own `AppearsIn` model (`src/models/edges.py:55`) already uses the canonical name. The loader Cypher was the divergent writer — claim confirmed independently, not trusted from the PR body.

### 2. Cypher edge-key vs parquet-column split — CORRECT
The split is exactly right:
- `_APPEARS_IN_QUERY` line 282: edge **property key** → `hadith_number_in_book` (graph property, now canonical).
- The bound value `row.hadith_number` is the Cypher parameter name, populated from the batch dict key `"hadith_number"` at `_load_appears_in` line 336 = `_val(row, "hadith_number")`, which reads the parquet **staging column** `hadith_number` (schema-defined at `src/parse/schemas.py:27`, `pa.field("hadith_number", ...)`).

Renaming the source-column reference would have been wrong; leaving it is correct. The property-key-only change is the right surgical scope.

### 3. Consistency sweep — loader confirmed as lone divergent writer
Full-tree grep at head: every other `hadith_number` occurrence in `src/` is the source-data field (parse staging dicts, `schemas.py` parquet field, `enrich_metadata` lookup keys) — none write or read the APPEARS_IN **edge property**. `_APPEARS_IN_CHECK` (lines 286-294) only probes endpoint-node existence by ID, never the edge property, so the rename has no idempotency interaction there. No `r.hadith_number` edge-property reader exists anywhere in `src/`, `queries/`, or `validate.py`. The #65 "lone divergent writer" premise holds — verified, not assumed.

### 4. Migration Cypher — SOUND and idempotent
```
MATCH ()-[r:APPEARS_IN]->()
WHERE r.hadith_number IS NOT NULL AND r.hadith_number_in_book IS NULL
SET r.hadith_number_in_book = r.hadith_number
REMOVE r.hadith_number
```
Idempotent: after the first run every migrated edge has the new key set and old key removed, so the `hadith_number_in_book IS NULL` guard makes a re-run a no-op. Edges with a legitimately-null legacy value are skipped by `IS NOT NULL` (nothing to carry). Correctly flagged as a post-merge runtime item for the environment owner (not in-PR), and documented clearly enough to run as-is. The loader's `MERGE` carries all three positional props in the pattern, so without this migration an old-key edge would co-exist with a new-key edge on re-load — which is precisely why the REMOVE step matters; the documentation captures this.

### 5. CI — all green including the strongest signal
statusCheckRollup at head: Lint, Type Check, Test, **Integration Tests (real Neo4j)**, and the Pre-commit ⇄ CI sync-drift gate all SUCCESS. The real-Neo4j integration job passing is the load-bearing confirmation that the renamed property writes and reads back correctly end-to-end. New regression test `test_appears_in_edge_uses_canonical_property_key` locks the canonical key and forbids the bare legacy key.

### Verdict
Approved. Minimal, correct, well-scoped property-key alignment; the edge-key/source-column split is exactly right, the lone-writer premise is verified, and the migration Cypher is sound and idempotent. No tech debt introduced.

parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: Integration test does not assert the APPEARS_IN edge property key by name — see point 3.

## Secondary review — data-quality lens (PR #68, head 9192370)

Reviewed against the artifact (`gh api contents` at head 9192370 + `gh pr diff`), not the PR-body framing. Canonical key `hadith_number_in_book` confirmed in org `ontology/domain.yaml` (APPEARS_IN attrs) and isnad-graph `src/models/edges.py:57` at `main`.

### 1. Test quality — negative assertion is precise (PASS)
The regression test asserts:
- positive: `"hadith_number_in_book: row.hadith_number" in _APPEARS_IN_QUERY`
- negative: `"hadith_number:" not in _APPEARS_IN_QUERY`

I checked the false-pass concern (`hadith_number_in_book` contains the substring `hadith_number`). It does NOT false-pass: the negative substring is **colon-anchored** (`hadith_number:`). In the canonical query the characters following `hadith_number` are `_in_book:`, and the value `row.hadith_number` is newline-terminated — so the literal `hadith_number:` is genuinely absent. The legacy form `hadith_number: row.hadith_number` ends in `hadith_number:` and WOULD trip the assertion. The colon choice correctly separates legacy from canonical. Good.

### 2. Mixed-state / data-shape risk — documented; consumer-break risk LOW
PR body documents the mixed-state window and ships the one-time migration Cypher (post-merge runtime item per #65). I verified the consumer side at isnad-graph `main` for break-on-mixed-keys:
- `queries/validation/collection_coverage.cypher` — `OPTIONAL MATCH (h)-[:APPEARS_IN]->(c)` then `count(h)`. Relationship-pattern traversal only; never reads the edge property.
- `src/api/routes/admin/reports.py` — `WHERE NOT (h)-[:APPEARS_IN]->(:Collection)` / `OPTIONAL MATCH … -[:APPEARS_IN]->`. Pattern/existence only.
- No live query reads `r.hadith_number` or `r.hadith_number_in_book` by name. The only by-name reader is the `AppearsIn` Pydantic model, which already uses the canonical key.

Conclusion: no consumer COALESCEs because no consumer reads the scalar property. During the mixed window, legacy-key edges yield `None` for `hadith_number_in_book` at the model boundary (data-completeness gap closed by the migration), but nothing breaks. Mixed-state risk is correctly characterized as a runtime migration item, not a code blocker.

### 3. Integration test does NOT guard this regression (TechDebt)
`tests/integration/test_graph_loading.py::test_appears_in_edges` runs the real loader against Neo4j but only asserts `len(appears_in)==1` and `MATCH ()-[r:APPEARS_IN]->() RETURN count(r) >= 0` — a trivially-true count. It never reads the edge property key, so it would pass under EITHER key name. The regression is guarded solely by the unit string-match test in `test_load_edges.py`. That unit test is sufficient to gate this PR (the query is a module constant; the string assertion is exact), so this is non-blocking. Recommend a follow-up: have the integration test read back `r.hadith_number_in_book IS NOT NULL` (or the property keys via `keys(r)`) so the real-graph path also asserts the canonical key by name.

### 4. CI — all green
`statusCheckRollup`: Lint, Type Check, Test, Integration Tests, Pre-commit ⇄ CI sync-drift gate all SUCCESS at head 9192370.

Approving. The one-line change is correct, the unit test gates the regression precisely, and the mixed-state/migration handling is sound. TechDebt is the integration test's by-name gap (follow-up, not a fix).

PR #74, opened 2026-06-11
test(da#69): assert APPEARS_IN edge property key in real-Neo4j integration test

### LINKED ISSUE #69: test: integration test should assert APPEARS_IN edge property key by name (PR #68 follow-up)
## Summary

PR #68 renamed the APPEARS_IN edge property key to canonical `hadith_number_in_book`, guarded by a unit test (string-match on the Cypher query). The real-Neo4j integration test `test_appears_in_edges` only asserts `count(r) >= 0` — it never reads the property key, so it would pass under either key name. The regression protection rests solely on the unit string-match.

## Proposed fix

Extend the integration test to read back `keys(r)` (or assert `r.hadith_number_in_book IS NOT NULL` on a loaded edge) so the canonical key is verified end-to-end against a real graph.

## Origin

TechDebt attestation by Oyunbileg Batbayar (QA/Data Quality) on PR #68 review (P4W1, 2026-06-02).

## Acceptance Criteria

- [ ] Integration test reads the edge property key by name from Neo4j
- [ ] Test fails if the loader writes the legacy `hadith_number` key

parametrization:
## Summary

`test_appears_in_edges` (real-Neo4j integration test) previously closed with a bare `MATCH ()-[r:APPEARS_IN]->() RETURN count(r) >= 0`. That assertion is **always true** and never reads the edge's property key, so it passed under either the old (`hadith_number`) or new (`hadith_number_in_book`) name. After PR #68 renamed the APPEARS_IN edge property to the ig#935-canonical `hadith_number_in_book`, the rename was guarded **only** by a unit string-match on the Cypher source — no regression protection rested on the actual graph.

This PR extends the integration test to read the edge back from the live graph and assert on the property **key** itself:

- `hadith_number_in_book` MUST appear in `keys(r)`
- it MUST carry a populated (non-null) value, not merely a declared-but-null key
- the pre-#68 `hadith_number` key MUST NOT linger

## Acceptance (per #69)

- `test_appears_in_edges` **fails** if the canonical edge property key is missing/renamed.
- `test_appears_in_edges` **passes** on the current canonical key.

## Test scope & verification

- Change is confined to `tests/integration/test_graph_loading.py`.
- ruff format/lint: clean (pinned 0.15.11).
- mypy is `src/`-scoped in both CI and pre-commit, so `tests/` is unaffected (a pre-existing line-279 typing note on the base branch is out of scope).
- Unit suite green locally: `525 passed, 3 skipped` (`pytest tests/ -m "not integration"`).
- pre-commit ⇄ CI sync-drift gate: OK.
- The live-Neo4j assertion executes via testcontainers in CI's `Integration Tests` job; it cannot run locally (no Docker daemon in this env) — runtime-gated, validated by collection + the static readback query shape against the loader's `_APPEARS_IN_QUERY`.

Closes #69

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,7 +1,16 @@
-"""Integration test fixtures using testcontainers."""
+"""Integration test fixtures using testcontainers.
+
+These tests REQUIRE a working Docker daemon: testcontainers spins up real
+Neo4j / Postgres containers. When Docker is unreachable — e.g. a dev box with
+no daemon, or a CI runner that cannot pull the testcontainers reaper image from
+Docker Hub — the ``neo4j_container`` fixture ``pytest.skip``s the leg rather
+than erroring, so a missing-infra environment yields a clean SKIP instead of a
+red ERROR that masks real signal (#69).
+"""
 
 from __future__ import annotations
 
+import docker.errors
 import pytest
 from testcontainers.neo4j import Neo4jContainer
 from testcontainers.postgres import PostgresContainer
@@ -13,10 +22,23 @@ NEO4J_TEST_PASSWORD = "testpassword123"
 
 @pytest.fixture(scope="session")
 def neo4j_container():
-    """Start a real Neo4j container for integration tests."""
+    """Start a real Neo4j container for integration tests.
+
+    Requires Docker. Skips (does not error) when Docker is unavailable so the
+    suite degrades gracefully off-CI / on a runner that cannot reach Docker Hub.
+    """
     container = Neo4jContainer("neo4j:5-community", password=NEO4J_TEST_PASSWORD)
-    with container as neo4j:
+    try:
+        # ``start`` is where Docker is actually contacted (daemon connect, image
+        # pull, container create); guard only this so a real test-body failure
+        # later is never mistaken for an infra skip.
+        neo4j = container.start()
+    except docker.errors.DockerException as exc:
+        pytest.skip(f"Docker/Neo4j unavailable — skipping integration tests: {exc}")
+    try:
         yield neo4j
+    finally:
+        container.stop()
 
 
 @pytest.fixture
--- a/tests/integration/test_graph_loading.py
+++ b/tests/integration/test_graph_loading.py
@@ -158,9 +158,18 @@ SAMPLE_HADITHS = [
     },
 ]
 
+# NOTE: ``collection_id`` here MUST be the corpus-qualified ``{corpus}:{name}``
+# form, because that is the key the APPEARS_IN edge loader reconstructs from each
+# hadith's ``source_corpus`` + ``collection_name`` (see load_edges.py
+# ``_load_appears_in``: ``col:{corpus}:{name}``) and matches against the
+# Collection node id (load_nodes.py: ``col:{collection_id}``). Plain ids like
+# "bukhari" produce node ``col:bukhari`` while the edge loader looks for
+# ``col:sunnah:bukhari`` — the endpoints never match, so ZERO APPEARS_IN edges
+# are created and the prior ``count(r) >= 0`` assertion silently passed on an
+# empty graph (#69).
 SAMPLE_COLLECTIONS = [
     {
-        "collection_id": "bukhari",
+        "collection_id": "sunnah:bukhari",
         "name_ar": "صحيح البخاري",
         "name_en": "Sahih al-Bukhari",
         "compiler_name": "Muhammad ibn Ismail al-Bukhari",
@@ -170,7 +179,7 @@ SAMPLE_COLLECTIONS = [
         "source_corpus": "sunnah",
     },
     {
-        "collection_id": "muslim",
+        "collection_id": "sunnah:muslim",
         "name_ar": "صحيح مسلم",
         "name_en": "Sahih Muslim",
         "compiler_name": "Muslim ibn al-Hajjaj",
@@ -180,7 +189,7 @@ SAMPLE_COLLECTIONS = [
         "source_corpus": "sunnah",
     },
     {
-        "collection_id": "tirmidhi",
+        "collection_id": "sunnah:tirmidhi",
         "name_ar": "سنن الترمذي",
         "name_en": "Jami at-Tirmidhi",
         "compiler_name": "Abu Isa al-Tirmidhi",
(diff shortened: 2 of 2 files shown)


parametrization:
### CI note — Integration Tests failure is an infra flake, not this change

The `Integration Tests` job (which is `continue-on-error: true` and **non-gating**) failed, but the failure is unrelated to this PR's one-file test change:

- **All 10** integration tests in the suite **errored identically at fixture setup** — including `test_graded_by_edges` and the node-loading tests this PR never touches. My `test_appears_in_edges` assertions never executed.
- Root cause: testcontainers could not pull the `testcontainers/ryuk:0.8.1` reaper image — `docker.errors.APIError: 500 ... Get "https://registry-1.docker.io/v2/": context deadline exceeded`. Docker Hub was unreachable from the runner.

Gating checks are green: **Lint ✓, Test ✓, Type Check ✓, Pre-commit ⇄ CI sync-drift gate ✓**. PR is `mergeable: true` (state `unstable` solely due to the advisory integration job).

A re-run once Docker Hub is reachable will let the new live-Neo4j assertions execute. Flagging rather than silently merging past it.

parametrization:
### Update — hardening surfaced a real masked bug; pushed `ff6917e`

Reviewers, scope grew slightly beyond the original one-line assertion. Hardening `test_appears_in_edges` to read back the property key from the **real graph** (the #69 ask) immediately exposed that **zero APPEARS_IN edges were ever persisted** by this test module — the old `count(r) >= 0` was passing on an empty graph.

Root cause (this module's own fixtures, pre-existing): `SAMPLE_COLLECTIONS` used plain `collection_id`s (`bukhari`) → Collection node `col:bukhari`, but `_load_appears_in` rebuilds the endpoint key from each hadith as `col:{source_corpus}:{collection_name}` = `col:sunnah:bukhari`. The endpoints never matched → loader created 0 edges while still returning an APPEARS_IN result object → `len(appears_in) == 1` and `count(r) >= 0` both passed vacuously.

This commit therefore:
1. **Assertion** (the ask): per-edge `keys(r)` readback — canonical `hadith_number_in_book` present + non-null, legacy `hadith_number` absent; plus an edge-count cross-check against the loader's `created`.
2. **Fixture fix**: corrected `SAMPLE_COLLECTIONS` ids to the corpus-qualified `sunnah:{name}` form so the 3 sample hadiths produce 3 real edges. The sibling `test_real_data_flow::test_full_pipeline_sample` already gets matching ids because it parses through `lk_corpus.run`, so this aligns the hand-rolled fixtures with the parser's real convention.
3. **Skip-guard**: `neo4j_container` now `pytest.skip`s on `docker.errors.DockerException` (daemon down / Docker Hub unreachable) instead of erroring — addresses the earlier infra-flake red and lets the suite degrade gracefully off-CI.

Verified against a live Neo4j 5.x container locally: **integration suite 13 passed** (was 12 passed + 1 fail before the fixture fix), unit suite 525 passed / 3 skipped, ruff clean, sync-drift gate OK.

Note: this is a test-fixture correctness fix, not a product-loader change. If the team prefers the fixture fix tracked/reviewed separately from the assertion, happy to split — flagging rather than burying it.

parametrization:
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved

Reviewed the diff at head ff6917e via REST, and verified every load-bearing claim against the loader source at the same sha rather than the PR-body framing.

**Endpoint-key fix is correct.** In load_edges.py `_load_appears_in`, each edge target is reconstructed as `col:{source_corpus}:{collection_name}` (so `col:sunnah:bukhari`), while load_nodes.py builds the Collection node id as `col:{collection_id}`. The prior fixtures used plain `bukhari` → node `col:bukhari`, which never matched the edge target `col:sunnah:bukhari`, so zero APPEARS_IN edges were created and the old `count(r) >= 0` assertion passed on an empty graph. Re-qualifying the three sample collections to `sunnah:bukhari` / `sunnah:muslim` / `sunnah:tirmidhi` aligns the endpoints. Confirmed.

**Edge-count expectation matches the data.** The three sample hadiths are two bukhari plus one muslim, each carrying a populated positional number (1, 2, 1) written as a real parquet column. All three match a loaded collection node, so `appears_in[0].created == 3` and the `len(rows) == 3` read-back are both exact, not lower-bound. Good — this replaces an always-true check with one that fails if the graph comes back empty.

**Property-key assertions match the loader.** The MERGE persists the canonical book-position key sourced from the row's positional number; the diff asserts that key is present and non-null on every edge, and that the pre-#68 key does not linger. This is precisely the regression a bare existence count could never catch — it guards the ig#935-canonical rename at the real-graph layer, not just via a Cypher string-match unit test.

**Conftest skip-guard is sound and an improvement.** Narrowing the guarded region to the start call (where the daemon is actually contacted) and moving the body into a separate try/finally means a genuine test-body failure can no longer be swallowed as an infra skip, while cleanup still runs. The caught DockerException is the right base class, and `docker` is a transitive testcontainers dependency so the import is safe. The Integration Tests job failing here is the known Docker-Hub reaper-pull flake, not this change — the assertion logic itself is correct.

No must-fix items. Strong, well-scoped test-hardening PR that closes a real silent-pass gap.

TechDebt: none.

parametrization:
### Deferred live-leg tracking

The real-Neo4j assertion here runs in CI via testcontainers and is now skip-guarded against a Docker-unreachable runner. The broader "live legs that require Docker/Neo4j" tracking for P4W2 is **noorinalabs-main#631** — referencing it per the wave convention. No blocker on this PR.

parametrization:
Requestor: Kwesi Boateng
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved

Reviewed the diff at head ff6917e (read via REST, not the PR body) and ran the integration tests against a real Neo4j 5 testcontainer locally — `TestEdgeLoading::test_appears_in_edges` passes, and the full `TestEdgeLoading` + `TestNodeLoading` classes are 9 passed in ~21s.

The change does three correct things:

1. Replaces the vacuous `count(r) >= 0` (always true — passes on an empty graph and regardless of the property name) with real assertions: the loader reports `created == 3`, the edges are read back from the graph, and every edge must carry the canonical `hadith_number_in_book` key, populated (non-null), with the pre-#68 `hadith_number` key absent. This is exactly the regression protection #69 asked for, and it now exercises a real persisted edge rather than asserting nothing.

2. Fixes the `SAMPLE_COLLECTIONS` ids from bare `bukhari`/`muslim`/`tirmidhi` to the corpus-qualified `sunnah:bukhari` etc. This is the right key: the edge loader reconstructs the collection endpoint as `col:{source_corpus}:{collection_name}` while the node loader stores `col:{collection_id}`, so the bare form produced `col:bukhari` vs `col:sunnah:bukhari`, zero matching edges, and the old assertion silently passed on an empty graph. I independently confirmed this same `col:sunnah:<name>` keying on the live staging graph while loading riyadussalihin (da#73), so the corpus-qualified form is correct against the real loader, not just the fixture.

3. The Docker-unavailable skip-guard is well-scoped: it wraps only `container.start()` (the point where the daemon is actually contacted), so an infra-absence yields a clean SKIP while a genuine test-body failure later still surfaces as a real failure. Good discipline — it degrades gracefully without masking signal.

I treated the Integration Tests job red on this PR as the known non-gating Docker-Hub ryuk-pull flake (continue-on-error), not a defect of the change; my local real-Neo4j run is the substantive evidence that the assertion logic is sound.

One complementary-coverage note (not blocking, and correctly out of scope for this key-name PR): the sample hadiths here carry non-null `hadith_number` (1, 2, 1), so the `hadith_number_in_book is not None` assertion is satisfied by populated values and this test does not exercise the null-`hadith_number` path. That path is a real loader issue I hit on the live staging load (da#73): `_APPEARS_IN_QUERY` puts `hadith_number_in_book` inside the relationship MERGE pattern, and Neo4j rejects a MERGE with a null property — which fires for every scraped hadith until da#72 lands. Worth a follow-up test once the loader MERGE is made null-safe, but it belongs with that fix, not here.

Approving.

TechDebt: none

parametrization:
### ✅ Verified locally against a real Neo4j (testcontainers)

Re-ran `test_appears_in_edges` at head `ff6917e` against a live Neo4j 5-community via testcontainers (Docker 29.5.3, ryuk pulled cleanly) — **1 passed**. The `keys(r)` / `hadith_number_in_book` assertions actually executed against the real graph (3 APPEARS_IN edges, canonical key present + non-null, legacy `hadith_number` absent), not merely collected. The Docker-unreachable skip-guard on the `neo4j_container` fixture stays in place so CI runners that flake on the Docker Hub ryuk pull degrade to a clean SKIP rather than a masking red.

Issue #72, opened 2026-06-11
bug(sunnah_scraper): hadith_number not extracted → latent source_id collisions in chapter-grouped collections


parametrization:
## Symptom

`sunnah_scraper` does not extract `hadith_number` — verified live against `riyadussalihin` book 1: **0/47** hadiths had a parsed number (all `None`). The number selector in `_extract_hadith_from_row` does not match sunnah.com's current layout for this collection.

## Why it matters

`parse.sunnah_scraped` builds `source_id = sunnah:<collection>:<book>:<chapter>:<hadith_no|0>`. With `hadith_number` missing, every hadith uses `:0`, so uniqueness relies entirely on `chapter_number`. For riyadussalihin (fine-grained chapters) this happened to yield 0 collisions — but **incidentally**. A collection that groups many hadiths under one chapter heading would collapse them to a single `source_id`, silently **merging distinct hadiths at dedup**.

## Evidence
Live scrape→parse→read-back (da#71 verification comment, 2026-06-10): 47 rows, schema-valid, bilingual text present, but `hadith_number` None for all; 0 collisions for this collection only because chapters disambiguated.

## Fix

In `src/acquire/sunnah_scraper.py` `_extract_hadith_from_row`, repair the `HADITH_NUMBER_SELECTORS` (or extraction logic) to capture the hadith number from the current sunnah.com markup, across the `SCRAPE_COLLECTIONS` set. Add a regression assertion that a known collection yields non-null `hadith_number` for a sample.

## Acceptance Criteria

- [ ] `_extract_hadith_from_row` returns a non-null `hadith_number` for ≥1 verified collection (e.g. riyadussalihin) sample.
- [ ] No `source_id` collisions for a chapter-grouped collection sample.
- [ ] Regression test covering the selector against a saved fixture page.

## Refs
da#71 (sunnah API key / scraper-default tracker, where this was found), main#139 (first-light E2E), main#628 (P4W2).

parametrization:
Requestor: Fatima Okonkwo
Requestee: Alejandra Reyes-Fuentes
RequestOrReplied: Request

**Wave 2 Kickoff — Phase 4**

This issue is assigned to you for p4-wave-2.
- Peer reviewer: Nikolaos Papadopoulos
- Secondary reviewer: Oyunbileg Batbayar
- Branch from: `deployments/phase-4/wave-2`
[...]

PR #75, opened 2026-06-11
fix(sunnah_scraper): extract hadith_number from current sunnah.com markup (#72)

parametrization:
## Summary

Fixes `sunnah_scraper` failing to extract `hadith_number` against sunnah.com's current layout (da#72), and routes the **in-book ordinal** into staging `hadith_number` so the downstream APPEARS_IN `hadith_number_in_book` edge prop is semantically correct (coordinated with da#77; folds in da#78, closed superseded).

### Background
Live riyadussalihin book 1 parsed **0/47** numbers (all `None`) because `_extract_hadith_from_row` used `.hadith_reference .hadith_num`, a selector the site no longer emits. With a missing number, `parse.sunnah_scraped` keys `source_id = sunnah:<coll>:<book>:<chapter>:<hadith_no|0>` on `:0`, collapsing uniqueness onto `chapter_number` — a chapter-grouped book then merges distinct hadiths at dedup.

### Semantic contract (da#77 / Oyunbileg's da#74-review catch)
`src/models/edges.py` documents `hadith_number_in_book` as "Hadith number within the book" (the IN-BOOK ordinal), and `load_edges.py` maps it straight from staging `hadith_number`. So `hadith_number` must carry the in-book ordinal (1, 2, …), not the collection-wide reference (680).

## Fix
The `table.hadith_reference` block carries both numbers:
```
Reference         : Riyad as-Salihin 680   (collection-wide)
In-book reference : Book 1, Hadith 1        (book + in-book ordinal)
```
- `_extract_in_book_ref` parses the in-book ordinal + book number from the "In-book reference" row → ordinal into `hadith_number`, book into `book_number`.
- `_extract_collection_ref_number` is the non-null fallback (sticky label / reference-link href / legacy `.hadith_num` span) used only when the in-book row is absent, so source_id keying always gets a per-collection-unique value.
- `_scrape_book_page` prefers the parsed book number, falling back to the page URL's book number.

The in-book ordinal is unique within a book, so `source_id = sunnah:<coll>:<book>:<chapter>:<ordinal>` stays collision-free.

## Tests
- `TestExtractInBookRef`, `TestExtractCollectionRefNumber`, `TestExtractHadithFromRowCurrentMarkup` — ordinal extraction, collection-ref fallback, and the full-record path on current markup.
- `TestRiyadussalihinFixture` — saved current-markup fixture (`tests/test_acquire/fixtures/riyadussalihin_book1_sample.html`) asserts the in-book ordinals (1, 2, 3, …) are parsed and that `source_id`s are unique.

17/17 scraper tests pass; 156/156 acquire+parse; ruff (pinned 0.15.11) format+lint and `mypy --strict` clean over `src/` and the test file. CI all 5 green on head `52ec39a`.

## Acceptance criteria
- [x] non-null `hadith_number` for the riyadussalihin sample (now the in-book ordinal)
- [x] no `source_id` collisions for a chapter-grouped sample (verified live, 47/47 unique)
- [x] regression test covering the selector against a saved fixture page

## Coordination
Changes the riyadussalihin sample's `source_id`s to `...:<in_book_ordinal>`. Kwesi (da#73) re-runs his loaded sample against this head at the wrapup rebase. da#77 (loader) needs no extra mapping once `hadith_number` holds the ordinal.

Closes #72

## TechDebt
None.

### DIFF
--- a/src/acquire/sunnah_scraper.py
+++ b/src/acquire/sunnah_scraper.py
@@ -35,14 +35,26 @@ USER_AGENT = "isnad-graph/1.0 (hadith-research)"
 # If sunnah.com redesigns, add new selectors at the front of each list.
 #
 # Hadith number: sunnah.com's current layout no longer exposes a bare
-# ``.hadith_num`` span.  The collection-wide reference number now lives in the
-# ``.hadith_reference_sticky`` label (e.g. "Riyad as-Salihin 680") and in the
-# reference link inside ``table.hadith_reference`` (``href="/<coll>:680"``).
-# We prefer the sticky label, fall back to the reference-link href, then to the
-# legacy span selectors for any collection still on the old markup.
+# ``.hadith_num`` span.  Both numbers we care about live in the
+# ``table.hadith_reference`` block at the bottom of each container, e.g.
+#     Reference         : Riyad as-Salihin 680   (collection-wide)
+#     In-book reference : Book 1, Hadith 1        (book + in-book ordinal)
+# We want the IN-BOOK ordinal (1) for ``hadith_number`` because the downstream
+# APPEARS_IN edge maps it into ``hadith_number_in_book`` (documented as "Hadith
+# number within the book" — da#77).  The in-book ordinal is unique within a
+# book, so ``source_id = sunnah:<coll>:<book>:<chapter>:<ordinal>`` stays
+# collision-free.  If the "In-book reference" row is absent we fall back to the
+# collection-wide reference (sticky label / link href) so source_id keying still
+# gets a non-null, per-collection-unique value, then to the legacy span.
+HADITH_REFERENCE_SELECTORS = ["table.hadith_reference"]
 HADITH_NUMBER_STICKY_SELECTORS = [".hadith_reference_sticky"]
 HADITH_NUMBER_LINK_SELECTORS = ["table.hadith_reference tr td a[href]"]
 HADITH_NUMBER_SELECTORS = [".hadith_reference .hadith_num", ".hadithNar498"]
+
+# In-book reference row: "In-book reference : Book 1, Hadith 1".
+_IN_BOOK_REF_RE = re.compile(
+    r"In-book reference\s*:?\s*Book\s+(\d+)\s*,\s*Hadith\s+(\d+)", re.IGNORECASE
+)
 ARABIC_TEXT_SELECTORS = [".arabic_hadith_full", ".text_details .arabic_text_details"]
 ENGLISH_TEXT_SELECTORS = [".english_hadith_full", ".text_details .english_hadith_full"]
 GRADE_SELECTORS = [".hadith_grade", ".hadith-grade"]
@@ -95,13 +107,27 @@ def _fetch_page(client: httpx.Client, url: str) -> BeautifulSoup | None:
         return None
 
 
-def _extract_hadith_number(row: Tag) -> int | None:
-    """Extract the collection-wide hadith reference number from a container.
+def _extract_in_book_ref(row: Tag) -> tuple[int | None, int | None]:
+    """Extract ``(book_number, in_book_ordinal)`` from the reference table.
 
-    The reference number (e.g. 680 for "Riyad as-Salihin 680") is unique across
-    the whole collection, so it is the strongest disambiguator for the downstream
-    ``source_id`` — it keeps records distinct even when book/chapter parsing is
-    imperfect.  Tried in order of robustness against the current sunnah.com markup:
+    Reads the "In-book reference : Book N, Hadith M" row of
+    ``table.hadith_reference``.  Returns ``(None, None)`` when that row is
+    absent (e.g. a collection whose pages omit it).
+    """
+    ref_tag = select_first(row, HADITH_REFERENCE_SELECTORS)
+    if ref_tag is None:
+        return None, None
+    match = _IN_BOOK_REF_RE.search(ref_tag.get_text(" ", strip=True))
+    if not match:
+        return None, None
+    return int(match.group(1)), int(match.group(2))
+
+
+def _extract_collection_ref_number(row: Tag) -> int | None:
+    """Extract the collection-wide reference number (e.g. 680).
+
+    Fallback for ``hadith_number`` when the in-book ordinal is unavailable: it
+    is unique per collection, so source_id keying still gets a non-null value.
 
     1. ``.hadith_reference_sticky`` label — trailing number ("... 680").
     2. ``table.hadith_reference`` reference link — ``href="/<coll>:680"``.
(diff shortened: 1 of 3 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved

Reviewed the diff at head 1662e37 via REST (files API), plus the downstream coupling in src/parse/sunnah_scraped.py and src/parse/base.py.

What this fixes (da#72) and why it is correct:
- sunnah.com's current markup dropped the bare hadith-number span, so `_extract_hadith_from_row` was emitting a null number for every row on the live site. The new `_extract_hadith_number` recovers it from the `.hadith_reference_sticky` label, falls back to the `table.hadith_reference` link href, then to the legacy span — ordered most-to-least robust, which is the right precedence.
- The regexes are sound: trailing `(\d+)\s*$` on the sticky text isolates the number from "Riyad as-Salihin 680", and `[:#](\d+)\s*$` on the href handles both `/coll:680` and `...#680`. The legacy branch is preserved unchanged, so collections still on the old markup keep working.

Data-integrity coupling (my QA lane, and the part that matters most here):
- The recovered number flows downstream into both the curated hadith_number field AND the source_id, which `sunnah_scraped.py` builds as corpus:collection:book:chapter:number. Pre-fix, that number was null, collapsing to a trailing zero for every record and producing source_id collisions across an entire collection; post-fix the collection-wide reference makes each record distinct. The fixture-backed test that asserts source_id uniqueness over a real saved page is exactly the right guard, and it exercises the genuine end-to-end selector path rather than a synthetic stub. Test coverage (sticky, href-fallback, legacy, full-record, fixture uniqueness) is thorough. Full CI is green at head including the Integration Tests job.

One follow-up worth tracking (non-blocking, out of da#72 scope): the value chosen is the collection-wide reference number (680), but the same fixture also carries a true in-book number ("In-book reference: Book 1, Hadith 1"). Downstream this lands on the APPEARS_IN edge property named hadith_number_in_book (see the loader and da#69/PR#74), so that property will hold the collection-wide reference rather than the in-book ordinal. For source_id uniqueness the collection-wide reference is the correct disambiguator and the right call for this fix; the naming-vs-semantics reconciliation of that edge property is a separate concern that predates this PR (the field was simply null before). Suggest a tracking issue to decide whether the edge property should carry the in-book ordinal, the collection-wide reference, or both — no change needed here.

Approving: da#72 acceptance (non-null number on current markup, source_id uniqueness preserved) is met and well-guarded.

parametrization:
RequestOrReplied: Approved

Reviewed the diff at head 1662e37 (read via gh api pulls/files, not the PR framing) and ran the suite at HEAD.

Solid, well-scoped fix for da#72. The hadith number is recovered from sunnah.com's current markup with a three-tier ordered fallback — sticky reference label, then the reference-link href, then the legacy bare-number span — which matches how the rest of the scraper degrades against markup churn via select_first (it iterates selectors in order and returns the first hit). The trailing-digit regex and the href separator regex both behave correctly on the real label/href shapes I probed (Riyad as-Salihin 680, Sahih al-Bukhari 7563, /coll:680, /foo#42), and return None cleanly when there is no number.

Test coverage is strong: the three tiers are each isolated, the full-record path proves the pre-fix None bug is gone, and the saved riyadussalihin fixture is real current-markup so it guards the live selectors. The fixture test asserts source-id uniqueness off the collection-wide reference number, which is the right disambiguator. 14/14 pass at HEAD locally.

One forward-looking note (NOT a change request on this PR): the source ids this produces are corpus-prefixed (sunnah followed by riyadussalihin and the parts), exactly the contract my main#139 E2E harness consumes. That is correct here. Be aware it intersects a separate downstream defect I just found — the streaming normalize worker prepends the corpus a second time to an already-corpus-prefixed source id, so a hadith keyed off your refnum would land double-prefixed in Neo4j relative to the batch loader. That is a normalize-side bug, tracked separately; nothing to do on this PR, flagging only so the two of us key the same id once it is fixed.

No correctness defects, no security concerns (read-only scrape). Approving.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved

Re-reviewed at head 52ec39a (fresh verdict — the amend invalidated the prior approval). This resolves the semantic mismatch I raised on the da#74 thread, and it does so correctly. CI is green on all five checks at this head.

The fix, verified against the actual diff and the committed fixture (not just the PR framing):
- `_extract_hadith_from_row` now sources hadith_number from the IN-BOOK ordinal first (`_extract_in_book_ref` reads "In-book reference : Book N, Hadith M" off `table.hadith_reference`), and only falls back to the collection-wide reference (`_extract_collection_ref_number`: sticky label, then link href, then legacy span) when that row is absent. That is exactly the right precedence: the in-book ordinal is what the downstream APPEARS_IN edge property hadith_number_in_book promises ("Hadith number within the book"), so the edge value is now honest, while the collection-wide ref is retained purely as a non-null source_id-keying fallback.
- I independently parsed the committed fixture with the new regex: the three containers yield in-book pairs (1,1) (1,2) (1,3) while the sticky labels carry 680/681/682 — confirming hadith_number gets the ordinal (1,2,3), not the collection ref. The book number is now also parsed from the same reference row, with a clean fall-through to the page-URL book number in `_scrape_book_page` when the row is absent.
- source_id stays collision-free: the in-book ordinal is unique within a book, so book:chapter:ordinal keys every record distinctly. The end-to-end fixture test asserts both that the parsed numbers are the sequential ordinals [1..N] and that the generated source_ids are unique, and the fallback path has its own coverage (collection ref used, book None) — good separation of the two extraction concerns.

Cross-check against the adjacent work, no negative interaction: the downstream effect is that staging hadith_number now carries the in-book ordinal, which flows into the hadith_number_in_book edge property and makes my da#69/da#74 read-back assertion semantically correct as well as green (my #74 fixture is independent of this scraper, so nothing there moves). It is also orthogonal to da#77's null-safe MERGE rewrite.

Approving: da#72 acceptance (non-null number on current markup, source_id uniqueness preserved) is met, AND the edge-property semantics are now correct — the exact gap I flagged is closed. Nice turnaround.

TechDebt: none

parametrization:
RequestOrReplied: Approved

Re-reviewed the amended diff at head 52ec39a (read via gh api pulls/files), ran the suite, and verified the downstream contract at source. Approving.

The semantic fix is exactly right and resolves what Oyunbileg flagged in da#74. `hadith_number` now carries the IN-BOOK ordinal (1, 2, 3 ...) rather than the collection-wide reference (680, 681 ...), so the downstream APPEARS_IN `hadith_number_in_book` edge prop is finally honest. I confirmed the contract at the destination: src/models/edges.py documents `hadith_number_in_book` as "Hadith number within the book", and src/graph/load_edges.py routes `row.hadith_number` into it — so the value this PR produces is precisely what that field promises. This is the value-meaning convergence the da-cluster has been coordinating (it also keeps the streaming and batch ingest paths in lockstep, since both read the same source column into the same edge property).

The extraction split reads cleanly: `_extract_in_book_ref` pulls book + ordinal from the "In-book reference : Book N, Hadith M" row, and `_extract_collection_ref_number` keeps the prior sticky-label / link-href / legacy-span fallback chain as the non-null source_id-keying backstop when the in-book row is absent. The in-book regex is robust on the variants I probed (spaces, missing colon, lowercase, multi-digit) and correctly does NOT match the sibling "Reference" row or a malformed row missing the Hadith segment. Book number falls back to the page-URL book number when the row is absent, which is the right precedence.

Test coverage is thorough: in-book extraction, the None-pair-when-absent case, the collection-ref fallback chain (sticky / href / legacy), the full-record path proving `hadith_number` is the ordinal not the ref, the no-in-book-row fallback, and the saved riyadussalihin fixture asserting the ordinals come out as 1..N and source_ids stay unique. 17/17 pass at HEAD; all five CI checks green on 52ec39a.

One orthogonal note for downstream awareness, not a change request here: load_edges.py:282 still carries `hadith_number_in_book` inside the MERGE pattern — that null-MERGE-abort is da#77 (Kwesi's separate fix, already converging on MERGE-on-pair + coalesce). This PR only changes the value's meaning, which is correct and independent of that edge-shape fix. No correctness or security concerns; read-only scrape.

TechDebt: none

Issue #73, opened 2026-06-11
feat: data-first vertical slice — load verified scraper sample into staging Neo4j + view in frontend


parametrization:
## Goal (data-first — owner priority 2026-06-10)

Get **real hadith data visible in the product fast**, ahead of the full multi-source pipeline. Thin vertical slice: take the **already-verified** `sunnah_scraper` sample → run it through parse → **load into the staging Neo4j** → confirm it's queryable and **visible in the isnad-graph frontend graph explorer** on staging.

## Why now

Verification (da#71, 2026-06-10) proved the keyless scraper produces real, schema-valid, bilingual hadith data (47 hadiths from one riyadussalihin book, 16-field `HADITH_SCHEMA` Parquet, 0 collisions). No API key needed. So a minimal load is the fastest path to "seeing data" — the W2 / P4 end-state #1 thrust.

## Scope (thin slice, not the full pipeline)

1. Acquire a bounded scraper sample (1–2 collections, e.g. riyadussalihin) via `sunnah_scraper`.
2. Parse → staging Parquet (`parse.sunnah_scraped`).
3. Run the **graph-load** stage (`src/graph/`) into the **staging Neo4j** (HADITH / COLLECTION nodes + edges).
4. Verify: nodes queryable via Cypher; the data renders in the isnad-graph frontend (graph explorer / search) on staging.

## Out of scope (later W2/W3)

- Full multi-source pipeline E2E (that's main#139).
- Entity resolution / enrichment quality passes.
- API-sourced sunnah data (needs key — da#71).

## Acceptance Criteria

- [ ] A bounded scraper sample is loaded into staging Neo4j (node counts reported).
- [ ] The loaded hadiths are queryable via Cypher (sample query + counts in the PR).
- [ ] The data is visible in the isnad-graph frontend on staging (screenshot / trace).
- [ ] Reproducible: a documented command/script runs the slice end-to-end against staging.

## Refs
main#601 (P4 end-state #1), main#139 (full E2E), da#71 (scraper verification), main#628 (P4W2). Depends on da#72 (hadith_number fix) for clean keying but can run before it on the 0-collision sample.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Kwesi Boateng
RequestOrReplied: Request

**Wave 2 Kickoff — Phase 4**

This issue is assigned to you for p4-wave-2.
- Peer reviewer: Jean-Claude Habimana
[...]

PR #76, opened 2026-06-11
feat(first-light): data-first vertical slice — scrape→parse→load riyadussalihin (#73)

parametrization:
## Summary

Thin **data-first vertical slice** (P4W2 keystone) — gets real hadith data visible in the product fast, ahead of the full multi-source pipeline (main#139). Takes the verified keyless `sunnah_scraper` sample (riyadussalihin), runs it through parse (`parse.sunnah_scraped`) and the Phase-3 graph loader into the **staging Neo4j**, with a documented Cypher verification query and frontend steps.

Scoped to **one collection** by design. Not the full pipeline.

## What's here

- **`scripts/first_light/run_slice.py`** — reproducible one-command driver (acquire → parse → load). Default replays the committed real sample offline; `--live` re-scrapes; `--dry-run` skips Neo4j. Loads into whatever `NEO4J_*` resolves to, so it targets the **staging Neo4j** when run from inside the cluster.
- **`scripts/first_light/sample/sunnah_scraped/riyadussalihin.json`** — the **real** 47-hadith sample (live-scraped from sunnah.com, riyadussalihin book 1), committed so the slice is reproducible offline and in CI without network.
- **`docs/first-light-slice.md`** — runbook: staging load command, verification Cypher, expected counts, frontend graph-explorer steps, da#72 caveat.
- **`tests/test_graph/test_first_light_slice.py`** — drives parse + node/edge load over the real fixture (in-process mock client).
- `.cspell/project-words.txt` — add `matn`, `riyadussalihin`.

## Verified locally (real data, this PR's fixture)

```
parse : 47 hadith rows, 1 collection, 0 source_id collisions, schema-valid, bilingual
nodes : 47 Hadith + 1 Collection
edges : 47 APPEARS_IN (hadith → col:sunnah:riyadussalihin), 0 missing endpoints
grading: 0 (riyadussalihin grades not in scraped HTML)
```

The node-loader Collection id (`col:sunnah:riyadussalihin`) and the edge-loader's derived target id match exactly — no dangling edges. Locked in by the test.

Verification Cypher (printed + run by the script after a load):

```cypher
MATCH (c:Collection {id: 'col:sunnah:riyadussalihin'})
OPTIONAL MATCH (h:Hadith)-[:APPEARS_IN]->(c)
RETURN c.name_en AS collection, count(h) AS hadith_count,
       collect(h.matn_en)[0..3] AS sample_matn_en;
```

## da#72 caveat (latent, not active here)

`hadith_number` is unextracted, so `source_id` carries `0` for the hadith-number segment. On **this** sample there are **0 collisions** — each riyadussalihin book-1 hadith sits in its own synthetic chapter (101..147), keeping `source_id` distinct. In chapter-grouped collections the `0` would cause collisions, and the loader's `MERGE` on `id` would coalesce hadiths into one node. Scaling to those collections should wait on **da#72** (owned by Alejandra Reyes-Fuentes); the script prints a loud `WARNING` if it ever sees a collision.

## Test Plan

- [x] `uv run python scripts/first_light/run_slice.py --dry-run` — acquire+parse on committed sample (47 hadiths, 0 collisions)
- [x] `pytest tests/test_graph/test_first_light_slice.py` — 3 passed
- [x] Full non-integration suite: 527 passed, 3 skipped
- [x] ruff lint + format, `mypy src/` clean, markdownlint + cspell clean, sync-drift gate OK
- [ ] **Runtime (post-merge, from inside the cluster):** run the slice against the staging Neo4j (`NEO4J_*` → staging) and capture a frontend graph-explorer screenshot of the riyadussalihin collection + its 47 hadiths.

> **Why the staging-Neo4j load + frontend screenshot is a runtime step:** the staging Neo4j (`bolt://neo4j:7687`) and the frontend host are reachable only from inside the cluster — not from the CI/dev sandbox (no route, no Docker locally). The data→Cypher mapping is fully verified on the **real** sample in CI via the mock-client load test; only the actual cluster write + render are deferred. Per the runtime-gate-scoping convention, that's a post-merge operational step, not a PR-time gate.

Closes #73


parametrization:
Requestee: Kwesi Boateng
RequestOrReplied: Approved

Reviewed at head `510dff9` against the parse/load source it exercises (`src/parse/sunnah_scraped.py`, `src/graph/load_nodes.py`, `src/graph/load_edges.py`) — not the PR-body framing. As Data Architect I weighed the parse→load contract, source_id derivation, node/edge keying, loader idempotency, and the legitimacy of the runtime carve-out. Approving.

## Contract correctness — verified end-to-end

The keystone invariant for this slice is that the APPEARS_IN edge target id equals the Collection node id, and I traced both derivations in source rather than trusting the test reconstruction:

- Collection node id (`load_nodes._load_collections`): `col:` + `generate_source_id("sunnah", "riyadussalihin")` = `col:sunnah:riyadussalihin`.
- APPEARS_IN edge target id (`load_edges._load_appears_in`): `col:` + `f"{source_corpus}:{collection_name}"` = `col:sunnah:riyadussalihin`.

These converge because the parse step sets `source_corpus="sunnah"` (constant) and `collection_name=json_path.stem` ("riyadussalihin"), and the collection row's `collection_id` is `generate_source_id("sunnah","riyadussalihin")` — the same two segments the edge loader rejoins. The keying contract holds, so the 47 APPEARS_IN edges land on the one loaded Collection node with zero dangling endpoints. `source_id` shape `sunnah:riyadussalihin:1:<chapter>:0` is correct; the 47 hadiths occupy distinct synthetic chapters 101..147, so the unextracted-hadith_number `0` segment does not collide on this sample. Loader idempotency holds — Hadith/Collection MERGE on `id`, APPEARS_IN MERGEs the relationship — so re-running the slice is safe, as the doc claims.

## da#72 (PR #75) coupling — invariant under the shift

PR #75 shifts riyadussalihin source_ids from `...:0` to `...:<refnum>`. Counts and uniqueness are invariant: 47 rows, 47 distinct ids either way (the `:0` suffix was already non-colliding here because chapter is the distinguishing segment on this sample), and `test_sample_parses_to_distinct_source_ids` asserts the prefix `sunnah:riyadussalihin:1:` and distinctness, not the trailing segment — so it survives that change. Good forward-compat.

## Runtime carve-out — legitimately gated, and honestly disclosed

The live staging-Neo4j write and frontend render are correctly deferred to a cluster-internal operational step (bolt endpoint reachable only inside the cluster), tracked in main#631 — this is runtime acceptance, not a hidden gap. I specifically credit the doc's transparent disclosure of the Phase-3 edge-loader null-property bug: because `hadith_number` is null for every scraped hadith, the real `isnad-ingest load` aborts on the APPEARS_IN MERGE (`null property value for 'hadith_number_in_book'`), and the staging load was completed with a null-safe edge variant. That is disclosed, scoped out (no loader code changed here), and tracked against main#139/#69/#74 — the right call for a thin slice. Surfacing it rather than burying it is exactly the discipline I want to see.

## One architectural note (non-blocking, for the follow-up that touches the loader)

The mock `_AllEndpointsExistMock` echoes every `*_exists` flag true unconditionally, so `test_appears_in_edges_match_collection`'s `missing_endpoints == 0` assertion does not by itself prove the edge's derived `collection_id` matches a loaded Collection — the mock would report `collection_exists` true even if the two derivations diverged. The real guard against keying drift is the separate literal-string assertion (`edge_collection_id == node_collection_id == "col:sunnah:riyadussalihin"`), which reconstructs the ids in the test body rather than asserting the value the loader actually puts in the batch. That is acceptable for a thin slice and the contract is correct today, but when the loader-null-property fix lands (the PR that does touch `_APPEARS_IN_QUERY`), the regression guard should be hardened so the mock keys endpoint existence off the actual loaded node-id set rather than a blanket true — otherwise a future keying drift between node and edge derivation would slip past CI.

TechDebt: APPEARS_IN test mock (`_AllEndpointsExistMock`) reports all endpoints existing unconditionally; node/edge collection-id match is only asserted via test-body string reconstruction, not against the loader's emitted batch. Harden to a node-id-aware mock when the loader null-property fix (main#139/#69/#74) is implemented.

parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Kwesi Boateng
RequestOrReplied: Approved

Reviewed the diff at head 510dff9 via REST (files API): run_slice.py, the committed riyadussalihin sample, the slice test, docs, and cspell additions. Cross-checked the parse/load coupling against src/parse/sunnah_scraped.py and the edge loader.

This is a clean, well-scoped vertical slice. The dual acquire modes (offline sample replay as default, live re-scrape behind --live) plus --dry-run make it reproducible by reviewers and CI without network, while the real-load path is clearly fenced as an in-cluster operational step. The committed sample is real, bilingual, and schema-valid, and the docs record the actual staging load (47 hadiths verified live on the staging Neo4j on 2026-06-10), which is the right runtime-gate evidence for the data-first goal. Full CI is green, including the Integration Tests, markdown-lint, cspell, and lychee jobs.

The data-integrity coupling (my QA lane) is exactly where I focused, and it holds:
- The slice test drives the REAL load_all_nodes / load_all_edges over the REAL parsed sample with an in-process endpoints-exist mock — a sound no-live-DB proof that still exercises the genuine loader logic rather than a synthetic stub.
- test_appears_in_edges_match_collection independently asserts the edge-loader-derived collection id (col:sunnah:riyadussalihin) equals the node-loader id. That is the same collection-id-keying invariant I just had to repair in the test_graph_loading fixtures over on da#69/PR#74 — good to see it pinned here for this sample so a future divergence is caught in CI.
- On the cross-PR coupling the lead flagged with da#75: this sample carries an unextracted (null) number for all 47 records, but each record sits in its own distinct synthetic chapter (101..147), so the source_id stays distinct per record and there are zero collisions on THIS sample. The slice's da#72 collision-warning path is therefore a dormant safety net here, not an active condition, and the test comment documents that precisely. Counts are invariant under da#75: once the scraper fix lands and the sample is re-scraped, the source_id suffix shifts from the trailing zero to the real reference number, but cardinality stays 47 and the test asserts the stable sunnah:riyadussalihin:1: prefix (not the suffix), so there is no merge-order hazard between the two PRs.

Two non-blocking observations, no change required:
- When the sample is eventually regenerated post-da#75, refresh the committed JSON and re-confirm the 47 / 0-collision counts so the docs table and EXPECTED_HADITHS stay truthful; worth a one-line follow-up rather than a silent staleness.
- The run_slice collision-warning prints but does not fail; that is the right call for a thin slice that MERGEs on id, and it is clearly commented.

Approving: da#73 acceptance (real sample → parse → load → query, end-to-end, with deterministic counts and the live staging verification) is met and well-guarded.

TechDebt: none

parametrization:
## Live run confirmed — real data in Neo4j (the "see real data" moment)

Docker is now enabled locally, so the deferred-to-runtime leg is now **in-PR evidence**. I ran the **real loader** (`scripts/first_light/run_slice.py` → `load_all`, not `--dry-run`) end to end against a live Neo4j 5 container, with the **da#72 scraper fix applied** so `hadith_number` is populated.

**Live graph (from the running Neo4j):**

```text
Nodes loaded : 48   (47 Hadith + 1 Collection)
Edges loaded : 47   (APPEARS_IN)
APPEARS_IN edges carry real hadith_numbers: 680, 681, 682, … 726
```

Documented verification Cypher, run live:

```text
MATCH (c:Collection {id: 'col:sunnah:riyadussalihin'})
OPTIONAL MATCH (h:Hadith)-[:APPEARS_IN]->(c)
RETURN c.name_en, count(h)
=> "riyadussalihin", 47   (with real bilingual matn: "Ibn 'Umar (May Allah be pleased with them) reported:…")
```

**The loaded graph**, rendered directly from the live Neo4j (red = Collection, green = 47 Hadith, edge labels = real hadith_number):

first-light graph

Committed at `scripts/first_light/evidence/firstlight_graph.png`; regenerate with `uv run --with matplotlib --with networkx python scripts/first_light/render_graph.py`.

### Important: the real loader needs da#72 first

Running the real loader on the **pre-da#72** sample (null `hadith_number`) **aborts** on the APPEARS_IN stage — reproduced through the real code path:

```text
neo4j.exceptions.ClientError: Cannot merge the following relationship because of
null property value for 'hadith_number_in_book': (h)-[:APPEARS_IN {…: null}]->(c)
```

`_APPEARS_IN_QUERY` puts `hadith_number_in_book` inside the relationship `MERGE` pattern, and Neo4j rejects a MERGE with a null property. With da#72 (#75) the value is non-null and the load succeeds (above). This confirms the merge order (da#72 first, then rebase this PR) is correct, and that the loader's null-handling is a real gap the mock suite missed — flagged separately for a coordinated fix (touches main#139's MERGE contract + #69/#74 edge-key tests). The product graph-explorer view (authenticated session) is the only remaining piece; this PNG is the direct-from-Neo4j equivalent.

Refs main#631.

Issue #77, opened 2026-06-11
bug(graph/load_edges): APPEARS_IN MERGE aborts on null hadith_number_in_book — real isnad-ingest load fails on scraped data


parametrization:
## Symptom

The real `isnad-ingest load` aborts on the APPEARS_IN stage for any hadith with a null `hadith_number`:

```
Cannot merge the following relationship because of null property value for 'hadith_number_in_book'
```

Surfaced 2026-06-11 during the da#73 first-light staging load (Kwesi Boateng): loading the real scraped riyadussalihin sample into staging Neo4j failed until a null-safe variant was used by hand.

## Root cause

`src/graph/load_edges.py` `_APPEARS_IN_QUERY` puts `hadith_number_in_book: row.hadith_number` **inside the MERGE pattern**. Neo4j refuses to MERGE a relationship with a null property. Because the scraper does not extract `hadith_number` (da#72), every scraped hadith carries null → the load aborts. The in-process mock load suite never caught this (the mock just counts batch rows) — a test-mock-masks-production-failure case.

## Fix (proposed, needs coordination)

MERGE on the (hadith, collection) pair, then `SET r.hadith_number_in_book = row.hadith_number` (null-safe), or `COALESCE` the key. This changes edge-dedup semantics, so it must stay consistent with:
- **main#139** (ig#62) Nikolaos's asserted Parquet→Neo4j MERGE contract
- **da#69 / PR #74** APPEARS_IN edge-key assertions (Oyunbileg)
- **da#72** (Alejandra) — once hadith_number is extracted, fewer nulls, but the MERGE-with-null pattern is still wrong for any null

## Acceptance

- [ ] `isnad-ingest load` completes the APPEARS_IN stage with null `hadith_number` present
- [ ] Edge dedup semantics preserved (no duplicate APPEARS_IN per hadith→collection)
- [ ] Edge-key tests (#69/#74) + main#139 contract updated consistently

PR #79, opened 2026-06-11
fix(graph/load_edges): null-safe APPEARS_IN MERGE + coalesce-preserve (#77)

parametrization:
## Summary

The APPEARS_IN edge loader MERGEd the relationship with its positional properties **inside the MERGE pattern**, so Neo4j aborted the entire edge stage with a null-property error whenever `hadith_number` was null. The scraper leaves `hadith_number` null until da#72, so the real `isnad-ingest load` failed on all scraped data (surfaced by da#73's live staging load).

```text
neo4j.exceptions.ClientError: Cannot merge the following relationship because of
null property value for 'hadith_number_in_book': (h)-[:APPEARS_IN {…: null}]->(c)
```

## The fix

```cypher
MERGE (h)-[r:APPEARS_IN]->(c)
SET r.book_number          = coalesce(row.book_number, r.book_number),
    r.chapter_number       = coalesce(row.chapter_number, r.chapter_number),
    r.hadith_number_in_book = coalesce(row.hadith_number, r.hadith_number_in_book)
```

An APPEARS_IN edge's identity is the **(hadith, collection) pair** — a hadith appears in a collection at exactly one position — so the positional values are *attributes* of that single edge, not part of its key. `SET` after the MERGE is null-safe and keeps dedup correct (one APPEARS_IN per hadith→collection).

The `SET` uses the **streaming ingest path's coalesce-preserve contract** (`ingest-platform workers/ingest/processor.py _build_edge_cypher`: `r.<f> = coalesce(row.<f>, r.<f>)`), so the batch and streaming Parquet→Neo4j paths converge byte-for-byte on idempotency and null-handling — an explicit-null preserves any value already on the edge rather than clearing it.

## Contract-change note (for the two affected contracts)

- **Edge property NAME (`hadith_number_in_book`) and value source (`row.hadith_number`) are UNCHANGED.** Only the mechanism changes: MERGE-key → `SET` (coalesce-preserve). The `AppearsIn` model (`src/models/edges.py`) is untouched.
- **#74 (Oyunbileg)**: the unit string-contract test in `tests/test_graph/test_load_edges.py` moves to the SET form (`r.hadith_number_in_book = coalesce(row.hadith_number, r.hadith_number_in_book)`); the legacy bare `hadith_number:` guard stays. The #74 read-back integration test (key present + non-null on a non-null sample) re-runs **green unchanged** against a real Neo4j.
- **ig#62 / main#139 (Nikolaos)**: the batch path now produces the same APPEARS_IN edge contract the streaming path already asserts — no harness change needed for this PR.

## Out of scope — folded decision, deferred to da#78

The semantic question Oyunbileg raised (the staging `hadith_number` will hold the **collection-wide reference** number, e.g. 680, not the in-book ordinal) is decided here: `hadith_number_in_book` must carry the genuine **in-book ordinal**; the collection-wide ref stays in `source_id` only. This PR ships correct **structure** — the prop maps from `row.hadith_number` and is simply null until the scraper provides the in-book ordinal (null-safe handles it; a correct-null beats a wrong-value). The scraper/schema half is **da#78** (parser lane), a fast-follow — not folded here.

## Test plan

- [x] New real-Neo4j regression test: a null `hadith_number` now **creates** the APPEARS_IN edge (was aborting) instead of failing the MERGE.
- [x] Unit guard that the MERGE line carries no inline property map (the null-unsafe form can't be reintroduced).
- [x] Live Neo4j: coalesce-preserve verified — set `hadith_number_in_book=5`, re-load the same edge with null → value preserved (5), single edge.
- [x] `TestEdgeLoading` (3) + `test_load_edges.py` (19) pass against a real Neo4j; full non-integration suite 525 passed; ruff + `mypy src/` clean; sync-drift gate OK.

Closes #77

### DIFF
--- a/src/graph/load_edges.py
+++ b/src/graph/load_edges.py
@@ -272,15 +272,22 @@ def _load_narrated(
 # 3. APPEARS_IN — hadith -> collection
 # ---------------------------------------------------------------------------
 
+# Identity of an APPEARS_IN edge is the (hadith, collection) PAIR — a hadith
+# appears in a given collection at exactly one position, so the positional
+# values (book / chapter / hadith number) are *attributes* of that single edge,
+# not part of its key. They are therefore SET after the MERGE, never inside the
+# MERGE pattern (da#77): Neo4j refuses to MERGE a relationship with a null
+# property value, and the scraper leaves ``hadith_number`` null until da#72, so a
+# keyed-on-properties MERGE aborts the whole load on real scraped data. SET is
+# null-safe and keeps dedup correct (one APPEARS_IN per hadith->collection).
 _APPEARS_IN_QUERY = """\
 UNWIND $batch AS row
 MATCH (h:Hadith {id: row.hadith_id})
 MATCH (c:Collection {id: row.collection_id})
-MERGE (h)-[:APPEARS_IN {
-    book_number: row.book_number,
-    chapter_number: row.chapter_number,
-    hadith_number_in_book: row.hadith_number
-}]->(c)
+MERGE (h)-[r:APPEARS_IN]->(c)
+SET r.book_number = row.book_number,
+    r.chapter_number = row.chapter_number,
+    r.hadith_number_in_book = row.hadith_number
 """
 
 _APPEARS_IN_CHECK = """\
--- a/src/graph/load_edges.py
+++ b/src/graph/load_edges.py
@@ -278,16 +278,22 @@ def _load_narrated(
 # not part of its key. They are therefore SET after the MERGE, never inside the
 # MERGE pattern (da#77): Neo4j refuses to MERGE a relationship with a null
 # property value, and the scraper leaves ``hadith_number`` null until da#72, so a
-# keyed-on-properties MERGE aborts the whole load on real scraped data. SET is
-# null-safe and keeps dedup correct (one APPEARS_IN per hadith->collection).
+# keyed-on-properties MERGE aborts the whole load on real scraped data.
+#
+# The SET uses the streaming ingest path's coalesce-preserve contract
+# (ingest-platform ``workers/ingest/processor.py _build_edge_cypher``:
+# ``r.<f> = coalesce(row.<f>, r.<f>)``) so the two ingest paths converge byte-for-
+# byte on idempotency: a row with an explicit-null positional value preserves any
+# value already on the edge rather than clearing it. This keeps dedup correct
+# (one APPEARS_IN per hadith->collection) and is null-safe.
 _APPEARS_IN_QUERY = """\
 UNWIND $batch AS row
 MATCH (h:Hadith {id: row.hadith_id})
 MATCH (c:Collection {id: row.collection_id})
 MERGE (h)-[r:APPEARS_IN]->(c)
-SET r.book_number = row.book_number,
-    r.chapter_number = row.chapter_number,
-    r.hadith_number_in_book = row.hadith_number
+SET r.book_number = coalesce(row.book_number, r.book_number),
+    r.chapter_number = coalesce(row.chapter_number, r.chapter_number),
+    r.hadith_number_in_book = coalesce(row.hadith_number, r.hadith_number_in_book)
 """
 
 _APPEARS_IN_CHECK = """\
--- a/tests/integration/test_graph_loading.py
+++ b/tests/integration/test_graph_loading.py
@@ -453,6 +453,57 @@ class TestEdgeLoading:
             # ... and the pre-#68 ``hadith_number`` key MUST NOT linger.
             assert "hadith_number" not in row["keys"]
 
+    def test_appears_in_edges_load_with_null_hadith_number(
+        self, neo4j_client: Neo4jClient, tmp_path: Path
+    ) -> None:
+        """da#77 regression: a null ``hadith_number`` must NOT abort the load.
+
+        Scraped hadiths carry null ``hadith_number`` until da#72. The previous
+        ``MERGE (h)-[:APPEARS_IN {hadith_number_in_book: row.hadith_number}]->(c)``
+        aborted the whole edge stage with a Neo4j null-property error. The fix
+        SETs the positional props after the MERGE, so the edge is created with a
+        null ``hadith_number_in_book`` instead of erroring. Runs against a real
+        Neo4j — the mock load suite cannot enforce the MERGE-null rule.
+        """
+        staging = tmp_path / "staging"
+        staging.mkdir()
+        curated = tmp_path / "curated"
+        curated.mkdir()
+
+        null_hadith = dict(SAMPLE_HADITHS[0])
+        null_hadith["source_id"] = "bukhari:nullnum"
+        null_hadith["hadith_number"] = None  # the da#77 trigger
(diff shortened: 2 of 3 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Kwesi Boateng
RequestOrReplied: Approved

Reviewed the full diff at head d41de43 via REST (loader + both test files). All five CI checks are green and the change is loader-only plus tests. This fixes a real load-aborting bug, and the null-regression test follows the SET-null-removes-property behavior correctly.

The loader fix is right, and better than the version we discussed:
- The MERGE is now keyed on the (hadith, collection) pair with the positional values SET afterward, which is the correct edge identity — a hadith appears in a collection at one position, so book/chapter/number are attributes, not key. That removes the null-property MERGE abort on real scraped data (the da#77 bug).
- Using coalesce(row.x, r.x) in the SET, rather than a bare assignment, is the right call: it matches the streaming ingest path's coalesce-preserve idempotency contract, so a later null row preserves an already-stored value instead of clearing it, and a later real value enriches a previously-null edge. The two ingest paths converge on idempotency. Good cross-path consistency, and it neatly sidesteps the "SET to null clears the property" footgun on re-runs.

The null-regression test (in the integration suite, against a real Neo4j) asserts the right things — exactly the edge-existence framing rather than a present-but-null key:
- appears_in.created == 1 and one row back: the load SUCCEEDED, no abort.
- r.hadith_number_in_book reads back None AND the key is absent from keys(r) — correctly capturing that Neo4j does not persist a property SET to null. That is precisely the behavior I verified on a live Neo4j, so the assertion is accurate, not aspirational.
- Non-null positional props still present (book_number in keys), and the fixture is self-contained with the corpus-qualified col:sunnah:bukhari collection id so the endpoints match independent of the shared fixture. That keying detail is the same one that bit the shared fixture earlier, so good to see it handled here.
- It does not touch test_appears_in_edges, so there is no behavioral overlap with that test.

Unit-side guards are solid too: the SET-form key assertion plus the new no-inline-property-map guard on the MERGE line lock the null-unsafe keyed form out of reintroduction.

One merge-sequencing note (not a code objection, no change requested here): this branch is cut from the pre-da#69 wave-2 HEAD — the diff context still shows the old count(r) >= 0 close of test_appears_in_edges, which my da#74 (head ff6917e, already open) replaces with a stronger read-back. Because da#74 rewrites the tail of test_appears_in_edges and this PR inserts a new sibling method immediately after that same line, a textual merge conflict at that boundary is likely depending on merge order. Whoever merges second should expect to resolve it by keeping BOTH the hardened test_appears_in_edges body and this new test_appears_in_edges_load_with_null_hadith_number method. Flagging so it is resolved deliberately rather than surprising the second merge. Happy to handle the resolution on whichever side lands second.

Approving: the loader fix is correct and idempotency-safe, the null path is proven against a real Neo4j with accurate assertions, and the guards prevent regression. Thanks for folding in the SET-null detail.

TechDebt: none

parametrization:
Requestee: Kwesi Boateng
RequestOrReplied: Approved

Reviewed the diff at head d41de43 (read via gh api pulls/files) and verified the contract claim against both code paths at HEAD. Approving.

The fix is exactly right. The APPEARS_IN query moves the positional props out of the MERGE relationship pattern — `MERGE (h)-[r:APPEARS_IN]->(c)` keyed only on the (hadith, collection) pair, then SET the three positional attrs after the MERGE. That resolves the real null-property MERGE abort (Neo4j refuses to MERGE a relationship with a null property, and scraped hadiths carry a null hadith_number until da#72), and it keeps dedup correct because the pair is the edge identity — a hadith appears in a collection at exactly one position.

The coalesce-preserve contract matches the streaming ingest path byte-for-byte where it counts. I confirmed against the streaming `_build_edge_cypher` at HEAD: streaming SETs `r.<f> = coalesce(row.props.<f>, r.<f>)`, this batch query SETs `r.<f> = coalesce(row.<f>, r.<f>)` — same preserve semantics (an explicit-null positional value preserves any value already on the edge rather than clearing it), the only difference being the read key, which differs only because the streaming worker reads from a props dict while the batch loader reads flat columns. So both ingest paths now converge on idempotency AND null-handling for APPEARS_IN. This needs no change on the main#139 / ig#62 side, as previously coordinated.

Test coverage is strong and addresses the test-mock-masks-production-failure root cause head-on: the new integration test loads a null-hadith_number row against a REAL Neo4j and asserts the edge is created (created == 1) with hadith_number_in_book absent rather than the load aborting — which is exactly the failure the mock suite could not catch. The two static guards in the unit suite are a nice regression lock: one pins the canonical coalesce SET form, the other asserts the MERGE line carries no inline property map so the null-unsafe keyed form can't be reintroduced. All five CI checks green on d41de43.

No correctness or security concerns. Clean, well-scoped loader fix.

TechDebt: none

parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Kwesi Boateng
RequestOrReplied: Approved

Re-confirm: Approved holds at 9c96a77 (post-#74-merge rebase).

Verified against my prior approval head d41de43, not on framing:
- Reviewed substance is byte-identical — src/graph/load_edges.py and tests/test_graph/test_load_edges.py have unchanged md5 between d41de43 and 9c96a77. The loader fix (property-less MERGE + `coalesce(row.hadith_number, r.hadith_number_in_book)` SET) and the unit guards are exactly what I approved.
- Keep-both landed cleanly in the rebased integration file: the hardened test_appears_in_edges (now from #74's merge, with `appears_in[0].created == 3`) and test_appears_in_edges_load_with_null_hadith_number coexist as separate methods, and both coupling comments (the per-edge-loop note + the SAMPLE_HADITHS note) are present in their respective regions — no conflict, as predicted.
- All five CI checks green at 9c96a77; mergeable state clean.

The only delta from the approved state is the additive join of #74's already-merged content plus the kept comments — no logic change. Approval stands. Good to merge.

parametrization:
Requestee: Kwesi Boateng
RequestOrReplied: Approved

re-confirm: Approved holds at 9c96a77.

I independently diffed d41de43..9c96a77 over my reviewed surfaces:
- src/graph/load_edges.py (the _APPEARS_IN_QUERY coalesce fix) — empty diff, byte-identical.
- tests/test_graph/test_load_edges.py (the two static guards) — empty diff, byte-identical.
- the test_appears_in_edges_load_with_null_hadith_number method — md5-identical across both heads (547797be…, 50 lines unchanged).

One precision note so the record is accurate: tests/integration/test_graph_loading.py is NOT empty-diff at the file level — the rebase onto the post-#74 base pulled in Oyunbileg's #74 hardening (SAMPLE_COLLECTIONS corpus-qualified to col:sunnah:bukhari etc., and test_appears_in_edges strengthened from count(r)>=0 to created==3 + per-edge keys(r) read-back). That is #74's content, not mine, and it is sound — it coexists with my null-regression test (the new comment block even documents the null-vs-#74-read-back interaction). My reviewed substance is unchanged, so the Approved stands; I just don't want "EMPTY diff on that file" on the record when the file did change via the #74 merge.

CI green on 9c96a77 (Test / Type Check / Lint / Integration Tests / sync-drift / close-issues all success). Good to merge.

Issue #82, opened 2026-06-12
T0-A: collision-safe canonical source_id scheme across all sources


parametrization:
**Tier:** T0 (foundation — sequence first) · **Suggested owner:** Jean-Claude Habimana (Architect)
Part of epic #81.

## Goal
Define a **collision-safe canonical `source_id` scheme** spanning all sources. Extend the existing `<corpus>:<collection>:<book>:<chapter>:<num>` pattern so IDs are stable and unique across every adapter.

## Why this is foundational
Every Tier-1 light-up and Tier-2 new source depends on a stable ID. Two known live hazards must be designed out:
- **Double-prefix** bug (streaming vs batch load produced `hdt:sunnah:sunnah:…` → same hadith as two nodes).
- **collection-ref vs in-book-ordinal** ambiguity (sunnah.com exposes a collection-wide reference number AND an in-book ordinal; `APPEARS_IN.hadith_number_in_book` means the in-book ordinal).

## Scope
- [ ] Document the canonical `source_id` grammar in `src/parse/schemas.py` (or a sibling) with per-corpus rules.
- [ ] Single normalization helper used by all parsers (no per-parser ad-hoc prefixing).
- [ ] Unit tests asserting no double-prefix and cross-source uniqueness.

## Dependencies
None — blocks T1/T2 landing cleanly.

parametrization:
## Canonical `source_id` contract (da#82) — build to this

Implemented in `src/parse/identity.py` (the single source of truth for every ingest path — batch loaders here AND the streaming/normalize path in `noorinalabs-isnad-ingest-platform`). Keystone light-ups (#85 Thaqalayn, #89 Sanadset, #92 Itqan) and all Tier-1/2 sources should conform to the grammar below.

### 1. `source_id` grammar (staging key)
```
<corpus>:<collection>[:<part>...]
```
- `<corpus>` — one of `SOURCE_CORPORA`, derived from the `SourceCorpus` enum (`lk, sanadset, thaqalayn, sunnah, fawaz, open_hadith, muhaddithat`). It is the namespace that makes ids collision-safe across sources. **`sunnah_scraped` intentionally shares the `sunnah` corpus** so the scraped and API records for the same hadith dedup to one node.
- `<collection>` — per-corpus collection slug (e.g. `bukhari`).
- `<part>...` — positional disambiguators: `book:chapter:ordinal` for collection sources, or a per-row index for dataset sources.
[...]

Issue #92, opened 2026-06-12
N1a: Itqan adapter — narrator profiles (115,735) → Narrator nodes + bios


parametrization:
**Tier:** T2 (new source — Itqan, PR 1 of 3) · **Suggested owner:** Jean-Claude Habimana + Ivana Horvat
Part of epic #81. **HIGHEST-VALUE source — keystone for isnad-graph#963 / #965.**

## Goal
New **Itqan** (github R3GENESI5/Itqan) adapter — **narrator profiles**: 115,735 profiles from 22 classical texts, 72.6% graded (jarh-wa-ta'dil). Map → `Narrator` nodes + `NARRATOR_BIO_SCHEMA` (generation, trust grade, birth/death years/locations).

Per owner, Itqan is **pre-split into 3 PRs** (this = narrators; #N1b = isnad chains; #N1c = name variants).

## Acceptance
- [ ] New `src/acquire/<src>.py` + `src/parse/<src>.py` producing canonical staging Parquet with `sect` + `source_corpus`.
- [ ] `SourceCorpus` enum value added (in `src/models/enums.py`).
- [ ] Parse + acquire-side tests.
- [ ] End-to-end load into a **local `neo4j:5` container** asserting real nodes/edges (local Docker confirmed working).
- [ ] Reachability + licensing note.

## Dependencies
T0-A (source_id scheme), T0-B (conformance gate).

Issue #89, opened 2026-06-12
L6: light up sanadset end-to-end (narrator keystone, sunni-heavy)


parametrization:
**Tier:** T1 (light up existing adapter) · **Suggested owner:** Alejandra Reyes-Fuentes
Part of epic #81. **KEYSTONE — narrators for isnad-graph#963 / #965.**

## Goal
Light up **sanadset** (sunni-heavy, Sanadset 650K — 650,986 narrator records / 926 books) end-to-end. Adapter exists (`src/acquire/sanadset.py`, `src/parse/sanadset.py`), parse-validated but never loaded. Highest-volume narrator source among the existing adapters.

## Acceptance (per owner — live local Neo4j load at PR-time)
- [ ] Adapter runs `acquire → parse → resolve → load` end-to-end into a **local `neo4j:5` container** (local Docker confirmed working — Server 29.5.3, image present).
- [ ] Test asserts **real nodes/edges** land with correct `sect` + `source_corpus` (not just parse/conformance).
- [ ] Add the missing **acquire-side test** (only `base` + `sunnah_scraper` have one today).
- [ ] Reachability + licensing note in the PR.
- [ ] Staging-cluster load (cluster-internal, ssh `docker exec cypher-shell`) tracked as a **separate post-merge verify**, not a PR gate.

## Dependencies
T0-A (source_id scheme) should merge first.

Issue #85, opened 2026-06-12
L2: light up thaqalayn end-to-end into the graph (SHIA keystone)


parametrization:
**Tier:** T1 (light up existing adapter) · **Suggested owner:** Kwesi Boateng
Part of epic #81. **KEYSTONE — Shia side of isnad-graph#964 (cross-sect parallels).**

## Goal
Light up **thaqalayn** (shia, the Four Books) end-to-end into the graph. Adapter exists (`src/acquire/thaqalayn.py`, `src/parse/thaqalayn.py`), parse-validated but **never loaded**. This is the first Shia corpus loaded — the gap was wiring, not finding a Shia source.

## Acceptance (per owner — live local Neo4j load at PR-time)
- [ ] Adapter runs `acquire → parse → resolve → load` end-to-end into a **local `neo4j:5` container** (local Docker confirmed working — Server 29.5.3, image present).
- [ ] Test asserts **real nodes/edges** land with correct `sect` + `source_corpus` (not just parse/conformance).
- [ ] Add the missing **acquire-side test** (only `base` + `sunnah_scraper` have one today).
- [ ] Reachability + licensing note in the PR.
- [ ] Staging-cluster load (cluster-internal, ssh `docker exec cypher-shell`) tracked as a **separate post-merge verify**, not a PR gate.

## Dependencies
T0-A (source_id scheme) should merge first.

PR #102, opened 2026-06-12
da#82 (T0-A): collision-safe canonical source_id scheme

parametrization:
## da#82 (T0-A): collision-safe canonical `source_id` scheme

Keystone foundation for epic #81 — the single source of truth for `source_id` and graph node-id construction across all ingest paths. Per-source light-ups (#85, #89, #92) build to this contract (posted on #82).

### What
New `src/parse/identity.py` owns the identity grammar and prefixing rules:
- **source_id**: `<corpus>:<collection>[:<part>...]` — corpus exactly once, namespaced from the `SourceCorpus` enum (single source of truth, no drift). `generate_source_id()` now fails fast on an unknown corpus.
- **node ids**: one prefixing function per type (`hadith_node_id`, `collection_node_id`, `narrator_node_id`, `chain_node_id`, `grading_node_id`).

### Two known live identity hazards, designed out
1. **Double-prefix (main#139)** — `hadith_node_id()` is the ONE rule both the batch loaders and the streaming/normalize path call. Idempotent on `hdt:` and *collapses* a doubled leading corpus (`hdt:sunnah:sunnah:... → hdt:sunnah:...`), so the same hadith → exactly one node id regardless of producer. The scattered `f"hdt:{x}" if not x.startswith("hdt:")` idiom is refactored out of `load_nodes.py` and `load_edges.py` into these helpers.
2. **collection-ref vs in-book-ordinal (da#77)** — documents + asserts that staging `hadith_number` is the in-book ordinal that flows to `APPEARS_IN.hadith_number_in_book`, not the collection-wide reference.

### Collision-safety
`source_id` is corpus-namespaced; same `(collection, book, chapter, num)` across all corpora yields distinct ids (unit-tested). `sunnah_scraped` intentionally shares the `sunnah` corpus for dedup.

### Tests / acceptance (owner bar: LIVE local neo4j:5)
- `tests/integration/test_source_id_collision.py` — live `neo4j:5` container: same hadith via batch shape + doubled-corpus shape → **one Hadith node**; idempotent reload; in-book ordinal on `APPEARS_IN.hadith_number_in_book`. **3/3 green locally.**
- `tests/test_parse/test_identity.py` — grammar, idempotency, double-prefix collapse, batch-vs-streaming convergence, cross-source uniqueness, validators.
- Full suite: 566 unit tests pass; mypy + ruff clean.

Closes #82
Refs #81

### DIFF
--- a/src/graph/load_edges.py
+++ b/src/graph/load_edges.py
@@ -15,6 +15,12 @@ from typing import Any
 
 import pyarrow.parquet as pq
 
+from src.parse.identity import (
+    collection_node_id,
+    grading_node_id,
+    hadith_node_id,
+    narrator_node_id,
+)
 from src.utils.logging import get_logger
 from src.utils.neo4j_client import Neo4jClient
 
@@ -242,7 +248,7 @@ def _load_narrated(
 
     batch: list[dict[str, Any]] = []
     for hid, (_pos, nid) in first_narrators.items():
-        full_hid = f"hdt:{hid}" if not hid.startswith("hdt:") else hid
+        full_hid = hadith_node_id(hid)
         batch.append({"narrator_id": nid, "hadith_id": full_hid})
 
     if not batch:
@@ -334,12 +340,12 @@ def _load_appears_in(
             if not sid or not cname:
                 skipped += 1
                 continue
-            hid = f"hdt:{sid}" if not sid.startswith("hdt:") else sid
+            hid = hadith_node_id(sid)
             # Collection IDs in staging use "{corpus}:{name}" format (e.g. "lk:bukhari").
             # Build the same key so we match the Collection nodes that were loaded.
             corpus = row.get("source_corpus", "")
             raw_cid = f"{corpus}:{cname}" if corpus else cname
-            cid = f"col:{raw_cid}" if not raw_cid.startswith("col:") else raw_cid
+            cid = collection_node_id(raw_cid)
             batch.append(
                 {
                     "hadith_id": hid,
@@ -426,8 +432,8 @@ def _load_parallel_of(
             skipped += 1
             continue
         # Ensure lower ID -> higher ID for consistent directionality
-        full_a = f"hdt:{id_a}" if not id_a.startswith("hdt:") else id_a
-        full_b = f"hdt:{id_b}" if not id_b.startswith("hdt:") else id_b
+        full_a = hadith_node_id(id_a)
+        full_b = hadith_node_id(id_b)
         if full_a > full_b:
             full_a, full_b = full_b, full_a
         batch.append(
@@ -511,8 +517,8 @@ def _load_studied_under(
             skipped += 1
             continue
         # Ensure narrator prefix
-        full_from = f"nar:{from_id}" if not from_id.startswith("nar:") else from_id
-        full_to = f"nar:{to_id}" if not to_id.startswith("nar:") else to_id
+        full_from = narrator_node_id(from_id)
+        full_to = narrator_node_id(to_id)
         batch.append({"from_id": full_from, "to_id": full_to})
 
     if not batch:
@@ -588,8 +594,8 @@ def _load_graded_by(
             if not sid:
                 skipped += 1
                 continue
-            hid = f"hdt:{sid}" if not sid.startswith("hdt:") else sid
-            gid = f"grd:{sid}"
+            hid = hadith_node_id(sid)
+            gid = grading_node_id(sid)
             batch.append({"hadith_id": hid, "grading_id": gid})
 
     if not batch:
--- a/src/graph/load_nodes.py
+++ b/src/graph/load_nodes.py
@@ -15,6 +15,12 @@ from typing import Any
 import pyarrow.parquet as pq
 import yaml
 
+from src.parse.identity import (
+    chain_node_id,
+    collection_node_id,
+    grading_node_id,
+    hadith_node_id,
+)
 from src.utils.logging import get_logger
 from src.utils.neo4j_client import Neo4jClient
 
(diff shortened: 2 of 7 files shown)


parametrization:
RequestOrReplied: Approved
TechDebt: none

Secondary review, read at head `66fd129` via `gh api`/`gh pr diff` (no checkout).

**Grammar / fail-fast (base.py).** `generate_source_id()` raises `ValueError` on a corpus outside `SOURCE_CORPORA`, and that set is built straight from the `SourceCorpus` enum (`frozenset(c.value for c in SourceCorpus)`) — so the namespace cannot drift from the enum. Delimiter is the shared `ID_DELIMITER`. Good.

**Double-prefix idempotency (identity.py).** `hadith_node_id()` routes through `bare_source_id()` → strips a leading `hdt:` then `_collapse_double_corpus()` repairs `sunnah:sunnah:... → sunnah:...`, gated on the repeated leading segment being a *known* corpus, and warns on collapse so the upstream main#139 producer stays visible. Idempotent on both `hdt:` and a doubled corpus. `chain_node_id`/`grading_node_id` share the same canonicalization, so every hadith-derived id agrees on one identity.

**Both loaders route through the helpers.** Confirmed every construction site in `load_nodes.py` and `load_edges.py` now calls `hadith_node_id`/`collection_node_id`/`narrator_node_id`/`chain_node_id`/`grading_node_id`; the scattered `f"hdt:{x}" if not x.startswith(...)` idiom is fully removed (grep clean at head). The one residual `startswith("nar:")` at `load_nodes.py:114` is a pre-existing *validation* guard on an already-prefixed `canonical_id`, not a construction idiom — correctly out of scope.

**collection-ref vs in-book-ordinal (da#77).** Schema comment on `HADITH_SCHEMA` documents `hadith_number` = in-book ordinal flowing to `APPEARS_IN.hadith_number_in_book`, not the collection-wide ref. Proven live: `test_in_book_ordinal_on_appears_in_edge` uses ordinal 7 deliberately distinct from book_number 1 so a conflation would be observable on the edge — asserts ordinal==7, book==1.

**Live proof.** `test_source_id_collision.py` runs the REAL batch loaders against `neo4j:5`: batch shape + doubled-corpus shape of the same hadith → `count(h)==1` and `h.id == "hdt:sunnah:bukhari:1:1:1"`; plus idempotent-reload (==1). Skip-guarded off-CI. `gh pr checks 102` green 5/5 (Integration Tests 49s, Test, Lint, Type Check, sync-drift gate).

**Scope.** Tight to the id scheme + the two loader refactors (+ base.py fail-fast, schema doc, two test files). No creep.

Non-blocking note (NOT requesting a change): `_collapse_double_corpus` would over-collapse a hypothetical id whose collection slug equals its corpus (`sunnah:sunnah:...`). The known corpus/collection sets have no such collision today and the docstring scopes it to known corpora, so this is a contained, well-reasoned design choice — flagging only for the record. The ingest-platform streaming-half (normalize calling `hadith_node_id`) is the separate cross-repo follow-up, correctly excluded here.

parametrization:
Requestor: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

## Review — PR #102 (da#82, T0-A: collision-safe canonical source_id scheme)

APPROVED. Reviewed at head `66fd129` via `gh api`/`gh pr diff` (no parent checkout), then ran the suite in an isolated worktree on the PR head. This is the foundation the per-source light-ups (#85/#89/#92) build to, so I verified both catalogued hazards are closed at the graph level, not just asserted.

### Hazard 1 — double-prefix (main#139): closed, live-verified
`hadith_node_id()` is now the single prefixing rule both ingest paths call. `bare_source_id()` strips a leading `hdt:` then `_collapse_double_corpus()` removes a repeated leading corpus, so the batch shape (`sunnah:bukhari:...`) and the streaming-bug shape (`sunnah:sunnah:bukhari:...`) of the same hadith converge to one id. Confirmed the scattered `f"hdt:{x}" if not x.startswith("hdt:")` idiom is fully refactored out of both `load_nodes.py` and `load_edges.py` into the helpers (hadith, collection, narrator, chain, grading). The collapse only fires on a repeated leading segment that is a known corpus, and the edge cases are tested (non-corpus lead and a legit repeated deeper segment are left untouched). Live proof: `test_double_prefix_collapses_to_one_node` loads both shapes through the real batch loaders into a neo4j:5 container and asserts exactly one Hadith node — passed.

### Hazard 2 — collection-ref vs in-book ordinal (da#77): closed
Documented on `HADITH_SCHEMA` and enforced by `test_in_book_ordinal_on_appears_in_edge`, which asserts `APPEARS_IN.hadith_number_in_book == 7` (the in-book ordinal) distinct from `book_number == 1`, so a conflation is observable — passed live.

### Collision-safety + fail-fast
`SOURCE_CORPORA` is derived from the `SourceCorpus` enum (no drift), and `generate_source_id()` now raises on an unknown corpus — exactly the namespacing that keeps the same `(collection, book, chapter, num)` distinct across corpora (`test_same_coords_distinct_per_corpus`). The intentional `sunnah_scraped`→`sunnah` shared namespace for dedup is documented.

### Verification run (head 66fd129, isolated worktree)
- 108 unit tests pass (`test_identity.py`, `test_schemas.py`, `tests/test_graph/`), incl. batch-vs-streaming convergence and a `test_naive_concat_would_diverge` that proves the bug exists absent the fix.
- 3/3 live `neo4j:5` integration tests pass (16s) — the owner's live-local acceptance bar is met, and the test skip-guards cleanly when Docker is unreachable.
- `ruff check` + `ruff format --check` clean; `mypy src/` clean (63 files).

No blocking issues. The double-prefix collapse logging a warning (rather than silently repairing) is the right call — it converges the id while still surfacing a non-canonical upstream producer.

Issue #103, opened 2026-06-12
Narrator/edge nodes carry sect_affiliation + source_corpus in Neo4j (resolve→loader)


parametrization:
**Tier:** T3-adjacent (resolve → canonical-schema → loader) · Part of epic #81. Spun out of #83 (T0-B) per investigation.

## Background
While scoping #83 (sect/source_corpus conformance gate) I found that **Narrator and edge graph nodes carry no `sect_affiliation`/`source_corpus` for ANY source**, not just muhaddithat. The chain is:
- `Narrator` nodes load from `narrators_canonical.parquet` (resolve-stage output). `NARRATORS_CANONICAL_SCHEMA` (`src/resolve/schemas.py`) has **no** sect/source_corpus column, and `_NARRATOR_MERGE` (`src/graph/load_nodes.py`) sets neither.
- Per-source `NARRATOR_BIO_SCHEMA` / `NETWORK_EDGE_SCHEMA` carry a free-text `source` field (and it isn't enum-aligned: sanadset bios use `source="kaggle_narrators"`, muhaddithat uses `"muhaddithat"`).

So "narrator nodes are sect/corpus-tagged in the graph" is a real gap, but it is a **multi-layer resolve→loader change**, not a parser tweak — which is why #83 was correctly scoped down to the parser-level conformance gate only.

## Why this is its own issue
Narrator-level sect is `SectAffiliation` (sunni/shia/**neutral/unknown**) — a **resolution-time** determination (a narrator, esp. a female muhaddithat, can transmit in both Sunni and Shia chains), NOT the collection-level binary `Sect`. So this belongs in the resolve/enrich stage, and it **overlaps**:
- **#82** (canonical identity/schema work — Jean-Claude) — schema additions
- **#99** (cross-source narrator-identity resolution — Ivana) — where sect_affiliation is actually derived/aggregated across a narrator's source corpora

## Scope
- [ ] Add `source_corpus` (+ a `source_corpora` list for multi-source narrators) and `sect_affiliation` to `NARRATORS_CANONICAL_SCHEMA`.
- [ ] Populate them in the resolve stage (aggregate the corpora a narrator was seen in; derive `sect_affiliation`, default `unknown`/`neutral` when unattested).
- [ ] `_NARRATOR_MERGE` sets `n.source_corpus`/`n.sect_affiliation`.
[...]

Issue #81, opened 2026-06-11
epic: multi-source hadith ingestion (Sunni + Shia) with cross-source transform/normalization


parametrization:
## Goal
Ingest **multiple hadith sources** with **both Sunni and Shia representation**, and ensure the pipeline **robustly transforms + normalizes disparate source schemas** into the canonical staging model (HADITH_SCHEMA / NARRATOR_MENTION_SCHEMA / NETWORK_EDGE_SCHEMA) with correct `source_corpus` + `sect` tagging.

This directly unblocks the graph being empty of narrators / events / parallels (isnad-graph#963, #965, #964) — those need real multi-source narrator + isnad + parallel data.

## Known sources (already configured in `src/acquire/`, reachability verified 2026-06-11)
| Module | Source | Sect | Type | Reachable |
|---|---|---|---|---|
| `sunnah_api.py` | Sunnah.com API (api.sunnah.com/v1) | Sunni | REST | ⚠️ 403 — needs API key (da#71, never granted) |
| `sunnah_scraper.py` | Sunnah.com web scraper | Sunni | scrape | ✅ keyless (in use for first-light) |
| `lk_corpus.py` | LK-Hadith-Corpus (ShathaTm) — 6 books | Sunni | GitHub CSV | ✅ |
| `open_hadith.py` | Open-Hadith-Data (mhashim6) — 9 books incl Six Books | Sunni | GitHub | ✅ |
| `fawaz.py` | fawazahmed0/hadith-api | Sunni (multi) | jsDelivr JSON | ✅ |
| `sanadset.py` | Sanadset 650K (Mendeley 5xth87zwb5) — 650,986 narrator records / 926 books | Sunni-heavy | dataset | ✅ |
| `muhaddithat.py` | muhaddithat/isnad-datasets — female narrators | both | GitHub CSV | ✅ |
| `thaqalayn.py` | **ThaqalaynAPI (MohammedArab1)** — the Four Books + more | **Shia** | GitHub/REST | ✅ (thaqalayn-api.net up) |

So Shia representation already exists (Thaqalayn) — the gap is wiring + normalizing them all, not finding a Shia source.

## New candidate sources (from web exploration 2026-06-11 — evaluate + add)
| Source | Sect | Why |
|---|---|---|
[...]

PR #104, opened 2026-06-12
da#83 (T0-B): provenance conformance gate (non-empty + in-domain sect/source_corpus/source)

### LINKED ISSUE #83: T0-B: sect + source_corpus staging conformance gate + fix muhaddithat tagging
**Tier:** T0 (foundation — sequence first) · **Suggested owner:** Kwesi Boateng
Part of epic #81.

## Goal
Add a **parser-level `sect` + `source_corpus` conformance gate** wired into CI, so a parser cannot silently ship invalid provenance.

## Scope (revised after investigation — Option A)
Investigation showed the original premise ("muhaddithat dropped sect/source_corpus tagging") was off: muhaddithat emits `NARRATOR_BIO_SCHEMA` + `NETWORK_EDGE_SCHEMA` (which have a `source` field, correctly set to `"muhaddithat"`), not `HADITH_SCHEMA`/`COLLECTION_SCHEMA` (whose `sect`/`source_corpus` are non-nullable). So there is **no parse-time sect bug to fix in muhaddithat**, and forcing a binary `sect` onto narrator rows would be semantically wrong (narrator sect = resolution-time `SectAffiliation`).

The real, valuable gap: non-nullable schemas prevent *null*, but a parser could still ship an **empty string** (`sect=""` passes a non-nullable column) or a **wrong-domain** value (`source_corpus="sunna"` typo, outside the `SourceCorpus` enum) and nothing catches it.

- [ ] Conformance gate test: for every parser in `src/parse/`, the provenance values on each emitted schema are **non-empty AND in-domain**:
  - `HADITH_SCHEMA` / `COLLECTION_SCHEMA` → `sect ∈ Sect` + `source_corpus ∈ SourceCorpus`
  - `NARRATOR_MENTION_SCHEMA` → `source_corpus ∈ SourceCorpus`
  - `NARRATOR_BIO_SCHEMA` / `NETWORK_EDGE_SCHEMA` → `source` non-empty
- [ ] Assert every `src/parse/` parser emits ≥1 provenance-bearing schema.
- [ ] Wire into the existing CI `pytest` job (no new job needed).

## Acceptance (revised — parser-level, NO DB needed)
- [ ] Conformance gate test passes and fails loudly on an injected empty/wrong-domain provenance value.
- [ ] CI runs it on every PR.
- [ ] muhaddithat passes via its `source` field — no sect hack.

## Split-out
The "Narrator/edge **nodes** carry sect_affiliation + source_corpus in Neo4j" goal (the resolve→canonical-schema→loader change, overlapping #82/#99) is tracked separately in **#103** — that's where the live-Neo4j node-tagging proof lives.

parametrization:
## da#83 (T0-B): provenance conformance gate

Parser-level gate ensuring every parser emits `sect` / `source_corpus` / `source` provenance that is **non-empty and in the enum value-domain**. Foundation for epic #81; complements #82.

### The real gap (re-scoped after investigation)
The original premise was "muhaddithat dropped sect/source_corpus tagging." Investigation (posted on #83) showed otherwise: muhaddithat emits the narrator-bio / network-edge schemas, which carry a free-text `source` (correctly set to `"muhaddithat"`), not `sect`/`source_corpus` — so there's no parse-time sect bug, and stamping a binary `sect` onto narrator rows would be semantically wrong (narrator sect = resolution-time `SectAffiliation`).

The genuine gap: `sect`/`source_corpus` are non-nullable, so a *null* is already impossible — **but an empty string passes a non-nullable column**, and a **wrong-domain** value (`source_corpus="sunna"` typo, a `sect` outside the enum) was never caught anywhere.

### What
- `provenance_violations(table)` in `src/parse/base.py` — checks non-null + non-empty + in-domain (`sect ∈ Sect`, `source_corpus ∈ SourceCorpus`; `source` non-empty) for whichever provenance columns a table carries.
- Enforced **fail-fast in `write_parquet()`** — the single chokepoint every parser writes through — so the invariant holds for every current and future parser, not just a test that can be forgotten.
- `tests/test_parse/test_provenance_conformance.py` — validator (empty / whitespace / null / wrong-domain), the `write_parquet` invariant (rejects + does not write), every canonical schema is provenance-bearing, and every per-source parser emits a provenance-bearing schema.

### Scope note
Node-level `sect_affiliation` + `source_corpus` on Narrator/edge nodes in Neo4j (a resolve→canonical-schema→loader change, overlapping #82/#99) is split out to **#103** — that's where the live-Neo4j node-tagging proof lives. This PR is parser-level, no DB needed (acceptance updated on #83).

### Test plan
- Full unit suite green; the new `write_parquet` enforcement breaks **no** existing parser fixtures (235 parse+graph tests pass post-rebase onto #102).
- `ruff check` + `ruff format --check` clean; `mypy src/` clean (63 files).

Closes #83
Refs #81

### DIFF
--- a/src/parse/base.py
+++ b/src/parse/base.py
@@ -6,9 +6,11 @@ from pathlib import Path
 from typing import Any
 
 import pyarrow as pa
+import pyarrow.compute as pc
 import pyarrow.csv as pcsv
 import pyarrow.parquet as pq
 
+from src.models.enums import Sect, SourceCorpus
 from src.parse.identity import ID_DELIMITER, SOURCE_CORPORA
 from src.utils.logging import get_logger
 
@@ -21,17 +23,74 @@ __all__ = [
     "safe_int",
     "safe_str",
     "validate_enum_fields",
+    "provenance_violations",
 ]
+
+# Provenance columns every staging schema is expected to populate, mapped to the
+# closed value-domain each must stay within. ``sect`` / ``source_corpus`` are the
+# enum-namespaced provenance on the hadith/collection/mention schemas; ``source``
+# is the free-text provenance label on the narrator-bio / network-edge schemas
+# (no enum domain, but it must still be non-empty). A table carrying none of
+# these columns (a non-canonical intermediate) is simply not provenance-checked.
+_PROVENANCE_DOMAINS: dict[str, frozenset[str] | None] = {
+    "sect": frozenset(s.value for s in Sect),
+    "source_corpus": frozenset(s.value for s in SourceCorpus),
+    "source": None,
+}
+
+
+def provenance_violations(table: pa.Table) -> list[str]:
+    """Return provenance-tagging violations for *table* (empty list == clean).
+
+    The conformance gate for da#83: a non-nullable schema column already rejects
+    a *null* ``sect``/``source_corpus`` at construction, but an **empty string**
+    slips through a non-nullable column, and a **wrong-domain** value
+    (``source_corpus="sunna"`` typo, a ``sect`` outside the enum) is never caught
+    anywhere. This checks both, for whichever provenance columns the table
+    carries, so every parser's output is tagged consistently and in-domain.
+    """
+    problems: list[str] = []
+    columns = set(table.column_names)
+    for column, domain in _PROVENANCE_DOMAINS.items():
+        if column not in columns:
+            continue
+        values = table.column(column)
+        if values.null_count:
+            problems.append(f"{column}: {values.null_count} null value(s)")
+        non_null = values.drop_null()
+        if not len(non_null):
+            continue
+        empty = pc.sum(pc.equal(pc.utf8_trim_whitespace(non_null), "")).as_py() or 0
+        if empty:
+            problems.append(f"{column}: {empty} empty/whitespace value(s)")
+        if domain is not None:
+            out_of_domain = sorted(
+                {
+                    v
+                    for v in pc.unique(non_null).to_pylist()
+                    if v is not None and v.strip() != "" and v not in domain
+                }
+            )
+            if out_of_domain:
+                problems.append(f"{column}: value(s) outside {sorted(domain)}: {out_of_domain}")
+    return problems
 
 
 def write_parquet(table: pa.Table, path: Path, schema: pa.Schema | None = None) -> Path:
     """Write PyArrow Table to Parquet with Snappy compression.
 
-    Validate against schema if provided. Create parent dirs. Log stats.
+    Validate against schema if provided, then enforce the provenance-tagging
+    contract (:func:`provenance_violations`) as a fail-fast invariant so no
+    parser can ship staging data with an empty or out-of-domain ``sect`` /
+    ``source_corpus`` / ``source``. Create parent dirs. Log stats.
     """
     path.parent.mkdir(parents=True, exist_ok=True)
     if schema is not None:
         table = table.cast(schema)
+    violations = provenance_violations(table)
+    if violations:
+        msg = f"provenance conformance failed for {path.name}: " + "; ".join(violations)
+        raise ValueError(msg)
     pq.write_table(table, path, compression="snappy")
     size_mb = path.stat().st_size / (1024 * 1024)
     logger.info("wrote_parquet", path=str(path), rows=table.num_rows, size_mb=round(size_mb, 2))
--- a/tests/test_parse/test_provenance_conformance.py
+++ b/tests/test_parse/test_provenance_conformance.py
@@ -1,0 +1,198 @@
+"""Provenance-tagging conformance gate (da#83, T0-B).
+
+Every parser must emit ``sect`` / ``source_corpus`` / ``source`` provenance that
+is non-empty and in the enum value-domain. A non-nullable schema column already
+rejects a *null* value at construction — the gap this gate closes is the
+**empty string** (which passes a non-nullable column) and the **wrong-domain**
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: parser-coverage test is a static source-substring grep, not a per-parser execution of real fixture output through write_parquet — follow-up could feed each parser's emitted table through the gate end-to-end.

QA review of PR #104 at head a4c70fb (da#83, Option A). Reviewed via gh api contents at head — no checkout.

Enforcement behavior — verified
- provenance_violations flags all four failure modes per the columns a table actually carries: null (drop_null + null_count), empty string and whitespace-only (utf8_trim_whitespace == ""), and out-of-domain (value ∉ Sect / ∉ SourceCorpus). Empties are excluded from the out-of-domain comprehension via the strip-non-empty guard, so a single bad value is reported once, not double-counted. Clean separation.
- Per-column scoping matches schemas.py at head — HADITH and COLLECTION carry non-nullable sect + source_corpus; NARRATOR_MENTION carries source_corpus; NARRATOR_BIO and NETWORK_EDGE carry only source (non-empty, no enum domain). The gate keys off column presence, so it never demands sect on bio/edge rows. muhaddithat (emits bio/edge) passes on its source label and is not forced into a sect — confirmed correct, which was the original mis-premise this PR re-scoped away from.
- Fail-fast is wired at the single write_parquet chokepoint every parser flows through, so the invariant holds for current and future parsers, not a forgettable per-call test. The raise precedes pq.write_table; test_rejected_table_is_not_written asserts both the ValueError AND not dest.exists(). Both halves of the invariant are covered.

Test rigor — adequate (~20 tests)
- Validator edge cases (clean, empty, whitespace, null, wrong-domain sect, wrong-domain source_corpus, empty source, no-provenance-columns intermediate) all present.
- write_parquet invariant covered for the valid-writes, empty-reject, wrong-domain-reject, and not-written cases.
- "every canonical schema is provenance-bearing" parametrized across all five schemas.
- "every per-source parser emits a provenance-bearing schema" present.

One gap, non-blocking (the TechDebt line)
- test_parser_modules_reference_a_canonical_schema checks inspect.getsource(module) for a schema-name substring; it does not import-and-run each parser nor push real parser output through write_parquet. A future parser that imports a schema name but emits an untagged table would pass this tripwire. The runtime chokepoint still protects production, and the synthetic write_parquet tests cover the enforcement mechanics, so this is a coverage tripwire weakness rather than a correctness hole — tracked above, not a blocker. Also note the _NON_PARSER_MODULES denylist (includes enrich_metadata in the parser set) is a manual list that can drift; the assert-non-empty guard mitigates a vacuous pass.

CI — gh pr checks 104 all green (Lint, Type Check, Test, Integration Tests, Pre-commit ⇄ CI sync-drift gate). Only src/parse/base.py and tests/test_parse/test_provenance_conformance.py change; the gate rides the existing pytest job with no workflow change.

parametrization:
Requestee: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

## Review — PR #104 (da#83 provenance conformance gate)

Reviewed at head a4c70fb via gh api/diff (no checkout) + independent run in a throwaway worktree.

### Correctness — provenance_violations(table)
- Logic is correct for each column a table actually carries: null_count flag, empty/whitespace via pc.utf8_trim_whitespace == "", and in-domain (sect ∈ Sect, source_corpus ∈ SourceCorpus) over unique non-null values; `source` non-empty only (domain None). Empty values are reported once (as empty), not double-counted as out-of-domain — clean.
- Domains derived from the enums (no drift), mirroring the SOURCE_CORPORA pattern from #102.

### The deliberate call — fail-fast in write_parquet (durability/enforcement tradeoff)
- This is the right call and the strong version of the charter enforcement hierarchy (enforcement at the single chokepoint > a test that can be forgotten). It holds for every current AND future parser.
- Verified it breaks NO legitimate existing write path — I enumerated every write_parquet caller (parse + resolve + enrich) and checked the provenance values they emit:
  - All 7 parsers emit enum-aligned literals; fawaz's dynamic sect only ever returns "shia"/"sunni" (both in-domain).
  - Resolve: narrators_canonical and merge_log schemas carry NO provenance column (not checked); the line-695 "unknown" source_corpus default feeds counters only, never the written merge_log table. narrator_mentions_resolved carries source_corpus but inherits valid corpus from already-gated parse output.
  - enrich_metadata round-trips HADITH_SCHEMA, preserving source_corpus/sect.
- Independent run at head: 325 passed / 3 skipped across tests/test_parse + tests/test_graph + tests/test_resolve (resolve suite confirms the write paths are unbroken); ruff check + format --check clean; mypy clean.

### Scope
- muhaddithat correctly passes on its free-text `source` field (no sect hack); node-level sect_affiliation/source_corpus split to #103 — agree, that's a resolve→canonical→loader change with a live-Neo4j proof, out of scope for this parser-level gate.

### Non-blocking observations (advisory, no change required)
1. provenance_violations assumes the provenance columns are string-typed; a same-named non-string column routed through write_parquet would raise inside utf8_trim_whitespace rather than return a clean violation message. Safe today (closed schema set); a one-line type-guard would harden it for future non-canonical tables.
2. The "every parser emits a provenance-bearing schema" test matches schema NAMES as source substrings — would pass on a mere comment/mention and miss an aliased import. Fine as a structural guard.

Verdict: Approved. Solid foundation, correct, well-tested, and the fail-fast chokepoint is the durable enforcement choice.

PR #114, opened 2026-06-12
da#100 (X2): cross-sect PARALLEL_OF detection across Sunni + Shia

### LINKED ISSUE #100: X2: cross-sect PARALLEL_OF detection across Sunni + Shia corpora
**Tier:** T3 (cross-source normalization) · **Suggested owner:** Ivana Horvat / Jean-Claude Habimana
Part of epic #81. **KEYSTONE — isnad-graph#964 (parallels, both sects).**

## Goal
**Cross-sect `PARALLEL_OF` detection** — detect parallel hadiths across Sunni and Shia corpora once both are loaded, materializing `PARALLEL_OF` edges.

## Acceptance
- [ ] Parallel detection runs across ≥1 Sunni + the Shia (Thaqalayn) corpus.
- [ ] `PARALLEL_OF` edges loaded into local `neo4j:5`; sample manually validated.

## Dependencies
**L2** (thaqalayn / Shia) + ≥1 Sunni light-up (L1/L3/L4/L5/L6).

parametrization:
## da#100 (X2): cross-sect PARALLEL_OF detection

Detect parallel hadiths across Sunni and Shia corpora — **intra-sunni, intra-shia, AND cross-sect** — and materialize `PARALLEL_OF` edges. The data side of isnad-graph#964 (Browse Parallels must span all three relationship kinds, not exclusively cross-sect). Part of epic #81; builds on the #82 identity contract.

### Detector — `src/resolve/parallels.py` (new)
Deterministic, dependency-light lexical detector: token-set **Jaccard over the normalized matn** (Arabic via `normalize_arabic`, else English lowercased), tiered into `VariantType` by the same thresholds `dedup.py` uses. `cross_sect` is read from the authoritative `sect` column, not inferred from corpus. Writes the identical `PARALLEL_LINKS_SCHEMA` the graph loader already consumes.

**Why a second detector alongside `dedup.py`:** `dedup.py` is the high-recall production path (sentence-transformer embeddings + FAISS) but it needs the model + compute and *degrades to an empty output* when those are unavailable — so it cannot deterministically prove cross-sect detection on the loaded slice or in CI. This one runs offline, in CI, and on the live slice. Recall grows as the embedding path and more corpora land; both write the same schema so the graph contract is unchanged.

### Loader fix — `src/graph/load_edges.py`
`PARALLEL_OF` now keys the edge on the `(hadith_a, hadith_b)` **pair** — bare-edge `MERGE` then `SET … coalesce(…)` — instead of putting `similarity_score`/`variant_type`/`cross_sect` **inside** the MERGE pattern. The old pattern (da#77 / main#139) both trips Neo4j's null-property-in-MERGE refusal and mints a **second** edge between the same pair on any re-run with a changed score. Now idempotent + null-safe — the same contract as `APPEARS_IN`.

### Scope / data-volume note
Per the issue, the deliverable is the **mechanism + a real-data proof**, not a large corpus. Detection is exhaustive O(n²) (ample for the loaded slice; the embedding/FAISS path is the answer at full-corpus scale, documented in the module). The integration test proves all three edge kinds on a controlled slice; recall grows as Itqan (#92a) and other light-ups land.

### Tests / acceptance (owner bar: LIVE local neo4j:5)
- `tests/integration/test_parallel_of_lightup.py` — live `neo4j:5`: intra-sunni AND cross-sect PARALLEL_OF edges materialize (`created==2`, `0` missing endpoints), are queryable the ig#964 way (returns both intra and cross), the cross-sect edge spans `{sunni, shia}`, and a reload with a **changed score** keeps **one** edge (the da#77/main#139 duplicate-edge guard). **2/2 green locally.**
- `tests/test_resolve/test_parallels.py` — tokenization (Arabic/English), Jaccard, tiering, cross_sect-from-sect, symmetric dedup, threshold, empty input.
- Full suite: 613 passed / 3 skipped (heavy-dep dedup) non-integration; ruff + mypy clean.

Closes #100
Refs #81

### DIFF
--- a/src/graph/load_edges.py
+++ b/src/graph/load_edges.py
@@ -383,15 +383,22 @@ def _load_appears_in(
 # 4. PARALLEL_OF — from parallel_links.parquet
 # ---------------------------------------------------------------------------
 
+# Identity of a PARALLEL_OF edge is the (hadith_a, hadith_b) PAIR — the similarity
+# score / variant tier / cross_sect flag are *attributes* of that one edge, not
+# part of its key. They are therefore SET after a bare-edge MERGE, never inside
+# the MERGE pattern (da#77 / main#139): Neo4j refuses to MERGE a relationship with
+# a null property in the pattern, and keying on the properties would mint a SECOND
+# edge between the same pair whenever a re-run produced a slightly different score
+# (dedup is re-run as more corpora land). coalesce-preserve keeps re-runs
+# idempotent and null-safe — exactly the APPEARS_IN contract.
 _PARALLEL_OF_QUERY = """\
 UNWIND $batch AS row
 MATCH (h1:Hadith {id: row.id_a})
 MATCH (h2:Hadith {id: row.id_b})
-MERGE (h1)-[:PARALLEL_OF {
-    similarity_score: row.score,
-    variant_type: row.variant_type,
-    cross_sect: row.cross_sect
-}]->(h2)
+MERGE (h1)-[r:PARALLEL_OF]->(h2)
+SET r.similarity_score = coalesce(row.score, r.similarity_score),
+    r.variant_type = coalesce(row.variant_type, r.variant_type),
+    r.cross_sect = coalesce(row.cross_sect, r.cross_sect)
 """
 
 _PARALLEL_OF_CHECK = """\
--- a/src/resolve/parallels.py
+++ b/src/resolve/parallels.py
@@ -1,0 +1,202 @@
+"""Deterministic lexical PARALLEL_OF detection (da#100).
+
+Detects parallel hadiths — **intra-sunni, intra-shia, and cross-sect** — by
+normalized-token similarity over the matn, and materializes
+``parallel_links.parquet`` (``PARALLEL_LINKS_SCHEMA``): the same artifact the
+graph ``PARALLEL_OF`` loader (``src/graph/load_edges.py``) already consumes. This
+directly feeds isnad-graph#964 (Browse Parallels, which must span all three
+relationship kinds, not exclusively cross-sect).
+
+Why a second detector alongside ``src/resolve/dedup.py``
+--------------------------------------------------------
+``dedup.py`` is the high-recall PRODUCTION path: multilingual
+sentence-transformer embeddings + a FAISS index. It needs the model (a network
+download) and non-trivial compute, and it *degrades to an empty output* when
+those deps are unavailable — so it cannot deterministically prove cross-sect
+detection on the loaded slice or in CI. This module is the **deterministic,
+dependency-light** detector: pure-Python normalized-token similarity that runs
+offline, in CI, and on the live slice, so the mechanism is provable now. Recall
+grows as the embedding path and more corpora land; both detectors write the
+identical ``PARALLEL_LINKS_SCHEMA``, so the downstream loader and the graph
+contract are unchanged.
+
+Matching signal
+---------------
+Token-set Jaccard over the normalized matn — Arabic via
+:func:`src.utils.arabic.normalize_arabic` when an Arabic matn is present, else the
+English matn lowercased. Scores are tiered into :class:`VariantType` by the same
+thresholds ``dedup.py`` uses. ``cross_sect`` is read from the hadith ``sect``
+column (authoritative), not inferred from the corpus name.
+
+Scale note
+----------
+The current pairing is exhaustive O(n²), which is correct and ample for the
+loaded slice. For full-corpus volume the same signal blocks on shared rare
+tokens (an inverted index over significant tokens) before scoring — the
+embedding/FAISS path in ``dedup.py`` is the production answer at that scale.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+import pyarrow as pa
+import pyarrow.parquet as pq
+
+from src.models.enums import VariantType
+from src.resolve.schemas import PARALLEL_LINKS_SCHEMA
+from src.utils.arabic import is_arabic, normalize_arabic
+from src.utils.logging import get_logger
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+logger = get_logger(__name__)
+
+__all__ = ["detect_parallels", "run"]
+
+# Tier thresholds — aligned with dedup.py ``_classify_pair`` so the two detectors
+# label a pair the same way.
+_VERBATIM_MIN = 0.90
+_CLOSE_PARAPHRASE_MIN = 0.80
+# Default minimum similarity to record a pair at all. Lexical Jaccard is stricter
+# than cosine on embeddings, so the floor sits below the thematic tier ceiling;
+# configurable per call.
+_DEFAULT_THRESHOLD = 0.50
+
+
+def _classify_pair(score: float) -> VariantType:
+    """Classify a similarity score into a variant tier (matches dedup.py)."""
+    if score >= _VERBATIM_MIN:
+        return VariantType.VERBATIM
+    if score >= _CLOSE_PARAPHRASE_MIN:
+        return VariantType.CLOSE_PARAPHRASE
+    return VariantType.THEMATIC
+
+
+def _tokenize(matn_ar: str | None, matn_en: str | None) -> frozenset[str]:
+    """Normalized token set for a hadith matn.
+
+    Prefers the Arabic matn (normalized via ``normalize_arabic``); falls back to
+    the English matn lowercased. Returns an empty set when neither is usable, in
+    which case the hadith is excluded from pairing.
+    """
+    if matn_ar and is_arabic(matn_ar):
+        normalized = normalize_arabic(matn_ar)
+    elif matn_en and matn_en.strip():
+        normalized = matn_en.lower()
+    else:
+        return frozenset()
+    return frozenset(tok for tok in normalized.split() if tok)
+
+
+def _jaccard(a: frozenset[str], b: frozenset[str]) -> float:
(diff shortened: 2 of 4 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

QA review of PR #114 at head a90231a (da#100 cross-sect PARALLEL_OF). Reviewed via gh api contents / pulls files (REST — GraphQL quota is down) with no checkout.

Point 1 — live neo4j:5 materializes both an intra and a cross-sect edge, and the idempotency assertion exercises the MERGE-key fix. Verified
- tests/integration/test_parallel_of_lightup.py seeds two near-duplicate Sunni hadiths (intra) plus a Shia/Sunni pair (cross-sect), runs the real detect_parallels then load_all_nodes/load_all_edges into the live testcontainers neo4j:5 (same conftest whose pytest.skip is scoped only to container.start, so a real failure is never an infra skip). It asserts the loader result created == 2 and missing_endpoints == 0, then reads back two PARALLEL_OF rows whose cross_sect flags sort to [False, True], and that the cross-sect edge genuinely spans {sunni, shia}.
- The idempotency leg loads the same pair at score 0.70 / close_paraphrase, reloads it at 0.95 / verbatim, and asserts exactly one edge remains with the score updated to 0.95 and the variant to verbatim. That genuinely exercises the fix: under the old property-in-MERGE pattern the changed score would have minted a second edge (count 2) and the assertion would fail; the bare MERGE + SET coalesce keeps one edge and updates its attributes. The score/variant update assertions also prove the SET path is live, not a no-op.

Point 2 — queryable the ig#964 way returns both intra and cross, not exclusively cross. Verified
- The test runs the Browse-Parallels-style read (MATCH ()-[r:PARALLEL_OF]->() with a CASE split on r.cross_sect) and asserts total == 2, cross == 1, intra == 1. That is a direct reproduction-and-fix of the exact ig#964 symptom (the view previously surfaced only cross-sect): both relationship kinds are now returned.

Point 3 — the Jaccard detector unit tests cover the edge cases. Verified
- Empty: detect_parallels writes an empty table when no pair clears threshold; _jaccard returns 0.0 when either set is empty; _tokenize returns an empty set for None/None and for ""/whitespace, and such hadiths are dropped from pairing.
- Threshold boundary: _classify_pair is asserted at every tier edge (0.90 and 0.95 verbatim, 0.80 and 0.85 close_paraphrase, 0.79 and 0.50 thematic), and the detect-level threshold is exercised both ways (Jaccard ~0.14 below 0.5 yields zero rows; lowering to 0.1 yields one thematic row).
- cross_sect-from-sect: an intra pair (sunni/sunni) flags False and a Shia/Sunni pair flags True, read from the sect column, not the corpus.
- Arabic vs English tokenization: Arabic is preferred and normalize_arabic collapses diacritic variants to the same token set; English is the lowercased fallback when no Arabic matn is usable.
- Also covered: Jaccard math (identical 1.0, disjoint 0.0, partial 0.5), symmetric dedup to a single canonically-ordered row, and the written variant_type equals the StrEnum value (VariantType.VERBATIM.value / .THEMATIC.value) — I checked VariantType is a StrEnum so str(VariantType.X) is the value, and dedup.py uses the identical str(_classify_pair(...)) serialization, so the two detectors label a pair identically.

Point 4 — it does not break the existing FAISS/dedup path. Verified
- src/resolve/dedup.py is unchanged (not among the four changed files). parallels.py is a new standalone module and is NOT wired into any resolve run_all registry in this PR (no __init__.py change), so the embedding/FAISS path runs exactly as before and the two detectors never co-run here.
- The only shared change is the PARALLEL_OF loader query, and it is backward-compatible: dedup.py emits the same PARALLEL_LINKS_SCHEMA, so its output now loads through the same idempotent, null-safe bare-MERGE-then-SET path — which also closes the latent duplicate/null-property bug for dedup's own output. The detector canonically orders id_a < id_b, matching the directional MERGE, so re-runs match the existing edge rather than minting a reversed twin.

Non-blocking observation (not debt incurred here, a note for the future wiring): both detectors write to the same staging/parallel_links.parquet path. As long as only one is invoked per resolve pass (the case today — parallels is called explicitly, dedup via run_all) there is no conflict, but whoever later registers parallels into run_all should decide how the two outputs compose (merge vs. last-writer-wins) so one does not silently overwrite the other. Mentioning so it is on the record; nothing to change in this PR.

CI — all five check-runs are green (Type Check, Lint, Test, Integration Tests, Pre-commit sync-drift gate). Scope is tight: a 12-line loader fix plus a new detector and its tests; no schema or contract change.

parametrization:
RequestOrReplied: Approved
TechDebt: none

## Review — PR #114 (da#100): cross-sect PARALLEL_OF across Sunni + Shia

Reviewed at head `a90231abfced12d043bd9281cd25161bc9048783` via `gh api` contents + git-over-SSH (no checkout; GraphQL quota was exhausted, REST had headroom). CI green at head: Integration Tests (live `neo4j:5`), Test, Type Check, Lint, Pre-commit⇄CI sync-drift all pass. Approved on substance — all four review points verified; one non-blocking follow-up noted at the end.

### 1. Loader fix — both properties hold (highest-scrutiny)
The new `_PARALLEL_OF_QUERY` is a bare-edge `MERGE (h1)-[r:PARALLEL_OF]->(h2)` followed by `SET r.* = coalesce(row.*, r.*)`, replacing the old properties-inside-MERGE pattern.
- (a) Null-property-in-MERGE refusal avoided: similarity_score / variant_type / cross_sect are no longer in the MERGE pattern, so Neo4j never refuses the merge on a null attribute (the da#77 / main#139 class of bug). They are SET after the bare merge.
- (b) Idempotent on a changed score: the bare MERGE keys solely on the (h1, h2) pair, so a re-run matches the one existing edge instead of minting a second; the coalesce SET then updates the (non-null) new score in place. Belt-and-suspenders against a reverse-direction duplicate: BOTH the loader (`if full_a > full_b: swap`) AND the detector (`id_a, id_b = (id_i, id_j) if id_i < id_j else ...`) canonically order the pair low→high, so the same pair can never land as two oppositely-directed edges. Endpoints are validated via `_PARALLEL_OF_CHECK` (OPTIONAL MATCH, missing counted, only both-exist rows written), so the test's `missing_endpoints == 0` is a real assertion.
`tests/integration/test_parallel_of_lightup.py::test_reload_with_changed_score_is_idempotent` proves (b) directly: loads the same pair at score 0.70/close_paraphrase then 0.95/verbatim and asserts `count(r) == 1` with score updated to 0.95 — the exact regression the old pattern failed.

### 2. Detector (parallels.py) — correct and complementary
- Deterministic token-set Jaccard over the normalized matn: Arabic via `normalize_arabic` when `is_arabic`, else English lowercased (`_tokenize`). Pure-Python, offline, no model.
- Tiered into `VariantType` by `_classify_pair` (VERBATIM ≥0.90, CLOSE_PARAPHRASE ≥0.80, else THEMATIC), thresholds aligned with dedup.py.
- `cross_sect` read from the authoritative hadith `sect` column (`bool(sect_i and sect_j and sect_i != sect_j)`), not inferred from corpus name.
- Writes `parallel_links.parquet` against the EXISTING `PARALLEL_LINKS_SCHEMA` (hadith_id_a, hadith_id_b, similarity_score, variant_type, cross_sect) — byte-compatible with what `_load_parallel_of` reads. Confirmed field-name parity between `_write_links` and the loader's `row.get(...)`.
- Genuinely COMPLEMENTS, does not replace, dedup.py: this PR does not touch dedup.py (changed files are load_edges.py + parallels.py + 2 tests). The docstring is accurate that dedup.py is the high-recall FAISS path that degrades to empty without the model, and this is the dependency-light detector that proves the mechanism deterministically. Both emit the identical schema, so the loader/graph contract is unchanged.

### 3. Live proof — real, not mocked
`test_intra_and_cross_sect_edges_materialize` seeds 2 near-duplicate Sunni hadiths + a Shia/Sunni cross-sect pair, runs the REAL `detect_parallels` -> `load_all_nodes` -> `load_all_edges` into a live `neo4j:5`, and asserts: PARALLEL_OF `created == 2`, `missing_endpoints == 0`; read-back `cross_flags == [False, True]`; the cross-sect edge spans `{sunni, shia}`; and the isnad-graph#964 read returns BOTH (`total==2, cross==1, intra==1`) — i.e. parallels-regardless-of-sect, not exclusively cross. The fixture is robust: cross-group Jaccard falls below threshold, so exactly the two intended pairs clear it. CI Integration is green, so this ran against a real container.

### 4. Scope — appropriate and documented
Mechanism + real-data proof. The O(n²) exhaustive pairing is documented as correct-and-ample for the loaded slice with a clear scale path (shared-rare-token inverted index, and dedup.py's FAISS as the production-scale answer). Recall growing as corpora land is stated; low parallel yield at current volume is expected.

### Non-blocking follow-up (recommend filing, not gating)
`src/resolve/__init__.py` `run_all` orchestrates NER -> disambiguate -> dedup and is NOT changed by this PR, so an orchestrated `make resolve` run still calls only `dedup.run` (FAISS) — the deterministic `detect_parallels` is reached via direct call (the integration test and any live-slice runbook step), not via the pipeline. For the no-model / CI / sandbox pipeline to actually emit these links (the environments this detector exists to serve), recommend a fast follow-up to wire `detect_parallels` into the resolve orchestrator — e.g. as a complement, or as the fallback when the embedding model is unavailable. The merged code carries no internal debt (hence TechDebt: none); this is feature-reachability wiring, best tracked as its own issue. Flagging so the live product graph gets populated through the normal pipeline rather than only via a manual call.

Verdict: Approved. Loader fix, detector, and live cross-sect proof are all correct and the loader change also de-risks the existing dedup.py path.

Issue #108, opened 2026-06-12
test(thaqalayn): cover the production github-format parse path (_extract_hadiths_github) — fixtures only exercise API format (follow-up from #105)


parametrization:
## Context

Follow-up flagged during review of PR #105 (da#85 Thaqalayn light-up, now merged). Non-blocking — the load proof stands — but a real production-format gap to close before the Thaqalayn keystone backs live isnad-graph#964 data.

## Gap

Both the e2e and parser unit tests for `thaqalayn` use **API-format fixtures**, but production acquisition is a **git clone** → the parser's separate `_extract_hadiths_github` branch. Neither test suite exercises that github-format parse path. The load layer is format-agnostic over the staging schema, so the merged load proof is valid; but the github-format → staging parse path is currently unproven by fixtures.

## Ask

Add a **github-format fixture** that drives `_extract_hadiths_github` into the canonical staging schema, asserting the same provenance (sect=shia, source_corpus=thaqalayn) + identity-contract conformance as the API-format path. This proves the production parse format before the keystone feeds real ig#964 Browse-Parallels data.

## Refs
- PR #105 / da#85 (the merged Thaqalayn light-up).
- Epic da#81; relates to source-hardening.

parametrization:
Merged to wave branch deployments/phase-4/wave-4 via PR #122.

PR #105, opened 2026-06-12
test(da#85): light up thaqalayn (Shia) end-to-end into Neo4j

parametrization:
## da#85 — light up the Thaqalayn adapter end-to-end (SHIA keystone)

The `thaqalayn` adapter (`src/acquire/thaqalayn.py`) + parser (`src/parse/thaqalayn.py`) already existed and were parse-validated, but the corpus had **never been loaded into the graph**. This is the first **Shia** corpus to land end-to-end — the keystone that unblocks cross-sect Browse Parallels (isnad-graph#964, whose view needs real `sect=shia` hadith in the graph).

No production code changed — the wiring was already correct (parser builds ids via `generate_source_id`, loaders glob `hadiths_*`/`collections_*`). What was missing was **proof it actually loads**. This PR adds that proof.

### What's here
- **`tests/integration/test_thaqalayn_lightup.py`** — live `parse -> load_nodes -> load_edges` against a real `neo4j:5-community` container (skip-guarded via the shared `neo4j_client` fixture when Docker is down, #69). Asserts, reading back from the live graph:
  - real Hadith + Collection nodes persist tagged `sect=shia` + `source_corpus=thaqalayn` (not just parse/conformance);
  - node ids are single-prefixed `hdt:thaqalayn:…` and not double-prefixed (the da#82 / main#139 identity contract), via `is_double_prefixed` + `validate_source_id`;
  - every Shia hadith gets an `APPEARS_IN` edge to its Shia collection (0 missing endpoints);
  - a `sect=shia` lookup is answerable — the concrete capability isnad-graph#964 keys on.
- **`tests/test_acquire/test_thaqalayn.py`** — closes the acquire-side test gap (only `base` + `sunnah_scraper` had one): clone happy path, idempotent skip, under-minimum guard, with the network mocked.

### Live load counts (real neo4j:5-community, this PR)
```
5 hadiths  (Al-Kafi ×3 + Man La Yahduruhu al-Faqih ×2)
2 collections
5 APPEARS_IN edges, 0 missing endpoints
sect=shia, source_corpus=thaqalayn on every node
```
`5 passed in 16.45s` against the real container; full new-file suite (acquire + parser + integration) green; ruff (v0.15.11) format+lint and mypy strict clean.

### Reachability + licensing
[...]

PR #122, opened 2026-06-12
test(thaqalayn): cover _extract_hadiths_github production parse path (#108)

parametrization:
Closes #108

## What & why

Follow-up from review of PR #105 (da#85 Thaqalayn light-up). Production Thaqalayn acquisition is a `git clone` of the ThaqalaynAPI repo, so the parser runs its **GitHub-format branch** (`_extract_hadiths_github`). But both the existing parser-unit (`test_thaqalayn_parser.py`) and e2e (`test_thaqalayn_lightup.py`) fixtures seed only **API-format** JSON (the `data`-wrapper branch). The production parse path was therefore unproven by fixtures — a bug there would slip through, before the Shia keystone feeds real isnad-graph#964 Browse-Parallels data.

## What the test covers

New `TestThaqalaynGithubFormat` + a `github_clone/` fixture (no top-level `book_*.json`, so format detection takes the GitHub branch — mirrors `src/acquire/thaqalayn.py`'s clone layout). It seeds both GitHub-format shapes `_extract_hadiths_github` handles:
- **bare array** of hadith objects, including a non-hadith metadata dict that `_looks_like_hadith` must filter out;
- **wrapper dict** whose hadiths live under a `"hadiths"` key.

Assertions hold the GitHub path to the **same staging contract as the API path**:
- `HADITH_SCHEMA` / `COLLECTION_SCHEMA` conformance;
- row count = 3 (2 bare-array, non-hadith dict filtered; + 1 wrapper-dict);
- both shapes' hadiths reach staging;
- provenance `sect=shia` + `source_corpus=thaqalayn` on every hadith and collection row;
- **identity contract**: `validate_source_id(sid) == []` and `not is_double_prefixed(sid)` for every `source_id` (the da#82 / main#139 contract the live light-up asserts);
- grade-array → JSON-string serialization preserved.

## Bug found?

No. The GitHub path produces correct, identity-conformant staging rows. Test-only — no production code changed.

## Test Plan
- `ENVIRONMENT=test uv run pytest tests/test_parse/test_thaqalayn_parser.py -q` → 11 passed (4 existing API + 7 new GitHub).
- `ruff@0.15.11 format --check` + `check` clean; `mypy` (strict) clean.

Reviewers: @Kwesi Boateng (primary), @Nikolaos Papadopoulos (secondary).

### DIFF
--- a/tests/test_parse/test_thaqalayn_parser.py
+++ b/tests/test_parse/test_thaqalayn_parser.py
@@ -7,6 +7,7 @@ from pathlib import Path
 
 import pyarrow.parquet as pq
 
+from src.parse.identity import is_double_prefixed, validate_source_id
 from src.parse.schemas import COLLECTION_SCHEMA, HADITH_SCHEMA
 from src.parse.thaqalayn import run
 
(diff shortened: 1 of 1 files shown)


parametrization:
Requestor: Kwesi Boateng
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved

Reviewed at head 082406c7 against the production parser source. This is a clean, genuine fill of the da#108 gap.

What I verified (not just the PR framing):

- **Drives the github path, not the API path.** The fixture is seeded under `thaqalayn/github_clone/dataset/` with no top-level `book_*.json`, so `run()` discovers via the `github_clone` rglob branch. Per-file, `_detect_format` returns the github extractor for both shapes — the bare list (not a dict) and the wrapper dict (no top-level `data` list, only `hadiths`). Both genuinely reach `_extract_hadiths_github`.
- **The non-hadith filter is really asserted.** The bare array carries `{"_meta": "license", "version": "2.0"}`, whose keys share nothing with `_HADITH_FIELD_INDICATORS`, so `_looks_like_hadith` drops it. The `num_rows == 3` assertion (2 bare + 1 wrapper) only holds if that filter fires — a broken filter yields 4. This is a meaningful assertion, not a coincidence.
- **Identity and provenance match the da#82 / main#139 contract.** Each row's `source_id` is checked with `validate_source_id(...) == []` and `not is_double_prefixed(...)`, and every row carries the shia sect plus the thaqalayn corpus on both hadiths and collections.
- **Grade serialization is correct.** Single string grade preserved as-is, the two-element array round-trips through json, matching `_normalize_grade`.
- **Truly test-only.** One file, 161 additions, 0 deletions, no production parser touched.
- **Fixture is hand-authored and public-domain.** Synthetic Arabic snippets generated at test time under tmp_path; nothing scraped from upstream is committed.

CI is green across all five checks (Test, Type Check, Lint, Integration, sync-drift gate).

Nice touch using generated-in-tmp_path fixtures rather than committing a github_clone tree. Approved.

parametrization:
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved

Secondary review, data-quality lens (da#108).

**Load-bearing against a real regression — yes.** I traced the fixture to production acquisition. `src/acquire/thaqalayn.py` clones the ThaqalaynAPI repo into `raw/thaqalayn/github_clone` and discovers files via `rglob("*.json")`; `parse/thaqalayn.run()` falls to the GitHub branch only when no top-level `book_*.json` exists and then walks `github_clone` the same way. The fixture seeds `github_clone/dataset/*.json` with no `book_*.json`, so detection genuinely takes `_extract_hadiths_github` (the gap PR #105 left, where only the API `data`-wrapper branch was covered). A regression in detection, extraction, or `_looks_like_hadith` filtering would now fail these tests. Both extractor shapes are covered — the bare array and the `"hadiths"` wrapper dict — which are the two real branches in `_extract_hadiths_github`, and the trailing non-hadith metadata dict proves the filter drops it (num_rows == 3, not 4).

**Schema assertions are strict enough.** `table.schema == HADITH_SCHEMA` and `== COLLECTION_SCHEMA` are pyarrow Schema equality (field names, types, and nullability), not a row-count proxy. Provenance is asserted on every row (sect=shia, source_corpus=thaqalayn) for both hadiths and collections, and the identity contract is the same one the live light-up uses — `validate_source_id == []` and `not is_double_prefixed` — confirming GitHub-path `source_id`s are canonical and not double-prefixed. Helpers resolve (identity.py 202/218); all five CI checks pass at head 082406c.

**In-book ordinal vs collection-ref** is not in play here — that distinction lives on the graph loader's APPEARS_IN edge, not the Thaqalayn staging schema, and these tests correctly do not assert on `hadith_number` values.

Non-blocking follow-up nits (optional, do not gate this merge):
- `_normalize_grade` has two untested branches — the single-element-array form (collapses to the scalar) and the semicolon-split form (`"a; b"` to a JSON array). The multi-element array and plain scalar are both covered; adding the other two would fully pin the grade normalizer.
- No fixture hadith carries an `isnad`/`isnad_ar` field, so `isnad_raw_ar` is always None and the `thaqalayn_isnad_rate` accounting path stays at zero. Fine given Thaqalayn rarely separates isnad, but one isnad-bearing row would exercise that counter.

These are coverage extensions, not correctness gaps. The PR does exactly what da#108 asked and closes the production-path gap honestly (test-only, no source change). Approved.

Issue #113, opened 2026-06-12
tech-debt: static CI guard for generate_source_id call-site corpus (prevent #89-style crash)


parametrization:
## Tech debt: static guard for `generate_source_id` call-site corpus

`generate_source_id()` (since #82) raises on an unknown corpus — correct, but it fails **at runtime** only when the real acquisition dir is present, so a caller passing a non-`SourceCorpus` first arg (e.g. a bio `source` label like `"kaggle_narrators"`) aborts the whole parse and is masked in unit tests that build no raw dir. This was the #89 sanadset crash.

An AST sweep confirmed the current blast radius is exactly **one** site (`sanadset.py:241`, fixed in #106) — not systemic. But nothing **prevents** a future caller from reintroducing it.

### Proposal (low priority)
Add a small CI test that statically asserts every `generate_source_id(<literal-or-module-const>, ...)` call-site across `src/parse/*.py` passes a first arg that resolves to a `SourceCorpus` value (the AST scan I used to bound the blast radius — ~20 lines). Bio/`source` provenance keys (free-text, not corpus-namespaced) should build via the direct `ID_DELIMITER` path instead.

This is **orthogonal** to #83's provenance gate (#104), which deliberately allows free-text `source` values and so does not catch this.

### Acceptance
- [ ] CI test fails if a `generate_source_id` call-site uses a non-`SourceCorpus` literal/const first arg.
- [ ] Doc note: bio/`source` keys use the direct `ID_DELIMITER` path, not `generate_source_id`.

Refs #81, #82, #89. Surfaced during da#87.

parametrization:
Merged to wave branch deployments/phase-4/wave-4 via PR #123 (static source_id-corpus CI guard).

PR #123, opened 2026-06-12
tech-debt(ci): static guard for generate_source_id call-site corpus (#113)

parametrization:
Closes #113.

## What this catches

`generate_source_id()` fails fast on an unknown corpus, but only **at runtime** (when the real acquisition dir is present). A caller passing a non-corpus first arg — e.g. a free-text bio `source` label like `"kaggle_narrators"` — aborts the whole parse and is masked by unit tests that build no raw dir. That was the **da#89** sanadset crash. An AST sweep confirms the live blast radius is currently zero call-sites, but nothing prevented reintroduction.

This adds a static guard that asserts every `generate_source_id` call-site passes a statically-provable `SourceCorpus` first arg, catching the bug at **CI time** instead of at runtime parse.

## Changes

- **`scripts/check_source_id_corpus.py`** — standalone, importable AST guard + CLI (exit 1 on violation; defaults to scanning `src/parse`). The authoritative valid set is `SourceCorpus` itself (imported, not re-listed).
- **`tests/test_parse/test_source_id_corpus_guard.py`** — runs in CI's existing `Test` pytest job, so **no new CI job and no pre-commit ⇄ CI sync-drift entry is needed** (the gate rides a workflow already mirrored in `.pre-commit-config.yaml`; sync-drift gate verified green locally). Asserts the live `src/parse` tree is clean (with a non-vacuous-green sanity check) + positive/negative discrimination cases.
- **`src/parse/identity.py`** — doc note only (the canonical `source_id`-grammar home, per `base.py`'s docstring pointer): bio/`source` free-text keys build via `ID_DELIMITER.join(...)`, never `generate_source_id`. **Acceptance criterion #2.**

## Accepted first-arg forms
- string literal that is a `SourceCorpus` value (`"lk"`)
- module-level string constant resolving to one (`SOURCE_CORPUS = "fawaz"`, `_SOURCE_CORPUS = "sanadset"`)
- enum-member access `SourceCorpus.LK` / `SourceCorpus.LK.value`

Everything else (unknown literal/constant, unknown enum member, or a non-static runtime expression) is a violation.

## Access forms covered (lint-gate-cover-all-syntactic-forms)
- bare `generate_source_id(...)` after `from src.parse.base import generate_source_id`
- dotted `<module>.generate_source_id(...)` (e.g. `base.generate_source_id(...)`)

**Form deliberately not chased** (documented in the script): a corpus value rebound through a local (function-scope) variable or imported under an alias — it reads as a non-static expression and is flagged, which is the safe default.

## #81 sequencing note
This guard **reads** `src/parse/base.py` and `src/models/enums.py` via call-site analysis but does **not** edit either. The only `src/` change is a docstring note in `src/parse/identity.py`. If #81 changes `generate_source_id`'s signature or location, the guard's `FUNC_NAME`/scan may need a trivial rebase — but there is no code-line conflict with `base.py`/`enums.py`.

## Test Plan
- `python scripts/check_source_id_corpus.py` → `OK ([src/parse] clean)`, rc=0
- `ENVIRONMENT=test uv run pytest tests/test_parse/test_source_id_corpus_guard.py -q` → 17 passed
- `uv run ruff format --check` + `uv run ruff check src/ tests/` → clean
- `uv run mypy src/` → Success, no issues in 68 files
- `python3 .claude/lib/pre_commit_ci_sync.py . --ci .github/workflows/ci.yml` → OK (mirror intact)

Reviewers: @Jean-Claude Habimana (architect, primary) + @Oyunbileg Batbayar (QA, secondary).

### DIFF
--- a/scripts/check_source_id_corpus.py
+++ b/scripts/check_source_id_corpus.py
@@ -1,0 +1,263 @@
+#!/usr/bin/env python3
+"""Static guard for ``generate_source_id`` call-site corpus arguments.
+
+``src.parse.base.generate_source_id(corpus, collection, *parts)`` fails fast
+(raises ``ValueError``) when *corpus* is not a known :class:`SourceCorpus`
+value. That check only fires **at runtime**, when the real acquisition dir is
+present — so a caller that passes a non-corpus first arg (e.g. a free-text bio
+``source`` label like ``"kaggle_narrators"``) aborts the whole parse and is
+masked in unit tests that build no raw dir. That was the da#89 sanadset crash.
+
+This guard catches the same class of bug **at CI time**: it AST-scans every
+``generate_source_id`` call-site and asserts the first positional argument
+statically resolves to a :class:`SourceCorpus` value. Free-text provenance keys
+(bio ``source`` values, etc.) must build their ids via the direct
+``ID_DELIMITER.join(...)`` path instead — see ``sanadset.py`` ``bio_id``.
+
+Run standalone (defaults to scanning ``src/parse``)::
+
+    python scripts/check_source_id_corpus.py
+    python scripts/check_source_id_corpus.py src/parse/sanadset.py
+
+Exit code is 1 if any violation is found, 0 otherwise. The same scan functions
+are exercised by ``tests/test_parse/test_source_id_corpus_guard.py``, which runs
+in CI's existing pytest job (so no new CI job / pre-commit ⇄ CI sync entry is
+required — the gate rides the ``Test`` workflow already mirrored in
+``.pre-commit-config.yaml``).
+
+Access forms covered (per the lint-gate-cover-all-syntactic-forms lesson):
+
+* bare call after ``from src.parse.base import generate_source_id`` —
+  ``generate_source_id(...)``
+* dotted/attribute call — ``base.generate_source_id(...)`` / any
+  ``<module>.generate_source_id(...)``
+
+First-arg forms accepted as valid:
+
+* a string literal whose value is a ``SourceCorpus`` value (``"lk"``)
+* a *module-level* string constant that resolves to one (``SOURCE_CORPUS = "fawaz"``)
+* an enum member access ``SourceCorpus.LK`` or ``SourceCorpus.LK.value``
+
+Anything else (an unknown literal/constant, an unknown enum member, or a
+non-static runtime expression such as a bare parameter) is a violation:
+it cannot be statically proven corpus-safe, which is exactly the reintroduction
+risk this guard exists to block.
+
+Form deliberately **not** chased: a corpus value rebound through a local
+(non-module-level) variable or imported from another module under an alias.
+Current call-sites never do this, and resolving arbitrary dataflow is out of
+scope for a line-of-defence static gate; such a first arg simply reads as a
+non-static expression and is flagged, which is the safe default.
+"""
+
+from __future__ import annotations
+
+import argparse
+import ast
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+_REPO_ROOT = Path(__file__).resolve().parent.parent
+if str(_REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(_REPO_ROOT))
+
+from src.models.enums import SourceCorpus  # noqa: E402  (after sys.path bootstrap)
+
+FUNC_NAME = "generate_source_id"
+ENUM_NAME = "SourceCorpus"
+
+#: Valid first-arg string values (the ``SourceCorpus`` *values*, e.g. ``"lk"``).
+VALID_CORPORA: frozenset[str] = frozenset(c.value for c in SourceCorpus)
+#: Valid enum *member* names (e.g. ``LK`` in ``SourceCorpus.LK``).
+VALID_MEMBERS: frozenset[str] = frozenset(c.name for c in SourceCorpus)
+
+
+@dataclass(frozen=True)
+class Violation:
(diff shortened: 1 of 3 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Tarek Mansour
RequestOrReplied: Approved

QA / data-quality secondary review of da#113 (static CI guard for `generate_source_id` call-site corpus). Approving — the guard is correct, well-tested, and provably non-vacuous.

## Verified locally at HEAD (e674d92), not just CI-green

**Coverage of the 17 tests — all four QA axes are met:**

- Positive AND negative discrimination — accepted forms (string literal, module constant, leading-underscore constant matching `sanadset.py`'s `_SOURCE_CORPUS`, enum member, enum `.value`, dotted) each have a no-violation test; rejected forms (bad literal = the exact `"kaggle_narrators"` da#89 shape, bad module constant, unknown enum member, non-static runtime expr, undeclared name, function-local constant) each have a one-violation test.
- Both access forms policed — bare `generate_source_id(...)` and dotted `base.generate_source_id(...)` are each covered on both the valid and the invalid side (`test_valid_dotted_access_form` / `test_dotted_access_form_invalid_corpus_is_flagged`). Satisfies the lint-gate-cover-all-syntactic-forms lesson.
- CLI exit codes — witnessed `rc=0` on the clean live tree and `rc=1` on a synthetic bad file (with a clear remediation message pointing at `ID_DELIMITER.join`); locked by `test_main_returns_zero_on_clean_tree` / `test_main_returns_one_on_dirty_file`.
- Live-tree regression gate — `test_live_parse_package_is_clean` asserts the real `src/parse` package is clean, and `test_guard_actually_finds_the_call_sites` blocks a vacuous green from a rename.

**Non-vacuous proof (the key QA check):** I injected a da#89-shape call-site (`generate_source_id("kaggle_narrators", ...)`) into a real `src/parse` module and the regression gate went red with a precise violation; reverting restored green. The gate genuinely fires on the bug class it exists to block. 15 real call-sites across 8 parse modules (all using `SOURCE_CORPUS`/`_SOURCE_CORPUS` module constants) are exercised.

All 17 tests pass locally; all 5 CI checks green.

## Non-blocking observations (optional follow-up, not gating)

- A keyword-form call `generate_source_id(corpus="...")` lands on the `<no positional args>` branch and is flagged as a violation — a safe default, but that branch has no dedicated test. A one-line test would close the last coverage seam.
- Scan scope is `src/parse` only (the default). `generate_source_id` is defined in `src.parse.base`, so any future caller in another package (e.g. `src/graph`) would be unguarded. The docstring documents this as intentional; fine as a line-of-defence scope.

parametrization:
Requestee: Tarek Mansour
RequestOrReplied: Approved

Architect-lens review of da#113 (static CI guard for generate_source_id call-site corpus). Verified at head e674d92 against the live src/parse tree.

## What I verified

**Call-site detection covers both access forms (bare + dotted).** `_is_target_call` matches `ast.Name` with id == "generate_source_id" (bare, after `from src.parse.base import generate_source_id`) and `ast.Attribute` with attr == "generate_source_id" (dotted `base.generate_source_id(...)`). Both forms are exercised by the suite (valid + invalid positive cases for each). This satisfies the lint-gate-cover-all-syntactic-forms lesson — every current import is the unaliased bare name (confirmed across fawaz/lk_corpus/sunnah_api/sunnah_scraped/thaqalayn), so the dotted+bare pair covers 100% of real call-sites.

**First-arg validation is correct against the real enum.** VALID_CORPORA / VALID_MEMBERS are derived from SourceCorpus at import time, so the guard tracks the enum (lk/sanadset/thaqalayn/sunnah/fawaz/open_hadith/muhaddithat/itqan) without a hardcoded duplicate. I walked all nine live call-sites: literal `"lk"` (lk_corpus), module constants SOURCE_CORPUS/_SOURCE_CORPUS resolving to fawaz/open_hadith/sanadset/sunnah/thaqalayn — every one resolves to a valid corpus. `_module_str_consts` handles both Assign and AnnAssign at module scope, which is why the underscore-prefixed sanadset constant resolves.

**The not-chased forms fail SAFE.** A corpus rebound through a function-local var or aliased-import first arg reads as a non-static expression and is flagged (false-positive, not false-negative) — `test_local_constant_does_not_count_as_module_constant` locks this in. For a line-of-defence gate against reintroducing the da#89 crash shape, erring toward flagging is the right default, and it is documented.

**The test is non-vacuous.** `test_live_parse_package_is_clean` is the real regression gate (asserts live src/parse clean); `test_guard_actually_finds_the_call_sites` blocks a vacuous green from a rename matching nothing; and the rejection cases (the exact "kaggle_narrators" da#89 shape as both literal and module constant, unknown enum member, runtime expression, undeclared name) prove a bad call-site WOULD fail — for both bare and dotted forms.

**Riding the existing Test job is sound.** The test lives in tests/test_parse/ and runs in the existing pytest job — no new CI job, so the pre-commit sync-drift gate stays green (and did). CI is 5/5 green (Lint, Type Check, Test, Integration, sync-drift).

## One non-blocking note

The docstring describes the not-chased "aliased import" case purely as a first-arg concern. Worth a one-line acknowledgement in a future touch that the guard keys on the literal function name, so a hypothetical `from src.parse.base import generate_source_id as gsid` would make a call-site invisible (a miss, not a flag). This is moot today — zero aliased imports and zero call-sites outside the scanned src/parse — and an engineer deliberately aliasing to smuggle a bad corpus is outside the reintroduction threat model. No change required.

Scope-tight, well-documented, fails safe, and the regression gate is real. Approved.

Issue #91, opened 2026-06-12
L8: sunnah_api adapter-ready + skip-guarded live leg (blocked on da#71 key)


parametrization:
**Tier:** T1 (light up existing adapter) · **Suggested owner:** Kwesi Boateng
Part of epic #81.

## Goal
Make the **sunnah_api** (sunni, api.sunnah.com/v1) adapter **adapter-ready + skip-guarded** (per owner decision). Adapter exists (`src/acquire/sunnah_api.py`, `src/parse/sunnah_api.py`).

## Constraint
**Blocked on da#71** — the API key was never granted (403). Build/validate the adapter fully; **skip-guard the live HTTP leg** behind the missing `SUNNAH_API_KEY` so CI is green and the live load runs once the key arrives (owner action). Do NOT block this on the key.

## Acceptance (adjusted for the blocked live leg)
- [ ] Adapter + parser validated; parse/conformance tests pass.
- [ ] Live acquire + local-Neo4j load **skip-guarded** behind `SUNNAH_API_KEY` (xfail/skip when absent).
- [ ] Reachability + licensing note documenting the da#71 dependency.

## Dependencies
T0-A; da#71 (owner key) for the live leg only.

parametrization:
Merged to wave branch deployments/phase-4/wave-4 via PR #126.

PR #126, opened 2026-06-12
test(sunnah_api): skip-guarded live leg + parse coverage (#91)

parametrization:
Closes #91

Makes the **sunnah_api** adapter *adapter-ready + skip-guarded* per the owner decision: CI stays green keyless, while the live `api.sunnah.com/v1` leg lights up once the **da#71** API key is provisioned (owner action — not a blocker here).

## What this adds
`tests/test_acquire/test_sunnah_api.py` — two legs:

1. **Keyless graceful-skip** (always runs, no network) — asserts `sunnah_api.run()` returns `None` and writes nothing when `SUNNAH_API_KEY` is absent. This locks the contract CI relies on so the source is adapter-ready without failing the build.
2. **Live API smoke** (skip-guarded) — `@pytest.mark.skipif(not os.getenv("SUNNAH_API_KEY"), reason="da#71: ...")`. Fetches the real `/collections` endpoint and runs the **real parser** to a schema-conforming `collections_sunnah.parquet`. Bounded to `/collections` (not a full `run()`) to avoid a multi-thousand-hadith pull in a smoke test. Skips cleanly off-key.

## Skip-guard mechanism
`@pytest.mark.skipif(not os.getenv("SUNNAH_API_KEY"), reason=...)` on the live test class — mirrors the repo's degrade-cleanly-when-infra-absent idiom (cf. the Docker `pytest.skip` in `tests/integration/conftest.py`). The keyless leg additionally proves the adapter's own graceful-skip (`run() -> None`) so the no-key path is exercised on every CI run.

## Live-leg attestation (da#71)
The live `api.sunnah.com/v1` leg is **intentionally deferred to da#71** — the API key was never granted (403 keyless), an EXTERNAL/owner dependency. It is not a blocker for this PR:
- Parse/conformance coverage already exists in `tests/test_parse/test_sunnah_parser.py` (fixture-driven, always runs).
- The dependency is already documented: `src/adapters.py` `sunnah` row carries `reachable=False` + a `license_note` citing da#71, and `docs/adapters.md` lists `sunnah` as `reachable: no`. Left **as-is** (scope: do not flip the registry flag until the key lands).

## Test Plan
- `ENVIRONMENT=test uv run pytest tests/test_acquire/test_sunnah_api.py tests/test_parse/test_sunnah_parser.py -v` → 6 passed, 1 skipped (live leg skips, no key).
- `ENVIRONMENT=test uv run pytest tests/test_acquire/ tests/test_messaging/ tests/test_adapters.py -q` → 115 passed, 1 skipped (no cross-test contamination from the settings cache_clear).
- `ruff check` / `ruff format --check` / `mypy` on the new file → all clean.
- **Post-key (da#71, owner):** run the suite with `SUNNAH_API_KEY` set — the live `/collections` smoke executes against the real API and parses to a conforming Parquet.

Reviewers: @Tarek Mansour (devops, primary) + @Dilara Erdogan (pipeline manager, secondary).

### DIFF
--- a/tests/test_acquire/test_sunnah_api.py
+++ b/tests/test_acquire/test_sunnah_api.py
@@ -1,0 +1,97 @@
+"""Tests for the Sunnah.com API acquire downloader (da#91).
+
+Two legs:
+
+* **Keyless graceful-skip** (always runs, no network) — locks the contract CI
+  relies on: with ``SUNNAH_API_KEY`` absent, ``run()`` returns ``None`` and
+  writes nothing, so the source is *adapter-ready* without failing the build.
+* **Live API smoke** (skip-guarded behind ``SUNNAH_API_KEY``) — exercises the
+  real ``api.sunnah.com/v1`` leg + the real parser on live data, but only when a
+  key is present.  The key was never granted (403 keyless, **da#71**), so this
+  leg ``skip``s off-key and lights up once the owner provisions the key.  See
+  the ``sunnah`` row in ``src/adapters.py`` (``reachable=False``) and
+  ``docs/adapters.md``.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+from pathlib import Path
+
+import pyarrow.parquet as pq
+import pytest
+
+from src.acquire import sunnah_api
+from src.config import get_settings
+from src.parse.schemas import COLLECTION_SCHEMA
+from src.parse.sunnah_api import run as parse_run
+
+_HAS_KEY = bool(os.getenv("SUNNAH_API_KEY"))
+_SKIP_REASON = "da#71: SUNNAH_API_KEY not granted — live Sunnah.com API leg skip-guarded"
+
+
+class TestSunnahApiKeylessSkip:
+    """The keyless graceful-skip path — always runs, never touches the network."""
+
+    def test_run_returns_none_without_key(
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        # Ensure the key is absent regardless of the ambient environment, then
+        # drop the cached Settings singleton so the absence is observed.
+        monkeypatch.delenv("SUNNAH_API_KEY", raising=False)
+        get_settings.cache_clear()
+
+        raw_dir = tmp_path / "raw"
+        result = sunnah_api.run(raw_dir)
+
+        # Graceful skip: no output dir, no network, no writes.
+        assert result is None
+        assert not (raw_dir / "sunnah").exists()
+
+
+@pytest.mark.skipif(not _HAS_KEY, reason=_SKIP_REASON)
+class TestSunnahApiLive:
+    """The live API leg — runs only when a real SUNNAH_API_KEY is present."""
+
+    def test_live_collections_acquire_and_parse(self, tmp_path: Path) -> None:
+        """Fetch the real collections endpoint and parse it to conforming Parquet.
+
+        Bounded on purpose: this hits ``GET /collections`` (a few dozen rows)
+        rather than calling ``sunnah_api.run()``, which would pull every hadith
+        of every collection.  The full production load is the pipeline's job
+        (``make acquire`` / ``make parse``); this leg just proves the key
+        authenticates and real data flows through the parser cleanly.
+        """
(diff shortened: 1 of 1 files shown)


parametrization:
Requestor: Dilara Erdogan
Requestee: Kwesi Boateng
RequestOrReplied: Approved

Secondary review (pipeline-manager lens, diff + CI only).

**All three da#91 acceptance criteria genuinely met — closing #91 here is correct.**
- Parse/conformance (pre-existing) — `tests/test_parse/test_sunnah_parser.py` imports `src.parse.sunnah_api.run` and asserts both HADITH and COLLECTION schema conformance across 5 tests. Not this PR, already green.
- Live acquire skip-guarded (this PR) — new `tests/test_acquire/test_sunnah_api.py`. Keyless leg always runs with no network, locks the graceful-skip contract (`run()` returns None, writes nothing). Live leg is `skipif(not _HAS_KEY)` and bounded to `GET /collections`, proving the real auth+parse path without pulling every hadith.
- Reachability + licensing note (already in registry) — the `sunnah` row in `src/adapters.py` carries `reachable=False` plus the da#71 "403 keyless, never granted" note.

**Live-leg deferral is documented, not silently dropped.** Module docstring, the skip marker, and `_SKIP_REASON` all cite da#71 and the owner-key dependency.

**Scope is clean.** One test file, +97/-0, no registry flag flip — the adapter stays `reachable=False` until the owner provisions the key, which is the right call.

**Latent-break check passed.** I confirmed every live-leg-only symbol exists at PR head — `sunnah_api.BASE_URL`, `sunnah_api.run`, `fetch_json_paginated` in `src/acquire/base.py`, and `get_settings().sunnah_api_key`. The test's `X-API-Key` header mirrors the real acquire path, so the skip-guarded leg will light up correctly the moment a key arrives rather than erroring on a stale import.

CI green across all 5 checks (Lint, Type Check, Test, Integration Tests, sync-drift gate).

Approved.

parametrization:
Requestor: Tarek Mansour
Requestee: Kwesi Boateng
RequestOrReplied: Approved

Reviewed by diff + CI + hand-verified source signatures. All 5 checks are green (Lint, Type Check, Test, Integration Tests, sync-drift gate).

What I verified, point by point:

1. Skip-guard idiom is correct. The class-level `@pytest.mark.skipif(not _HAS_KEY, ...)` keeps CI green keyless and lights the live leg up the moment a real key is provisioned (da#71). Reason string cites the issue. Good.

2. Keyless leg genuinely locks the graceful-skip contract. I read `src/acquire/sunnah_api.py` — `run()` reads `get_settings().sunnah_api_key`, and on falsy key it logs and returns `None` BEFORE any `ensure_dir`. So the two assertions (`result is None`, `not (raw_dir/"sunnah").exists()`) are exactly the right contract, and the `monkeypatch.delenv` + `cache_clear()` combo makes the absence deterministic regardless of ambient env.

3. Bounding the live leg to GET /collections (rather than a full `run()` that pulls every hadith of every collection) is the right call for a smoke leg. The docstring rationale is accurate.

4. No cross-test contamination of concern. Both tests clear the lru-cached settings at the START (defensive read), and monkeypatch auto-restores env after the keyless test. In CI (no ambient key) there is no observable leak.

One thing worth flagging for the record, NOT a blocker. The live leg is exercised by neither lane in automation — it is skip-guarded off keyless, and CI runs `uv run mypy src/`, which excludes `tests/`, so the signatures in that leg get no static check either. I therefore hand-verified them against branch source and they all line up:
- `sunnah_api.BASE_URL` is `https://api.sunnah.com/v1`; the test composes `.../collections`, matching the production call in `run()`.
- `fetch_json_paginated(base_url, *, headers, limit=50)` — the test's positional-url + `headers=` + `limit=` call is valid and mirrors production usage.
- `parse_run(raw_dir, staging_dir)` skips missing `*_hadiths.json` and writes `collections_sunnah.parquet` under `COLLECTION_SCHEMA` when collections exist, so `collections_parquet in output_files` and the schema/row assertions hold.

Since those are runtime-only and will surface loudly the first time the key lands, that residual is acceptable for a skip-guarded smoke leg.

Nice, tight change. Approving.

Issue #117, opened 2026-06-12
tech-debt(resolve): wire detect_parallels + bio_promote into run_all orchestrator (cross-sect edges + promoted narrators not in the pipeline path)


parametrization:
## Context
Follow-up from PR #114 (da#100 cross-sect PARALLEL_OF, merged). The deterministic `detect_parallels` (src/resolve/parallels.py) is currently reached ONLY via direct call (its test / a live-slice runbook), NOT via the resolve orchestrator `src/resolve/__init__.py` run_all, which still calls only `dedup.run` (FAISS — degrades to empty without the model).

## Problem
An orchestrated `make resolve` run in CI / no-model environments produces ZERO cross-sect PARALLEL_OF edges into the product graph, because the deterministic detector is not in the pipeline. So ig#964 only lights up if someone runs detect_parallels directly.

## Fix
Wire `detect_parallels` into the resolve orchestrator (run_all) — as a complement to dedup, or as the no-model fallback when the embedding model/deps are unavailable. Decide composition vs last-writer-wins for the shared `staging/parallel_links.parquet` output (both detectors write it; Oyunbileg flagged this on #114). Reconciliation-invariant note: disambiguate OVERWRITES narrators_canonical while bio_promote MERGEs, so order disambiguate before bio_promote when wiring (Alejandra, #99).

## Refs
PR #114 / da#100; epic #81; relates to da#101 (bio-merge) + the run_all wiring.

parametrization:
Merged to wave branch deployments/phase-4/wave-4 via PR #127.

Issue #109, opened 2026-06-12
Backfill canonical_narrator_id into resolved mentions → wire NARRATED edges + Chains


parametrization:
## Gap: canonical_narrator_id is never backfilled into resolved mentions → no NARRATED edges / empty Chains

Surfaced by the da#89 sanadset light-up (PR #106). The narrator-resolution pipeline
produces canonical **Narrator nodes**, but the narrator→hadith **wiring never
materializes** because the resolved-mention rows the edge/chain loaders read are
never updated with the canonical id.

### Mechanism (precise)
- `src/resolve/ner.py` writes `narrator_mentions_resolved.parquet` with
  `canonical_narrator_id = None` (placeholder).
- `src/resolve/disambiguate.py::run` resolves each mention and writes
  `narrators_canonical.parquet` + a `merge_log.parquet` (`mention_id → canonical_id`),
  but it does **NOT** write the resolved `canonical_narrator_id` back into
  `narrator_mentions_resolved.parquet`.
- The graph loaders read `canonical_narrator_id` straight off the resolved mentions:
  - `src/graph/load_edges.py::_load_narrated` keys the first-narrator→hadith
    `NARRATED` edge on `row.get("canonical_narrator_id")` → always `None` → **0 edges**.
  - `src/graph/load_nodes.py::_load_chains` collects `narrator_ids` from
    `m.get("canonical_narrator_id")` → **empty chains** (`chain_length=0`,
    `is_complete=False`).

Net: Narrator nodes exist and are searchable (the da#89 keystone — delivered), but
they are **disconnected** from hadiths/chains.

### Fix direction
Backfill `canonical_narrator_id` (and `confidence`) onto the resolved-mention rows
during/after disambiguation — either rewrite `narrator_mentions_resolved.parquet`
in place from the `merge_log`, or have `_load_narrated` / `_load_chains` join the
`merge_log` on `mention_id`. Prefer the former so the resolved-mentions artifact is
self-contained for all downstream consumers.

### Acceptance
- Live `neo4j:5` integration test: after `parse → ner → disambiguate → load`, a
  resolved mention yields a real `(:Narrator)-[:NARRATED]->(:Hadith)` edge and a
  `Chain` with non-empty `narrator_ids` / `chain_length > 0`.
- Idempotent reload (no duplicate edges).

### Scope / overlap
This is the cross-source resolve concern, deliberately kept **out** of the da#89
[...]

PR #127, opened 2026-06-12
tech-debt(resolve): wire detect_parallels + bio_promote into run_all, disambiguate-first ordering (#117)

parametrization:
Closes #117

## What

The resolve orchestrator `run_all` (`src/resolve/__init__.py`) ran only `NER -> disambiguate -> dedup`. Two stages that exist but were never in the pipeline path produced nothing in an orchestrated `make resolve`:

- **`detect_parallels`** (deterministic lexical PARALLEL_OF, `parallels.py`) — so a no-model/CI run emitted **zero cross-sect edges** (dedup's FAISS path degrades to empty without the embedding model).
- **`bio_promote`** (bio-direct canonical promoter, `bio_promote.py`) — so bio-only narrators were never promoted into `narrators_canonical.parquet` in an orchestrated run.

## Wiring + ordering

New order: **`NER -> disambiguate -> bio_promote -> (dedup + detect_parallels)`**.

- **`bio_promote` runs AFTER `disambiguate`** (the critical da#99 footgun): `disambiguate` **OVERWRITES** `narrators_canonical.parquet` while `bio_promote` **MERGEs** into it — the reverse order clobbers promoted bio-only narrators. Both stages receive the curated `output_dir`, so they reconcile against one canonical artifact (da#112). `bio_promote` runs regardless of NER/disambiguation status (merge-safe against whatever canonical table already exists, or none).
- **Composition over last-writer-wins** for the shared `staging/parallel_links.parquet`: both detectors write it. `run_all` captures each detector's table and **unions** them, deduped by canonical `(hadith_id_a, hadith_id_b)`, with the **semantic** (dedup) row winning on collision. This guarantees the deterministic cross-sect edges survive even when dedup degrades to empty — directly fixing the issue's "zero cross-sect edges in CI" symptom. (This is the explicit "composition vs last-writer-wins" decision the issue asked for, and the resolution to Oyunbileg's #114 double-write flag.)

## Scope

Orchestration wiring + ordering + composition only. **No changes** to `bio_promote`/`detect_parallels` internals or the `PARALLEL_LINKS`/canonical schemas.

## Files touched (for #109 serial-merge)

- `src/resolve/__init__.py` — **the only shared file with #109.** My edits are confined to: module docstring; `ResolveMetrics` (+`bio_promote_files`/`parallels_files` fields and the `total_files` sum); two new helpers `_read_parallel_links` + `_compose_parallel_links`; and `run_all` (imports, results dict keys, the new bio_promote step, and the dedup+parallels+compose steps). I did **not** touch `disambiguate.run` or `disambiguate.py`. If #109 adds a backfill inside `disambiguate.run` (before the loaders), it composes cleanly with this — my wiring preserves disambiguate running before bio_promote. (Rebased on the current wave tip; clean, no concurrent edit to this file.)
- `tests/test_resolve/test_run_all.py` — new file (no collision).

## Test Plan

`tests/test_resolve/test_run_all.py`:
1. `test_run_all_invokes_bio_promote_and_parallels_disambiguate_first` — `run_all` invokes both `bio_promote` and `detect_parallels`; asserts `NER < disambiguate < bio_promote` call order.
2. `test_run_all_promoted_bio_only_narrator_survives` — real `disambiguate` + `bio_promote`; a promoted bio-only narrator (no mention) survives `run_all` **alongside** a disambiguate-produced narrator (so the overwrite really happened and the merge landed on top) — proves disambiguate-before-bio_promote. Also asserts the composed `parallel_links.parquet` carries a cross-sect edge with **no embedding model**.

Local: `tests/test_resolve/` 117 passed, 3 skipped (Docker-gated). `ruff format --check` + `ruff check` clean; `mypy src/resolve/` clean.

Reviewers: @Ivana Horvat (ML/NLP, primary) + @Jean-Claude Habimana (architect, secondary).

### DIFF
--- a/src/resolve/__init__.py
+++ b/src/resolve/__init__.py
@@ -1,7 +1,17 @@
 """Phase 2: Entity Resolution pipeline.
 
-Dependency-aware orchestrator: NER -> disambiguate -> dedup.
-Dedup runs independently of NER/disambiguation (only needs hadith text).
+Dependency-aware orchestrator:
+``NER -> disambiguate -> bio_promote -> (dedup + detect_parallels)``.
+
+Ordering invariants (da#117):
+- ``disambiguate`` OVERWRITES ``narrators_canonical.parquet`` while
+  ``bio_promote`` MERGEs into it, so ``bio_promote`` MUST run *after*
+  ``disambiguate`` or promoted bio-only narrators get clobbered (da#99).
+- ``dedup`` (semantic embeddings, degrades to empty without the model) and
+  ``detect_parallels`` (deterministic lexical, offline/CI) both write the shared
+  ``staging/parallel_links.parquet``. run_all runs both and *composes* their
+  outputs (union, deduped by canonical hadith pair) so an orchestrated run in a
+  no-model environment still emits cross-sect PARALLEL_OF edges (da#100/da#114).
 """
 
 from __future__ import annotations
@@ -9,8 +19,12 @@ from __future__ import annotations
 import traceback
 from dataclasses import dataclass, field
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 from src.utils.logging import get_logger
+
+if TYPE_CHECKING:
+    import pyarrow as pa
 
 logger = get_logger(__name__)
 
@@ -33,7 +47,9 @@ class ResolveMetrics:
     parallel_cross_sect: int = 0
     ner_files: list[Path] = field(default_factory=list)
     disambiguate_files: list[Path] = field(default_factory=list)
+    bio_promote_files: list[Path] = field(default_factory=list)
     dedup_files: list[Path] = field(default_factory=list)
+    parallels_files: list[Path] = field(default_factory=list)
 
     def summary(self) -> str:
         """Return a human-readable summary string."""
@@ -55,7 +71,13 @@ class ResolveMetrics:
                     f"    cross-sect             : {self.parallel_cross_sect}",
                 ]
             )
-        total_files = len(self.ner_files) + len(self.disambiguate_files) + len(self.dedup_files)
+        total_files = (
+            len(self.ner_files)
+            + len(self.disambiguate_files)
+            + len(self.bio_promote_files)
+            + len(self.dedup_files)
+            + len(self.parallels_files)
+        )
         lines.append(f"  Output files             : {total_files}")
         return "\n".join(lines)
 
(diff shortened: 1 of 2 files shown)


parametrization:
RequestOrReplied: Approved

Secondary review, architect lens — verified via diff + module signatures at the PR head ref + CI status (no local suite run).

**CI** — 5/5 green (Lint, Type Check, Test, Integration Tests, sync-drift gate).

**(a) Composition vs last-writer-wins — correct call.** Both detectors materialize the same shared `staging/parallel_links.parquet` under `PARALLEL_LINKS_SCHEMA`, and `detect_parallels` runs last, so plain last-writer-wins would either drop dedup's semantic edges or — worse — let the no-model empty dedup table clobber the deterministic edges. The union (deterministic inserted first, semantic second into the dict so the production embedding signal wins on a key collision) is the right way to keep deterministic cross-sect edges alive in CI/no-model runs while not discarding semantic ones. Crucially, both detectors canonically order pairs to `hadith_id_a < hadith_id_b` (parallels collapses symmetric dupes at the `id_i < id_j` swap; dedup does the same "canonical ordering to eliminate symmetric duplicates"). That means the `(hadith_id_a, hadith_id_b)` dedup key references the same logical edge across both producers, so the union does NOT double-count a pair found by both detectors. Composed table is rebuilt with the exact 5-column schema and types — non-null constraints are satisfied because every source row came from a schema-validated table.

**(b) run_all ordering — robust.** disambiguate -> bio_promote -> dedup -> parallels -> compose. bio_promote correctly sits after the overwriting disambiguate (da#99 footgun) and runs regardless of NER/disambig status (merge-safe). `output_dir` (curated dir) is threaded consistently to disambiguate and bio_promote, so both reconcile against one canonical artifact the loader reads. Signatures all line up: `promote_bios_to_canonical(staging, output)` returns Path|None (handled by `[promoted] if promoted is not None else []`); `parallels.run`/`dedup.run` both return list[Path]. The capture-before-overwrite pattern (`_read_parallel_links` after each detector) is the right mechanic given the shared artifact.

**(c) Stays pure orchestration.** No edits to bio_promote/parallels/dedup internals and no schema change — `_compose_parallel_links` imports and reuses `PARALLEL_LINKS_SCHEMA` rather than redefining it. The composition is genuinely orchestration glue. Minor, non-blocking: it fully materializes both link sets in Python (`to_pylist`) and rewrites; fine at parallel_links cardinality, just noting it scales with link count, not a concern at current corpus size.

**(d) #109 serial-merge surface — low risk, coordinate at merge.** This PR's `__init__.py` changes are localized to the run_all step body (new dedup/parallels/bio_promote steps), the two new module-level helpers, and the metrics dataclass — none of it touches disambiguation logic. #109 is disambiguate.py-centric. Whichever lands first, the other should rebase cleanly; the only shared file is `__init__.py` and the regions don't overlap (run_all orchestration block vs. whatever #109 needs in the disambiguate call). Recommend serial merge with a quick rebase verify on the second PR, not a blocker here.

Approving. Clean, well-documented orchestration wiring; the composition correctness (canonical pair ordering on both sides + semantic-wins) is the load-bearing detail and it holds.

parametrization:
Requestor: Ivana Horvat
RequestOrReplied: Approved

Reviewed the diff at PR head + the real signatures/contracts of every wired stage. CI green (5/5: Lint, Type Check, Test, Integration, sync-drift). LGTM — the wiring is correct, the ordering footgun is honored, and the composition is deterministic.

What I verified against source (not just the PR body):

1. Ordering footgun (da#99) honored. `disambiguate.run` writes `narrators_canonical.parquet` to `output_dir` (overwrite, disambiguate.py:915); `bio_promote.promote_bios_to_canonical(staging_dir, output_dir)` reads that same `output_dir` artifact and MERGEs (bio_promote.py:92-94, returns `Path | None`). run_all runs bio_promote as Step 3, strictly after disambiguate (Step 2), both handed the curated `output_dir`. Reversed order would clobber bio-only narrators; this order does not. Signature matches the wiring exactly (`[promoted] if promoted is not None else []`).

2. Composition is correct and deterministic — and the dedup key actually aligns. The concern with a literal `(hadith_id_a, hadith_id_b)` dedup key is symmetric pairs ((h1,h2) vs (h2,h1)) surviving as duplicate edges. Confirmed both detectors emit canonically-ordered pairs (smaller id first): parallels (parallels.py:140) and dedup (dedup.py:232-236, `if hid_a >= hid_b: swap`). So the keys collide as intended; union + dedup yields one edge per logical pair. Insertion order (deterministic first, semantic second) makes the semantic row win on collision, as specified. Schema rebuild uses the exact `PARALLEL_LINKS_SCHEMA` (5 non-nullable fields, schemas.py:69-77) — values come from already-conforming tables, so no null-violation risk.

3. #114 double-write fixed for the no-model case. dedup writes `parallel_links.parquet` into staging even when degraded (empty table, dedup.py:289-292), so `semantic_links` is a real 0-row table, not None; parallels then overwrites; compose unions them so the deterministic cross-sect edges survive. The CI-green `test_run_all_promoted_bio_only_narrator_survives` exercises exactly this path (real dedup, no embedding model) and asserts a `cross_sect` edge persists — strong evidence the fallback works, not just a claim.

4. run_all contract preserved. Same `(raw_dir, staging_dir, output_dir) -> dict[str, list[Path]]`; changes are purely additive (`bio_promote`/`parallels` keys, two helpers, ResolveMetrics fields). Per-stage `try/except` matches the existing dedup pattern.

5. Tests assert both invariants. `test_run_all_invokes_..._disambiguate_first` asserts `ner < disambiguate < bio_promote` call order via monkeypatched recorders; the second test proves the merge-after-overwrite (bio-only id present alongside a disambiguate-produced narrator, `len(ids) >= 2`) plus the composed cross-sect edge.

Non-blocking nit (no change required): `_compose_parallel_links` rebuilds the table row-by-row through Python lists (`to_pylist`) — fine at current scale, but if `parallel_links` grows large this materializes every row in memory. A `pyarrow.concat_tables` + group-dedup would scale better. Out of scope for this wiring PR; flag only if a future profiling pass shows it hot.

#109 serial-merge note: confirmed — both PRs touch `src/resolve/__init__.py`. This PR confines its edits to the module docstring, ResolveMetrics, the two new helpers, and `run_all`; it does NOT touch `disambiguate.run`/`disambiguate.py`. If #109 adds a backfill inside `disambiguate.run` it composes cleanly, but if #109 also edits the `run_all` body there's textual-conflict surface — recommend serial merge (whichever lands second rebases on the first and re-confirms NER->disambiguate->bio_promote order survives).

PR #128, opened 2026-06-12
feat(resolve): backfill canonical_narrator_id into mentions → NARRATED edges fire (#109)

parametrization:
Closes #109.

## Problem
The narrator-resolution pipeline produced canonical **Narrator nodes** but the narrator→hadith **wiring never materialized**: `src/resolve/ner.py` writes `narrator_mentions_resolved.parquet` with `canonical_narrator_id = None`, and `disambiguate.run` — which knows each mention's canonical id and writes `merge_log.parquet` — never wrote that id back onto the mention rows. The graph loaders (`load_edges.py::_load_narrated`, `load_nodes.py::_load_chains`, in noorinalabs-isnad-graph) read `canonical_narrator_id` straight off the mentions, so they saw only `None` → **0 NARRATED edges, empty Chains** from real data.

This was masked because the graph-loader unit fixtures hand-inject `canonical_narrator_id`, and no integration test ran ner→disambiguate→(load contract) end-to-end.

## Fix (backfill mechanism)
`disambiguate.run` now captures `mention_id → (canonical_id, confidence)` for **every** mention that received a canonical narrator — both bio-matched mentions **and** the da#99 self-canonicalized (fallback) mentions — and rewrites `narrator_mentions_resolved.parquet` in place via `_backfill_mention_canonical_ids`:
- Streams row-group by row-group (mirrors `_iter_mention_batches`) → memory bounded for the 3.3M-row scale.
- Idempotent: re-running yields the same file (no duplicate rows / stable ids).
- Self-contained artifact: every downstream consumer now reads the resolved id directly off the mention.

Covering the fallback path (not just `merge_log`) matters: da#99 made non-bio mentions real Narrator nodes too; backfilling only bio-matched mentions would leave those positions unwired and chains incomplete.

## New non-mocked integration test
`tests/test_resolve/test_canonical_backfill_integration.py` runs the **real** `ner → disambiguate` pipeline on fixtures and:
- Asserts NER-alone yields `canonical_narrator_id = None` → **0** derived NARRATED edges (the bug state).
- After `disambiguate.run`, asserts every resolved mention carries a `nar:<uuid5>` id matching the canonical narrator nodes, and that a real NARRATED edge (count > 0) + a non-empty Chain (`chain_length = 2`) now derive from the backfilled mentions.
- Does **not** hand-inject `canonical_narrator_id` (the anti-pattern that masked #109).
- The graph loaders live in noorinalabs-isnad-graph, so the test replicates their documented contract (read `canonical_narrator_id` off the first-narrator-per-hadith / per-chain) rather than standing up Neo4j.

**Proof it fails-without / passes-with:** with the backfill call disabled, `test_disambiguate_backfills_canonical_id_into_mentions` FAILS (canonical ids stay None → 0 edges); restored, it PASSES.

## Edge-count proof
Fixture: one hadith, two chain positions, two bio candidates. After disambiguate → 2/2 mentions backfilled (`backfill_mentions_complete backfilled=2`), 1 NARRATED edge (first narrator → hadith), chain narrator_ids length 2. `disambiguate_summary` shows `total_resolved=2`, `bio_coverage_pct=100`.

## Verification
- `ruff format --check` + `ruff check` clean
- `mypy` (strict) clean on both files
- `pytest tests/test_resolve tests/test_pipeline` → **138 passed, 3 skipped**

## Files touched
- `src/resolve/disambiguate.py` (backfill helper + capture map + call in `run`)
- `tests/test_resolve/test_canonical_backfill_integration.py` (new)

**Note:** `src/resolve/__init__.py` is **untouched** → no serial-merge conflict with #117 (the backfill lives inside `disambiguate.run`, which `run_all` already calls).

## Test Plan (runtime, post-merge)
Live `neo4j:5` end-to-end (`parse → ner → disambiguate → load`) asserting a real `(:Narrator)-[:NARRATED]->(:Hadith)` edge + Chain on the staging box — deferred because the graph loaders live in noorinalabs-isnad-graph (cannot run a real `load` from this repo).

Reviewers: @Ivana Horvat (ML/NLP, primary — owns the da#99 fallback path this wires) + @Jean-Claude Habimana (architect, secondary).

### DIFF
--- a/src/resolve/disambiguate.py
+++ b/src/resolve/disambiguate.py
@@ -24,7 +24,11 @@ from rapidfuzz.distance import Levenshtein
 
 from src.parse.base import safe_str, write_parquet
 from src.parse.identity import make_canonical_id
-from src.resolve.schemas import AMBIGUOUS_NARRATORS_SCHEMA, NARRATORS_CANONICAL_SCHEMA
+from src.resolve.schemas import (
+    AMBIGUOUS_NARRATORS_SCHEMA,
+    NARRATOR_MENTIONS_RESOLVED_SCHEMA,
+    NARRATORS_CANONICAL_SCHEMA,
+)
 from src.utils.arabic import normalize_arabic
 from src.utils.logging import get_logger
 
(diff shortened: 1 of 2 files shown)


parametrization:
RequestOrReplied: Approved

Secondary review through the resolve-flow lens (diff + CI only; no local suite run).

**Composition with #117 (run_all wiring) — conflict-free.** The backfill lives entirely inside `disambiguate.run`, and `src/resolve/__init__.py` is untouched by this PR. My #117 orchestrator already calls `disambiguate.run` at `__init__.py:166`, so the backfill rides along automatically — no `resolve/__init__.py` overlap, no serial-merge conflict, merge order does not matter.

**Read/write path is a closed loop.** `run()` consumes mentions via `_iter_mention_batches(output_dir)` → `output_dir/narrator_mentions_resolved.parquet`; `_backfill_mention_canonical_ids` rewrites that same artifact in place. That is the file NER writes the `None` placeholders into and the file the isnad-graph edge/chain loaders read `canonical_narrator_id` straight off of. Correct target.

**Both resolved branches captured (no under-wire).** `resolved_map` is populated on the bio-matched path (`float(best.score)`) and on the da#99 self-canonicalized fallback (`None`). A merge_log-only backfill would have wired only bio-backed positions; this wires the full chain.

**Streaming rewrite is memory-safe + idempotent + schema-preserving.** Row-group-by-row-group, writes `.backfill.tmp` then atomic `replace()`, `writer.close()` in `finally`. Re-running re-derives `resolved_map` from bio candidates (keyed on mention names, not on the canonical column it just rewrote), so the result is deterministic; the idempotency test asserts no row duplication or id drift. Unresolved rows retain their NER placeholder and prior confidence.

**Typing/schema sound.** `confidence` is `float32` nullable; backfill emits `pa.array(new_conf, type=pa.float32())` (score→float32 fine, `None` preserved). `mention_id` is `str` on both the capture side (`str(mention.get("mention_id", ""))`) and the lookup side (`resolved.get(str(mid))`) — keys align.

**Confidence semantics don't break downstream.** NARRATED edges and Chains key off the non-null `canonical_narrator_id`, not off `confidence`; the `None` confidence on fallback mentions is a quality annotation only and does not affect edge construction.

CI: all five checks green (Integration 50s, Lint, Pre-commit⇄CI sync-drift gate, Test, Type Check).

No blockers. Approving.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved

I am the reviewer who originally proved this was a gap (disambiguate never rewrote `narrator_mentions_resolved`, so `canonical_narrator_id` stayed None and real data produced zero NARRATED edges, masked only by hand-injected loader fixtures). I verified the fix against the diff and CI head — it genuinely closes that gap.

What I checked

1. Backfill covers every resolved mention, not just bio-matched ones. `resolved_map` is populated in BOTH branches of `run()`: the bio-matched path writes `(canonical_id, float(best.score))`, and the da#99 self-canonicalized/fallback path writes `(fallback_id, None)`. A merge_log-only backfill would have left fallback narrators unwired and broken mid-chain positions; keying off `resolved_map` instead avoids that. Confirmed the full chain wires, not just bio-backed positions.

2. Join correctness / path consistency. `run()` reads its mentions from `output_dir/narrator_mentions_resolved.parquet` via `_iter_mention_batches`, and `_backfill_mention_canonical_ids` rewrites that same file. The map key is `str(mention_id)` on the producer side and `resolved.get(str(mid))` against the on-disk `mention_id` column on the consumer side — both `str()`-normalized, so the join is sound. Unresolved mentions keep the NER placeholder untouched.

3. Streaming + idempotent at scale. Rewrites row-group by row-group through a `.backfill.tmp` then `replace()`s in place (mirrors `_iter_mention_batches`), so peak memory stays bounded for the 3.3M-row file. Deterministic `resolved_map` makes a re-run byte-stable; `test_backfill_is_idempotent_on_reload` asserts row count, ids, and derived edges are unchanged across two `disambiguate.run` calls.

4. The new integration test is non-mocked and guards the original anti-pattern. `test_disambiguate_backfills_canonical_id_into_mentions` runs the real `ner.run` then `disambiguate.run` on fixtures and reads `canonical_narrator_id` back off the on-disk artifact — it does NOT hand-inject the canonical id (the exact masking pattern that hid the bug). It asserts NER-alone yields all-None ids and zero derived edges, then post-disambiguate yields two `nar:<uuid5>` ids and a NARRATED edge count of 1. Structurally this fails without the backfill: removing it leaves ids None, so `assert len(resolved) == 2` and the edge-count assertion both fail — the author's "fails without, passes with" claim holds.

CI: all five checks green at head 7d7cff9 (Integration Tests, Lint, sync-drift gate, Test, Type Check).

Two non-blocking observations (follow-up, not gating)
- The integration test exercises only the bio-matched path (both fixture names exact-match a bio candidate). The fallback/self-canonicalized branch is correct in code but not covered by a test case (a named mention with no bio candidate). A fallback-path assertion would harden the "whole chain wires" guarantee.
- The test replicates the isnad-graph loader's NARRATED/Chain contract locally rather than running the real loader (the loader lives in another repo, so a live Neo4j load cannot run here). This is documented in the test; the load-bearing assertion is the non-null `canonical_narrator_id` on the real artifact, which the real loader reads directly, so the proof is sound.

Verdict: Approved. The gap I found is truly closed — disambiguation now writes resolved canonical ids back onto the mention rows for both bio-matched and fallback narrators, so real scraped data will materialize NARRATED edges instead of zero.

PR #129, opened 2026-06-12
feat(graph): Narrator nodes carry sect_affiliation + source_corpus (#103)

parametrization:
## Summary

`Closes #103`. Narrator (and edge) graph nodes carried **no** `sect_affiliation`/`source_corpus` for any source — only Hadith/Collection nodes did. This wires sect/corpus provenance through the full chain `resolve → canonical-schema → loader` so a Narrator node in Neo4j is sect- and corpus-tagged.

## Changes

**Schema** (`src/resolve/schemas.py`)
- `NARRATORS_CANONICAL_SCHEMA` gains `source_corpus` (scalar primary), `source_corpora` (multi-source list), `sect_affiliation` — **all nullable** so the other `narrators_canonical` producers (#109 / #117) are not forced to populate them.

**Derivation** (new `src/resolve/sect_affiliation.py`)
- `derive_sect_affiliation(corpora)` → narrator-level `SectAffiliation` (`sunni`/`shia`/`neutral`/`unknown`). Aggregation: only-Sunni→`sunni`, only-Shia→`shia`, **both→`neutral`** (a muhaddithat transmitting across Sunni *and* Shia chains), nothing-mapped→`unknown`. The corpus→tradition map mirrors the per-corpus `SECT` the parsers stamp; `fawaz` (per-collection sect) and `itqan` (mixed rijal DB) are intentionally unmapped → `unknown` rather than a guess.
- `normalize_corpus()` reconciles the **bio-provenance inconsistency** the issue flags (sanadset bios carry `source="kaggle_narrators"`, not a `SourceCorpus` value) so the scalar `Narrator.source_corpus` stays inside the enum domain; genuinely-unknown provenance → dropped.

**Producers** (`disambiguate.py`, `bio_promote.py`)
- `disambiguate` accumulates each mention's `source_corpus`; `bio_promote` each bio's normalized `source`. `bio_promote` is **merge-safe** — it unions onto corpora a prior `disambiguate` run already recorded (`_load_existing_canonical` setdefaults `source_corpora`), preserving the two-producer ordering contract.

**Loader** (`src/graph/load_nodes.py`)
- `_NARRATOR_MERGE` SETs `n.source_corpus` / `n.source_corpora` / `n.sect_affiliation` (explicit SET, no blanket `SET n += row`). Legacy/unpopulated rows default to `""` / `[]` / `"unknown"`.

**Provenance gate** (`src/parse/base.py`)
- Adding `source_corpus` to the canonical schema newly subjected `narrators_canonical.parquet` to `write_parquet`'s `provenance_violations` gate. The gate's null-check now **honors schema nullability**: a null is a violation only for a **non-nullable** provenance column (parser staging, where the tag is mandatory); the resolve-derived **nullable** canonical `source_corpus` permits null ("no attributable corpus"). Empty-string + wrong-domain checks still apply to present values for both — so a bad corpus on a canonical narrator is still caught.

## How `sect_affiliation` is derived
From the set of corpora a canonical narrator was actually observed in (mentions + promoted bios), each normalized to a `SourceCorpus` value, each mapped to its collection-level tradition, then aggregated (both traditions ⇒ `neutral`). It is a **resolution-time** determination, distinct from the collection-level binary `Sect` on Hadith/Collection nodes.

## Test Plan
- `tests/test_resolve/test_sect_affiliation.py` — derivation table + `normalize_corpus` (alias/unknown/empty).
- `tests/test_resolve/test_schemas.py` — canonical schema carries the three fields.
- `tests/test_resolve/test_bio_promote.py` — promoted bios carry source_corpus/_corpora + derived sect.
- `tests/test_graph/test_load_nodes.py` — loader SETs all three (populated + legacy-default).
- `tests/test_parse/test_provenance_conformance.py` — null allowed in nullable column, still flagged in non-nullable; domain/empty still enforced.
- `tests/integration/test_graph_loading.py::test_narrator_properties` — **live `neo4j:5` acceptance**: a loaded Narrator node carries `sect_affiliation="neutral"`, `source_corpus="sunnah"`, `source_corpora=["sunnah","thaqalayn"]`. *(This is the live-Neo4j node-tagging proof #83 originally named, relocated here per the issue.)*
- Full suite: 709 passed, 3 skipped. `ruff format --check` + `ruff check` clean; `mypy src/` clean.

## Coordination (serial-merge)
- Files touched: `src/resolve/schemas.py`, `src/resolve/disambiguate.py`, `src/resolve/bio_promote.py`, `src/graph/load_nodes.py`, `src/parse/base.py` (+ the new `sect_affiliation.py`). The new schema field is **nullable** → #109 (disambiguate/load_edges) and #117 (resolve/`__init__`) producers need not populate it; if they write `narrators_canonical`, a null `source_corpus` is now legal under the refreshed provenance gate.
- da#90 (muhaddithat e2e) can tag at the parser side and rely on this schema as the canonical home for narrator sect/corpus.

Reviewers: @Kavitha Sundaramurthy (primary), @Oyunbileg Batbayar (QA).

### DIFF
--- a/src/graph/load_nodes.py
+++ b/src/graph/load_nodes.py
@@ -100,7 +100,10 @@ SET n.name_ar           = row.name_ar,
     n.aliases           = row.aliases,
     n.external_id       = row.external_id,
     n.source_ids        = row.source_ids,
-    n.mention_count     = row.mention_count
+    n.mention_count     = row.mention_count,
+    n.source_corpus     = row.source_corpus,
+    n.source_corpora    = row.source_corpora,
+    n.sect_affiliation  = row.sect_affiliation
 """
 
 
@@ -159,6 +162,13 @@ def _load_narrators(
                 # ``MATCH (n:Narrator) WHERE any(s IN n.source_ids WHERE s STARTS WITH 'itqan:')``.
                 "source_ids": _val(row, "source_ids", []),
                 "mention_count": _val(row, "mention_count"),
+                # Sect/corpus provenance (da#103). ``source_corpus`` defaults to ""
+                # so the property is always present even for legacy canonical files
+                # written before these columns existed; ``sect_affiliation`` defaults
+                # to ``unknown`` for the same reason.
+                "source_corpus": _val(row, "source_corpus", ""),
+                "source_corpora": _val(row, "source_corpora", []),
+                "sect_affiliation": _val(row, "sect_affiliation", "unknown"),
             }
         )
 
--- a/src/parse/base.py
+++ b/src/parse/base.py
@@ -55,7 +55,12 @@ def provenance_violations(table: pa.Table) -> list[str]:
         if column not in columns:
             continue
         values = table.column(column)
-        if values.null_count:
+        # A *nullable* provenance column is resolve-derived (e.g. the canonical
+        # narrator master's source_corpus, da#103) where "no attributable corpus"
+        # is legitimately null; only a non-nullable parser-staging column must be
+        # fully populated. Empty-string and out-of-domain checks below still apply
+        # to whatever values ARE present, for both kinds.
+        if values.null_count and not table.schema.field(column).nullable:
             problems.append(f"{column}: {values.null_count} null value(s)")
         non_null = values.drop_null()
         if not len(non_null):
--- a/src/resolve/bio_promote.py
+++ b/src/resolve/bio_promote.py
@@ -28,6 +28,11 @@ import pyarrow.parquet as pq
 from src.parse.base import safe_str, write_parquet
 from src.parse.identity import make_canonical_id
 from src.resolve.schemas import NARRATORS_CANONICAL_SCHEMA
+from src.resolve.sect_affiliation import (
+    derive_sect_affiliation,
+    normalize_corpus,
+    primary_corpus,
+)
 from src.utils.arabic import normalize_arabic
 from src.utils.logging import get_logger
 
@@ -58,6 +63,9 @@ def _load_existing_canonical(path: Path) -> dict[str, dict[str, Any]]:
         if cid:
             row.setdefault("source_ids", [])
             row.setdefault("aliases", [])
+            # Preserve corpora a prior producer (disambiguate) already recorded so
+            # the bio-direct merge unions onto them rather than dropping them (da#103).
+            row.setdefault("source_corpora", [])
             rows[cid] = row
     return rows
 
@@ -104,6 +112,9 @@ def promote_bios_to_canonical(
             if sources is not None and source not in sources:
                 skipped_source += 1
                 continue
+            # Normalize the free-text bio provenance to a SourceCorpus value for
+            # sect/corpus tagging (da#103); unknown provenance → None (dropped).
+            corpus = normalize_corpus(source)
 
             name_ar = safe_str(row.get("name_ar"))
             norm = safe_str(row.get("name_ar_normalized")) or (
@@ -132,12 +143,18 @@ def promote_bios_to_canonical(
                     "external_id": safe_str(row.get("external_id")),
                     # mention_count stays 0: a bio is provenance, not a chain hit.
                     "mention_count": 0,
+                    # Corpus provenance for sect derivation (da#103). Finalized into
+                    # source_corpus + sect_affiliation after the merge loop.
+                    "source_corpora": [corpus] if corpus else [],
                 }
             else:
                 rec = canonical_map[cid]
                 src_ids = rec.setdefault("source_ids", [])
                 if bio_id and bio_id not in src_ids:
                     src_ids.append(bio_id)
+                corpora = rec.setdefault("source_corpora", [])
+                if corpus and corpus not in corpora:
+                    corpora.append(corpus)
                 # Back-fill only fields the existing record is missing. Keep the
                 # raw Parquet value (already a clean str / int) — do NOT coerce,
                 # or an int column like death_year_ah gets stringified.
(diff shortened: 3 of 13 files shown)


parametrization:
RequestOrReplied: Approved

Architect-lens review of da#103 (narrator nodes carry sect/corpus provenance). Verified via diff, the provenance gate source at PR head, and CI only — no local suite run.

## The sensitive change — shared provenance gate in src/parse/base.py

Safe and correctly scoped. No standards/Aino sign-off needed. The relaxation reads `values.null_count and not table.schema.field(column).nullable`, so it keys off the column's actual declared nullability rather than a blanket allow. The non-nullable parser-staging columns (sect, source_corpus on the hadith/edge schemas) still reject null because those schemas declare nullable=False — enforcement there is unchanged. Only the new narrators_canonical columns, declared nullable in schemas.py, may now legitimately carry null ("no attributable corpus"). Critically, the empty-string check and the out-of-domain check below the branch run unconditionally on the present (`non_null`) values for both kinds, so a typo like the "sunna" value or any out-of-enum string is still flagged regardless of nullability. Both sides of the branch carry a regression test — `test_null_provenance_flagged` proves a non-nullable column still rejects null, and `test_null_allowed_in_nullable_column` proves a nullable column permits null while still catching the wrong-domain value. No previously-enforced adapter is silently let through, because the only schema turned nullable in this PR is narrators_canonical's three new fields.

## Other architect checks

1. Nullable was the right call. Sibling producers (#109/#117) need not populate, so they leave null rather than being forced into a migration or a fake empty-string; load_nodes defaults at load time (empty scalar, empty list, "unknown") so the Neo4j property is always present even for legacy canonical files. Clean split between parquet-null and load-default.

2. derive_sect_affiliation logic is sound — only-Sunni maps sunni, only-Shia maps shia, both-traditions maps neutral, nothing-mapped maps unknown. fawaz (per-collection sect) and itqan (mixed rijal DB) are intentionally unmapped and contribute unknown rather than a guess, which is documented in the module docstring. The mixed case (itqan plus thaqalayn yields shia, itqan dropping out) is a defensible consequence and is pinned by a parametrized test.

3. Two-producer merge honors the da#99 ordering without clobbering. disambiguate rebuilds source_corpora fresh from the accumulated set in _build_canonical_table; bio_promote preserves any corpora a prior producer recorded (_load_existing_canonical setdefault, then unions in the merge loop, then re-derives scalar plus affiliation in the finalize pass). When disambiguate runs first, bio_promote reads its corpora and unions onto them — no overwrite. This PR does not touch run_all ordering (still a separate open item per the existing note), but it does not regress it; the merge producer correctly unions rather than overwriting.

4. _NARRATOR_MERGE explicitly SETs all three properties on the node (no blanket SET n += row), confirmed in load_nodes.py and asserted by test_narrator_sets_sect_and_corpus.

5. Rebase over #109 is clean — single commit, base deployments/phase-4/wave-4, mergeable clean, rebaseable.

## CI

All five checks green, including Integration Tests (live neo4j, 52s) whose test_narrator_properties asserts the loaded node carries source_corpus, source_corpora, and sect_affiliation. Type Check, Lint, Test, and the pre-commit sync-drift gate also pass.

Well-tested on both branches of the gate, well-documented domain rationale, tightly scoped. Approving.

TechDebt: none. run_all producer ordering (disambiguate before bio_promote) remains a pre-existing open item tracked separately and is not introduced or worsened here.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved

QA / data-quality review (diff + CI only, no local suite run).

**Live node-tagging proof is genuine.** `test_narrator_properties` uses the real `neo4j:5-community` testcontainer fixture (`tests/integration/conftest.py`, not a mock), runs the production `load_all_nodes`, then issues an actual `MATCH (n:Narrator {id:'nar:001'}) RETURN properties(n)` read-back and asserts all three tags on the returned node (`source_corpus=="sunnah"`, sorted `source_corpora==["sunnah","thaqalayn"]`, `sect_affiliation=="neutral"`). The staging row genuinely carries those values. CI "Integration Tests" is green at 52s, which confirms the container leg executed rather than skipping the infra path — so this is a real end-to-end node-tagging acceptance, not a synthetic one.

**Derivation matrix is complete.** `test_sect_affiliation.py` parametrizes every aggregation branch — only-Sunni (`["sunnah"]`, `["sunnah","open_hadith"]`), only-Shia (`["thaqalayn"]`), both-traditions to neutral (`["sunnah","thaqalayn"]`, `["muhaddithat","thaqalayn"]`), nothing-mapped to unknown (`[]`, `["itqan"]`, `["fawaz"]`), plus the multi-corpus list cases and the unmapped-corpus-drops case (`["itqan","thaqalayn"]` to shia). The bio-provenance alias (`kaggle_narrators` to sanadset to sunni) and empty-string filtering are both covered.

**Scalar stays in the SourceCorpus domain.** `normalize_corpus` only returns a value that is in `_VALID_CORPORA` (or the alias-mapped `SANADSET.value`), else `None`; `primary_corpus` selects from already-normalized corpora; `bio_promote` re-normalizes defensively at finalize. So `Narrator.source_corpus` never leaks an out-of-enum string, keeping the #113 guard and the `write_parquet` provenance gate satisfied. The gate change correctly gates the null-check on `field.nullable` while keeping empty-string and out-of-domain checks active for present values — verified by the new `test_null_allowed_in_nullable_column` (clean passes, `"sunna"` flagged) and the tightened `test_null_provenance_flagged`.

**No silent all-null narrators.** `bio_promote` populates from each bio source and is asserted end-to-end (`test_bio_promote_tags_sect_and_corpus` to muhaddithat to sunni); the loader populates plus legacy-defaults are asserted; the live node carries real values.

One non-blocking coverage note for follow-up (not a merge blocker): the `disambiguate` producer path (the `corpus=` plumb into `_upsert_canonical` plus `_build_canonical_table`) has no test asserting it emits a populated, non-null corpus from a valid mention source — the existing disambiguate tests pass `source_corpus="test"`, which normalizes to None. The path reuses the same unit-tested helpers and is low risk, but a single assertion that a real-corpus mention yields a populated scalar would close the last gap. Worth a small follow-up rather than holding this PR.

CI: all five checks green (Integration Tests, Lint, sync-drift gate, Test, Type Check).

Issue #96, opened 2026-06-12
N3: halimbahae/Hadith adapter (sunni, 9 books, diacritics)


parametrization:
**Tier:** T2 (new source) · **Suggested owner:** Nikolaos Papadopoulos
Part of epic #81.

## Goal
**halimbahae/Hadith** (GitHub) — 9 books incl. Bukhari/Muslim with Arabic diacritics. New `src/acquire/halimbahae.py` + parser.

## Acceptance
- [ ] New `src/acquire/<src>.py` + `src/parse/<src>.py` producing canonical staging Parquet with `sect` + `source_corpus`.
- [ ] `SourceCorpus` enum value added (in `src/models/enums.py`).
- [ ] Parse + acquire-side tests.
- [ ] End-to-end load into a **local `neo4j:5` container** asserting real nodes/edges (local Docker confirmed working).
- [ ] Reachability + licensing note.

## Dependencies
T0-A (source_id scheme), T0-B (conformance gate).

parametrization:
Merged via PR #131 (halimbahae Sunni adapter, 9 books).

Issue #95, opened 2026-06-12
N2: Bihar al-Anwar (via Thaqalayn) — deepen Shia coverage


parametrization:
**Tier:** T2 (new source) · **Suggested owner:** Kwesi Boateng
Part of epic #81.

## Goal
**Bihar al-Anwar** (via ThaqalaynAPI) — largest Shia collection (~100k items). Deepen Shia coverage beyond the Four Books. Extend the `thaqalayn` adapter or add `src/acquire/bihar.py`.

## Dependency
Best after **L2** (thaqalayn light-up) proves the Shia path end-to-end.

## Acceptance
- [ ] New `src/acquire/<src>.py` + `src/parse/<src>.py` producing canonical staging Parquet with `sect` + `source_corpus`.
- [ ] `SourceCorpus` enum value added (in `src/models/enums.py`).
- [ ] Parse + acquire-side tests.
- [ ] End-to-end load into a **local `neo4j:5` container** asserting real nodes/edges (local Docker confirmed working).
- [ ] Reachability + licensing note.

## Dependencies
T0-A (source_id scheme), T0-B (conformance gate).

parametrization:
Merged via PR #134 (real Bihar al-Anwar adapter vs hubeali.com; 394 live bilingual hadiths; NEW SourceCorpus.BIHAR, Itqan licensing posture). Full multi-volume crawl + kitab source_id fix tracked in da#136.

PR #35, opened 2026-05-05
P3W4 → main (noorinalabs-data-acquisition)

parametrization:
## P3W4 → main (noorinalabs-data-acquisition)

Wave-bootstrap merge of `deployments/phase-3/wave-4` → `main` for noorinalabs-data-acquisition.

## W4 PRs included

| PR | Title | Closes |
|---|---|---|
| #34 | tech-debt(docs): CLAUDE.md branching-strategy backslash → slash | #33 |

## Single-Reviewer Exception (Wave-Bootstrap Only)

Per charter § Single-Reviewer Exception, this PR is labeled `wave-bootstrap` and qualifies for single-reviewer merge.

PR #131, opened 2026-06-12
feat: halimbahae Sunni adapter (9 books) (#96)

parametrization:
Closes #96 · epic #81 (multi-source ingestion) · P4W4 · Tier T2 (new source)

## What
Net-new ingest adapter for **halimbahae/Hadith** (github halimbahae/Hadith) — a Sunni corpus of **9 books** (the Six Books + Muwatta Malik, Musnad Ahmad, Sunan al-Darimi) shipped with full Arabic **diacritics (tashkeel)**. Builds acquire + parse from scratch and registers it.

## New corpus + registry row (serial-merge surface)
Shared files touched with #95/#97 — exact additions:
- **`src/models/enums.py`** — `SourceCorpus.HALIMBAHAE = "halimbahae"` (appended after `ITQAN`). `SOURCE_CORPORA` in `parse.identity` derives from the enum, so the new value is automatically valid for `generate_source_id` and the `check_source_id_corpus` guard.
- **`src/adapters.py`** — one appended `SourceAdapter` row: `slug="halimbahae"`, `corpus=HALIMBAHAE`, `sect=Sunni`, acquire/parse module `halimbahae`, `reachable=True`, ODbL 1.0 + DbCL 1.0 license note.
- **`tests/test_adapters.py`** — `"halimbahae"` appended to `EXPECTED_SLUGS` (run-order pin).

Distinct value/slug from the sibling adapters; rebase between merges and the da#81 coverage invariant (`covered_corpora() == set(SourceCorpus)`) stays green.

## Diacritic handling
`full_text_ar` preserves the diacritised matn **verbatim**; only the stray bidi control marks (U+200E/U+200F the corpus scatters between words) and redundant whitespace are removed. This is consistent with `src/utils/arabic.py`, where `normalize_arabic`/`strip_diacritics` strip tashkeel **only** for the canonical name-matching path — never for stored hadith text. Parser test asserts tashkeel survives while bidi marks do not.

## Files
- `src/acquire/halimbahae.py` — shallow-clone, assert >=9 `mushakkala` CSVs, manifest + `raw.landed` emit, idempotent.
- `src/parse/halimbahae.py` — headerless `"num","text"` CSVs -> `HADITH_SCHEMA` + `COLLECTION_SCHEMA` (one Collection per book), `source_corpus`/`sect` tagged, corpus-namespaced `source_id`.
- Tests: parser, acquire (mocked clone), live `neo4j:5` light-up.

## Test Plan
- `ruff@0.15.11 format --check` + `ruff check` — clean.
- `mypy` (strict) on new modules — clean.
- `pytest tests/test_parse tests/test_acquire tests/test_adapters.py tests/test_models` — **347 passed, 1 skipped**.
- `pytest tests/integration/test_halimbahae_lightup.py` against **live neo4j:5** (local Docker) — **2 passed**: Hadith+Collection nodes land sunni/halimbahae-tagged; APPEARS_IN edges resolve each hadith to its own book (6 edges, no danglers).

## Reviewers
@Tarek Mansour (devops, primary) · @Dilara Erdogan (pipeline manager, secondary)

## Caveats
- halimbahae mirrors the same upstream data shape as Open-Hadith-Data (`*_mushakkala_mufassala.utf8.csv`); per the issue it is a **distinct `halimbahae` corpus** (own namespace -> own nodes), not folded into `open_hadith`.
- The batch Hadith loader does not copy matn text onto the node (`full_text_ar` not in `_HADITH_MERGE`, a pre-existing #35 reconciliation that `open_hadith` shares) — so the diacritic-preservation contract is asserted at the **parser** level, not on the graph node.


parametrization:
Requestor: Dilara Erdogan
RequestOrReplied: Approved

Secondary review (pipeline-manager lens, diff + CI only).

## Registry plug-in — clean
The new `SourceAdapter` row resolves lazily through `SOURCE_REGISTRY` like every sibling. Both `src/acquire/halimbahae.py` and `src/parse/halimbahae.py` expose `run`, matching the adapter defaults (`acquire_fn`/`parse_fn` both default to `run`), so `import_module("src.acquire.halimbahae")` / `import_module("src.parse.halimbahae")` resolve with no override needed. Adding `SourceCorpus.HALIMBAHAE` plus an adapter carrying that corpus keeps `covered_corpora()` whole, and `EXPECTED_SLUGS` is updated so the run-order pin stays green. `run_all` picks it up with zero extra wiring.

## Distinct corpus — correct modeling call
Keeping halimbahae its own corpus rather than folding it into `open_hadith` is right. The two share an upstream CSV shape but are different collections with different provenance; a shared namespace would collide `source_id`s and conflate where a hadith actually came from. Own corpus to own namespace to own nodes preserves provenance integrity, and issue #96 specifies exactly that.

## Live light-up — genuine
The integration test parses real committed diacritised Arabic slices (Muwatta Malik + Sunan al-Darimi, 3 each) through the real `load_all_nodes` / `load_all_edges`. The parser mints `collection_id="halimbahae:<book>"`; the loader applies the canonical `col:` prefix to land `col:halimbahae:<book>`, matching `EXPECTED_PER_COLLECTION`. Result is 6 Hadith + 2 Collection nodes, every one sunni/halimbahae tagged, and 6 APPEARS_IN edges each routing a hadith to its own book with no danglers. `hadith_number` is always populated (`safe_int(...)` or the `i+1` fallback), so APPEARS_IN never trips the null-property MERGE abort, and that number is the in-book ordinal, which is the correct edge semantics.

## #35 matn caveat — acceptable, tracked
`full_text_ar` is not copied onto the graph node by the batch loader. That is pre-existing reconciliation behavior shared with `open_hadith`, tracked under #35, and the diacritic-preservation contract is correctly asserted at the parser level instead. Not a new regression.

## CI
All required checks green (Lint, Test, Type Check, sync-drift gate). The Integration Tests job is `continue-on-error` and ships no neo4j service, so the live light-up skips there via the Docker skip-guard. The genuine live proof is the local `neo4j:5` run, consistent with the established pattern.

Ship it.

parametrization:
Requestor: Tarek Mansour
RequestOrReplied: Approved

Reviewed via diff + CI only (no local suite run), per assignment. PR #131 (da#96 halimbahae Sunni adapter). Approving — clean, well-tested, serial-merge-friendly.

## What I verified

**(a) acquire + parse produce schema-valid, sunni / halimbahae-tagged staging parquet** — `src/parse/halimbahae.py` builds rows against `HADITH_SCHEMA` and emits one `Collection` per book against `COLLECTION_SCHEMA`, tagging `source_corpus="halimbahae"` and `sect="sunni"` on every row. The dict-comprehension materialisation against the schema would KeyError on any missing field, and `test_hadith_schema_conforms` asserts `table.schema == HADITH_SCHEMA` — both green in the CI Test job. The acquire module shallow-clones, asserts the 9-book diacritised-CSV floor, writes a manifest and emits `raw.landed`; idempotent.

**(b) #81 coverage invariant stays green** — `test_every_corpus_has_an_adapter` asserts `covered_corpora() == set(SourceCorpus)`. The PR adds all three sides — the `HALIMBAHAE` enum value, the `SourceAdapter` row, and the `EXPECTED_SLUGS` run-order pin — so the invariant, `test_canonical_run_order`, and `test_declared_sect_agrees_with_parser_constant` (adapter `Sect.SUNNI` vs parser `SECT="sunni"`) all hold. CI Test green.

**(c) #113 source_id corpus guard passes** — the parser uses the module-constant form `SOURCE_CORPUS = "halimbahae"` then `generate_source_id(SOURCE_CORPUS, collection, n)`. The `scripts/check_source_id_corpus.py` AST guard accepts that form (`test_valid_module_constant`), and `"halimbahae"` is now a real `SourceCorpus` member so it is also value-valid. `test_live_parse_package_is_clean` passes — no da#89-class free-text-key crash. Not the failing-fast bio-key path.

**(d) diacritic handling correct and consistent with src/utils/arabic** — `_clean_preserving_diacritics` strips only the bidi-control family (U+200E/F, U+202A–E, U+2066–9) and collapses whitespace via `clean_whitespace`; it never calls `strip_diacritics`/`normalize_arabic`, so tashkeel survives into `full_text_ar`. This matches the arabic-utils contract where diacritics are stripped only for the canonical name-matching path. `test_diacritics_preserved_bidi_marks_stripped` asserts tashkeel remains (`first_text != strip_diacritics(first_text)`, `حَدَّثَنَا` present) while U+200E/F are gone and no doubled spaces survive.

**(e) committed fixtures are real-data, no licensing issue** — two 3-row slices (Muwatta Malik + Sunan al-Darimi) of genuine classical matn with the real halimbahae bidi-mark + tashkeel shape. The text is centuries-old public-domain religious content; upstream halimbahae/Hadith is ODbL 1.0 + DbCL 1.0, documented in the adapter `license_note`. Unlike the Itqan no-upstream-license case, the license here is explicit. No blocker.

## Extra checks (clean)

- **APPEARS_IN null-property loader bug** (the known MERGE-on-null `hadith_number_in_book` abort): not triggered here — `hadith_num = safe_int(...) or (i+1)` is always a non-null int, and the live light-up asserts 6 APPEARS_IN edges resolving each hadith to its own book with no danglers. The CI **Integration Tests** job ran this against live `neo4j:5` and passed.
- **Double-corpus id** (`hdt:corpus:corpus:...`): not present — staging `source_id` is single-corpus `halimbahae:<book>:<n>`, graph node ids land as `col:halimbahae:<book>` per the integration assertions.
- All five CI checks green: Lint, Type Check, Test (347 passed/1 skipped), Integration Tests, Pre-commit/CI sync-drift gate.

## Non-blocking nits (no change required)

- Acquire counts CSVs by `"mushakkala" in name`, while the parser matches `mushakkala|tashkeel|diacritics`. Parser is a strict superset, so anything acquire accepts the parser also reads — harmless, but worth a one-line comment if you touch it again.
- `safe_int(n) or (i+1)` would fall back to the row index for a literal hadith number of 0; hadith numbering is 1-based so this never fires.

## Serial-merge surface (for #132 / #95)

Three trailing-append-only edits, distinct value/slug, rebase-clean:
- `src/models/enums.py` — `HALIMBAHAE = "halimbahae"` appended to `SourceCorpus` (after `ITQAN`).
- `src/adapters.py` — one appended `SourceAdapter` row, `slug="halimbahae"`, `corpus=SourceCorpus.HALIMBAHAE`, `sect=Sect.SUNNI`, `reachable=True`.
- `tests/test_adapters.py` — `"halimbahae"` appended to `EXPECTED_SLUGS`.

Issue #97, opened 2026-06-12
N4: Multi-IsnadSet (MIS) adapter — multiple isnad per Sahih Muslim hadith


parametrization:
**Tier:** T2 (new source) · **Suggested owner:** Kavitha Sundaramurthy
Part of epic #81.

## Goal
**Multi-IsnadSet (MIS)** — multiple isnad per Sahih Muslim hadith; strong for chain modeling. New `src/acquire/mis.py` + parser.

## Acceptance
- [ ] New `src/acquire/<src>.py` + `src/parse/<src>.py` producing canonical staging Parquet with `sect` + `source_corpus`.
- [ ] `SourceCorpus` enum value added (in `src/models/enums.py`).
- [ ] Parse + acquire-side tests.
- [ ] End-to-end load into a **local `neo4j:5` container** asserting real nodes/edges (local Docker confirmed working).
- [ ] Reachability + licensing note.

## Dependencies
T0-A (source_id scheme), T0-B (conformance gate).

parametrization:
Merged via PR #132 (Multi-IsnadSet adapter; STUDIED_UNDER allowlist interim for da#133).

PR #130, opened 2026-06-12
feat(itqan): emit isnad chains→edges (#93) + name variants→aliases (#94)

parametrization:
## Summary
Extends the Itqan rijal adapter (bios shipped in da#92a/PR#110) to emit the two remaining per-profile slices. Both extend `src/parse/itqan.py`, so they ship together to avoid a self-conflict.

- **#93 — isnad chains → isnad edges:** each profile's `teachers`/`students` id lists become student→teacher `STUDIED_UNDER` edges (`NETWORK_EDGE_SCHEMA`), deduped on the id pair across reciprocal perspectives.
- **#94 — name variants → identity aliases:** each profile's `namings` array becomes name-variant alias rows (new `NARRATOR_ALIAS_SCHEMA`) that `bio_promote` unions onto the matching canonical Narrator's `aliases`.

## Outputs added (staging Parquet, `sect`/`source_corpus` tagged)
| File | Schema | Contents |
|---|---|---|
| `network_edges_itqan.parquet` | NETWORK_EDGE_SCHEMA | student→teacher STUDIED_UNDER edges (design ~100,656) |
| `narrator_aliases_itqan.parquet` | NARRATOR_ALIAS_SCHEMA | name variants → aliases (design ~217,762) |

`run()` now returns `[bios, edges, aliases]` (a `ParseOutput` list — `run_all` flattens it).

## Key design decisions
- **No NARRATOR_MENTION rows.** Itqan is a rijal DB with no hadith-scoped chains (no `source_hadith_id`/`position_in_chain`); the teacher/student adjacency maps to STUDIED_UNDER, mirroring muhaddithat. The issue's mention-schema reference doesn't apply to this source.
- **STUDIED_UNDER loader fix (shared):** generalized from the hardcoded `network_edges_muhaddithat.parquet` to glob `network_edges_*.parquet`, and now resolves endpoints by name → `make_canonical_id` instead of `nar:<external_id>`. Canonical Narrator nodes are keyed by `uuid5(normalized-name)`, so the old external-id keying matched no node and silently dropped every edge — this also fixes that latent gap for muhaddithat. No existing test asserted the old behavior (the only prior STUDIED_UNDER test was graceful-skip).
- **Aliases feed the identity layer, not the matcher.** `bio_promote` attaches variants to the canonical record/node (`n.aliases`); making the disambiguator *key* on aliases for cross-source collapse is #X1's job (#94 "feeds" #X1). An alias never creates a narrator.

## Files touched
[...]

PR #132, opened 2026-06-12
feat: Multi-IsnadSet adapter (multiple isnad per Sahih Muslim hadith) (#97)

parametrization:
Closes #97 · Part of epic #81. Net-new T2 source: **Multi-IsnadSet (MIS)** — Sahih Muslim with the *multiple* isnad chains per hadith made explicit.

## Why MIS is different
Most corpora give one chain per hadith. MIS (Mendeley `10.17632/gzprcr93zn.2`, Farooqi et al., *Data in Brief* 2024) is the platform's **multi-isnad** source: a single Sahih Muslim hadith carries several parallel asanid (the tahwil / ح split). The whole value is preserving those as **distinct** transmission chains — collapsing them would throw away exactly what MIS exists to model.

## How multiplicity is preserved
The parser emits **`NETWORK_EDGE_SCHEMA`** (directed `from→to` narrator pairs), not `NARRATOR_MENTION_SCHEMA`. Deliberate and load-bearing:
- A hadith's N isnads contribute **N independent edge-chains** that coexist as distinct rows. Parallel chains that share endpoints but diverge in the middle keep their chain-specific links.
- The mention schema is keyed only by `source_hadith_id` + `position_in_chain` with **no per-chain discriminator** — feeding multi-isnad through it would merge the parallel asanid into one chain. NETWORK_EDGE sidesteps that.
- Each isnad is grouped by `(hadith, sanad)` and walked into consecutive pairs; edge `hadith_id` == the hadith's `source_id` (corpus-namespaced join), so edges tie to their Hadith node.

Proven end-to-end: the live `neo4j:5` lightup loads a 3-isnad hadith and asserts the chain-specific edges (`A→D`, `E→B`) land as real graph relationships — edges that only exist if the chains were **not** collapsed. 7 staging edges → 7 relationships, 5 distinct hadith-1 pairs.

## Changes
- **`src/models/enums.py`** — `SourceCorpus.MIS = "mis"`.
- **`src/adapters.py`** — register the `mis` `SourceAdapter` row (sect=`sunni`, reachable, CC BY 4.0 note).
- **`src/acquire/mis.py`** — download the two Mendeley `.xlsx` workbooks; file ids resolved at runtime from the dataset files API (keyless-public, same mechanism proven for `sanadset`).
- **`src/parse/mis.py`** — `hadiths_mis.parquet` (HADITH_SCHEMA) + `network_edges_mis.parquet` (NETWORK_EDGE_SCHEMA). Tolerant column detection; sequence + explicit-edge DetailsInfo layouts; xlsx + csv readers. `SOURCE_CORPUS="mis"` / `SECT="sunni"`.
- **`pyproject.toml` / `uv.lock`** — add `openpyxl>=3.1` (xlsx engine) + its mypy override.
- **`tests/test_adapters.py`** — `EXPECTED_SLUGS += "mis"` (coverage invariant).
- Tests: `test_parse/test_mis_parser.py` (multiplicity / schema / tagging / source_id grammar, both layouts, both readers), `test_acquire/test_mis.py` (no-network), `tests/integration/test_mis_lightup.py` (live neo4j:5, skip-guarded).

## SERIAL-MERGE NOTE (for #95 Kavitha / #96 Alejandra)
Shared files touched — all **append-only**, no reorders:
- `src/models/enums.py` — appended one enum value `MIS = "mis"`.
- `src/adapters.py` — appended one `SourceAdapter` row (slug `mis`) at the end of `SOURCE_REGISTRY`.
- `tests/test_adapters.py` — appended `"mis"` to `EXPECTED_SLUGS`.
- `pyproject.toml` / `uv.lock` — one dep (`openpyxl>=3.1`) + one mypy override; if a sibling also touches the lock, re-run `uv lock` after merge.

## Edge counts (fixture)
3-isnad hadith → A→B, B→C, A→D, D→C, E→B (5 distinct pairs; 6 raw edges incl. duplicate B→C across two chains) + 1-isnad hadith X→Y = **7 edges**, all schema-valid and tagged `source=mis`.

## Scope boundary / follow-up
The production `network_edges → graph` loader (`load_edges._load_studied_under`) is hardcoded to `network_edges_muhaddithat.parquet`, and isnad-transmission edges are semantically `TRANSMITTED_TO`, not `STUDIED_UNDER`. Wiring MIS's network_edges into the production edge loader is a **graph-layer decision** (glob vs. a new isnad branch) deliberately left out of this acquisition PR; the lightup loads MIS's own staging rows directly to give honest live E2E evidence now. Recommend a graph-repo follow-up issue.

## Test Plan
- [x] `pytest -m "not integration and not ml"` — 701 passed
- [x] `pytest -m integration tests/integration/test_mis_lightup.py` — 1 passed against live `neo4j:5`
- [x] `ruff check` + `ruff format --check` clean
- [x] `mypy` clean
- [x] `scripts/check_source_id_corpus.py src/parse` — clean (MIS uses module-const `SOURCE_CORPUS`)

Reviewers: @Jean-Claude Habimana (architect, primary) · @Nikolaos Papadopoulos (secondary)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ dependencies = [
     "tenacity>=8.2",
     "rapidfuzz>=3.6",
     "kafka-python>=2.3.1",
+    "openpyxl>=3.1",
 ]
 
 [project.scripts]
@@ -109,6 +110,10 @@ ignore_missing_imports = true
 
 [[tool.mypy.overrides]]
 module = ["kafka", "kafka.*"]
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = ["openpyxl", "openpyxl.*"]
 ignore_missing_imports = true
 
 [tool.pydantic-mypy]
--- a/src/acquire/mis.py
+++ b/src/acquire/mis.py
@@ -1,0 +1,111 @@
+"""Acquire the Multi-IsnadSet (MIS) dataset from Mendeley Data.
+
+MIS (``Multi-IsnadSet``, Farooqi et al., *Data in Brief* 2024 —
+DOI ``10.17632/gzprcr93zn.2``) models **Sahih Muslim** as a directed narrator
+graph in which a single hadith carries its *multiple* isnad chains explicitly
+(2,092 narrator nodes / 77,797 Sanad-Hadith edges). It is the dedicated
+multi-isnad source for the platform: where most corpora give one chain per
+hadith, MIS preserves the parallel asanid (the ``ح`` / *tahwil* split) that make
+Sahih Muslim valuable for chain modelling.
+
+The dataset ships as two Excel workbooks under the Mendeley file set:
+
+* ``Hadith_SahihMuslim_CoreInfo.xlsx``               — one row per hadith
+  (book, hadith number, matn, narrator sequence, isnad count).
+* ``Hadith_SahihMuslim_DetailsInfo_Sanad_Narrators.xlsx`` — one row per
+  (hadith, sanad, narrator) position — the data the parser walks into per-sanad
+  chains.
+
+Reachability: ``data.mendeley.com`` serves the file set over HTTPS with **no
+auth / no API key** — the same keyless-public mechanism this repo already proves
+for ``sanadset`` (Mendeley ``5xth87zwb5``). The file ids are resolved at runtime
+from the dataset files API rather than hardcoded, so a new dataset version is
+picked up without a code change.
+
+Licensing: Mendeley Data's default license for this dataset is **CC BY 4.0**
+(attribution) — redistribution-safe with credit to the depositors.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+from src.acquire.base import download_file, ensure_dir, fetch_json, write_manifest
+from src.messaging import emit_raw_new_for_manifest
+from src.utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+# Mendeley Data — Multi-IsnadSet (DOI: 10.17632/gzprcr93zn.2)
+_MENDELEY_DATASET_ID = "gzprcr93zn"
+_MENDELEY_VERSION = 2
+_MENDELEY_FILES_API = f"https://data.mendeley.com/api/datasets/{_MENDELEY_DATASET_ID}/files?version={_MENDELEY_VERSION}"
+
+# We only want the spreadsheet workbooks; the file set may also carry PDFs / docs.
+_WANTED_SUFFIXES = (".xlsx", ".xls")
+
+
+def _resolve_files() -> list[dict[str, Any]]:
+    """Return the Mendeley file-set entries for the wanted spreadsheet workbooks.
+
+    Each entry is normalized to ``{"name", "url", "sha256"}``. The Mendeley files
+    API returns objects shaped ``{"filename", "content_details": {"download_url",
+    "sha256_hash", ...}}``; we tolerate a couple of historical key spellings.
+    """
+    raw = fetch_json(_MENDELEY_FILES_API)
+    entries = raw if isinstance(raw, list) else raw.get("results") or raw.get("files") or []
+
+    wanted: list[dict[str, Any]] = []
+    for entry in entries:
+        if not isinstance(entry, dict):
+            continue
+        name = entry.get("filename") or entry.get("name") or ""
+        if not name.lower().endswith(_WANTED_SUFFIXES):
+            continue
+        details = entry.get("content_details") or entry
+        url = details.get("download_url") or details.get("url")
+        if not url:
+            continue
+        wanted.append(
+            {
+                "name": name,
+                "url": url,
+                "sha256": details.get("sha256_hash") or details.get("sha256"),
+            }
+        )
+    return wanted
+
+
+def run(raw_dir: Path) -> Path:
+    """Download the MIS Excel workbooks into ``raw_dir/mis``.
+
+    Idempotent: ``download_file`` skips a non-empty existing workbook, so a
+    re-run only fetches what is missing. Fails fast if the file set carries no
+    spreadsheet workbook (a dataset-shape change we want to notice loudly).
+    """
+    dest = ensure_dir(raw_dir / "mis")
+
+    files = _resolve_files()
+    if not files:
+        msg = (
+            f"MIS Mendeley file set ({_MENDELEY_DATASET_ID} v{_MENDELEY_VERSION}) "
+            "exposed no .xlsx workbook"
+        )
+        raise AssertionError(msg)
+
+    downloaded: list[Path] = []
+    for entry in files:
+        path = download_file(
+            entry["url"],
+            dest / entry["name"],
+            expected_sha256=entry["sha256"],
+            timeout=600.0,
+        )
+        downloaded.append(path)
+
+    logger.info("mis_acquired", workbooks=len(downloaded), names=[p.name for p in downloaded])
+
+    write_manifest(dest, downloaded)
+    emit_raw_new_for_manifest(source="mis", local_dir=dest, files=downloaded)
+    return dest
--- a/src/adapters.py
+++ b/src/adapters.py
@@ -228,6 +228,23 @@ SOURCE_REGISTRY: tuple[SourceAdapter, ...] = (
         ),
         description="halimbahae/Hadith — 9 Sunni books with full Arabic diacritics (tashkeel).",
     ),
+    SourceAdapter(
+        slug="mis",
+        corpus=SourceCorpus.MIS,
+        sect=Sect.SUNNI,
+        acquire_module="mis",
+        parse_module="mis",
+        reachable=True,
+        license_note=(
+            "Multi-IsnadSet (MIS) — Mendeley Data 10.17632/gzprcr93zn.2 (Farooqi et al., "
+            "Data in Brief 2024). CC BY 4.0; keyless public download over the same "
+            "data.mendeley.com mechanism proven for sanadset."
+        ),
+        description=(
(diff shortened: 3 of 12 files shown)


parametrization:
Requestee: Kwesi Boateng
RequestOrReplied: Approved

Secondary (data-engineer lens) review of da#97 — diff + CI only, local suite not run.

## Multiplicity math — holds
Sequence-layout fixture, hadith 1 = three isnads:
- isnad A `A->B->C` yields (A,B),(B,C)
- isnad B `A->D->C` yields (A,D),(D,C)
- isnad E `E->B->C` yields (E,B),(B,C)
- hadith 2 single isnad `X->Y` yields (X,Y)

Six h1 rows (the two `(B,C)` from isnad-A and isnad-E are BOTH retained — no cross-chain dedup) + one h2 row = 7 staging rows. Five DISTINCT h1 directed pairs {(A,B),(B,C),(A,D),(D,C),(E,B)}. The chain-specific links `A->D` and `E->B` only exist because the parallel asanid are walked per `(hadith, sanad)` group rather than collapsed — `_parse_edges_sequence` groups on `(hadith_no, sanad_no)`, sorts by position, walks consecutive pairs. Math is correct and the asserted counts (`len==7`, distinct==5) match.

## Schema + tagging — valid
Edge table built straight from NETWORK_EDGE_SCHEMA (6 fields: from/to name, hadith_id, source, from/to external_id). Required non-null fields always populated; `_emit_edge` drops incomplete/self-loop pairs. `source="mis"` tags every edge; `sect=sunni` rides on the Hadith node (no sect field on the edge schema, correct). `source_id` grammar is corpus-namespaced and the enum `SourceCorpus.MIS` was added, so `generate_source_id` will not fail-fast on the `mis` namespace (clears the da#82 non-corpus-key trap). CI `Type Check` + `test_edge_schema_and_tagging` green confirm field-set parity.

## Excel parse — robust
openpyxl `read_only`+`data_only`, alnum-normalized headers, tolerant candidate-key sets for both the sequence and explicit-edge layouts, xlsx+csv readers, empty-row + ragged-row guards (`min(len(headers), len(raw))`). Fail-loud `ValueError` on missing hadith-number / from-to columns with the resolved headers in the message. Defensible against the real Mendeley shape; the live `neo4j:5` lightup (CI Integration green) is honest E2E.

## Download — keyless + idempotent
File ids resolved at runtime from the Mendeley files API (no hardcoded ids, version-bump tolerant), tolerant key fallbacks, `.xlsx`-only selection with PDF skip proven in `test_downloads_only_xlsx_workbooks`, fail-fast AssertionError when the set exposes no workbook. `download_file` skips a non-empty existing file → re-run idempotent. Real network fetch is a runtime-acceptance item, not PR-blocking.

## Edge-type concern — I concur with the deferral, with one forward caveat for the architect
Agreed: MIS isnad-transmission edges are semantically a distinct relationship from muhaddithat STUDIED_UNDER teaching edges; folding them into `load_edges._load_studied_under` (hardcoded to `network_edges_muhaddithat.parquet`) would conflate two different edge types. Scoping the production-loader wiring to a graph-repo follow-up is the right call and the lightup loading MIS's own staging rows is honest evidence now.

One caveat to carry into that wiring decision (NOT a blocker here — staging correctly retains the rows): NETWORK_EDGE_SCHEMA has no per-isnad discriminator, so the duplicate `(B,C)` from isnad-A and isnad-E are byte-identical rows. The integration proof uses `CREATE` (gets 7 relationships). If the eventual production edge loader uses `MERGE` on (from,to,hadith) the way the node/APPEARS_IN loaders do, those shared sub-path duplicates collapse and the transmission MULTIPLICITY that MIS exists to model is undercounted at load. The follow-up issue should specify CREATE-or-chain-keyed-MERGE so the count survives into the graph.

CI: Lint / Type Check / Test / Integration / sync-drift all green.

Approved.

parametrization:
Requestee: Kwesi Boateng
RequestOrReplied: Approved

## Architect review — da#97 Multi-IsnadSet adapter

Strong, well-reasoned PR. The multi-isnad modeling is correct, the evidence is honest, and the production-loader scope boundary is drawn in exactly the right place. Approving. The one substantive issue (a cross-PR latent mislabel with #130) is genuinely deferred, not papered over — see the coordination note below, which I'm escalating to the wave lead in parallel.

### What I verified

- **Multiplicity genuinely preserved.** `_parse_edges_sequence` groups by `(hadith, sanad)` and walks each sanad into its own consecutive-pair chain, so N isnads yield N independent edge-chains. The live lightup proves it end-to-end — a 3-isnad hadith lands 7 relationships with the chain-specific pairs A->D and E->B present, which only exist if the parallel asanid were not collapsed. This is the load-bearing claim and it holds.
- **NETWORK_EDGE over the mention schema is the right call.** The mention schema is keyed by `(source_hadith_id, position_in_chain)` with no per-chain discriminator, so routing multi-isnad through it would merge parallel chains at each position — the exact failure mode this avoids. Emitting directed `from -> to` pairs sidesteps that cleanly.
- **Source provenance is sound.** Mendeley 10.17632/gzprcr93zn.2, CC BY 4.0, keyless-public over the same `data.mendeley.com` files-API path already proven for sanadset. File ids resolved at runtime (version-tolerant), wanted-suffix filtered, fail-fast when no workbook is present. Good.
- **Dependency change is clean.** `openpyxl>=3.1` added to deps plus a mypy ignore-missing-imports override, with a matching uv.lock bump. Standard re-lock caveat if a sibling PR also touches the lock.
- **source_id guard (#113) passes.** `generate_source_id(SOURCE_CORPUS, ...)` keys on the module constant `"mis"`, which is now a valid SourceCorpus member, so the corpus-const check is satisfied.
- **CI** — all five checks green (Lint, Type Check, Test, Integration, sync-drift gate).

### Cross-PR concern — my call (real, but correctly deferred)

(a) **Yes, this is a real latent mislabel once both land.** #130 rewrites `_load_studied_under` to glob `network_edges_*.parquet` and load every match as STUDIED_UNDER. MIS writes `network_edges_mis.parquet`, whose rows are isnad-transmission pairs (semantically TRANSMITTED_TO), and `mis` is a registered SourceAdapter, so a full parse populates that file into the shared staging dir. If `load_all_edges` ever runs over a staging dir containing it, the MIS edges silently become STUDIED_UNDER — wrong relationship type, and very likely reversed direction too (the studentship loaders expect student->teacher; the isnad walk emits transmission order). The MockNeo4jClient and the fact that nothing wires MIS into production today are what keep this latent.

(b) **Yes — NETWORK_EDGE_SCHEMA should carry an edge-relation discriminator.** The durable fix is a relation/edge-type column on the schema (e.g. studied_under vs transmitted_to) that the glob-loader reads to route each file to the correct relationship, instead of the loader assuming every `network_edges_*` file is studentship. That removes the filename-coupling and makes the producer declare its own semantics. A cheaper interim, if speed matters, is to narrow #130's glob to a known-studentship allowlist (muhaddithat, itqan) rather than an open glob — but the schema field is the right long-term answer.

(c) **#132 is safe to merge as-is.** It does not itself wire MIS into the production edge loader; the lightup loads MIS's own staging rows via a dedicated `ISNAD_TRANSMITTED` rel, so #132 in isolation cannot mislabel anything. The mislabel is purely an emergent property of #130's open glob meeting #132's new file. The scope boundary and follow-up recommendation in the PR body are correct.

What I'm not comfortable leaving implicit: the follow-up must land **before** MIS staging is included in any production `load_all_edges` run, not "eventually." Whichever of #130/#132 merges second leaves the mislabel armed. I'd treat the edge-type-field follow-up as a same-wave gate and coordinate the merge sequencing with #130.

### Minor (non-blocking) notes

- Edge `hadith_id` carries the raw `source_id` form (`mis:sahih_muslim:1:1`) while the Hadith node id is the `hdt:`-prefixed form. It's a property tag here, not a join, so harmless today — but the PR body's "edges tie to their Hadith node" is slightly optimistic; a future Hadith-join would need to reconcile the prefix. Worth a one-line comment.
- `expected_sha256` is `None` in the no-network tests; production relies on the Mendeley API actually returning `sha256_hash`. Fine, just noting the verification is best-effort if upstream omits it.

Nothing here blocks. Approved.

parametrization:
Requestee: Kwesi Boateng
RequestOrReplied: Approved

## Re-affirm at post-rebase head (25c7d697d)

Re-reviewed the rebase + the folded-in glob allowlist (commit 25c7d69). MIS core is unchanged from my prior Approved; the only new surface is the `_load_studied_under` narrowing, and it resolves the cross-PR concern cleanly. Re-affirming Approved.

- **(a) Allowlist correctly excludes mis while keeping the studentship sources.** `_STUDIED_UNDER_SOURCES = {muhaddithat, itqan}` gated through `_is_studied_under_file`; for `network_edges_mis.parquet` the stripped slug `mis` neither equals a source nor starts with `muhaddithat_`/`itqan_`, so it is dropped before the batch. The new `test_excludes_non_studentship_source` proves the mis pair never enters the MERGE batch while the muhaddithat edge does (created == 1), and the existing two-source glob test still yields created == 2. So the change is additive-safe: muhaddithat + itqan behavior is preserved.
- **(b) Chunked `<slug>_NNN` matching is sound.** `slug == src or slug.startswith(f"{src}_")` resolves a future sharded write (e.g. `network_edges_muhaddithat_001`) while a distinct corpus never does. One theoretical edge — a future corpus slug that literally begins `itqan_`/`muhaddithat_` would false-positive — but slugs are a controlled finite SourceCorpus set, and da#133 (the edge-relation field that keys on data rather than filename) is the durable supersede. Acceptable for the interim allowlist.
- **(c) CI green at head** — Lint, Type Check, Test, Integration, sync-drift gate all pass on run 27447244085.

The comment block correctly frames this as the cheap interim and points at da#133 as the durable fix. Good to merge. Approved.

PR #134, opened 2026-06-12
feat(bihar): Bihar al-Anwar Shia coverage via hubeali (#95)

parametrization:
Closes #95

## New-corpus vs extension decision: NEW `SourceCorpus.BIHAR`
Bihar al-Anwar is a distinct collection from the Four Books, `corpus` is the
identity/dedup namespace, and its real source isn't Thaqalayn — so tagging it
`thaqalayn` would be a provenance lie. New corpus it is.

## Source pivot (da#95) — owner-approved
The issue assumed Bihar is reachable "via Thaqalayn." Verification proved that
false: the ThaqalaynAPI GitHub repo carries 33 books and live `thaqalayn.net`
27 — the Four Books + ~20 secondary texts, **no Bihar** — and no openly-licensed
structured Bihar JSON exists anywhere. The owner approved pivoting to
**hubeali.com**'s bilingual (Arabic + English) "Read Online" edition, the one
openly-readable per-hadith copy.

## What this adds
- **`SourceCorpus.BIHAR = "bihar"`** (`src/models/enums.py`).
- **One `SourceAdapter` row** (`src/adapters.py`): `slug=bihar`, `corpus=BIHAR`,
  `sect=SHIA`, `reachable=True`. The #81 coverage invariant + `EXPECTED_SLUGS`
  updated together.
- **`src/acquire/bihar.py`** — polite bounded scraper (robots.txt honoured,
  throttled, raw-HTML cached, idempotent) over
  `/books-library/bihar-al-anwaar/volume-<V>/part-<P>/...`. `DEFAULT_VOLUMES=(1,)`
  for a bounded PR-time crawl; widen via `volumes=` for a full data-load run.
- **`src/parse/bihar.py`** — segments the alternating AR/EN `<p>` run into
  chapters (`باب N`) and numbered hadiths (`<N>- ...`), emitting schema-valid
  parquet tagged `sect=shia` + `source_corpus=bihar`.
  - `source_id` grammar: `bihar:bihar-al-anwar:<volume>:<chapter>:<ordinal>`.
    The collection slug is **`bihar-al-anwar`, deliberately not `bihar`**, so the
    id is not collapsed by `identity.is_double_prefixed` (which fires on a
    repeated leading corpus segment).

## Licensing
Same owner-approved posture as Itqan (da#92a): no machine-readable upstream
license, non-profit use, hadith facts re-expressed in our own schema, cleanly
removable via the `source_corpus='bihar'` provenance tag. Recorded in the
adapter's `license_note`.

## Test Plan
- `tests/test_parse/test_bihar_parser.py` — parser exercised on a **real**
  (trimmed, verbatim) hubeali fixture: 8 hadiths, schema conformance, AR+EN
  extraction, `<N>-` marker stripping, sequential ordinals,
  `validate_source_id == []`, not double-prefixed, sect/corpus provenance.
- `tests/test_acquire/test_bihar.py` — mocked unit legs (robots-gating,
  content-link discovery, raw caching, idempotent re-run, no-pages→None) +
  **skip-guarded live leg** (`BIHAR_LIVE_TEST=1`) that crawls + parses the real
  site.
- `tests/test_adapters.py` — #81 coverage invariant passes with `bihar`.
- Full `tests/test_acquire` + `tests/test_parse`: **272 passed, 2 skipped**;
  mypy strict clean; ruff format + lint clean.
- **Real-data proof:** a live volume-1 crawl produces **394 bilingual hadiths
  across 7 chapters**, every row with non-null Arabic + English and valid,
  unique source_ids.
- **Runtime/data-load step (not PR-time):** the full ~100k-hadith multi-volume
  scrape + Neo4j load is a bounded-by-design follow-up (`volumes=range(...)`),
  documented in the acquire module; the live neo4j:5 E2E load is a runtime
  acceptance item.

## Serial-merge note (#95 / #96 / #97)
Rebased onto wave-4 after **#96 (halimbahae) merged**; the `EXPECTED_SLUGS` /
enum / registry conflicts were resolved keeping all entries (registry tail order
is now `... itqan, halimbahae, bihar`). #97 will hit the same three files —
trivial keep-all resolution.

Reviewers: @Jean-Claude Habimana (architect, primary) + @Oyunbileg Batbayar (QA, secondary).

### DIFF
--- a/src/acquire/bihar.py
+++ b/src/acquire/bihar.py
@@ -1,0 +1,151 @@
+"""Acquire Bihar al-Anwar (بحار الأنوار) from hubeali.com.
+
+Bihar al-Anwar — al-Majlisi's ~100k-hadith Shia encyclopedia — is NOT carried by
+the Thaqalayn data source (which is the Four Books + ~20 secondary texts only;
+verified da#95). hubeali.com's "Read Online" edition is the one openly-readable,
+per-hadith, bilingual (Arabic + English) copy. It is published as one large
+WordPress page per *volume / part*::
+
+    /books-library/bihar-al-anwaar/volume-<V>/                       (volume index)
+    /books-library/bihar-al-anwaar/volume-<V>/part-<P>/
+        bihar-al-anwaar-volume-<V>-part-<P>/                         (content page)
+
+This downloader is a *polite, bounded* scraper modelled on
+:mod:`src.acquire.sunnah_scraper`: it honours ``robots.txt`` (hubeali only
+disallows ``/wp-admin/``), throttles between requests, and caches each content
+page's raw HTML to ``data/raw/bihar/volume-<V>-part-<P>.html`` (idempotent — a
+cached page is skipped). :mod:`src.parse.bihar` parses those cached pages.
+
+``DEFAULT_VOLUMES`` is deliberately small (volume 1): a PR-time acquisition pulls
+a bounded, representative slice rather than the whole ~100k-hadith corpus. Pass
+``volumes=range(1, N)`` (or set it in a job) to widen the crawl for a real
+data-load run.
+
+Licensing (owner-approved da#95, same posture as Itqan da#92a): hubeali content
+carries no machine-readable upstream licence; use is non-profit, the hadith facts
+are re-expressed in our own schema, and everything is cleanly removable via the
+``source_corpus="bihar"`` provenance tag.
+"""
+
+from __future__ import annotations
+
+import re
+import time
+from collections.abc import Iterable
+from pathlib import Path
+from urllib.robotparser import RobotFileParser
+
+import httpx
+
+from src.acquire.base import ensure_dir, write_manifest
+from src.messaging import emit_raw_new_for_manifest
+from src.utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+BASE_URL = "https://hubeali.com"
+BOOK_ROOT = f"{BASE_URL}/books-library/bihar-al-anwaar"
+RATE_LIMIT_SECONDS = 1.0
+REQUEST_TIMEOUT = 60.0
+USER_AGENT = "isnad-graph/1.0 (hadith-research)"
+
+# A PR-time crawl pulls a bounded slice; widen for a full data-load run.
+DEFAULT_VOLUMES: tuple[int, ...] = (1,)
+
+# The deep "content" link on a volume index page, e.g.
+# ".../volume-1/part-1/bihar-al-anwaar-volume-1-part-1/". The shorter
+# ".../part-1/" part-index link is ignored — we want the page with the text.
+_CONTENT_LINK_RE = re.compile(
+    r"/books-library/bihar-al-anwaar/volume-(\d+)/part-(\d+)/"
+    r"bihar-al-anwaar-volume-\d+-part-\d+/?",
+    re.IGNORECASE,
+)
+
+
+def _check_robots_txt(client: httpx.Client) -> bool:
+    """Return True if ``robots.txt`` permits crawling the book pages."""
+    try:
+        resp = client.get(f"{BASE_URL}/robots.txt")
+        parser = RobotFileParser()
+        parser.parse(resp.text.splitlines())
+        allowed: bool = parser.can_fetch(USER_AGENT, f"{BOOK_ROOT}/volume-1/")
+        if not allowed:
+            logger.warning("bihar_robots_denied")
+        return allowed
+    except Exception:  # noqa: BLE001 — robots unreachable: proceed cautiously, like sunnah_scraper
+        logger.warning("bihar_robots_unavailable")
+        return True
+
+
+def _get(client: httpx.Client, url: str) -> str | None:
+    """GET *url*, returning the response text or ``None`` on any HTTP error."""
+    try:
+        resp = client.get(url)
+        resp.raise_for_status()
+        return resp.text
+    except httpx.HTTPError as exc:
+        logger.warning("bihar_fetch_error", url=url, error=str(exc))
+        return None
+
+
+def _discover_content_links(client: httpx.Client, volume: int) -> list[tuple[int, str]]:
+    """Find ``(part_number, url)`` content links on a volume index page."""
+    html = _get(client, f"{BOOK_ROOT}/volume-{volume}/")
+    if html is None:
+        return []
+    seen: dict[int, str] = {}
+    for match in _CONTENT_LINK_RE.finditer(html):
+        if int(match.group(1)) != volume:
+            continue
+        part = int(match.group(2))
+        url = match.group(0)
+        if not url.startswith("http"):
+            url = BASE_URL + url
+        seen.setdefault(part, url.rstrip("/") + "/")
+    return sorted(seen.items())
+
+
+def run(raw_dir: Path, volumes: Iterable[int] = DEFAULT_VOLUMES) -> Path | None:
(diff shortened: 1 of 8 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved

QA / data-quality review of da#95 — real Bihar al-Anwar adapter (hubeali.com). Verified by diff + files API + CI; local suite not run, per brief.

## What holds up well

- The committed fixture `bihar_volume-1-part-1_sample.html` is genuine hubeali markup, not a synthetic toy — it carries the real WordPress shape (`_Toc...` anchors, `<sup>asws</sup>` honorific glyphs, `&#8211;` entities, the `كتاب`/`باب` headings, the Quranic-verse preamble, and the alternating Arabic/English `<p>` run with real isnad chains). The parser is exercised against authentic source structure.
- Bilingual completeness is asserted, not just claimed — `test_real_bilingual_text_extracted` enforces `all(full_text_ar)` and `all(full_text_en)` across all 8 rows, and the `<N>- ` ordinal-marker strip is checked.
- Provenance is enforced on every row — `sect == shia` and `source_corpus == bihar` are asserted in two tests, plus on the collection row; the new `SourceCorpus.BIHAR` and the adapter wiring (`sect=Sect.SHIA`) are consistent.
- The double-prefix defense is sound and tested — collection slug `bihar-al-anwar` is deliberately distinct from the corpus token `bihar`, and `validate_source_id(sid) == []` plus `not is_double_prefixed(sid)` plus per-row uniqueness are all asserted.
- Robots gating, content-link discovery (deep content link vs shorter part-index link), raw-cache idempotency, and the empty-crawl `None` path are all covered; the live leg is correctly skip-guarded behind `BIHAR_LIVE_TEST=1` so CI never depends on a third-party site.
- All five CI checks are green (Lint, Type Check, Test, Integration Tests, sync-drift gate).

## One required pre-load follow-up (not blocking this bounded slice)

The source_id grammar is `bihar / bihar-al-anwar / <volume> / <chapter> / <ordinal>` — it omits both the `كتاب` (kitab) and the part. In Bihar al-Anwar the أبواب restart numbering from 1 inside each kitab, and a single volume contains several kitabs. `run()` extends rows across every cached page with no cross-page uniqueness guard, and `_parse_page` ignores the `كتاب` heading entirely (only `باب` advances `chapter_number`). So once the crawl crosses a kitab boundary the second kitab's bab-1 hadith-1 re-derives the same id as the first kitab's bab-1 hadith-1 (`bihar / bihar-al-anwar / 1 / 1 / 1`), yielding duplicate source_ids in the staging parquet — which a downstream MERGE-on-source_id load would silently collapse into one node.

This does not surface in the committed evidence because the fixture is a single kitab / single bab — exactly the toy-fixture-masks-a-real-defect shape we have been bitten by before (cf. the MockNeo4jClient and mock-injection masking lessons). And `DEFAULT_VOLUMES = (1,)` crawls ALL parts of volume 1, which spans multiple kitabs, so the real 394-hadith crawl is the path that would actually collide — that figure is live-only and not corroborated by CI (acceptable given the skip guard, but worth stating plainly).

I am approving because the delivered unit is acquire + parse-to-staging on a genuinely-tested bounded slice with correct provenance and a green pipeline, and the collision is a load-time integrity concern rather than a defect in this slice's committed artifact. Before any multi-kitab or full-volume crawl feeds the graph, the id must incorporate kitab (and ideally part) so distinct hadiths stay distinct, with a multi-kitab fixture added so the uniqueness assertion actually exercises a kitab boundary. Recommend a tracking issue for that pre-load hardening.

## Minor, non-blocking

- An English paragraph containing any embedded Arabic glyph would be routed into `full_text_ar` by `_is_arabic`; the real fixture does not trigger it, but it is a latent classification edge worth a comment.
- `chapter_name_en` is always null and `chapter_name_ar` stores the whole `باب ...` line including the number — fine for staging, just noting it.

Strong, well-documented adapter. Approved with the source_id-uniqueness item flagged as required before the full data load.

parametrization:
Requestor: Tarek Mansour
RequestOrReplied: Approved

Reviewed with a devops/scraper lens — diff + CI + the upstream contracts at PR head. Approving. This is a real adapter, not a stub, and the source-pivot reasoning (Thaqalayn provably lacks Bihar to hubeali bilingual Read-Online) is sound and owner-approved.

## Politeness and safety — good
- robots.txt honoured via RobotFileParser.can_fetch(USER_AGENT, ...); deny short-circuits to None, unreachable-robots falls back permissively with a warning, matching the house sunnah_scraper convention.
- Throttled ~1s before each content fetch; raw HTML cached to data/raw/bihar/volume-V-part-P.html with an idempotent size-guarded skip on re-run. follow_redirects, 60s timeout, descriptive UA. Client closed in finally.
- DEFAULT_VOLUMES=(1,) keeps the PR-time crawl bounded; volumes=range(...) widens for the real load. Correct shape.

## CI never depends on the third party — verified
The live leg is skip-guarded behind BIHAR_LIVE_TEST=1; the always-on legs patch httpx.Client.get with a URL router (robots-gate, content-link discovery, raw caching, idempotency, no-pages to None). All 5 checks green (Lint, Type Check, Test, Integration, sync-drift gate).

## Schema and provenance — verified
Parser tags sect=shia + source_corpus=bihar; write_parquet enforces the provenance contract (non-empty, in-domain) as a fail-fast, so an untagged row cannot ship. Fixture test asserts non-null AR+EN across all 8 rows, schema equality, and the collection row.

## source_id collision-safety — verified in code, not just asserted
I traced it upstream. SOURCE_CORPORA is frozenset(c.value for c in SourceCorpus), so adding BIHAR makes the leading segment a known corpus and validate_source_id passes. is_double_prefixed fires only when segments[0]==segments[1] and segments[0] is a known corpus; with the bihar-al-anwar collection slug, segment[1] differs, so it correctly returns False. Using the bare bihar slug WOULD have collapsed — the deliberate slug choice is the right call and is now defended by a test.

## #81 coverage invariant — green
Enum value, SourceAdapter row, and EXPECTED_SLUGS are updated together; covered_corpora()==set(SourceCorpus) holds. license_note carries the Itqan-posture text. Registry tail order is now ...itqan, halimbahae, bihar.

## Fixture is genuinely real
fixtures/bihar_volume-1-part-1_sample.html is verbatim hubeali markup — _Toc anchors, nbsp, sup-asws honorific glyphs, ftn footnote links, the real Quranic-verse chapter preamble — not a toy. The preamble is correctly dropped (cur_number is None until the first numbered marker), the N- ordinal marker is stripped, and ordinals 1..8 sequence under chapter 1.

## Non-blocking — for the full multi-volume data-load follow-up (NOT this PR)
1. Cross-kitab ordinal uniqueness. The grammar is bihar:bihar-al-anwar:<volume>:<chapter>:<ordinal> where chapter is the باب number, which restarts per كتاب. A volume holds several kutub, so two kutub in one volume can each carry "باب 1"/hadith 1 to identical source_ids. The parser detects the كتاب heading but does not fold it into the id. The bounded PR slice (one kitab span) cannot exercise this, and uniqueness IS proven within the fixture, so it is not a PR-time defect — but the full-volume load should either include a kitab segment or carry continuous chapter numbering, else a real load will collide. Worth a tracked follow-up before the runtime data-load step.
2. Minor — the volume-index discovery GET and the robots GET are not throttled (only content fetches are). Negligible at one index page per volume; fold a sleep before the index fetch into the same follow-up if convenient.

Neither blocks merge — both are full-corpus runtime concerns, and the PR explicitly defers the full scrape+load to a bounded follow-up. Mechanic correctness and provenance are right for the bounded slice. Approved.

Issue #90, opened 2026-06-12
L7: light up muhaddithat end-to-end (both sects) — depends on T0-B


parametrization:
**Tier:** T1 (light up existing adapter) · **Suggested owner:** Kavitha Sundaramurthy
Part of epic #81.

## Goal
Light up **muhaddithat** (both sects, female narrators — muhaddithat/isnad-datasets) end-to-end. Adapter exists (`src/acquire/muhaddithat.py`, `src/parse/muhaddithat.py`) but **currently does not tag sect/source_corpus**.

## Dependency
**Blocked on T0-B** (muhaddithat tagging fix + conformance gate) — that fix lands first, then this loads.

## Acceptance (per owner — live local Neo4j load at PR-time)
- [ ] Adapter runs `acquire → parse → resolve → load` end-to-end into a **local `neo4j:5` container** (local Docker confirmed working — Server 29.5.3, image present).
- [ ] Test asserts **real nodes/edges** land with correct `sect` + `source_corpus` (not just parse/conformance).
- [ ] Add the missing **acquire-side test** (only `base` + `sunnah_scraper` have one today).
- [ ] Reachability + licensing note in the PR.
- [ ] Staging-cluster load (cluster-internal, ssh `docker exec cypher-shell`) tracked as a **separate post-merge verify**, not a PR gate.

## Dependencies
T0-A (source_id scheme) should merge first.

parametrization:
Merged via PR #135 (muhaddithat end-to-end: cross-tradition sect=neutral + Arabic-name STUDIED_UNDER edge fix; live neo4j proof).

PR #135, opened 2026-06-12
feat(muhaddithat): end-to-end light-up, both-sects sect tagging (#90)

parametrization:
Closes #90 — light up the **muhaddithat** corpus end-to-end (both sects), proven against a live `neo4j:5` container.

## What was already fixed (verified on the wave branch, NOT redone)
- **#130** — `_load_studied_under` globs `network_edges_*.parquet` + name-resolves endpoints. Mechanism present.
- **#117** — `bio_promote` wired into `run_all` (muhaddithat bios flow to canonical). Verified.
- **#103** — `narrators_canonical` carries `source_corpus`/`source_corpora`/`sect_affiliation`.

## The two muhaddithat-specific gaps that remained (reproduced, then fixed)
Reproduced empirically with realistic upstream-shaped data (separate `displayname` + `arabicname` columns — the parser fixtures conflated them, which is why these slipped through):

1. **Sect tagging wrong for a both-sects source.** muhaddithat was mapped `corpus -> SUNNI`, so every muhaddithat narrator derived `sect_affiliation = sunni`. The upstream `variousnarrators.csv` carries **no per-narrator sect column** (`id, displayname, fullname, searchname, arabicname, gender, generation, info, hadiths, notes`) and the corpus spans female narrators recognized across **both** traditions.
   - **Fix:** `src/resolve/sect_affiliation.py` — new `CROSS_TRADITION_CORPORA`; `derive_sect_affiliation` contributes **both** Sunni and Shia for such a corpus -> a muhaddithat-only narrator is **`neutral`** (both), never a one-sided guess, never `unknown`. Aligns the narrator-level derivation with the adapter registry, which already declares `muhaddithat sect=None` (multi-sect).

2. **STUDIED_UNDER edges resolved to 0.** `bio_promote` mints the canonical id from the normalized **Arabic** name; `_load_studied_under` resolves endpoints the same way — but the edge map carried the Latin `displayname`, so every endpoint minted a **different** id -> all missing -> 0 edges. #130's glob fix can't help because the names don't reconcile.
   - **Fix:** `src/parse/muhaddithat.py` — feed the **Arabic** name into the edge id-map (fall back to display name only when no Arabic name exists).

## The muhaddithat sect decision
Source has no per-narrator sect -> cannot preserve per-narrator. Per the both-sects AC and the brief's fallback, muhaddithat is modeled as **cross-tradition -> `neutral`**, consistent with the registry's existing `sect=None` multi-sect declaration.

## E2E proof (live neo4j:5)
`tests/integration/test_muhaddithat_lightup.py` drives the real `parse -> run_all -> load` path (no hand-injected ids):
- Narrator nodes materialize with `source_corpus=muhaddithat` and `sect_affiliation=neutral`.
- `STUDIED_UNDER` edges fire with **0 missing endpoints**.
- Idempotent reload (nodes + edges).

Live run: **3 passed in 15.5s**. Unit suites for changed areas: **404 passed, 3 skipped**. ruff 0.15.11 (lint+format) clean; mypy clean.

## Reviewers
@Kavitha Sundaramurthy (primary), @Jean-Claude Habimana (secondary)

## Reachability + licensing
- **Reachability:** muhaddithat = a public GitHub clone (`muhaddithat/isnad-datasets`), reachable from the sandbox; the e2e load runs against a local `neo4j:5` container.
- **Licensing:** source data from the muslimscholars.info / muhaddithat project; facts are re-expressed in our own schema with `source_corpus=muhaddithat` provenance on every node, so the corpus is auditable and cleanly removable.

## Test Plan (post-merge)
- [ ] Staging-cluster load (cluster-internal, ssh `docker exec cypher-shell`) — separate verify, not a PR gate per #90 AC.

### DIFF
--- a/src/parse/muhaddithat.py
+++ b/src/parse/muhaddithat.py
@@ -107,8 +107,16 @@ def _parse_narrator_bios(narrators_path: Path) -> tuple[pa.Table, dict[str, str]
         gender = safe_str(gender_values[i])
         bio_text = safe_str(bio_values[i])
 
-        if display_name:
-            id_to_name[ext_id] = display_name
+        # Edges must reference the SAME name the canonical identity is minted
+        # from. ``bio_promote`` keys a narrator on ``make_canonical_id`` over the
+        # *normalized Arabic* name, and the ``STUDIED_UNDER`` loader resolves an
+        # edge endpoint the same way (``normalize_arabic`` → ``make_canonical_id``).
+        # So feed the Arabic name into the edge map; the Latin display name would
+        # mint a different id and every endpoint would miss (da#90). Fall back to
+        # the display name only when no Arabic name is available.
+        edge_name = name_ar or display_name
+        if edge_name:
+            id_to_name[ext_id] = edge_name
 
         rows.append(
             {
--- a/src/resolve/sect_affiliation.py
+++ b/src/resolve/sect_affiliation.py
@@ -21,6 +21,14 @@ corpora are intentionally left unmapped: ``fawaz`` (sect is per-collection, not
 per-corpus) and ``itqan`` (a mixed rijal database) — both contribute ``unknown``
 rather than a guess.
 
+``muhaddithat`` is a **cross-tradition** corpus (``CROSS_TRADITION_CORPORA``):
+the upstream isnad-datasets carry no per-narrator sect/madhhab column — only
+id/name/gender/generation — and the corpus deliberately spans female narrators
+recognized across **both** Sunni and Shia chains (da#90, epic #81). So rather
+than a single-tradition guess, a narrator observed in such a corpus contributes
+*both* traditions: a ``muhaddithat``-only narrator therefore derives ``neutral``
+(both), not ``sunni`` and not ``unknown``.
+
 ``normalize_corpus`` reconciles the **bio provenance inconsistency** #103 flags:
 ``narrators_bio_*`` rows carry a free-text ``source`` that is not always a
 ``SourceCorpus`` value (e.g. the sanadset narrator bios use
@@ -40,6 +48,7 @@ __all__ = [
     "derive_sect_affiliation",
     "normalize_corpus",
     "primary_corpus",
+    "CROSS_TRADITION_CORPORA",
     "SECT_BY_CORPUS",
 ]
 
@@ -51,15 +60,21 @@ _CORPUS_ALIASES: dict[str, str] = {
 }
 
 # Corpus → collection-level tradition. Unmapped corpora (``fawaz`` per-collection,
-# ``itqan`` mixed rijal DB) contribute nothing → ``unknown`` when nothing maps.
+# ``itqan`` mixed rijal DB) contribute nothing → ``unknown`` when nothing maps;
+# ``muhaddithat`` is handled separately as a cross-tradition corpus (see below).
 SECT_BY_CORPUS: dict[str, Sect] = {
     SourceCorpus.SUNNAH.value: Sect.SUNNI,
     SourceCorpus.OPEN_HADITH.value: Sect.SUNNI,
     SourceCorpus.LK.value: Sect.SUNNI,
     SourceCorpus.SANADSET.value: Sect.SUNNI,
-    SourceCorpus.MUHADDITHAT.value: Sect.SUNNI,
     SourceCorpus.THAQALAYN.value: Sect.SHIA,
 }
+
+# Cross-tradition corpora: no per-narrator sect signal in the source and the
+# corpus spans both traditions, so an observation contributes *both* Sunni and
+# Shia. A narrator seen only here aggregates to ``neutral`` (both) — never a
+# one-sided ``sunni``/``shia`` guess, never ``unknown`` (da#90, epic #81).
+CROSS_TRADITION_CORPORA: frozenset[str] = frozenset({SourceCorpus.MUHADDITHAT.value})
 
 
 def normalize_corpus(source: str | None) -> str | None:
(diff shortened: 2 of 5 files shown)


parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved

## Primary review — da#90 PR #135 (muhaddithat end-to-end light-up)

Verified by diff + files API at the branch head plus CI; did not run the local suite.

### (a) Cross-tradition sect fix — correct
`src/resolve/sect_affiliation.py` drops the wrong `MUHADDITHAT -> SUNNI` mapping from `SECT_BY_CORPUS` and routes muhaddithat through a new `CROSS_TRADITION_CORPORA` frozenset. In `derive_sect_affiliation`, a cross-tradition corpus contributes BOTH `Sect.SUNNI` and `Sect.SHIA`, so a muhaddithat-only narrator aggregates to a two-element set and resolves to `neutral` (both) — never a one-sided guess, never `unknown` (the `if not sects` unknown branch is unreachable for this corpus). The added parametrize cases confirm `["muhaddithat"] -> neutral`, and that mixing with `sunnah`/`thaqalayn` stays `neutral`. This aligns the narrator-level derivation with the adapter registry already declaring muhaddithat sect=None, which matches the both-sects acceptance criterion.

### (b) Arabic-name edge-id fix — genuinely reconciles
I traced both endpoints to the SAME identity function. bio_promote mints `cid = make_canonical_id(norm)` where `norm = name_ar_normalized or normalize_arabic(name_ar)`. The loader `_studied_under_endpoint(name, ...)` returns `make_canonical_id(normalize_arabic(name))`. The parser change sets `edge_name = name_ar or display_name` into `id_to_name`, and `_parse_network_edges` writes that into `from_narrator_name`/`to_narrator_name`. So both sides now compute `make_canonical_id(normalize_arabic(<Arabic name>))` and match exactly. Before the fix the edge map carried the Latin display name, minting a different id for every endpoint — hence 0 edges. The display-name fallback only triggers when no Arabic name exists, in which case bio_promote also skips that bio (no canonical node), so the fallback never produces a mismatched-but-present node.

### (c) Live neo4j:5 e2e is genuine
`tests/integration/test_muhaddithat_lightup.py` seeds raw CSVs and drives the real `muhaddithat.run -> run_all -> load_all_nodes + _load_studied_under` path with no hand-injected ids. It asserts nodes land with `source_corpus=muhaddithat` and `sect_affiliation=neutral`, that STUDIED_UNDER `created > 0` AND `missing_endpoints == 0` (was 0 created before), graph edge count equals `result.created`, and that reload is idempotent on both nodes and edges. The fixture mirrors real upstream shape (separate `displayname` + `arabicname`), which is exactly what exposes the id-reconciliation regression rather than papering over it. The loader is source-agnostic via `network_edges_*` glob, matching the parser output filename. CI Integration Tests job is green (52s runtime indicates the container actually spun up, not a skip).

### (d) bio_promote test correctly revised, not deleted
`tests/test_resolve/test_bio_promote.py::test_bio_promote_tags_sect_and_corpus` keeps all its assertions and flips the single sect expectation from `sunni` to `neutral` with an explanatory comment. The test still proves corpus/source tagging; only the now-correct sect value changed. No test was removed to pass.

### CI
All 5 checks green — Integration Tests, Lint, sync-drift gate, Test, Type Check.

Approved. Staging-cluster load remains a post-merge Test Plan item per the #90 acceptance criteria, not a PR gate.

TechDebt: none

parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved

Reviewed as owner of `src/resolve/sect_affiliation.py` (#103), with focus on the cross-tradition sect change and the Arabic-name edge reconciliation. Verified via diff, files API at PR head, and CI. CI is fully green (Integration, Lint, sync-drift gate, Test, Type Check).

## Sect-derivation lens

The `CROSS_TRADITION_CORPORA` split is correct and leaves my existing cases intact. `derive_sect_affiliation` still aggregates the same way at the tail (empty to `unknown`, `{SUNNI}` to `sunni`, `{SHIA}` to `shia`, otherwise `neutral`), so the four contract cases are preserved — only-Sunni corpora stay `sunni`, only-Shia (`thaqalayn`) stays `shia`, a Sunni+Shia mix stays `neutral`, and the empty case stays `unknown`. Muhaddithat is removed from `SECT_BY_CORPUS` and now contributes both traditions through the new `elif`/`if` branch, so a muhaddithat-only narrator aggregates to `neutral` rather than the pre-#90 one-sided guess. The two maps are disjoint, the `elif` prevents any double-handling, and `CROSS_TRADITION_CORPORA` is exported in `__all__`. `normalize_corpus` resolves the corpus to `SourceCorpus.MUHADDITHAT.value`, which lands in the cross-tradition set — coherent.

Multi-source interaction is the one place I looked hardest. A narrator observed in muhaddithat AND a single-sect source resolves to `{SUNNI, SHIA}` union that one sect, which is still `{SUNNI, SHIA}` and therefore `neutral`. That is a deliberate behavioral change (muhaddithat + a Sunni corpus was `sunni` before, now `neutral`), and I am satisfied it is the right call — it is consistent with the existing `sunnah + thaqalayn -> neutral` semantic, and muhaddithat genuinely attests cross-tradition recognition, so a narrator touching it is correctly cross-tradition. Worth keeping in mind that muhaddithat is now "dominant" in the sense that it always pulls a narrator to `neutral` and can never be overridden by a single-sect source — intentional and well-documented (da#90, epic #81), not a defect.

This is consistent with the source carrying no per-narrator sect column; removing it from the de-facto sect registry and routing it through the cross-tradition set is the right modeling of that.

## Arabic-name edge fix

Sound, and the id reconciliation holds end to end. The edge map now stores the raw Arabic name (`edge_name = name_ar or display_name`). The STUDIED_UNDER loader resolves each endpoint via `make_canonical_id(normalize_arabic(name))`, and bio_promote mints the node id via `make_canonical_id(name_ar_normalized)` where `name_ar_normalized = normalize_arabic(arabic_name)`. Both sides therefore collapse to `make_canonical_id(normalize_arabic(arabic_name))` — identical keys, so endpoints reconcile. Pre-fix the Latin display name was run through `normalize_arabic` and minted a divergent id, which is exactly why every endpoint missed and zero edges landed. The `name_ar or display_name` fallback is a reasonable best-effort for the no-Arabic-name edge case; muhaddithat upstream always carries Arabic names, so it does not bite here.

## Test updates

Correct revisions. The `test_sect_affiliation` table adds `(["muhaddithat"], "neutral")` and `(["muhaddithat", "sunnah"], "neutral")`, both matching the logic; the pre-existing `(["muhaddithat", "thaqalayn"], "neutral")` row is unchanged and still right. `test_bio_promote` flips the muhaddithat-only assertion from `sunni` to `neutral`, matching the new derivation. The new live integration test drives the real parse to resolve to load path with no hand-injected canonical id, asserts the `neutral` sect plus the `muhaddithat` provenance tag, asserts edges fire with zero missing endpoints, and checks reload idempotency — so an id-reconciliation regression re-breaks it rather than being papered over. It degrades to a clean skip off-Docker.

No blockers. The cross-tradition change is safe for multi-source narrators.

Issue #118, opened 2026-06-12
Fuzzy cross-source narrator clustering (recall increment on #99 exact-name pass)


parametrization:
## Fuzzy cross-source narrator clustering (recall increment on the exact-name pass)

Follow-up to **#99** (X1 cross-source narrator-identity resolution). #99 collapses
the same narrator across sources by **exact normalized-name** identity — high
precision, but it **under-merges** true variants that survive Arabic
normalization differently (kunya-only forms, nasab/nisba expansions, transliteration
drift, optional "ابن/بن", honorifics). Those land as separate `nar:` nodes for one
real person.

The `src/resolve/quality.py` pairwise precision/recall metric added in #99 is built
to quantify exactly this: the exact-name pass scores high precision but leaves
**recall** on the table. This issue is the recall increment.

### Scope
A fuzzy / blocked clustering pass over canonical narrators (or over mentions
pre-canonicalization) that merges high-confidence name variants of the same person
across sources, on top of — not replacing — the high-precision exact-name pass.
Candidate signals: rapidfuzz token-set ratio within a blocking key, plus the
temporal (death-year) and chain-position context the disambiguator already
computes; optionally consume **#94** (Itqan name-variant → alias data, 217,762
variants) as a variant dictionary to drive merges.

### Acceptance
- Measured precision/recall (via `quality.pairwise_quality`) on a labelled
  multi-source sample shows a **recall gain** over the exact-name-only baseline at
  a stated precision floor.
- Live `neo4j:5`: a known same-person variant pair across ≥2 sources collapses to
  one `Narrator` node.

### Notes / dependencies
- The single identity rule is `identity.make_canonical_id` (da#110) — any merge
  must route through it, never a parallel id scheme.
- **Pipeline ordering**: when this lands alongside `bio_promote`, the
  disambiguate-before-bio_promote ordering invariant (disambiguate overwrites
  `narrators_canonical`, bio_promote merges) is tracked in **#117** — coordinate
  there.
- Consumes variant data from **#94**.

Part of epic #81. Recall increment on #99.

parametrization:
Merged via PR #137 (fuzzy cross-source narrator clustering; over-merge guards: ≥2 shared significant tokens + clique re-partition; runtime precision validation tracked in da#138).

Issue #99, opened 2026-06-12
X1: cross-source narrator-identity resolution (NER + disambiguate + dedup)


parametrization:
**Tier:** T3 (cross-source normalization — "the heart") · **Suggested owner:** Ivana Horvat (ML/NLP)
Part of epic #81. **KEYSTONE — isnad-graph#963 (narrator search).**

## Goal
Run **cross-source narrator-identity resolution** — NER + disambiguate + dedup (`src/resolve/{ner,disambiguate,dedup}.py`) across multiple loaded sources so the same narrator from Sunnah, Sanadset, Itqan, Thaqalayn collapses to **one** `Narrator` node. The resolve code exists but has never run at multi-source scale.

## Acceptance
- [ ] Resolution runs across ≥2 loaded sources; duplicate narrators collapse to a single node.
- [ ] Measured dedup quality (precision/recall on a labelled sample) reported.
- [ ] Loads into local `neo4j:5`; asserts collapsed narrator count < naive union.

## Dependencies
≥2 of **L2 / L6 / N1a** loaded; consumes **#N1c** (Itqan name variants).

parametrization:
Done via #116 (cross-source narrator-identity resolution: bio-less exact-name fallback, shared nar:uuid5 identity, 3-source live-neo4j convergence test incl. Itqan; disambiguate-before-bio_promote ordering locked), merged to deployments/phase-4/wave-3. 2 approvals (Jean-Claude + Ivana), CI 5/5 green. Follow-up da#120 wires bio_promote into run_all.

Issue #94, opened 2026-06-12
N1c: Itqan adapter — name variants (217,762) → identity aliases


parametrization:
**Tier:** T2 (new source — Itqan, PR 3 of 3) · **Suggested owner:** Ivana Horvat (ML/NLP)
Part of epic #81. **Feeds cross-source narrator resolution (#X1).**

## Goal
**Itqan** adapter — **name variants**: 217,762 name variants → narrator-identity aliases feeding the resolve/disambiguation stage so the same narrator across sources collapses to one node.

## Dependency
**#N1a** (Itqan narrators). Directly enables **#X1**.

## Acceptance
- [ ] New `src/acquire/<src>.py` + `src/parse/<src>.py` producing canonical staging Parquet with `sect` + `source_corpus`.
- [ ] `SourceCorpus` enum value added (in `src/models/enums.py`).
- [ ] Parse + acquire-side tests.
- [ ] End-to-end load into a **local `neo4j:5` container** asserting real nodes/edges (local Docker confirmed working).
- [ ] Reachability + licensing note.

## Dependencies
T0-A (source_id scheme), T0-B (conformance gate).

parametrization:
Merged via PR #130 (Itqan name variants→aliases).

PR #137, opened 2026-06-12
feat(resolve): fuzzy cross-source narrator clustering — recall increment on #99 exact pass (#118)

parametrization:
Closes #118. Recall increment on the #99 exact-name cross-source pass.

## Problem
#99 collapses a narrator across sources only on a **byte-exact** normalized-name
match (`make_canonical_id(name_ar_normalized)`) — high precision, but it
under-merges true variants that survive Arabic normalization differently:
nasab/nisba expansion or truncation, kunya-only forms, an optional `ابن`/`بن`,
transliteration drift, honorifics. Those land as separate `nar:` nodes for one
real person — recall left on the table.

## Approach — `src/resolve/fuzzy_cluster.py`
A bio/canonical-level clustering pass (the analogue of `disambiguate`'s rapidfuzz
*mention* stage, but with both sides canonical narrators). It reads the
already-collapsed `narrators_canonical.parquet`, clusters high-confidence name
variants, and merges each cluster into one record. Runs **on top of**, never
replacing, the exact pass.

- **Signals:** rapidfuzz `token_set_ratio` over each record's `name_ar_normalized`
  **plus every `alias`** — the #94 Itqan name-variant feed (217,762 variants) is
  consumed as additional match keys, not just the canonical name. Token-set is
  order-/extra-token-insensitive, so `محمد بن اسماعيل` ↔ `محمد بن اسماعيل البخاري`
  scores high while a genuinely different name does not.
- **Blocking:** token inverted index (only records sharing a significant,
  non-connector name token are ever compared) → near-linear, not O(n²).

## Precision/recall guardrails (against over-merging distinct same-named narrators)
- Conservative **tunable threshold** (`_CLUSTER_RATIO_THRESHOLD = 90`) — the
  precision/recall knob the acceptance criteria call for.
- **Death-year guard:** when both records carry `death_year_ah` and they disagree
  by more than the tolerance, the merge is blocked even on a perfect name match
  (two scholars a century apart are not one person).
- **Gender guard:** two records with explicit, differing `gender` never merge.

## Identity & ordering (no re-architecture)
- Every merge routes through `identity.make_canonical_id` — never a parallel id
  scheme (da#110). The survivor is an existing canonical record, so its id is
  itself a prior `make_canonical_id` output: clustering re-keys nothing.
- Wired into `run_all` **after** `disambiguate → bio_promote` (da#117), as an
  additive step. It is idempotent (re-running on a clustered table is a no-op), so
  the disambiguate-overwrites/bio_promote-merges ordering invariant is untouched.
- When the mentions file is present, the absorbed-id → survivor remap is applied
  to `narrator_mentions_resolved.parquet` so graph NARRATED edges follow the merge
  instead of dangling on a dissolved node (#109).

## Test proof — `tests/test_resolve/test_fuzzy_cluster.py` (10 tests)
- **Recall gain:** `quality.pairwise_quality` recall rises over the exact-name
  baseline (`exact_q.recall < 1.0 < fuzzy_q.recall == 1.0`) at **precision = 1.0**.
- **Aliases (#94):** a variant matched only via an alias key clusters.
- **Precision guards:** conflicting death-year, conflicting gender, and genuinely
  distinct names that share a common token all stay split.
- **Mechanics:** survivor id routes through `make_canonical_id`; provenance
  unioned; absorbed spelling kept as an alias; mention_count summed; idempotency;
  mention remap; empty/missing no-ops.

## Verification
- `pytest tests/test_resolve/` → 152 passed, 3 skipped (pre-existing, semantic model).
- `ruff format --check` + `ruff check` clean; `mypy src/` → no issues (77 files).

## Test Plan (runtime, post-merge)
- **Live `neo4j:5`:** run the resolve pipeline over ≥2 loaded sources; assert a
  known same-person variant pair across sources collapses to one `Narrator` node
  (canonical count drops by the merged count) and its NARRATED edges resolve to
  the surviving node.

Reviewers: @Kavitha Sundaramurthy (primary), @Oyunbileg Batbayar (QA).

### DIFF
--- a/src/resolve/__init__.py
+++ b/src/resolve/__init__.py
@@ -1,12 +1,17 @@
 """Phase 2: Entity Resolution pipeline.
 
 Dependency-aware orchestrator:
-``NER -> disambiguate -> bio_promote -> (dedup + detect_parallels)``.
+``NER -> disambiguate -> bio_promote -> fuzzy_cluster -> (dedup + detect_parallels)``.
 
 Ordering invariants (da#117):
 - ``disambiguate`` OVERWRITES ``narrators_canonical.parquet`` while
   ``bio_promote`` MERGEs into it, so ``bio_promote`` MUST run *after*
   ``disambiguate`` or promoted bio-only narrators get clobbered (da#99).
+- ``fuzzy_cluster`` (da#118) runs *after* both: it clusters cross-source name
+  variants the exact-name pass under-merged, operating on the canonical set
+  those two produce. It re-keys nothing (merges route through
+  ``make_canonical_id``) and is idempotent, so adding it never perturbs the
+  disambiguate→bio_promote invariant above.
 - ``dedup`` (semantic embeddings, degrades to empty without the model) and
   ``detect_parallels`` (deterministic lexical, offline/CI) both write the shared
   ``staging/parallel_links.parquet``. run_all runs both and *composes* their
@@ -241,12 +246,13 @@ def run_all(raw_dir: Path, staging_dir: Path, output_dir: Path) -> dict[str, lis
     """
     logger.info("resolve_pipeline_start")
 
-    from src.resolve import bio_promote, dedup, disambiguate, ner, parallels
+    from src.resolve import bio_promote, dedup, disambiguate, fuzzy_cluster, ner, parallels
 
     results: dict[str, list[Path]] = {
         "ner": [],
         "disambiguate": [],
         "bio_promote": [],
+        "cluster": [],
         "dedup": [],
         "parallels": [],
     }
@@ -314,6 +320,31 @@ def run_all(raw_dir: Path, staging_dir: Path, output_dir: Path) -> dict[str, lis
         )
     except Exception:  # noqa: BLE001
         logger.error("resolve_step_failed", step="bio_promote", traceback=traceback.format_exc())
+
+    # Step 3.5: Fuzzy cross-source clustering (da#118) — recall increment on the
+    # exact-name pass. Merges high-confidence cross-source name variants the
+    # byte-exact collapse left split, operating ON the canonical set the prior two
+    # steps produced. Routes every merge through make_canonical_id and is
+    # idempotent, so it never perturbs the disambiguate→bio_promote ordering. The
+    # mentions file is remapped so graph NARRATED edges follow the merge (#109).
+    try:
+        logger.info("resolve_step", step="cluster", status="running")
+        canonical_path = output_dir / "narrators_canonical.parquet"
+        mentions_path = output_dir / "narrator_mentions_resolved.parquet"
+        cluster_metrics = fuzzy_cluster.cluster_canonical_narrators(
+            canonical_path,
+            mentions_path=mentions_path if mentions_path.exists() else None,
+        )
+        results["cluster"] = [canonical_path] if cluster_metrics.merged_records else []
+        logger.info(
+            "resolve_step",
+            step="cluster",
+            status="complete",
+            merged=cluster_metrics.merged_records,
+            clusters=cluster_metrics.multi_member_clusters,
+        )
+    except Exception:  # noqa: BLE001
+        logger.error("resolve_step_failed", step="cluster", traceback=traceback.format_exc())
 
     # Step 4: Dedup (semantic; degrades to an empty table without the embedding
     # model). Runs independently of NER/disambiguation. Capture its output before
--- a/src/resolve/fuzzy_cluster.py
+++ b/src/resolve/fuzzy_cluster.py
@@ -1,0 +1,510 @@
+"""Fuzzy cross-source narrator clustering — recall increment on the exact-name pass.
+
+The exact-name cross-source collapse (da#99, in ``disambiguate``/``bio_promote``)
+keys every canonical narrator on ``make_canonical_id(name_ar_normalized)``: two
+records merge **only** when their normalized names are byte-identical. That is
+high precision, but it leaves *recall* on the table — the same real person whose
+spelling survives Arabic normalization differently across sources (kunya-only
+forms, nasab/nisba expansion or truncation, an optional ``ابن``/``بن``,
+transliteration drift, honorifics) lands as two separate ``nar:`` nodes.
+
+This module is that recall increment. It runs **on top of** — never replacing —
+the exact-name pass: it reads the already-collapsed ``narrators_canonical.parquet``
+and clusters records that are high-confidence name variants of one person, then
+merges each cluster into a single canonical record. It is the bio/canonical-level
+analogue of the rapidfuzz mention stage in ``disambiguate`` (which matches
+*mentions* to *bios*); here both sides are canonical narrators.
+
+Signals (issue da#118)
+----------------------
+* **rapidfuzz ``token_set_ratio``** over each record's match keys — its
+  ``name_ar_normalized`` **plus** every ``alias`` (the da#94 Itqan name-variant
+  feed): token-set is order- and extra-token-insensitive, so a nasab expansion
+  (``محمد بن اسماعيل`` ↔ ``محمد بن اسماعيل البخاري``) scores high while a genuinely
+  different name does not.
+* **Token blocking** — only records sharing a significant (non-connector) name
+  token are ever compared, keeping the pass near-linear instead of O(n²).
+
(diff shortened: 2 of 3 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Ivana Horvat
RequestOrReplied: Approved

## Update — re-reviewed at ee427a6, both vectors closed, flipping to Approved
Ivana addressed both over-merge vectors inline (option a, `fuzzy_cluster` stays default-on). I re-ran the actual module at the PR head to confirm — not just trusting the new test assertions. Both are genuinely closed and recall is preserved. CI is fully green at ee427a6 (Test, Integration, Lint, Type Check, sync-drift). Runtime real-data precision validation is tracked as da#138. The original blocking detail is retained below as the record.

Blocker 1 (bare single-token subset) — `_can_merge` now requires `_MIN_SHARED_TOKENS = 2` shared significant (non-connector) tokens. Verified `_can_merge(محمد, محمد بن اسماعيل البخاري)` returns False (shares only the one token محمد) and the pair stays split, while a genuine variant sharing the nasab/nisba stem (احمد بن حنبل ↔ احمد بن حنبل الشيباني, 2 shared tokens) still merges — the guard did not cost legitimate recall.

Blocker 2 (transitive bridge) — new `_safe_partition` re-partitions each union-find group into `_can_merge`-cliques, and because membership requires compatibility with *every* existing member of a sub-cluster (the `all(...)` check), no guard-conflicting pair can co-occur in any sub-cluster. I drove the A(d256)–B(bridge, no death)–C(d320) case through all six processing permutations; A and C land in separate clusters in every order (sizes [1,2] each time), so the property is genuinely order-independent, not an artifact of one input order. The death-year/gender guards are now load-bearing across the whole cluster, not just adjacent pairs.

Minor non-blocker (multi-cluster global idempotency via accumulated alias keys) stands as a follow-up note; not gating.

---

_Original review (ChangesRequested at the prior head), retained as the record:_

QA / data-quality secondary review. CI is fully green (Test, Integration, Lint, Type Check, sync-drift all pass), the identity contract is respected (merge routes through `make_canonical_id`), mention remap is correct, idempotency holds for the simple case, and the synthetic-fixture precision caveat is honestly flagged in the PR body. The approach is sound and additive. I am blocking on one data-quality class — over-merge precision — because the pass is wired default-on into `run_all`, so it will run against the real graph, and two foreseeable over-merge vectors are statically identifiable (not "needs real data to observe") yet uncovered by the tests and unguarded by the code.

## Blocker 1 — `token_set_ratio` scores a subset at 100, so a bare common given-name absorbs any fuller name
`token_set_ratio` returns 100 whenever one record's token set is a subset of the other's (its designed behavior). I confirmed against the pinned rapidfuzz that `token_set_ratio("محمد", "محمد بن اسماعيل البخاري")` is 100.0. A canonical record carrying only a bare common given name (محمد / احمد / علي / عبد الله — entirely plausible from truncated or sparse mention-derived records) is a subset of every fuller name sharing that token, so it merges at 100. The two pairwise guards do not save this case — sparse bare-name records are exactly the ones with no `death_year_ah` and no explicit `gender`, so `_death_years_conflict` and `_genders_conflict` both return False and the only gate left is the name score, which is 100. `test_distinct_names_do_not_cluster` does NOT exercise this — سفيان الثوري vs سفيان بن عيينة each have a distinct non-shared token (neither is a subset of the other → 84.4, below threshold), so the test passes while the dominant Arabic over-merge mode goes untested.

## Blocker 2 — transitive union through a sparse bridge bypasses the pairwise guards entirely
Clustering is union-find (transitive), but `_can_merge` is evaluated per pair. A sparse bridge record B that is a subset of two well-attested but distinct narrators A and C unions all three even when the A–C pair is independently rejected. I confirmed empirically — for A=محمد بن اسماعيل البخاري, B=محمد بن اسماعيل, C=محمد بن اسماعيل الكوفي, the A–C edge is 84.4 (rejected, and would also hit a death-year guard if both carried years), but A–B is 100 and B–C is 100, so A and C land in one cluster through B. The death-year/gender guards give false confidence here — they fire on the A–C pair, but that pair never drives the union, and union-find never un-unions. So two distinct narrators with conflicting death years CAN be merged through a bridge, which is precisely the identity-corruption the guards are advertised to prevent. No test covers a 3-record transitive bridge with conflicting endpoints.

## Why this blocks rather than defers to the post-merge runtime Test Plan
The live-neo4j variant-pair-collapse check is the right runtime gate for recall, and I am fine deferring recall validation. But over-merge is identity corruption (two real people collapse to one `Narrator` node), it is hard to detect after the fact, and the synthetic precision=1.0 result actively masks both vectors above — the fixtures contain no bare-common-name subset and no transitive bridge. With the owner's data-first priority, real data must be correct, not merely present, before this feeds the production graph.

## Acceptable resolutions (either is fine)
- Add a guard plus tests — e.g. require a minimum count of shared significant (non-connector) tokens on the shorter side rather than accepting a pure single-token subset, and re-validate every member pair against the death-year/gender guards after union (or split a cluster that contains a guard-conflicting pair). Cover the bare-common-name subset case and the 3-record transitive-bridge-with-conflicting-endpoints case. OR
- Gate `fuzzy_cluster` behind a flag that defaults OFF in `run_all` until the post-merge live precision run validates it, so it cannot silently corrupt the graph before that check.

## Minor (non-blocking, note for follow-up)
Idempotency is asserted only for an isolated cluster. Because the survivor accumulates the absorbed spelling as an alias and aliases are match keys, a second pass over a larger table could bridge a third record that only matched the absorbed spelling — so the pass may not be globally run-once/run-twice stable. Worth a test on a multi-cluster table; not a blocker.

Happy to re-review promptly on either resolution.

parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved

Reviewed via diff plus CI at head f1576c5 (all five checks green — Lint, Type, Test, Integration, sync-drift gate). I verified the implementation against source at the PR head rather than the PR body, and confirmed each acceptance dimension.

## Runs on top of the exact pass, ordering invariant untouched
fuzzy_cluster is inserted at Step 3.5, after bio_promote (Step 3) and before dedup (Step 4). It never replaces disambiguate or bio_promote — it reads the canonical table those two already produced and clusters within it. The da#117 invariant (disambiguate overwrites, bio_promote merges, so bio_promote runs second) is physically unchanged. The step is additive and idempotent — the idempotency test confirms a second run merges zero records, and re-keying is a no-op because every survivor id is already a make_canonical_id output.

## Identity contract honored, provenance correctly combined
Survivor id routes through make_canonical_id(representative name_ar_normalized) — no parallel id scheme (da#110), with a safe fallback to the existing id when the normalized name is empty. Because the representative is itself a prior canonical record, survivor id equals its existing id, so mentions already on the representative are left untouched and only absorbed ids remap. On merge, source_ids and source_corpora are unioned (corpora normalized via normalize_corpus), mention_count is summed, absorbed spellings plus all members' aliases become aliases of the survivor, and source_corpus plus sect_affiliation are recomputed from the unioned corpora. I checked the full NARRATORS_CANONICAL_SCHEMA against _merge_cluster — all sixteen fields are carried, so no field is silently dropped even for singleton clusters.

## Blocking is correct and near-linear, not dropping pairs
Token-inverted-index blocking compares only records sharing a significant (non-connector, length two-plus) token, scored once per pair via dedup. Match keys include da#94 aliases, and _significant_tokens draws blocking tokens from those same keys, so an alias-bridged variant pair still shares a block token — the alias-driven-merge test exercises exactly this. Connector or honorific particles are excluded as blocking keys but still scored, so a block does not balloon on shared "bin". Precision guards (token_set_ratio threshold ninety, death-year tolerance, gender conflict) are layered and independently tested.

## Mentions remap is sound and guarded
_remap_mention_canonical_ids mirrors disambiguate._backfill_mention_canonical_ids faithfully — same streaming row-group rewrite, same temp-then-replace, same field-index lookup — so it is bounded-memory and idempotent. It is double-guarded: run_all passes mentions_path only when the file exists, and the function itself early-returns on an empty remap or a missing file. Absorbed-to-survivor remapping keeps NARRATED edges from dangling on a dissolved node (#109), confirmed by the remap test.

## Wiring
Correct point in run_all, reading and writing the same output_dir paths disambiguate and bio_promote use (the curated dir the graph loader reads). Failure is contained in the same try/except logging pattern as the sibling steps.

No blockers. Clean, well-documented, thoroughly tested. Approved.

PR #142, opened 2026-06-13
fix(graph): route network edges by declared relation, not filename (da#133)

### LINKED ISSUE #133: tech-debt(graph): NETWORK_EDGE_SCHEMA needs an edge-relation field so the loader routes STUDIED_UNDER vs TRANSMITTED_TO
**Surfaced during P4W4 review of #130 (Itqan) + #132 (MIS), architect call by Jean-Claude.**

`load_edges._load_studied_under` globs `network_edges_*.parquet` and loads EVERY row as **STUDIED_UNDER**, with no source/relation filter — NETWORK_EDGE_SCHEMA has no relation/type field. That is correct for the current producers (muhaddithat + itqan = teacher/student adjacency = STUDIED_UNDER), but **MIS (#132) emits `network_edges_mis.parquet` for isnad chains, which are semantically TRANSMITTED_TO, not STUDIED_UNDER.**

## Latent mislabel
Once a non-muhaddithat/non-itqan `network_edges_*.parquet` (e.g. MIS's) is present in a staging dir that `load_all_edges` runs over, its isnad edges load as STUDIED_UNDER — wrong relationship type in the graph.

## Why not gating now
- At #130 merge, only muhaddithat + itqan produce network_edges (both correctly STUDIED_UNDER).
- #132 (MIS) **defers** production graph-loading — it does NOT wire `network_edges_mis` into `load_all_edges`; its live E2E uses its own ISNAD_TRANSMITTED integration query. So the mislabel does not manifest at #132's current merge.

## Fix
Add an **edge-relation field to NETWORK_EDGE_SCHEMA** (default `STUDIED_UNDER` for muhaddithat/itqan back-compat) and route the loader by relation (STUDIED_UNDER vs TRANSMITTED_TO). 

## BLOCKER scope
This MUST land **before** MIS (or any isnad-chain source) is wired into the production `load_all_edges` path — it is a blocker on #132's eventual production-load step, NOT on #132's current deferred-load merge. Ref #130, #132.

parametrization:
## What & why

`load_edges._load_studied_under` globbed `network_edges_*.parquet` and routed files to **STUDIED_UNDER** by a hardcoded filename allowlist (`muhaddithat`, `itqan`). `NETWORK_EDGE_SCHEMA` had no relation/type field, so a non-allowlisted producer — notably **mis** (#132), whose rows are *isnad transmission* (`TRANSMITTED_TO`, opposite direction) — would have loaded as wrong-type, wrong-direction STUDIED_UNDER once wired into the production load path. This is the blocker the architect (Jean-Claude) carved out in #133.

## Change

- **`NETWORK_EDGE_SCHEMA`** gains an explicit, nullable **`relation`** field plus value constants (`EDGE_RELATION_STUDIED_UNDER`, `EDGE_RELATION_TRANSMITTED_TO`, `DEFAULT_EDGE_RELATION`).
- **Producers declare their relation**: `muhaddithat` + `itqan` → `STUDIED_UNDER`; `mis` → `TRANSMITTED_TO`.
- **Loader routes by the declared field** per-row, not by filename. The filename allowlist is demoted to a **back-compat default** consulted *only* for pre-#133 rows that carry no `relation` (legacy studentship file → STUDIED_UNDER; anything else → skipped). Once a row carries an explicit relation the allowlist is never consulted, so:
  - a new studentship source needs no allowlist edit, and
  - a transmission row can never be mislabeled even in a studentship-named file.
- Fixed the stale `muhaddithat._parse_network_edges` docstring that claimed TRANSMITTED_TO (it loads as STUDIED_UNDER).

## Tests

- `TestStudiedUnderRelationRouting`: non-allowlisted slug declaring STUDIED_UNDER **loads**; a `TRANSMITTED_TO` row in a `muhaddithat`-named file is **excluded**; a mixed-relation file splits correctly.
- Per-producer relation-stamping assertions (mis → TRANSMITTED_TO; muhaddithat/itqan → STUDIED_UNDER).
- Existing filename-fallback tests (incl. the original mis-exclusion proof) retained and green.

Full suite: **827 passed, 5 skipped**. `ruff format`/`ruff check`/`mypy src/` clean.

Closes #133

### DIFF
--- a/src/graph/load_edges.py
+++ b/src/graph/load_edges.py
@@ -22,6 +22,7 @@ from src.parse.identity import (
     make_canonical_id,
     narrator_node_id,
 )
+from src.parse.schemas import EDGE_RELATION_STUDIED_UNDER
 from src.utils.arabic import normalize_arabic
 from src.utils.logging import get_logger
 from src.utils.neo4j_client import Neo4jClient
@@ -485,11 +486,17 @@ def _load_parallel_of(
 # NETWORK_EDGE_SCHEMA is reused by producers whose edges mean DIFFERENT relations.
 # muhaddithat + itqan emit teacher↔student (studentship) pairs that ARE
 # STUDIED_UNDER. Others — e.g. `mis`, whose rows are *isnad transmission* pairs (a
-# different relation type AND the opposite direction) — must NOT be globbed in
-# here, or their edges would load as wrong-type, wrong-direction STUDIED_UNDER.
-# This filename allowlist is the cheap, safe interim; the durable fix is an
-# explicit edge-relation field on NETWORK_EDGE_SCHEMA so the loader keys on the
-# data, not on a slug — tracked in da#133.
+# different relation type AND the opposite direction) — must NOT load as
+# STUDIED_UNDER, or their edges would land as wrong-type, wrong-direction edges.
+#
+# The durable routing (da#133) is the per-row ``relation`` field on
+# NETWORK_EDGE_SCHEMA: a row is loaded here iff it DECLARES STUDIED_UNDER. The
+# filename allowlist below is no longer the primary gate — it survives only as the
+# back-compat inference for a pre-da#133 file whose rows carry no ``relation`` at
+# all (a legacy studentship file → STUDIED_UNDER; anything else → skipped). Once a
+# file carries an explicit ``relation`` the allowlist is never consulted, so a new
+# studentship source needs no allowlist edit and a transmission file named like a
+# studentship one (or vice-versa) is still routed correctly by its data.
 _STUDIED_UNDER_SOURCES: frozenset[str] = frozenset({"muhaddithat", "itqan"})
 
 
@@ -497,10 +504,25 @@ def _is_studied_under_file(path: Path) -> bool:
     """True if a ``network_edges_<slug>.parquet`` belongs to a studentship source.
 
     Matches the exact slug or a chunked ``<slug>_NNN`` variant, so a future
-    sharded write still resolves but a different corpus (``mis``) never does.
+    sharded write still resolves but a different corpus (``mis``) never does. Used
+    only as the back-compat default for relation-less (pre-da#133) rows.
     """
     slug = path.stem.removeprefix("network_edges_")
     return any(slug == src or slug.startswith(f"{src}_") for src in _STUDIED_UNDER_SOURCES)
+
+
+def _row_relation(row: dict[str, Any], *, legacy_default: str | None) -> str | None:
+    """Relation a NETWORK_EDGE row declares.
+
+    Honors an explicit ``relation`` value (da#133). A row carrying none — a
+    pre-da#133 parquet — falls back to *legacy_default*: the relation the file's
+    name implied under the old filename allowlist (``None`` when the name matched
+    no studentship source, so such legacy rows are skipped rather than mislabeled).
+    """
+    rel = row.get("relation")
+    if isinstance(rel, str) and rel.strip():
+        return rel
+    return legacy_default
 
 
 _STUDIED_UNDER_QUERY = """\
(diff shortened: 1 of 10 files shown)


parametrization:
## Architecture review — da#133 relation-keyed edge routing

Reviewed at head (`K.Sundaramurthy/0133-network-edge-relation-field`). This is the durable fix I carved out in #133, and it lands the contract cleanly.

**Schema & constants.** The new `relation` field on `NETWORK_EDGE_SCHEMA` is a nullable `pa.string()`, which is the right call — nullable is what lets a pre-da#133 parquet round-trip without a migration, and the docstring states that current producers always populate it. The three value constants live in `schemas.py` next to the schema, are exported in `__all__`, and `DEFAULT_EDGE_RELATION` aliases `EDGE_RELATION_STUDIED_UNDER` rather than re-spelling the literal. Typed, single-source-of-truth, no magic strings leaking into producers or loader.

**Producer stamping.** All three producers declare the correct relation at the row-construction site — itqan and muhaddithat emit `STUDIED_UNDER`, mis emits `TRANSMITTED_TO` — using the imported constants, not literals. The stale muhaddithat `_parse_network_edges` docstring (which wrongly claimed TRANSMITTED_TO) is corrected, and the mis docstring now explains why these rows must stay off the studentship loader. Good hygiene.

**Loader routing.** `_load_studied_under` now globs every `network_edges_*` file and filters per-row via `_row_relation`, keeping only rows that resolve to `STUDIED_UNDER`. The key architectural property holds: a row that declares `TRANSMITTED_TO` is excluded even inside a `muhaddithat`-named file, because the explicit field is consulted before the filename ever is. The filename allowlist is correctly demoted to a bounded back-compat inference — `_is_studied_under_file` feeds `legacy_default`, which is reached only when a row carries no relation, and yields `None` (skip) for any non-studentship slug. So a relation-less legacy mis-style file can never be silently promoted. The boundary is clearly commented and the back-compat path is the only place the slug set still matters.

**skipped semantics.** Non-`STUDIED_UNDER` rows are `continue`d without bumping `skipped`, with a comment that `skipped` tracks edges we meant to load but couldn't resolve, not edges of other relations. That's the correct distinction — a transmission row is not a failed studentship load.

**Tests.** `TestStudiedUnderRelationRouting` proves the three properties that matter: a non-allowlisted slug declaring `STUDIED_UNDER` loads (field beats filename), a `TRANSMITTED_TO` row in a `muhaddithat` file is excluded, and a mixed-relation file splits per-row. Per-producer stamping assertions cover mis/muhaddithat/itqan, and the original filename-fallback mis-exclusion test is retained, so the legacy path stays covered.

**Independent verification.** Built the head worktree and ran it myself: the 5 touched test modules pass (77 passed), `ruff check` on all changed sources clean, `mypy src/graph/load_edges.py src/parse/schemas.py` reports no issues. `Any` is imported in `load_edges.py` for the new `_row_relation` signature. No `--no-verify`.

Scope note (not a change request): this PR intentionally only teaches the STUDIED_UNDER loader to reject non-studentship rows; a dedicated TRANSMITTED_TO loader is out of scope until mis is wired into the production load path, which matches #133. The exclusion guarantee is what was needed now, and it is proven.

Nicely done — this is exactly the data-keyed routing the interim allowlist was standing in for.

---
RequestOrReplied: Approved

TechDebt: none

parametrization:
## Second review — pipeline-integration lens

Reviewed at head (`K.Sundaramurthy/0133-network-edge-relation-field`, all 10 files, CI fully green: Lint / Type Check / Test / Integration / sync-drift gate).

The change does exactly what da#133 carves out: the graph load now routes a NETWORK_EDGE by its declared per-row relation instead of a filename allowlist. I verified the end-to-end correctness from a pipeline standpoint:

- **Round-trip is real, not just schema-deep.** `relation` is a nullable string field on `NETWORK_EDGE_SCHEMA`; every current producer stamps it (`itqan`/`muhaddithat` → STUDIED_UNDER, `mis` → TRANSMITTED_TO), `_read_parquet_rows` returns it via `to_pylist()`, and `_load_studied_under` keys on it row-by-row through `_row_relation`. The test helper `_write_network_edges` builds arrays straight from `NETWORK_EDGE_SCHEMA`, so the parser-side and loader-side tests exercise the same column that production parquet carries.
- **No studentship regression.** The filename allowlist (`_is_studied_under_file`) is demoted to the legacy default consulted only for relation-less rows. A pre-da#133 file named for a studentship source still resolves to STUDIED_UNDER; a non-studentship legacy file still yields `None` and is skipped — identical to the old `edge_files` filter, and the original mis-exclusion proof is retained and green.
- **Back-compat default can't silently mislabel a future transmission source.** This was my main worry. It holds: once a row carries any explicit relation the allowlist is never consulted, and `mis` already declares TRANSMITTED_TO, so even a transmission row dropped into a `muhaddithat`-named file is excluded (`test_transmitted_to_relation_excluded_despite_studentship_filename`). The legacy default only ever fires when the column is entirely absent — i.e. genuinely old parquet, which by construction was studentship-only.
- **Mixed-relation files are covered** by `test_explicit_relation_routes_mixed_file` (a `mis`-named file with one STUDIED_UNDER and one TRANSMITTED_TO row splits correctly).
- **`skipped` metric stays clean** — other-relation rows `continue` without inflating the skip counter, which tracks unresolved studentship edges only. Good call; the inline comment documents it.
- **Docstring fix is in the correct direction.** `muhaddithat._parse_network_edges` previously claimed TRANSMITTED_TO while the allowlist already loaded it as STUDIED_UNDER; the PR makes the declared relation match the long-standing runtime behavior rather than changing it. No behavior change there.

Two forward-looking, non-blocking observations (do not gate this merge):

1. **`mis` TRANSMITTED_TO network edges have no consumer.** `_load_transmitted_to` sources from `narrator_mentions_resolved.parquet`, not `network_edges_*`. So once `mis` is wired into the load path its edges are correctly kept *off* STUDIED_UNDER but are not loaded as TRANSMITTED_TO anywhere — they're simply dropped by the studentship loader. That is the right scope for da#133 (prevent mislabeling) and matches `mis` being intentionally off the production load path today, but whoever wires `mis` (da#132) will need to add a network_edges→TRANSMITTED_TO load path; the declared field is the right hook for it.
2. **Merge order vs da#141.** da#141 is not an open PR yet, so there is no live conflict and #142 merges cleanly. Since da#133 is what gives the real-data load correct relation routing, land this first; if da#141 later opens touching the same `_load_studied_under` region, rebase it on top of #142.

Clean, well-tested, correctly scoped. No must-fix items.

---
Requestor: Dilara Erdogan
RequestOrReplied: Approved

TechDebt: none

Issue #144, opened 2026-06-13
bug(mis): adapter can't parse current Mendeley dataset — filenames drifted from CoreInfo/DetailsInfo parser tokens


parametrization:
Surfaced by Alejandra Reyes-Fuentes during da#141 (PR #143), out of that issue's scope (da#141 used muhaddithat + lk, not mis). The `mis` adapter can't parse its current Mendeley dataset version — the dataset filenames have drifted from the parser's expected `CoreInfo`/`DetailsInfo` tokens. This blocks mis (isnad-transmission) network-edge production. Not blocking #601 (the narrator graph lights up via muhaddithat+lk), but needed before mis TRANSMITTED_TO edges can flow. Provenance: da#141 caveat.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Kwesi Boateng
RequestOrReplied: Request

**Wave 1 Kickoff — Phase 5**

This issue is assigned to you for p5-wave-1.
- Peer reviewer: Tarek Mansour
- Secondary reviewer: Jean-Claude Habimana
- Branch from: `deployments/phase-5/wave-1`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: bug (per charter § Wave Planning & Priority)

Please begin implementation.

PR #149, opened 2026-06-14
fix(mis): parse current Mendeley v2 three-file graph export (#144)

parametrization:
Closes #144

## Problem
The upstream Multi-IsnadSet dataset (Mendeley DOI `10.17632/gzprcr93zn.2`) was restructured into a **three-file graph export**, breaking the `mis` adapter, which assumed the old two-file `CoreInfo` + `DetailsInfo_Sanad_Narrators` layout. The current file set is:

| file | role |
|------|------|
| `1_Hadith_SahihMuslim_HadithContent.xlsx` | core hadith rows |
| `2_Hadith_SahihMuslim_Narrators=Nodes for Graph.xlsx` | narrator node list (not consumed) |
| `3_Hadith_SahihMuslim_Isnad=Edges for Graph.xlsx` | explicit edge rows |

Three independent drift points each broke the parse:
1. **Core file unmatched** — `1_…HadithContent.xlsx` contains no `core` token, so `_find_file` raised `FileNotFoundError` before any parse.
2. **Wrong detail file** — the *Nodes* filename contains `narrator`, so the detail selector picked the node list instead of the *Edges* file.
3. **Unresolved edge columns** — columns are now `sourceNarratorNameEn`/`targetNarratorNameEn` (+ ID variants); the exact-match picker resolved neither, collapsing to the sequence layout and erroring.

## Fix
- Added `hadithcontent`/`content` core tokens and `edges`/`isnad` detail tokens **ahead of** `narrator`, so content + edges workbooks are selected and the node list is ignored.
- Added `source/target NarratorNameEn` name candidates, plus a new from/to external-id pair so the numeric narrator IDs flow into `{from,to}_external_id` for downstream disambiguation.
- Matn now reads from the vocalized `HadithText_Mushakkal` column.
- Updated both acquire + parse docstrings to the real current shape.

`mis` stays an **isnad-transmission** producer: every edge declares `TRANSMITTED_TO`, keeping it off the `STUDIED_UNDER` allowlist (da#133). `SanadNo` discriminates the parallel asanid, so multi-isnad multiplicity is preserved row-for-row. The legacy two-file layout still parses (tokens/candidates retained, its tests still green).

## Test Plan
- **Live trace** against the real downloaded v2 dataset: **7748 hadiths, 63642 edges, 3586 multi-isnad hadiths** (max 24 asanid/hadith); `source_corpus=mis`, `sect=sunni`, `relation` all `TRANSMITTED_TO`, every edge `hadith_id` joins to a hadith node, all `from/to_external_id` populated. Sample `source_id`: `mis:sahih_muslim:2:1`.
- **Unit tests** (`tests/test_parse/test_mis_parser.py`, new `TestMisCurrentMendeleyShape`): fixtures built in the real three-file shape (including the `Nodes` decoy) assert detail resolves to the *Edges* file (not *Nodes*), schema/tagging, matn from `HadithText_Mushakkal`, `TRANSMITTED_TO`, external IDs, multi-sanad multiplicity, and edge→hadith join. The fixture columns/filenames were taken directly from the live dataset headers.
- `ruff format --check` + `ruff check` (v0.15.11) clean; `mypy` clean on changed files; full `tests/test_parse/` (247) + `tests/test_acquire/test_mis.py` + `tests/integration/test_mis_lightup.py` pass.

### DIFF
--- a/src/acquire/mis.py
+++ b/src/acquire/mis.py
@@ -8,19 +8,24 @@ multi-isnad source for the platform: where most corpora give one chain per
 hadith, MIS preserves the parallel asanid (the ``ح`` / *tahwil* split) that make
 Sahih Muslim valuable for chain modelling.
 
-The dataset ships as two Excel workbooks under the Mendeley file set:
+The current dataset (v2) ships as **three** Excel workbooks under the Mendeley
+file set — a graph export (content + nodes + edges):
 
-* ``Hadith_SahihMuslim_CoreInfo.xlsx``               — one row per hadith
-  (book, hadith number, matn, narrator sequence, isnad count).
-* ``Hadith_SahihMuslim_DetailsInfo_Sanad_Narrators.xlsx`` — one row per
-  (hadith, sanad, narrator) position — the data the parser walks into per-sanad
-  chains.
+* ``1_Hadith_SahihMuslim_HadithContent.xlsx``         — one row per hadith
+  (``BookNo``, ``HadithNo``, matn, ``SanadCount``). The parser's "core" file.
+* ``2_Hadith_SahihMuslim_Narrators=Nodes for Graph.xlsx`` — narrator *node*
+  list (id ↔ name, generation, gender). Downloaded but not consumed for staging.
+* ``3_Hadith_SahihMuslim_Isnad=Edges for Graph.xlsx`` — one row per directed
+  transmission *edge* (``source*`` → ``target*`` narrator, keyed by
+  ``(BookNo, HadithNo, SanadNo)``). The parser's "detail" file.
 
-Reachability: ``data.mendeley.com`` serves the file set over HTTPS with **no
-auth / no API key** — the same keyless-public mechanism this repo already proves
-for ``sanadset`` (Mendeley ``5xth87zwb5``). The file ids are resolved at runtime
-from the dataset files API rather than hardcoded, so a new dataset version is
-picked up without a code change.
+We download every ``.xlsx`` in the file set, so all three arrive; the parser
+selects the content + edges workbooks by filename token. Reachability:
+``data.mendeley.com`` serves the file set over HTTPS with **no auth / no API
+key** — the same keyless-public mechanism this repo already proves for
+``sanadset`` (Mendeley ``5xth87zwb5``). The file ids are resolved at runtime from
+the dataset files API rather than hardcoded, so a new dataset version is picked
+up without a code change.
 
 Licensing: Mendeley Data's default license for this dataset is **CC BY 4.0**
 (attribution) — redistribution-safe with credit to the depositors.
--- a/src/parse/mis.py
+++ b/src/parse/mis.py
@@ -20,17 +20,28 @@ Two output files (both tagged ``sect=sunni`` / ``source_corpus=mis``):
 
 Input shape
 -----------
-Two Mendeley workbooks (``.xlsx``; ``.csv`` exports are also accepted for tests):
+Mendeley workbooks (``.xlsx``; ``.csv`` exports are also accepted for tests). The
+current dataset (DOI ``10.17632/gzprcr93zn.2``) ships a **three-file graph
+export** — the parser uses two of them:
 
-* **CoreInfo** — one row per hadith: a hadith number, an (optional) book number,
-  the matn text, and an (optional) narrator-sequence string.
-* **DetailsInfo (Sanad/Narrators)** — the per-position detail the chains are
-  walked from. Two layouts are supported:
-    - *sequence layout* (primary): ``(hadith, isnad/sanad, position, narrator)``
-      — grouped by ``(hadith, isnad)`` and walked into consecutive-pair edges, so
-      each sanad of a hadith becomes its own chain;
-    - *edge layout*: explicit ``(from_narrator, to_narrator[, hadith][, isnad])``
-      rows — each row is already one edge.
+* **HadithContent** (``1_Hadith_SahihMuslim_HadithContent.xlsx``) — one row per
+  hadith: ``HadithNo``, ``BookNo``, the matn (``HadithText_Mushakkal`` /
+  ``…_GhairMushakkal``), and per-hadith ``SanadCount``. This is the "core" file.
+* **Isnad=Edges** (``3_Hadith_SahihMuslim_Isnad=Edges for Graph.xlsx``, sheet
+  ``2-Relationship``) — one row per **edge**: a ``(BookNo, HadithNo, SanadNo)``
+  key plus ``sourceNarrator{ID,NameEn}`` → ``targetNarrator{ID,NameEn}``. Each
+  row is already one directed transmission edge; ``SanadNo`` discriminates the
+  parallel asanid, so the multi-isnad multiplicity survives row-for-row.
+* The third file (``2_…Narrators=Nodes for Graph.xlsx``) is a narrator *node*
+  list (id ↔ name) and is **not** consumed for staging — the edge file carries
+  the narrator names inline. The detail-file selector deliberately resolves to
+  the *Edges* workbook even though the *Nodes* filename also contains
+  "narrator".
+
+An older two-file layout (``CoreInfo`` + ``DetailsInfo_Sanad_Narrators`` in
+either a per-position *sequence* layout or an explicit *edge* layout) is still
+accepted — its filename tokens and column candidates are retained below so both
+the legacy and current dataset shapes parse.
 
 Column names are matched case/spacing/underscore-insensitively (the upstream
 spreadsheets are not perfectly normalized), with documented candidate sets.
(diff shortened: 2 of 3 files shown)


parametrization:
Requestor: Tarek Mansour
Requestee: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

Reviewed at confirmed head **`0e79a0029e70137958ef8da383bbc6c156a62ed8`** (`0e79a002`), base `deployments/phase-5/wave-1`. All 5 CI checks green at this head, **including Integration Tests** (`SUCCESS`, run 27484742209). Diff read from origin contents at the head SHA, not local.

All three drift points are correctly addressed and the load-bearing isnad-transmission invariant holds.

**(a) Core-file token additions — correct.** `_CORE_TOKENS` adds `hadithcontent` and `content` ahead of the legacy `core`. `1_…HadithContent.xlsx` resolves via `hadithcontent`; neither `content` nor `core` collides with the Nodes/Edges filenames (neither contains those substrings). Matn drift is also handled: `_MATN_KEYS` puts `hadithtextmushakkal` (vocalized) before `hadithtextghairmushakkal`, so `matn_ar` takes the vocalized column — asserted by `test_hadith_schema_matn_and_tagging`. `BookNo` (not `BookTitleEnAr`) flows into `source_id` because `_BOOK_KEYS` resolves `bookno` first.

**(b) Edges-over-Nodes decoy ordering — correct.** `_DETAIL_TOKENS = (isnadedges, edges, isnad, detailsinfo, sanad, narrator, details)` orders `edges`/`isnad` ahead of `narrator`. `_find_file` is token-priority ordered, so for the v2 set the `edges` token matches `3_…Isnad=Edges…` before `narrator` could ever match the `2_…Narrators=Nodes…` decoy. Confirmed the `=` in `Isnad=Edges` means the leading `isnadedges` token is a near-miss (no match) and `edges` is what actually wins — harmless, and the Nodes file shares none of the earlier tokens. `test_selects_content_and_edges_not_nodes` pins exactly this (`"Edges" in detail.name`, `"Nodes" not in detail.name`). Legacy `DetailsInfo_Sanad_Narrators` still resolves via `detailsinfo`/`sanad`. No core/detail cross-collision in either layout.

**(c) New edge columns — correct.** `_FROM_KEYS`/`_TO_KEYS` lead with `sourcenarratornameen`/`targetnarratornameen`, so the canonical `from`/`to` stays a NAME; the separate `_FROM_ID_KEYS`/`_TO_ID_KEYS` (`sourcenarratorid`/`targetnarratorid`) flow the stable numeric id into `{from,to}_external_id`. `test_edge_schema_relation_and_external_ids` verifies `from_external_id == "10"`, `to_external_id == "11"` for the A→B edge. Both-present `_FROM/_TO` keys correctly route to `_parse_edges_explicit`; the sequence layout's `Narrator` header matches no from/to key, so legacy data still routes to `_parse_edges_sequence`.

**mis stays an isnad-transmission producer (load-bearing).** `_emit_edge` sets `"relation": EDGE_RELATION_TRANSMITTED_TO` on every row. Verified in `schemas.py` at this head that `NETWORK_EDGE_SCHEMA` carries the `relation` field (da#133) and `EDGE_RELATION_TRANSMITTED_TO = "TRANSMITTED_TO"`. This is the producer-side declaration that keeps mis OFF the STUDIED_UNDER name-based allowlist once the relation-routing loader is wired (da#133 / memory `studied_under_network_edge_allowlist`). `test_edge_schema_relation_and_external_ids` and `test_edge_schema_and_tagging` both assert `relation == {"TRANSMITTED_TO"}` on the current AND legacy shapes. mis remains the *only* declared TRANSMITTED_TO producer here — it is never STUDIED_UNDER.

**Schema/builder soundness.** The 7 keys `_emit_edge` emits (`from_narrator_name`, `to_narrator_name`, `hadith_id`, `source`, `from_external_id`, `to_external_id`, `relation`) match the 7 `NETWORK_EDGE_SCHEMA` fields exactly, so the `{field.name: [e[field.name] …]}` table builder cannot KeyError. Multiplicity is preserved: `test_multiplicity_preserved_across_sanads` confirms the two sanads of hadith 1 survive as distinct chains (chain-specific `A→D`, `D→C` present), and edge `hadith_id`s join to real HADITH `source_id`s.

**No legacy regression.** `TestMisSequenceLayout` (2-file sequence) and `TestMisEdgeLayout` (2-file explicit-edge) are retained and pass against the same code paths, now additionally asserting the relation tag.

Approving — correct, well-tested, and the isnad-transmission invariant is intact.

## Throughline observations

Two patterns worth watching at the wave level — neither is a blocker here:

1. **da#133 relation-field adoption is a cross-producer migration.** This PR is one network-edge producer declaring `relation=TRANSMITTED_TO`. `DEFAULT_EDGE_RELATION` still falls back to `STUDIED_UNDER` for any pre-da#133 parquet that omits `relation`. If other edge-producers (e.g. isnad-transmission sources beyond mis) ship in this wave without setting `relation`, they will silently default to STUDIED_UNDER — the exact mis-routing failure da#133 exists to prevent. Worth a wave-level check that every isnad-transmission producer explicitly populates the field, and that the consuming loader actually routes by it rather than by the name allowlist.

2. **Upstream dataset-shape drift is a recurring adapter failure class** (this is the second Mendeley-shape break, after the CoreInfo/DetailsInfo era). The defenses here are good and reusable: runtime file-id/version resolution in `acquire` (no hardcoded URLs), tolerant alnum-normalized column matching with documented candidate sets, and token-priority file selection. If other acquire/parse adapters in the wave still hardcode file ids or exact column names, they carry the same latent break.

parametrization:
Requestee: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

Secondary review (architecture / schema-contract). Confirmed PR head **`0e79a0029e70137958ef8da383bbc6c156a62ed8`** (re-fetched twice from origin before and after the read). Reviewed every changed file at that ref via `gh api contents`, not local.

**Selector robustness — sound.** The detail-file token list `("isnadedges", "edges", "isnad", "detailsinfo", "sanad", "narrator", "details")` with `_find_file` iterating token-priority-first is the right mitigation for the Nodes/Edges filename collision: `2_…Narrators=Nodes…` and `3_…Isnad=Edges…` both contain "narrator"-ish tokens, but "edges"/"isnad" rank ahead of "narrator", so the current set resolves to the Edges workbook and the Nodes decoy loses (the new `test_selects_content_and_edges_not_nodes` proves it). It is filename-substring matching, so inherently brittle to a future upstream rename — but the author documents that explicitly, and the failure mode is benign: if a rename ever stripped the edges/isnad tokens, `_find_file` would fall through to "narrator" and grab the Nodes list, which has no `source*`/`target*` columns → routes to the sequence layout → raises `"No network edges extracted"`. That's a loud error, not silent corruption. Acceptable.

**Column disambiguation — clean.** `_pick` is exact normalized set-membership (not substring), so `_FROM_KEYS` ("sourcenarrator…") and `_FROM_ID_KEYS` ("sourcenarratorid") cannot cross-contaminate: the canonical `from`/`to` resolve to the *name* columns while the numeric narrator ids resolve separately into `{from,to}_external_id`. This is the schema-consistent split — human-readable name as canonical, stable numeric id for downstream disambiguation.

**Schema contract — held.** `_emit_edge` hardcodes `relation = EDGE_RELATION_TRANSMITTED_TO` for **both** the explicit and the sequence layouts, so MIS isnad-transmission edges are TRANSMITTED_TO and never STUDIED_UNDER — exactly the da#133 edge-relation discipline that keeps these off the studentship loader. Tests assert `table.schema == NETWORK_EDGE_SCHEMA` / `== HADITH_SCHEMA` and `relation == {"TRANSMITTED_TO"}`, and `test_edges_join_to_hadith_source_id` confirms edge `hadith_id ⊆ hadith source_id` (same corpus-namespaced grammar via the `id_map` join). No schema field is introduced by this PR — it consumes the existing `from_external_id`/`to_external_id`/`relation` columns.

**Multi-isnad multiplicity — preserved row-for-row.** Explicit layout emits one edge per row with no cross-sanad collapse; `seen_isnads[hadith_no].add(sanad_no)` only counts. `test_multiplicity_preserved_across_sanads` proves hadith 1's two parallel asanid keep their chain-specific A→D / D→C edges. The mechanism is unit-proven; the 3,586-multi-isnad / max-24-asanid figures are runtime-data claims not checkable in-sandbox, but the no-collapse mechanism is what produces them.

**No scope creep.** `src/acquire/mis.py` is a docstring-only change — behavior is unchanged: `_WANTED_SUFFIXES = (".xlsx", ".xls")` and the loop already downloads *every* matching workbook, so all three files arrive without a code change. parse + tests are the only behavioral surface.

Two non-blocking notes for a future touch (neither gates this PR):
1. The lead detail token `"isnadedges"` is effectively dead — the real filename is `Isnad=Edges` (with the `=`), so `"isnadedges" in name.lower()` never matches and `"edges"` is what actually selects file 3. Harmless, just slightly misleading as the priority-0 token.
2. Identical (from, to, hadith) transmission links shared by two parallel asanid are emitted once per sanad (row-for-row by design); de-dup is left to the loader's MERGE. Correct given the stated conservative-provenance intent — worth a one-line comment so it isn't read as an oversight later.

Approve.

## Throughline observations

This is a standalone adapter fix responding to upstream restructuring (Mendeley v2 two-file → three-file graph export). The one cross-cutting thread worth flagging: it threads the da#133 edge-relation field correctly (TRANSMITTED_TO), so as more isnad-transmission sources land this wave, the TRANSMITTED_TO-vs-STUDIED_UNDER typing discipline is the pattern to keep verifying per-adapter — the `network_edges_*.parquet` glob on the loader side allowlists by source, and a future adapter that forgets to set `relation` would silently inherit the wrong loader path. No wave-level defect pattern observed in this PR itself.

Issue #148, opened 2026-06-13
bug(graph-quality): integrity sweep — 86 null hadith_number_in_book, 51 hadith w/o NARRATED, 8 self-loops, 8 orphans, 15 lk-STUDIED_UNDER


parametrization:
## Summary

A sweep of the live #601 graph (isnad-graph#1018) surfaced several smaller node/edge integrity issues. Grouping them into one tracking issue; each is low-volume but real.

## Findings (live staging Neo4j, 2026-06-13)

| # | Finding | Count | Query |
|---|---------|-------|-------|
| 1 | `APPEARS_IN` edges with **null `hadith_number_in_book`** | 86 | `MATCH ()-[r:APPEARS_IN]->() WHERE r.hadith_number_in_book IS NULL RETURN count(r)` |
| 2 | `Hadith` nodes with **no `NARRATED` edge** (no narrator chain) | 51 | `MATCH (h:Hadith) WHERE NOT (h)<-[:NARRATED]-() AND NOT ()-[:NARRATED]->(h) RETURN count(h)` |
| 3 | `TRANSMITTED_TO` **self-loops** (narrator → self) | 8 | `MATCH (a:Narrator)-[r:TRANSMITTED_TO]->(a) RETURN count(r)` |
| 4 | **Orphan narrators** (no relationships at all) | 8 | `MATCH (n:Narrator) WHERE NOT (n)--() RETURN count(n)` — all `muhaddithat` bio-only (Abu Bakr, Fāṭimah, …) |
| 5 | `STUDIED_UNDER` edges involving the `lk` corpus | 15 | `11 lk→muhaddithat + 4 muhaddithat→lk` |

## Notes

- **#1** is the known APPEARS_IN-null-property class (cf. the `hadith_number_in_book` MERGE bug history); 86 edges slipped through with a null ordinal.
- **#4**: 8 famous companions exist as bio-only nodes (from the `muhaddithat` source) with zero isnad/hadith links — they won't appear in any traversal. Expected artifact of bio-promotion without mention-linking, but worth a decision (link or drop).
- **#5**: `STUDIED_UNDER` should carry *studentship* sources only; the `lk`-involved edges (15) may be isnad-transmission producers leaking into the studentship relation (cf. the STUDIED_UNDER allowlist / edge-relation-field concern, da#133). Verify these 15 are genuine studentship, not transmission.

## Suggested acceptance

[...]

PR #150, opened 2026-06-14
fix(graph): da#148 integrity sweep — TRANSMITTED_TO self-loops + normalized grade display

parametrization:
## Summary

Producer/load-layer fixes from the da#148 graph-integrity sweep. Scoped to the
items with a clean, testable producer fix so a re-load stays clean (not one-off
DB pokes). Broader, data-decision items are split into a tracked follow-up
below.

Addresses #148.

## What shipped

**Finding #3 — TRANSMITTED_TO self-loops (8 live `a→a`).**
`_build_chain_pairs` (the chain-derived TRANSMITTED_TO producer) now drops a
pair whose two **adjacent** mentions resolve to the same canonical narrator (a
repeated narrator in the raw chain, or two mentions the disambiguator collapsed
onto one node). This mirrors the guard the network-edge producers already apply
(`muhaddithat._parse_network_edges` skips `from_id == to_id`; `mis` skips
`from_name == to_name`) — the chain path was the one place missing it.

**Grade raw-Arabic → normalized display field.**
New `src/utils/grade.normalize_grade()` maps a raw grade string (Arabic
`صحيح`, transliterated `Da'if`, or descriptive `weak`) onto the `HadithGrade`
vocabulary. The **Grading** and **Hadith** node loaders now write a
`grade_normalized` property *alongside* the verbatim `grade`, so the UI and
Cypher filters get a stable key while the original is never lost. Unrecognised
input maps to `unknown` rather than guessing.

## Investigated — not a producer bug

**Finding #5 — 15 `lk`-involved STUDIED_UNDER edges.** The issue suspected an
isnad-transmission producer leaking into the studentship relation. There is **no
`lk` network-edges producer**: only `muhaddithat`/`itqan` emit STUDIED_UNDER
(they declare `relation=STUDIED_UNDER`) and `mis` emits TRANSMITTED_TO. The 15
edges are genuine `muhaddithat`/`itqan` studentship edges whose endpoints
**name-resolve** (`make_canonical_id(normalize_arabic(name))`) onto narrators
also present in the `lk` corpus — correct cross-corpus identity merge, not a
relation leak. No producer change needed; recorded here so the follow-up does
not re-investigate.

## Test Plan

- `tests/test_utils/test_grade.py` — `normalize_grade` table: Arabic (±
  diacritics) + English/transliterated forms, `li-ghayrihi` precedence over the
  bare grade, empty/None/garbage → `unknown`, result always a valid enum member.
- `tests/test_graph/test_load_edges.py::...::test_adjacent_duplicate_narrator_skipped`
  — adjacent duplicate narrator drops the self-loop, surviving pair intact, no
  `from==to` pairs emitted.
- `tests/test_graph/test_load_nodes.py::...::test_grading_carries_normalized_display_grade`
  — Grading node keeps raw `صحيح` and adds `grade_normalized == "sahih"`;
  `Da'if → daif`.
- Full suite green under the pre-push hook: `ENVIRONMENT=test uv run pytest
  tests/ -m "not integration"` → 808 passed, 5 skipped. ruff (check+format) and
  mypy-strict clean on changed files.

## Follow-up

Filing/needs a tracked issue (deferred — each needs a data decision or upstream
producer work, not a clean load-layer fix):

- **#1 — 86 null `hadith_number_in_book`.** The in-book ordinal is null for
  scraped sources that only carry the collection-wide ref (cf. the
  `hadith_number` collection-ref-vs-in-book-ordinal split). Needs per-scraper
  ordinal derivation where available + an explicit-null contract where genuinely
  unknown. Not safe to fabricate at the load layer.
- **#2 — 51 Hadith with no NARRATED.** Mention/NER coverage gap upstream of the
  load layer (no resolved chain → no position-0 narrator → no NARRATED). Needs
  producer-side mention extraction, not a load guard.
- **#4 — 8 orphan `muhaddithat` bio-only narrators** (Abu Bakr, Fāṭimah, …).
  Owner decision: link (mention-linking) or drop. Expected artifact of
  bio-promotion without mention-linking.
- **Grade normalization parity:** the ingest-platform streaming path
  (`workers/ingest`) does not yet emit `grade_normalized`; a cross-repo parity
  follow-up should mirror this so batch and stream converge.
- **Collection metadata gap** (blank Arabic name / riyadussalihin) — data
  enrichment, separate from these load-layer fixes.

### DIFF
--- a/src/graph/load_edges.py
+++ b/src/graph/load_edges.py
@@ -120,10 +120,27 @@ def _build_chain_pairs(
     resolved = [m for m in sorted_mentions if m.get("canonical_narrator_id")]
     pairs: list[dict[str, Any]] = []
     for i in range(len(resolved) - 1):
+        from_id = resolved[i]["canonical_narrator_id"]
+        to_id = resolved[i + 1]["canonical_narrator_id"]
+        # Drop self-loops: two ADJACENT mentions that resolve to the same
+        # canonical narrator (a repeated narrator in the raw chain, or two
+        # mentions the disambiguator collapsed onto one node) would otherwise
+        # MERGE a TRANSMITTED_TO edge from a narrator to itself (da#148 — 8 live
+        # self-loops). The network-edge producers already guard this
+        # (``muhaddithat._parse_network_edges`` skips ``from_id == to_id``,
+        # ``mis`` skips ``from_name == to_name``); applying it on the
+        # chain-derived path keeps a re-load self-loop-free everywhere.
+        if from_id == to_id:
+            logger.debug(
+                "transmitted_to_self_loop_skipped",
+                narrator_id=from_id,
+                hadith_id=resolved[i].get("hadith_id", ""),
+            )
+            continue
         pairs.append(
             {
-                "from_id": resolved[i]["canonical_narrator_id"],
-                "to_id": resolved[i + 1]["canonical_narrator_id"],
+                "from_id": from_id,
+                "to_id": to_id,
                 "position": resolved[i].get("position_in_chain", i),
                 "hadith_id": resolved[i].get("hadith_id", ""),
             }
--- a/src/graph/load_nodes.py
+++ b/src/graph/load_nodes.py
@@ -40,6 +40,7 @@ from src.parse.identity import (
     grading_node_id,
     hadith_node_id,
 )
+from src.utils.grade import normalize_grade
 from src.utils.logging import get_logger
 from src.utils.neo4j_client import Neo4jClient
 
@@ -202,6 +203,7 @@ SET n.matn_ar      = row.matn_ar,
     n.isnad_raw_ar = row.isnad_raw_ar,
     n.isnad_raw_en = row.isnad_raw_en,
     n.grade        = row.grade,
+    n.grade_normalized = row.grade_normalized,
     n.source_corpus = row.source_corpus,
     n.sect         = row.sect,
     n.collection_name = row.collection_name
@@ -247,6 +249,10 @@ def _load_hadiths(
                     "isnad_raw_ar": _val(row, "isnad_raw_ar"),
                     "isnad_raw_en": _val(row, "isnad_raw_en"),
                     "grade": _val(row, "grade"),
+                    # Normalized display grade alongside the verbatim raw value: a
+                    # corpus may store the grade as raw Arabic / mixed script,
+                    # which is not a stable key to filter or display on (da#148).
+                    "grade_normalized": normalize_grade(_val(row, "grade")),
                     "source_corpus": _val(row, "source_corpus", ""),
                     "sect": _val(row, "sect", ""),
                     "collection_name": _val(row, "collection_name", ""),
@@ -440,6 +446,7 @@ MERGE (n:Grading {id: row.id})
 SET n.hadith_id          = row.hadith_id,
     n.scholar_name       = row.scholar_name,
     n.grade              = row.grade,
+    n.grade_normalized   = row.grade_normalized,
     n.methodology_school = row.methodology_school,
     n.era                = row.era
 """
@@ -488,6 +495,10 @@ def _load_gradings(
                     "hadith_id": hadith_node_id(sid),
                     "scholar_name": _val(row, "collection_name", "unknown"),
                     "grade": grade,
+                    # Normalized display grade (da#148): the raw ``grade`` may be
+                    # Arabic / mixed-script free text; map it onto the HadithGrade
+                    # vocabulary so the UI and queries have a stable key.
+                    "grade_normalized": normalize_grade(grade),
                     "methodology_school": None,
                     "era": None,
                 }
--- a/src/utils/grade.py
+++ b/src/utils/grade.py
@@ -1,0 +1,63 @@
+"""Hadith authenticity-grade normalization.
+
+Source corpora record a hadith's grade as free text in whatever script the
+upstream used: Arabic (``صحيح``), transliterated English (``Sahih``,
+``Da'if``), or descriptive English (``authentic``, ``weak``). The graph stores
+that raw string verbatim so the original is never lost, but a raw Arabic /
+mixed-script value is not a stable key to filter, facet, or display on. This
+module maps any of those raw forms onto the canonical :class:`HadithGrade`
+vocabulary so a normalized *display* field can sit alongside the raw one
+(da#148 integrity sweep — "grade stored as raw Arabic, needs normalized
+display").
+
+The function is deliberately conservative: anything it does not recognise maps
+to :data:`HadithGrade.UNKNOWN` rather than guessing, so a normalized value is
+always a real enum member.
+"""
+
+from __future__ import annotations
+
+from src.models.enums import HadithGrade
+from src.utils.arabic import normalize_arabic
+
+__all__ = ["normalize_grade"]
+
+# Match order matters: the ``li-ghayrihi`` ("…by virtue of corroboration")
+# qualifiers are checked before the bare grade so ``صحيح لغيره`` does not match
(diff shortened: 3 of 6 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

Approve. Reviewed at head **a80932e8** (re-fetched + confirmed == headRefOid `a80932e8f5e27a77d4a34ff3c4a864f0d21d6d07`); base `deployments/phase-5/wave-1`, +181/−2. All 5 CI checks SUCCESS (Lint, Type Check, Test, Integration Tests, Pre-commit⇄CI sync-drift gate). Diff read at HEAD via `gh api contents?ref=a80932e8`.

**Test coverage — both new behaviors are tested, both directions:**
- **Self-loop guard** (`_build_chain_pairs`): `test_adjacent_duplicate_narrator_skipped` asserts the a→a adjacent pair (`nar:1`→`nar:1`) is dropped AND the distinct adjacency (`nar:1`→`nar:2`) survives intact, plus a blanket `all(from != to)`. No over-drop — the pre-existing `test_sorts_by_position` continues to assert distinct adjacent pairs are kept. The guard correctly mirrors the network-edge producers (`muhaddithat`/`mis`) that already had it.
- **`normalize_grade()`**: `test_grade.py` table covers Arabic (`صحيح` + diacritic variant `صَحِيح`), transliterated (`Sahih`/`Da'if`/`Daif`), descriptive English (`weak`/`authentic`/`fabricated`), `li-ghayrihi` precedence over the bare grade (rule-order verified), and None/`""`/whitespace/garbage/`42` → `unknown`. `test_return_value_is_always_a_valid_enum_member` guarantees the conservative contract. Verified independently: all referenced `HadithGrade` members exist in `enums.py`, and `normalize_arabic` only rewrites U+0600–U+06FF so the English needles pass through `normalize_arabic(raw).lower()` unchanged.
- **Loader wiring**: `test_grading_carries_normalized_display_grade` confirms raw `صحيح` is preserved verbatim AND `grade_normalized == "sahih"` is written alongside (`Da'if → daif`).

**Re-load idempotency:** confirmed this is a producer-level fix, not a one-off DB poke — the chain producer no longer emits `from==to` pairs and the loaders write `grade_normalized` deterministically, so a re-load from staging stays self-loop-free and carries the normalized field. (Cleanup of the 8 *already-live* self-loops happens on the next full re-load, not via in-place edge deletion — consistent with the "not DB pokes" scope and acceptable.)

**Deferred items are genuine, not testability dodges:** the 5 follow-ups (null `hadith_number_in_book`, 51 NARRATED gaps, 8 orphan bio-only narrators, ingest-platform `grade_normalized` parity, collection-metadata gap) each need a data decision or upstream producer/NER work that can't be unit-tested at the load layer, and all are tracked in da#153 (verified OPEN). Finding #5 (`lk` STUDIED_UNDER) is correctly closed out as a cross-corpus identity merge, not a relation leak — no producer change needed.

## Throughline observations

Wave-level pattern: this integrity sweep cleanly bifurcates findings into (a) producer/load-layer fixes with a deterministic test (shipped here) and (b) data-decision/upstream items punted to da#153. That split is the right discipline and worth carrying as the template for the remaining sweep work — keep the "clean producer fix + re-load stays clean" bar as the inclusion gate, and resist folding data-decision items into a scoped PR just to close the parent faster.

parametrization:
RequestOrReplied: Approved
TechDebt: none

Approved. Reviewed at confirmed head **a80932e8** (`headRefOid` a80932e8f5e27a77d4a34ff3c4a864f0d21d6d07), base `deployments/phase-5/wave-1`, +181/−2 over 6 files. Read the diff and every touched file from origin contents at that ref, not local. CI is 5/5 green (Lint, Type Check, Test, Integration Tests, sync-drift gate).

**Self-loop guard (`_build_chain_pairs`, load_edges.py) — correct.** It compares only the canonical ids of *adjacent* resolved mentions (`resolved[i]` → `resolved[i+1]`) and `continue`s when `from_id == to_id`. That drops exactly the a→a same-canonical adjacent pair (a repeated narrator or two mentions the disambiguator collapsed onto one node) and nothing else: a non-adjacent re-appearance (nar:1, nar:2, nar:1) still produces two distinct edges, so legitimate distinct-narrator adjacency is untouched. This matches the existing producer guards, which I verified at HEAD rather than trusting the comment: `src/parse/muhaddithat.py:215` skips `from_id == to_id` in `_parse_network_edges`, and `src/parse/mis.py:275` skips `from_name == to_name`. The chain-derived path was the one producer still missing the invariant; this closes it. The new test `test_adjacent_duplicate_narrator_skipped` pins the behavior (nar:1→nar:1 dropped, nar:1→nar:2 survives, plus a blanket `from_id != to_id` assertion).

**`normalize_grade()` (src/utils/grade.py) — faithful.** Arabic→HadithGrade mapping spot-checked against `src/models/enums.py` HadithGrade (sahih/hasan/daif/mawdu/sahih_li_ghayrihi/hasan_li_ghayrihi/unknown) — all rule targets are real members. Both input and the Arabic needles pass through the same `normalize_arabic()` pipeline (diacritics/alif/hamza/taa-marbuta/tatweel/whitespace), so diacritic variants like صَحِيح collapse correctly; the Latin side is covered by `.lower()` on the text. Match order is right: the `li-ghayrihi` qualifiers are checked before the bare grade, so صحيح لغيره resolves to SAHIH_LI_GHAYRIHI and never short-circuits to bare SAHIH (test `test_li_ghayrihi_wins_over_bare_grade` locks this). Conservative fallback to `HadithGrade.UNKNOWN.value` on empty/non-str/unrecognised, and `test_return_value_is_always_a_valid_enum_member` guarantees the output is always a real enum value.

**`grade_normalized` is additive, not a replacement.** Both `_HADITH_MERGE` (load_nodes.py:206) and `_GRADING_MERGE` (load_nodes.py:449) keep `n.grade = row.grade` and *add* `n.grade_normalized = row.grade_normalized`; verbatim raw is never overwritten. I traced the two Cypher `row.grade_normalized` references to their producing Python batches — hadiths compute it at L255, gradings at L501 — so there is no uncomputed/orphan reference (the diff hunk headers "_load_narrators"/"_load_chains" are git's nearest-preceding-def labels and are misleading; the lines actually live in the hadith and grading constants). `_NARRATOR_MERGE`/`_CHAIN_MERGE` don't carry a grade, so they're correctly left alone. `test_grading_carries_normalized_display_grade` asserts raw صحيح preserved and `grade_normalized == "sahih"`.

**Split is honest.** da#153 (OPEN) enumerates every deferred item — 86 null hadith_number_in_book, 51 hadith w/o NARRATED, 8 orphan bio-only narrators, grade_normalized parity in the ingest-platform streaming path, collection metadata — each explicitly deferred for a data decision or upstream/cross-repo work rather than fabricated at load time. It also records the "15 lk-STUDIED_UNDER" finding as investigated-and-not-a-bug (correct cross-corpus identity merge, no lk network-edge producer), consistent with the STUDIED_UNDER allowlist behavior. Nothing from da#148 is silently dropped.

## Throughline observations

Producer-parity is the wave-level pattern to watch in P5W1. Each graph-integrity invariant this PR establishes — the self-loop guard and grade normalization — has to hold on *every* edge/node producer, not just the one being fixed. This PR completes the self-loop invariant across all three chain/network producers, but da#153 item #4 correctly flags that `grade_normalized` now exists only on the batch load path here while the isnad-ingest-platform Kafka streaming worker still needs the mirror fix, or streamed loads will drift from batch loads. The recurring risk across this wave is a fix landing on one of the two load paths (batch vs. streaming) and silently diverging; reviewers should treat "did the streaming path get the same invariant?" as a standing checklist item for any integrity fix in P5W1, and da#153 #4 is the right place that's tracked.

Issue #146, opened 2026-06-13
bug(resolve/ner): 67% of lk-corpus narrators are un-segmented isnad chain strings — NER never split chains into narrators


parametrization:
## Summary

Exploratory bug-bash over the **live #601 narrator graph** on staging (47,199 narrators / 153,804 edges, isnad-graph#1018) found that the dominant `lk` corpus narrator records are **fundamentally mis-extracted**: a majority of `:Narrator` nodes are un-segmented isnad *chain strings*, not individual narrators, and they carry no English name and an opaque corpus code. The graph's node/edge **counts** are healthy (which is why the #601 load invariants passed), but the **entity quality** for the main corpus is poor.

## Evidence (live staging Neo4j, 2026-06-13)

**1. 67% of narrators are un-segmented isnad chains.**
```
MATCH (n:Narrator) WHERE n.name_ar CONTAINS 'حدثنا' OR n.name_ar CONTAINS 'أخبرنا' OR n.name_ar CONTAINS 'حدثني'
RETURN count(n)  -- 31,525  (of 47,199 = 67%)
```
`name_ar` length buckets: `<=30: 6,983` · `31–60: 5,114` · `61–120: 24,116` · `>120: 10,986`. A person's name is short; 35,102 nodes have `name_ar > 60` chars.

Representative node (a single `:Narrator`, `mention_count=1`):
> `name_ar = "حدثنا عبد الله بن مسلمه بن قعنب القعنبي، حدثنا عبد العزيز، يعني ابن محمد عن محمد، يعني ابن عمرو عن ابي سلمه، عن المغيره بن شعبه،"`

That is an **entire isnad chain** ("Abdullah b. Maslamah narrated to us … from al-Mughirah b. Shu'bah") stored as one narrator. The isnad-NER/segmentation step did not split the chain into its constituent narrators.

**2. 99.99% of `lk` narrators have no English name.**
```
source_corpus | total  | missing_name_en
lk            | 47,089 | 47,086
muhaddithat   |    110 |      0
```
The English-facing product (search, narrator list, graph explorer) would render blank/Arabic-only names for ~all narrators.

**3. Opaque `source_corpus = "lk"`** on 47,086 nodes — a 2-char code that does not match any known `SourceCorpus` enum value (sunnah / itqan / muhaddithat / bihar / mis). `external_id = NULL`, `source_ids = []` on these nodes, so provenance is unclear. What is `lk`, and is the corpus assignment correct?

[...]

PR #151, opened 2026-06-14
fix(da#146): segment voweled lk isnad chains into per-narrator mentions

parametrization:
## Summary

67% of lk-corpus narrators (`source_corpus="lk"`, ~31,525 rows) were **un-segmented isnad chain strings**: each whole Arabic chain became one "narrator" node, blocking the narrator surface (ig#1024) and Phase-5 criterion #1 (main#665).

### Root cause

`src.utils.arabic.extract_transmission_phrases` matched **diacritic-free** keyword patterns (`حدثنا`, `عن`, `قال` …) against **fully-voweled** isnad text (`حَدَّثَنَا …`). The patterns never matched real data, so it returned **zero** transmission phrases on every voweled chain. `_extract_arabic` then took its `if not phrases:` branch and emitted the *entire chain* as a single mention at `position 0`. The toy test fixtures used un-voweled Arabic (`حدثنا عمر بن الخطاب`), which *did* match — masking the bug.

Verified on the real Bukhari h1 isnad: raw → `phrases == []` → 1 blob mention; after normalization → 6 phrases.

### Fix (surgical — entirely in `extract_transmission_phrases`, positions preserved)

- **Diacritic-tolerant patterns**: each transmission term is compiled to tolerate an optional tashkeel/tatweel run between base letters, so voweled chains match while match positions still index into the caller's **raw** string — `_extract_arabic`'s slice logic is unchanged, and the fix benefits every Arabic source (thaqalayn, open_hadith), not just lk.
- **Singular/3rd-person variants** added (`حدثني`, `أخبرني`, `أنبأني`, `سمع`) that real chains alternate to, with **overlap resolution** (`سمعت` wins over the bare `سمع` at a shared start; longer match preferred).
- **Trailing Arabic comma/semicolon** (`،`/`؛`) stripped from extracted names so the exact-match disambiguation key isn't polluted.

A single voweled Bukhari h1 chain now yields **6 narrator mentions** (was 1 blob), which the existing `disambiguate` stage resolves into multiple canonical narrators. No change was needed in `lk_corpus.py` or `_extract_arabic` — they were already correct given a working phrase finder.

## Test Plan

- `tests/test_utils/test_arabic.py` — voweled-term matching, overlap-resolution (`سمعت` not double-counted), singular-variant recognition.
- `tests/test_parse/test_narrator_extraction.py::TestDa146VoweledSegmentation` — a fully-voweled chain segments into ≥4 mentions with sequential positions and clean (comma-free) names.
- `tests/test_resolve/test_lk_segmentation_e2e.py` — **end-to-end**: voweled lk CSV → `lk_corpus.run` → `ner.run` → `disambiguate.run`; asserts ≥4 mention rows and ≥4 canonical narrators from one chain (was 1), with the seeded bio exact-matching the segmented `سفيان`.
- Full suite: **837 passed, 5 skipped**. `ruff format`/`ruff check` clean; `mypy` clean on changed sources.

## Follow-up

This is the deterministic-splitter slice (the tractable, tested keystone). What real Arabic NER would add beyond this:

- **Name-boundary precision**: deterministic splitting still captures connective words into a span (e.g. `أَنَّهُ`/`يَقُولُ` trailing a name) — a stopword/honorific filter or a model-based NER would yield cleaner narrator surface forms. Tracked as a refinement, not a blocker.
- **`عن` substring false-positives**: `عن` can match inside a token; a morphology-aware tokenizer would avoid spurious splits.
- **Kunya/laqab joining** (`أبو …`, `ابن …`) across transmission boundaries.

Closes #146

### DIFF
--- a/src/parse/narrator_extraction.py
+++ b/src/parse/narrator_extraction.py
@@ -20,8 +20,10 @@ _EN_DELIMITERS: re.Pattern[str] = re.compile(
     re.IGNORECASE,
 )
 
-# Trailing punctuation to strip from extracted names.
-_TRAILING_PUNCT_RE: re.Pattern[str] = re.compile(r"[.,;:!?()]+$")
+# Trailing punctuation to strip from extracted names. Includes the Arabic comma
+# (، U+060C) and semicolon (؛ U+061B), which routinely terminate a narrator span
+# in voweled isnads (da#146) and would otherwise pollute the exact-match key.
+_TRAILING_PUNCT_RE: re.Pattern[str] = re.compile(r"[.,;:!?()،؛\s]+$")
 
 
 @dataclass(frozen=True)
--- a/src/utils/arabic.py
+++ b/src/utils/arabic.py
@@ -48,15 +48,55 @@ _ARABIC_CHAR_RE: re.Pattern[str] = re.compile(r"[\u0600-\u06FF]")
 # Transmission phrase patterns
 # ---------------------------------------------------------------------------
 
+# Optional run of Arabic diacritics (tashkeel U+064B–U+065F, superscript alef
+# U+0670) and tatweel/kashida (U+0640). Real isnad text is fully voweled
+# (e.g. ``حَدَّثَنَا``), so a bare keyword like ``حدثنا`` will never match unless
+# we tolerate diacritics interleaved *between* the base letters. Without this
+# the extractor found zero transmission phrases on every voweled chain and fell
+# back to emitting the whole isnad as a single un-segmented "narrator" (da#146).
+_OPT_DIAC: str = r"[\u064b-\u065f\u0670\u0640]*"
+
+# Base transmission terms keyed to their canonical label. Written with the
+# classical hamza-bearing forms (أ/إ) so they match raw, un-normalized isnad
+# text; positions returned therefore index into the caller's original string.
+# Singular (-ني) and 3rd-person (سمع) variants are included alongside the plural
+# (-نا) forms because real chains alternate between them ("حدثني" / "أخبرني" /
+# "سمع فلانٌ") — omitting them under-segments the chain (da#146).
+_TRANSMISSION_TERMS: dict[str, str] = {
+    "حدثنا": "haddathana",
+    "حدثني": "haddathani",
+    "أخبرنا": "akhbarana",
+    "أخبرني": "akhbarani",
+    "أنبأنا": "anba_ana",
+    "أنبأني": "anba_ani",
+    "سمعت": "samitu",
+    "سمع": "samia",
+    "عن": "an",
+    "قال": "qala",
+    "ناولني": "nawalani",
+    "كتب إلي": "kataba_ilayya",
+}
+
+
+def _compile_diacritic_tolerant(term: str) -> re.Pattern[str]:
+    """Compile *term* into a regex that tolerates interleaved diacritics.
+
+    Each base letter is followed by an optional diacritics/tatweel run, and
+    inter-word whitespace is matched flexibly. The compiled pattern matches the
+    fully-voweled form found in real isnads while preserving match positions in
+    the original (un-normalized) text.
+    """
+    parts: list[str] = []
+    for ch in term:
+        if ch.isspace():
+            parts.append(r"\s+")
+        else:
+            parts.append(re.escape(ch) + _OPT_DIAC)
+    return re.compile("".join(parts))
+
+
 TRANSMISSION_PATTERNS: dict[re.Pattern[str], str] = {
-    re.compile(r"حدثنا"): "haddathana",
-    re.compile(r"أخبرنا"): "akhbarana",
-    re.compile(r"سمعت"): "samitu",
-    re.compile(r"عن"): "an",
-    re.compile(r"قال"): "qala",
-    re.compile(r"أنبأنا"): "anba_ana",
-    re.compile(r"ناولني"): "nawalani",
-    re.compile(r"كتب\s+إلي"): "kataba_ilayya",
+    _compile_diacritic_tolerant(term): label for term, label in _TRANSMISSION_TERMS.items()
 }
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

Secondary (architecture/scope) review. Confirmed head **dcf966d1** (full `dcf966d193cda4a122bd6c3ec587507dd5ae0b6d`), base `deployments/phase-5/wave-1`, OPEN/CLEAN/MERGEABLE. Diff read at HEAD via origin contents API. All 5 CI checks green (Lint, Test, Type Check, Integration Tests, sync-drift gate).

**Right layer — confirmed a feature, not unscoped blast radius.** The fix lands in `src/utils/arabic.py` `TRANSMISSION_PATTERNS` / `extract_transmission_phrases`, the shared phrase finder consumed by `_extract_arabic`. The ontology already names this as the canonical Arabic-processing seam (ingestion.yaml `arabic_normalization: "...via src/utils/arabic.py"`). Diacritic tolerance benefits thaqalayn and open_hadith equally — fixing it in the lk adapter would have been the wrong, narrower layer and would have left the same latent blob bug in every other voweled source. Blast radius is bounded: the existing Arabic-source suite stays green (837 passed) because the patterns still match the same terms, only now with optional interleaved tashkeel.

**Position-preservation claim verified at the consuming call site.** `_extract_arabic` slices `text[end:next_start]` against the *raw* isnad (narrator_extraction.py:78), so match offsets must index into the un-normalized string. The compiled patterns keep the classical hamza-bearing base letters and only insert an optional diacritics/tatweel run *between* base letters — positions therefore still land in the caller's raw text and the slice logic is correctly left untouched. The `if not phrases:` whole-chain fallback (lines 58-61) is exactly the blob path, and it no longer fires on voweled chains.

**No over-match into name tokens.** `_OPT_DIAC` is restricted to the tashkeel block (U+064B-065F), superscript alef (U+0670) and tatweel (U+0640) — it can only ever consume diacritics/kashida, never a name's base consonants, so a transmission term's voweled variant cannot swallow an adjacent narrator token. The one genuine substring hazard (`عن` matching inside a larger word) is correctly *not* hidden — it is named in the Follow-up. The greedy longest-at-shared-start overlap resolution (sort by start asc / span-length desc, then non-overlapping keep) is the right call for the سمعت-vs-سمع collision and is pinned by `test_overlapping_variants_resolve_to_longer`.

**Honestly bounded scope.** The Follow-up section names precisely what real NER adds beyond the deterministic splitter — name-boundary/connective-word precision, `عن` substring false-positives, kunya/laqab joining — without blocking. This is the tractable, tested keystone: one voweled Bukhari h1 chain goes from 1 blob to 6 mentions, and the e2e test drives the full lk_corpus -> ner -> disambiguate path to >=4 canonical narrators. That is the correct deliverable for #146. One nit, non-blocking: the Follow-up says these residuals are "Tracked as a refinement" but cites no issue number — please file (or link) the real-NER refinement issue so the deferred precision work isn't lost.

## Throughline observations

**Fixture-masks-bug — wave-level pattern, recurring.** The root cause here is textbook: the toy test fixtures used *un-voweled* Arabic (`حدثنا عمر بن الخطاب`), which matched the diacritic-free patterns, so the suite was green while every real (fully-voweled) chain silently returned zero phrases and collapsed to a blob. This is the same class as MockNeo4jClient masking the APPEARS_IN null-property abort, the toy h-1 fixtures masking the hadith-id double-prefix, and the staging "real edges" that were local-only. The corrective Ivana applied is the right one and worth generalizing: this PR's e2e test seeds *real-shape* (voweled) data through the actual lk_corpus->ner->disambiguate stages rather than a synthetic fixture. Recommend the wave retro capture "Arabic/text-processing fixtures MUST use production-realistic (voweled) input" as a standing rule — synthetic-shape fixtures are the recurring failure surface in this corpus.

parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: short-particle patterns عن/قال/سمع lack word-boundary anchoring and over-segment real narrator names (مَعْن, عنبسة) and common words (يعني) — file a follow-up to anchor them on a non-Arabic-letter/whitespace boundary before disambiguate output is trusted on real lk data.

Reviewed at head dcf966d1 (confirmed via gh pr view: headRefOid dcf966d193cda4a122bd6c3ec587507dd5ae0b6d, base deployments/phase-5/wave-1, +260/-15 / 5 files). Diff read from origin contents at that ref. CI all green (Lint, Type Check, Test, Integration Tests, sync-drift gate).

**Raw-string offset integrity — verified, this is correct.** `_compile_diacritic_tolerant` builds `re.escape(ch) + _OPT_DIAC` per base letter and `finditer`s the pattern against the original `text`. `match.start()/end()` therefore index the un-normalized voweled string. In `_extract_arabic`, segments are sliced as `text[end:next_start]` from that same raw string and only THEN passed through `normalize_arabic`. There is no stripped intermediate copy that offsets are computed against, so `_extract_arabic`/`lk_corpus.py` correctly stay unchanged. The load-bearing claim holds.

**Voweled match — verified.** `test_matches_diacritized_term` pins حَدَّثَنَا matching, and the overlap resolver (`test_overlapping_variants_resolve_to_longer`) correctly prefers سمعت over the bare سمع at a shared start via the `(start asc, -span desc)` sort + `start >= last_end` sweep. Singular/3rd-person variants (حدثني/أخبرني/سمع) are covered.

**E2E — verified.** `test_lk_segmentation_e2e` drives a real fully-voweled Bukhari h1 chain through lk_corpus → ner → disambiguate and asserts ≥4 canonical narrators (was 1 blob), including the seeded Sufyan bio exact-matching the segmented سفيان mention. The mention-level `test_voweled_chain_does_not_collapse_to_blob` also pins ≥4 spans with none exceeding half the chain length. Good keystone coverage — it exercises the actual disambiguate stage, not a synthetic stub.

**No regression — confirmed via green Test job + retained fixtures.** The shared `extract_transmission_phrases`/`extract_narrator_mentions` is used by thaqalayn/open_hadith parsers; the pre-existing un-voweled tests (e.g. `حدثنا سفيان عن الزهري`) are retained and pass, and the diacritic-tolerant patterns are supersets of the bare forms, so un-voweled text still matches.

**عن-substring risk — real, NOT handled and NOT documented (the basis for the TechDebt line, not a merge blocker).** I reproduced the matcher locally: the short particles compile to `ع[diac]*ن`, `ق[diac]*ا[diac]*ل`, `س[diac]*م[diac]*ع` with no leading/trailing word-boundary guard, so:
- مَعْنُ (Ma'n b. Isa, a real narrator) → عن matches mid-word at offset 2, splitting the name.
- عَنْبَسَة (Anbasa, real narrator) → عن matches at offset 0, so عَنْ is consumed as a delimiter and "بسة" becomes the spurious narrator name.
- يَعْنِي (common isnad filler) and مَقَالَة → عن / قال match inside.
This is strictly a *precision* loss layered on a large recall *gain*: before this PR voweled chains produced one useless blob; now they segment correctly and only occasionally over-segment on these substring words. It adds noise canonical narrators but does not corrupt the genuine narrators in the chain, and the keystone's e2e fixture happens not to contain عن at all — which is exactly why CI stays green while the hazard is live on real data. The surgical fix is to anchor the short particles on a non-Arabic-letter / whitespace / string-edge boundary (the long forms حدثنا/أخبرنا are low-risk and can stay as-is). Approving because the core contract is met, CI is green, and there is no regression — but this needs a tracked follow-up before disambiguate output on real lk corpora is trusted.

## Throughline observations

Same fixture-masks-bug class that motivated this PR is still live one layer down. The original da#146 bug (un-voweled toy fixtures hid that real chains are fully voweled → one blob) is fixed, but the *replacement* test fixture (Bukhari h1) coincidentally contains no عن particle, so the green e2e masks the عن/قال short-particle over-segmentation on real data. Wave-level pattern: Arabic NER fixtures must be drawn from (or adversarially include) real corpus rows that contain the high-frequency عن particle and at least one عن/قال-substring narrator name (مَعْن، عنبسة), not hand-built minimal chains — otherwise each fix's test passes while the next precision gap stays invisible.

PR #152, opened 2026-06-14
test(resolve): da#138 fuzzy-cluster precision harness + nasab-reversal guard

### LINKED ISSUE #138: tech-debt(fuzzy-cluster): validate precision on a real labelled multi-source sample before it feeds the production graph
**Surfaced by Oyunbileg in #137 review (da#118 fuzzy cross-source narrator clustering). Runtime gate — the inline static-vector guards land in #137; this tracks the real-data precision validation.**

`src/resolve/fuzzy_cluster.py` (da#118) clusters cross-source name variants the exact #99 pass leaves split (rapidfuzz token_set_ratio + #94 aliases). #137 adds inline guards against the two statically-identifiable over-merge vectors (single-token-subset→100; transitive-bridge guard-bypass). But the precision=1.0 claim is on **synthetic fixtures** — there is no labelled multi-source gold sample loaded locally, so real-data precision is unproven.

## Why it matters
Over-merge = identity corruption (two real narrators → one Narrator node), hard to detect after the fact, and it propagates to NARRATED/STUDIED_UNDER edges. Data-first means the data must be CORRECT, not just present.

## Required before fuzzy clustering feeds the production graph
- [ ] Assemble (or sample) a **labelled multi-source narrator set** with known same-person variant pairs AND known distinct-but-similar pairs across ≥2 sources (Sanadset/Itqan/muhaddithat/Bihar/halimbahae/MIS).
- [ ] Measure real-data **precision + recall** of `fuzzy_cluster` against it; tune the threshold / significant-token rule conservatively (precision-first).
- [ ] Live neo4j:5 acceptance: a known same-person variant pair across ≥2 sources collapses to ONE Narrator node, with NARRATED edges following the merge (the #118 runtime AC).
- [ ] If real-data precision is inadequate at any reasonable threshold, gate `fuzzy_cluster` default-OFF in run_all until it is.

Ref #118, #137, #99. Companion runtime-gate to da#133 (edge-relation) + da#136 (Bihar kitab source_id).

parametrization:
## What & why (da#138 — W1 tech-debt intake 1/1)

Validate the **precision** of cross-source fuzzy narrator clustering
(`src/resolve/fuzzy_cluster`) and guard it against regression.

### Finding
Built a labeled gold set and ran it end-to-end through `cluster_assignment` +
`quality.pairwise_quality`. The clustering **falsely merged a nasab reversal** —
`محمد بن عبد الله` (Muhammad b. Abdullah) and `عبد الله بن محمد` (Abdullah b.
Muhammad), two different people — because `rapidfuzz.token_set_ratio` is
order-insensitive and scores that pair a perfect **100**. With no
death-year/gender metadata to corroborate, the merge went through:
**precision 0.667 / false-merge rate 0.333** on the set.

Raising the numeric threshold cannot fix a 100 score and would only cost recall,
so the fix is a guard, not a higher bar (per the brief: tighten only if evidence
supports, don't degrade recall blindly).

### Fix (recall-preserving)
A **token-order guard**: a merge requires the shared significant tokens to appear
in the **same relative order** in the two matched spellings. A genuine variant
only adds/drops tokens (kunya/nisba/nasab expansion), which preserves order; only
a reordering is rejected. Post-fix the gold set scores **precision 1.000
(false-merge rate 0.000) at recall 1.000** — legitimate cross-source variants
still merge.

### Changes
- `fuzzy_cluster.py`: `_significant_token_sequence`, `_token_order_consistent`,
  `_name_match`; `_can_merge` now routes through `_name_match`; dropped the
  now-dead `_name_similarity`; documented the guard.
- `quality.py`: `ClusterQuality.false_merge_rate` (`1 - precision`) as the shared
  over-merge metric.
- `tests/`: `cluster_precision_gold.json` labeled fixture +
  `test_cluster_precision.py` harness (precision floor, recall floor,
  reversal-stays-split, and a self-test that re-introduces the regression to
  prove the harness catches it).

## Test Plan
- `ENVIRONMENT=test uv run pytest tests/` → **833 passed, 5 skipped** (incl. the
  4 new harness tests and the 17 existing fuzzy_cluster/quality tests).
- `ruff@0.15.11 format --check` + `check` → clean; `mypy` → clean.
- pre-commit + pre-push hook stages → all Passed.
- Measured precision: pre-fix 0.667 (false-merge 0.333) → post-fix 1.000
  (false-merge 0.000), recall held at 1.000.

Closes #138

### DIFF
--- a/src/resolve/fuzzy_cluster.py
+++ b/src/resolve/fuzzy_cluster.py
@@ -33,6 +33,17 @@ Precision guards (against over-merging distinct same-named narrators)
   disagree by more than :data:`_DEATH_YEAR_TOLERANCE`, the merge is blocked even
   on a perfect name match — two scholars a century apart are not one person.
 * A **gender guard**: two records with explicit, differing ``gender`` never merge.
+* A **token-order guard** (da#138): ``token_set_ratio`` is order-*insensitive*, so
+  a pure nasab reversal of two **different** people — ``محمد بن عبد الله``
+  (Muḥammad son of ʿAbdullāh) ↔ ``عبد الله بن محمد`` (ʿAbdullāh son of Muḥammad) —
+  scores a perfect 100 and, with no death-year/gender to corroborate, would
+  falsely merge. The guard requires the shared significant tokens to appear in the
+  **same relative order** in the two matched spellings. A genuine variant only
+  *adds or drops* tokens (kunya/nisba/nasab expansion), which preserves the order
+  of the shared tokens, so recall is untouched; only a reordering is rejected.
+  This precision defect — and the validation harness that quantified its
+  false-merge rate — is the da#138 tech-debt work; raising the numeric threshold
+  cannot fix a perfect-100 reversal score and would only cost recall.
 
 Identity & ordering invariants
 -------------------------------
@@ -162,22 +173,45 @@ def _significant_tokens(keys: list[str]) -> set[str]:
     return tokens
 
 
-def _name_similarity(keys_a: list[str], keys_b: list[str]) -> float:
-    """Best token_set_ratio across the cross product of two records' match keys.
+def _significant_token_sequence(key: str) -> list[str]:
+    """Significant (non-connector) tokens of a single name string, in name order.
 
-    Taking the max over names + aliases means a record matches if *any* of its
-    spellings (including a da#94 variant) is a strong token-set match for any of
-    the other's — exactly the cross-source-variant recall this pass targets.
+    The ordered counterpart of :func:`_significant_tokens` — the token-order guard
+    compares relative order, so it needs the sequence, not the set.
     """
-    best = 0.0
+    return [tok for tok in key.split() if len(tok) >= 2 and tok not in _CONNECTOR_TOKENS]
+
+
+def _token_order_consistent(a: str, b: str) -> bool:
+    """True when the tokens shared by names ``a`` and ``b`` keep the same order.
+
+    A genuine variant only adds/removes tokens, so the shared tokens stay in the
+    same relative order; a nasab reversal of two different people permutes them.
+    Fewer than two shared significant tokens is left to :data:`_MIN_SHARED_TOKENS`
+    — this guard only adjudicates the high-overlap case a reversal can exploit.
+    """
+    seq_a = _significant_token_sequence(a)
+    seq_b = _significant_token_sequence(b)
+    shared = set(seq_a) & set(seq_b)
+    if len(shared) < 2:
+        return True
+    return [t for t in seq_a if t in shared] == [t for t in seq_b if t in shared]
+
+
+def _name_match(keys_a: list[str], keys_b: list[str], *, threshold: float) -> bool:
+    """True when some key pair is a strong AND token-order-consistent match.
+
+    A pair counts only if its ``token_set_ratio`` clears ``threshold`` *and* it
+    survives the da#138 token-order guard. Taking the best over the cross product
+    of each record's match keys (its name + da#94 aliases) keeps the alias-driven
+    recall — a record matches if *any* of its spellings is both a strong and
+    order-consistent match for any of the other's.
+    """
     for a in keys_a:
         for b in keys_b:
-            score = fuzz.token_set_ratio(a, b)
-            if score > best:
-                best = score
-                if best >= 100.0:
-                    return best
-    return best
+            if fuzz.token_set_ratio(a, b) >= threshold and _token_order_consistent(a, b):
+                return True
+    return False
 
 
 def _death_years_conflict(a: dict[str, Any], b: dict[str, Any]) -> bool:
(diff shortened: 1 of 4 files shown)


parametrization:
RequestOrReplied: Approved
TechDebt: none

Approved. Head confirmed `88502d4b236e956a852be655ef4b7213eede35f8` (== 88502d4b); diff read at that ref via origin contents; all 5 CI checks green (Lint, Type Check, Test, Integration Tests, Pre-commit ⇄ CI sync-drift gate).

**Fixture covers BOTH precision and recall — not a precision-only blind spot.**
- *Recall (same-narrator spelling variants):* bukhari (nasab/nisba expansion `محمد بن اسماعيل البخاري` ↔ `محمد بن اسماعيل`), ahmad, malik — nisba truncation pairs; umar exercises the da#94 **alias-driven** path (`ابو حفص العدوي القرشي` + alias `عمر بن الخطاب` ↔ `عمر بن الخطاب`), so recall through the alias match-key cross-product is in scope, not just direct-name.
- *Precision traps, each exercising a DIFFERENT guard:* sufyan al-Thawri vs sufyan b. ʿUyayna share only `سفيان` → caught by `_MIN_SHARED_TOKENS` (≥2); the two Yaḥyā b. Saʿīd (al-Qaṭṭān d.198 vs al-Anṣārī d.143) share 2 significant tokens and clear token_set_ratio → caught by the **death-year guard**; and the da#138 core trap — the nasab reversal `محمد بن عبد الله` ↔ `عبد الله بن محمد`, token_set_ratio a perfect 100 with no death-year/gender to corroborate → caught only by the new **token-order guard**. So precision is not satisfiable by any single guard; the set forces all three.

**`false_merge_rate` is correctly defined and guarded.** `1 - precision` == (predicted_pairs − tp)/predicted_pairs == false-merges/total-merges, the over-merge measure. Consistent at the empty-prediction boundary (precision defined 1.0 → rate 0.0). It is a *guarded* metric, not decorative: `test_clustering_precision_meets_baseline` pins `precision >= 1.0` and `false_merge_rate <= 0.0`, and `test_harness_detects_a_precision_regression` neutralizes `_token_order_consistent` and asserts `false_merge_rate > 0.0` — so a future loosening of clustering is caught from both directions.

**Harness measures the SHIPPED path, not a reimplementation.** The test drives the real `src.resolve.fuzzy_cluster.cluster_assignment` → `cluster_records` → `_can_merge` → `_name_match`/`_token_order_consistent`, and builds records via the genuine `make_canonical_id(normalize_arabic(...))` identity contract — the same decision logic the parquet IO wrapper `cluster_canonical_narrators` invokes. Skipping only the parquet IO is appropriate for a precision metric.

**The fix is the right shape.** `_token_order_consistent` short-circuits to True below 2 shared tokens (correctly deferring to `_MIN_SHARED_TOKENS` so it adjudicates only the high-overlap case a reversal exploits), and a genuine variant only adds/drops tokens — shared-token relative order is preserved — so recall is provably untouched (RECALL_FLOOR 1.0 holds). Raising the numeric threshold could not fix a perfect-100 reversal score, as the docstring argues; the order guard is the correct instrument. `test_gold_fixture_is_well_formed` additionally pins the no-distinct-name-collision and ≥1-true-merge invariants so the fixture can't silently rot into vacuity.

## Throughline observations

The resolve stage's precision is accreting guard-by-guard (death-year, gender, `_MIN_SHARED_TOKENS`, now token-order, plus `_safe_partition` for the union-find bridge case). This PR is the healthy capstone to that pattern: rather than a fifth ad-hoc guard, it adds a *measured* precision/recall floor over the shipped path with a meta-test proving the harness is sensitive. Worth flagging at wave level as the template for future resolve-stage guards — each new guard should land with a labeled fixture row that would regress without it, so the floor stays load-bearing rather than guard-count-driven. Not a blocker; an observation for the retro.

parametrization:
Requestor: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

Reviewed at head **88502d4b** (re-fetched + confirmed: `headRefOid` = 88502d4b236e956a852be655ef4b7213eede35f8, base `deployments/phase-5/wave-1`). Diff read at HEAD via `gh api contents?ref=88502d4b`. CI green — all 5 checks SUCCESS (Lint, Type Check, Test, Integration Tests, Pre-commit⇄CI sync-drift gate).

**Recall preservation (the load-bearing claim) — verified.** The harness pins recall, not just precision: `test_clustering_precision_meets_baseline` asserts `quality.recall >= RECALL_FLOOR` (=1.0) right next to the precision floor, so a guard that "passes" by refusing every merge would fail the test. The 17 pre-existing tests (12 in `test_fuzzy_cluster.py` + 5 in `test_quality.py`) stay green under the green Test job, and none reference the removed `_name_similarity` symbol. Mechanically the guard is recall-safe: `_token_order_consistent` compares only the *shared* significant tokens' relative order, and a genuine variant only adds/drops tokens (trailing nisba البخاري/الشيباني/الاصبحي, or an alias key) — which never permutes the shared subsequence. Confirmed against the fixture's variant pairs.

**Order guard is the right mechanism (vs a threshold bump) — agreed.** `token_set_ratio` is order-insensitive and scores the reversal a perfect 100; no numeric threshold can discriminate a 100 from a 100, and raising the bar only costs recall on legitimate variants. The author's reasoning holds. The guard correctly handles kunya/nasab structure: بن، ابن، ابو، ابي، ام، ال، بنت، عن are all in `_CONNECTOR_TOKENS`, so `_significant_token_sequence` drops the nasab particle and compares the meaningful tokens. For محمد بن عبد الله ↔ عبد الله بن محمد the shared significant tokens are {محمد, عبد, الله}; forward keeps محمد first, reversal puts it last → subsequences differ → merge rejected. The بن↔ابن cross-source spelling variation cannot break the guard since both are connectors and never enter the comparison. The `shared < 2 → True` early-return is safe: `_can_merge` already gates on `_MIN_SHARED_TOKENS` (=2) at the record level, and a per-key pair with <2 shared significant tokens isn't a reversal a permutation can exploit — documented as such.

**Dead-code removal — confirmed genuine.** `grep -c _name_similarity` over the file at HEAD = 0; no other caller in the tree and no test depends on it. (`gh search code` still surfaces it, but that index lags the default branch — the HEAD file is clean.)

`false_merge_rate` (`1 - precision`, guarded by `predicted_pairs`) is a clean, single-definition over-merge metric, and the `test_harness_detects_a_precision_regression` meta-test — neutralizing `_token_order_consistent` and asserting precision drops / false-merge rate rises — proves the harness actually bites. This is a precise, well-reasoned fix with the validation to back it. Approving.

## Throughline observations

No wave-level pattern observed — this PR is a standalone da#138 fix. Worth noting as a positive template for other resolve/quality work this wave: a validation harness that pins BOTH a precision and a recall floor AND ships a self-test re-introducing the regression to prove the harness catches it. That pattern is what makes "precision 1.000 at recall 1.000" a durable gate rather than a one-time measurement.

PR #170, opened 2026-06-14
fix(parallels): scale PARALLEL_OF detection to production corpus + empty-load conformance (da#160)

### LINKED ISSUE #160: bug(dedup): 0 PARALLEL_OF edges loaded — /compare Browse Parallels permanently empty; parallel-detection stage produced nothing
## Summary
The `/compare` "Browse Parallels" tab is permanently empty: staging Neo4j has **0** `(:Hadith)-[:PARALLEL_OF]->(:Hadith)` edges. The deduplication / parallel-detection pipeline that should populate them has produced nothing.

## Evidence (staging Neo4j, 2026-06-14)
- `MATCH ()-[r:PARALLEL_OF]->() RETURN count(r)` → **0**.
- Relationship inventory: TRANSMITTED_TO 52,182 · APPEARS_IN 34,028 · NARRATED 33,977 · GRADED_BY 33,478 · STUDIED_UNDER 186 · **PARALLEL_OF 0**.
- The `/parallels` API (`src/api/routes/parallels.py`) queries `PARALLEL_OF`, so the endpoint returns empty and the Browse tab shows only the "No parallel hadith pairs yet" state.

## Asks
- [ ] Determine whether the dedup/parallel-detection stage **never ran** vs **ran but loaded 0 edges**.
- [ ] Run parallel detection over the loaded matn corpus (34,028 Hadith nodes have matn) and load `PARALLEL_OF` edges with `similarity_score`, `variant_type`, `cross_sect`.
- [ ] Add a conformance/sanity counter so an empty PARALLEL_OF load is surfaced, not silent.

## Relationship to the data spine
Sibling to data-acquisition#158 (narrator segmentation). Parallel detection on matn is somewhat independent of narrator quality (it can work on text), so this may be fixable ahead of / alongside #158. Part of the P5 "data is real & queryable" criterion (noorinalabs-main#665). P5W3 backlog.

parametrization:
## Summary
`/compare` "Browse Parallels" was permanently empty: staging Neo4j had **0** `PARALLEL_OF` edges. This makes the deterministic parallel detector scale to the production corpus and adds an empty-load conformance counter so this regression cannot recur silently.

## Diagnosis — ran, but produced 0 rows (not an endpoint mismatch)
- `APPEARS_IN` loaded **34,028** edges using the *same* `hadith_node_id()` endpoint matching the `PARALLEL_OF` loader uses → hadith-id identity is sound. So the loader was fine; **`parallel_links.parquet` itself was empty.**
- Why empty:
  - The deterministic detector (`src/resolve/parallels.py`) used an exhaustive **O(n²)** scan — ~34k hadith ⇒ ~580M pairs, which cannot complete in a pipeline run. Its own "Scale note" already flagged that blocking was required for full-corpus volume.
  - The semantic path (`src/resolve/dedup.py`) **degrades to an empty table** when the embedding model is unavailable (CI / no-model load environments).
  - Net: the composed `parallel_links.parquet` is empty → `_load_parallel_of` loads 0, silently (INFO).

## Changes
- **`src/resolve/parallels.py`** — candidate pairs are now generated by **inverted-index blocking** on discriminative tokens (document frequency `<= max_block_df`, default 500) instead of the all-pairs scan.
  - Corpora no larger than the cap keep **every** token as a blocking key, so the result is **identical to the exhaustive scan** — the loaded slice and existing unit tests are bit-for-bit unaffected.
  - On the full corpus only stop-word-frequency tokens are skipped; genuine near-duplicates share rarer content tokens, so recall on the lexical near-duplicates this detector targets is preserved (the embedding path remains the high-recall production answer).
  - WARN when hadiths are present but no pair is emitted.
- **`src/graph/load_edges.py`** — conformance counter (issue ask #3): WARN (not silent INFO) when `parallel_links.parquet` is empty, when links are present but **0** edges load (endpoint mismatch / silent-failure mode), and a per-edge-type zero-count summary in `load_all_edges`.

## Tests (production-shaped fixtures)
- Duplicate clusters among unique distractors: proves PARALLEL_OF links are **emitted** at scale, that **blocking == the exhaustive scan** on the same corpus, and that the emitted links **load** as edges via the loader.
- Blocking unit tests (`_build_postings`, `_candidate_partners`) proving common tokens are skipped and rare tokens retained.
- Conformance-warning tests for the empty-file and all-endpoints-missing cases.
- Full suite: **878 passed, 5 skipped**; ruff (format+lint) and mypy clean.

## Not covered here (runtime / ops)
I could not exercise the live staging Neo4j from the sandbox (cluster-internal only). The post-merge verification is re-running the resolve→load pipeline on staging and confirming `MATCH ()-[r:PARALLEL_OF]->() RETURN count(r) > 0`; the new conformance WARN will make any future empty load loud in the run logs.

Closes #160


parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved

**Review: LGTM** — inverted-index blocking on discriminative tokens replaces the exhaustive O(n squared) scan that could not complete on the ~34k production corpus, and the empty/zero-edge load is now surfaced as a warning instead of passing silently.

Must-fix: None
TechDebt: None

Verified against the diff at PR head. The blocking-equals-exhaustive claim holds — a pair sharing no token has Jaccard 0, below any positive threshold, so it never becomes a record; with the cap above corpus size every token stays a blocking key and the candidate set reproduces the exhaustive pair set bit-for-bit. The default cap (500, a constant not a fraction) keeps every token eligible on the loaded slice and in CI, so results there are unchanged. The production-shaped test proves edges are actually emitted at scale (high-df scaffold tokens skipped, intra-cluster pairs still recovered) AND asserts equality with the max_block_df=100000 exhaustive reference. Conformance: per-zero-type warning plus parallel_of_no_edges, parallel_of_loaded_zero, and all-endpoints-missing are all covered with structlog capture. All 5 checks pass.

parametrization:
RequestOrReplied: Approved

**Review: LGTM** — inverted-index blocking is a sound, correct scale fix and the empty/zero-load conformance warnings close the silent-failure gap that left /compare blank.

What I verified against the diff at head bfae3be8 (reading the code, not the PR body).

- Blocking equivalence. `_candidate_partners` only emits pairs that share a token whose document frequency is at most `max_block_df`. Any pair sharing no token scores Jaccard 0, which is below any positive threshold and is rejected by the exhaustive scan anyway — so dropping those candidates loses nothing. When the corpus is no larger than the cap, every token's df is bounded by n and therefore eligible, so blocking enumerates every token-sharing pair and the surviving set is identical to the old all-pairs scan. The 47-hadith loaded slice and the unit corpora sit far under the 500 floor, so existing results are unchanged. The recall trade-off above the cap (genuine near-dupes that share only stop-word-frequency tokens) is real, openly documented, and delegated to the FAISS path — a reasonable call for a deterministic CI-safe complement.
- The cap=10 test (`test_emits_edges_and_matches_exhaustive_scan`) captures the blocked pair set as a value before the exhaustive run overwrites the parquet, so its blocked-equals-exhaustive assertion is a genuine comparison and passes.
- Conformance counters in load_edges. Empty links warn (parallel_of_no_edges), links-present-but-zero-loaded warns (parallel_of_loaded_zero), and load_all_edges emits a per-zero-type summary — exactly the da#160 PARALLEL_OF=0 regression made loud. Detector side mirrors this (warning when candidates exist but no pair clears the threshold).
- Ran the two new test files at head locally (55 passed) and confirmed all five CI checks are green.

Must-fix: None
TechDebt: In test_parallels.py::TestProductionShapedCorpus.test_default_cap_equivalent_to_exhaustive_on_slice, both detect_parallels calls write the same fixed parallel_links.parquet, so reading both paths after the second run compares the exhaustive output to itself — that one default-cap equivalence assertion is tautological (the real guard lives in the sibling cap=10 test). Worth tightening to capture the default-run pair set as a value before the overwrite, or write to distinct staging dirs.

Issue #174, opened 2026-06-16
Staging: load multi-source Sunni corpora (open_hadith, halimbahae, fawaz, sanadset, mis) → staging Neo4j


parametrization:
## Goal
Load the **already-coded** multi-source Sunni adapters onto **staging Neo4j** so `/admin/data/sources` surfaces more than the current two values (`lk` + `sunnah`). This is the direct fix for the owner-observed "only two ingestion sources in the UI."

## Context (CorpusSources investigation 2026-06-16)
The "two sources" is **not a code gap** — `src/adapters.py` SOURCE_REGISTRY has 12 adapters all wired into `run_all`. The UI shows two because only `lk` (33,981) + `sunnah` (47) were ever bulk-loaded to staging. The Kafka pipeline has never run on staging; loads to date are manual produce→graph-load (matches memory "Staging pipeline not wired").

## Scope
Acquire→parse→load these proven-E2E-locally Sunni adapters onto staging, `sect=sunni`:
- `open_hadith` (da#87/PR#111)
- `halimbahae` (da#96)
- `fawaz` (da#88/PR#119)
- `sanadset` (da#89/PR#106)
- `mis` (da#97)

## Acceptance
- `GET /admin/data/sources` on staging lists >2 source_corpus values.
- Each source's Hadith/Collection nodes present with correct `source_corpus` + provenance (cleanly removable).
- Spot-check counts logged.

## Notes
W5 = "expand on staging, then cut over full corpus" (owner 2026-06-16). This is the sunni-expansion half; prod cutover (deploy#470) loads the full expanded corpus after validation.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Alejandra Reyes-Fuentes
RequestOrReplied: Request

**Wave 5 Kickoff — Phase 5**

This issue is assigned to you for p5-wave-5.
- Peer reviewer: Kavitha Sundaramurthy
- Secondary reviewer: Kwesi Boateng
- Branch from: `deployments/phase-5/wave-5`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: feature (per charter § Wave Planning & Priority)

Please begin implementation.

parametrization:
[...]

PR #180, opened 2026-06-17
da#174: load multi-source Sunni corpora → staging Neo4j (containerized loader)

parametrization:
## Summary
da#174 — load the multi-source Sunni corpora to staging Neo4j so the UI surfaces more than the two existing sources (`lk` + `sunnah`).

The 5 Sunni adapters were already merged; the gap was a repeatable way to run the graph loader against the cluster Neo4j, which is reachable only on the internet-isolated `noorinalabs_backend` docker network (no host-published bolt port, no egress for a runtime pip install). This PR adds that transport and uses it to load staging.

### What's in this PR
- `Dockerfile.load` — self-contained loader image (deps + loader code baked at build time) so it runs attached to `noorinalabs_backend` without needing egress. The deployed API image can't be reused: it ships the neo4j driver but not pyarrow, which the Parquet loaders require.
- `scripts/load_staging.sh` — ships the node-bearing staging Parquet, builds the image, runs the loader on the backend network, and reads the Neo4j password from the running neo4j container's `NEO4J_AUTH` (no secret committed or shell-logged). Default is a **nodes-only** load: idempotent (MERGE on corpus-namespaced ids), per-source, and with no dependency on the shared narrator-resolve output — so it is conflict-free with the parallel da#175 / da#176 staging loads.

### Staging load result (read back from staging Neo4j)
`/admin/data/sources` (`MATCH (h:Hadith) ... group by source_corpus`) now returns **7** distinct sources (was 2):

| source_corpus | hadiths |
|---|---|
| sanadset | 650,986 |
| halimbahae | 62,169 |
| open_hadith | 62,169 |
| fawaz | 36,435 |
| lk (pre-existing) | 33,981 |
| mis | 7,748 |
| sunnah (pre-existing) | 47 |

Hadith created=819,507 (merged=77 — fawaz ara/eng same-hadith dedup), Collection=28, Grading=21,182, total_errors=0, validation_passed.

### Acceptance
- [x] `GET /admin/data/sources` lists >2 source_corpus values (7).
- [x] Each source's Hadith/Collection nodes present with correct `source_corpus` provenance (cleanly removable via `MATCH (n {source_corpus:'<corpus>'}) DETACH DELETE n`).
- [x] Spot-check counts logged (above).

### Notes
- **Nodes-only by design.** Narrator nodes + chain/transmission edges come from the shared *global* resolve (`narrators_canonical.parquet`), which is a single combined step feeding the prod cutover (deploy#470) — deliberately kept out of the 3 parallel staging loads to avoid clobbering the shared canonical set.
- `sanadset` / `mis` emit no `collections_*` Parquet (narrator-record / multi-isnad centric); their Hadith nodes still carry `source_corpus`, so they surface as sources.
- The known sanadset hadith-id double-prefix is handled by the identity layer's collapse guard (node ids stay single-prefixed); the high warning volume is log-bloat only, not a correctness issue.

### Tests / checks
ruff format + check clean, mypy clean (78 files), pytest 897 passed / 5 skipped / 45 deselected. No secrets committed.

Refs #174

### DIFF
--- a/Dockerfile.load
+++ b/Dockerfile.load
@@ -1,0 +1,37 @@
+# syntax=docker/dockerfile:1
+# One-off graph-load image for staging / prod ingest (da#174).
+#
+# Why a dedicated image: the target Neo4j (``bolt://neo4j:7687``) is reachable
+# only on the internet-isolated ``noorinalabs_backend`` docker network, so a
+# loader container attached to that network CANNOT ``pip install`` at run time.
+# This image therefore bakes every dependency + the loader code at BUILD time
+# (on a network with egress, e.g. the default bridge), and is then RUN attached
+# to ``noorinalabs_backend`` (no egress needed). Staging Parquet is bind-mounted
+# at run time under ``/app/data`` — nothing data-bearing is baked into the image.
+#
+# Driven by ``scripts/load_staging.sh`` (the runbook). The deployed isnad-graph
+# API image cannot be reused for this: it ships the neo4j driver but not pyarrow,
+# which the Parquet node/edge loaders require.
+FROM python:3.14-slim AS base
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends gcc libpq-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+FROM base AS builder
+WORKDIR /app
+COPY pyproject.toml uv.lock ./
+RUN pip install --no-cache-dir uv && uv sync --frozen --no-dev --no-install-project
+
+FROM base AS runtime
+WORKDIR /app
+COPY --from=builder /app/.venv /app/.venv
+COPY pyproject.toml uv.lock ./
+COPY src ./src
+COPY queries ./queries
+ENV PATH="/app/.venv/bin:$PATH"
+# Data dirs (data/staging, data/curated) are bind-mounted at run time.
+# Invoked via the module path (not the console script) so no project install is
+# required in the image. Default command is a nodes-only load; override the
+# command (e.g. `load` for full nodes+edges) at `docker run` time.
+ENTRYPOINT ["python", "-m", "src.cli"]
+CMD ["load", "--nodes-only"]
--- a/scripts/load_staging.sh
+++ b/scripts/load_staging.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+# Build + run the one-off graph-load image against a remote cluster's Neo4j.
+#
+# da#174 — multi-source staging graph load. The cluster Neo4j (bolt://neo4j:7687)
+# is reachable only on the internet-isolated `noorinalabs_backend` docker network,
+# so we cannot tunnel a host port to it and cannot pip-install inside a container
+# attached to that network. Instead we build a self-contained loader image
+# (Dockerfile.load) where deps + code are baked at build time, ship the staging
+# Parquet, and run the loader attached to `noorinalabs_backend`.
+#
+# Idempotent: the loaders MERGE on stable, corpus-namespaced ids, so re-running is
+# safe. Reversible: every node carries `source_corpus`, so a source is cleanly
+# removable with `MATCH (n {source_corpus:'<corpus>'}) DETACH DELETE n`.
+#
+# The Neo4j password is read from the running neo4j container's NEO4J_AUTH on the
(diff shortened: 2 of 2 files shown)


parametrization:
RequestOrReplied: Approved

Reviewed at HEAD 6135a4c (Dockerfile.load + scripts/load_staging.sh), read-only via gh api contents — no checkout in any shared tree. CI is green across Lint, Type Check, Test, Integration, and the sync-drift gate.

Loader correctness — sound. The build-time dependency bake (multi-stage, `uv sync --frozen --no-dev --no-install-project`, then run via `python -m src.cli` from /app) correctly sidesteps the no-egress backend network, and pyarrow lives in the default non-ml dependency set so the Parquet node/edge loaders have it. src/ and queries/ are copied; data dirs are bind-mounted, so nothing data-bearing is baked into the image.

Secrets — clean. The Neo4j password is fetched at run time from the container's NEO4J_AUTH on the remote host and injected only into the ephemeral `docker run` env; nothing is committed or left in this machine's shell history. The remote heredoc escaping (\$PW / \$()) keeps expansion cluster-side, and the SC2087 disable is correctly scoped to that intentional case.

nodes-only safety — good default. CMD and LOAD_ARGS default to `load --nodes-only`, and the rsync ships only hadiths_/collections_ parquet, so no narrator or edge files are even present; under strict=False the narrator/chain loaders skip cleanly. The load is idempotent (MERGE on corpus-namespaced ids) and reversible (source_corpus DETACH DELETE) as documented.

source_id/sect namespacing — not altered by this diff (it lives in the parsers); the step-4 read-back groups by source_corpus, which is the right provenance key.

Three non-blocking observations, none gating approval:
1) Because nodes-only skips edges, hadith to collection APPEARS_IN links are NOT created on the default path — collection-faceted views will need a full `LOAD_ARGS="load"` run. Worth confirming that matches the da#174 acceptance, or running the full load once it is safe alongside the narrator data.
2) The step-4 read-back reports source_corpus counts only; adding a sect breakdown would make the Sunni provenance explicit in the runbook output.
3) Image dependency correctness (pyarrow present without the ml group) is not exercised by CI's integration job, which runs against the local env rather than this image — a smoke `docker build -f Dockerfile.load` in CI would catch image-only drift later. Nice-to-have.

Approving — a clean, well-documented one-off loader with safe defaults and no secret exposure.

parametrization:
Requestor: Kwesi Boateng
RequestOrReplied: Approved

Secondary review (da#174). Read the diff and supporting code at head `6135a4c`.

**What I verified**

- **Nodes-only / zero-narrator safety holds — and not only via the `--nodes-only` default.** The real guard is the rsync include-filter in `load_staging.sh`: it ships only `hadiths_*.parquet` and `collections_*.parquet` (`--exclude='*'`), so `narrators_canonical.parquet` never reaches the remote. Even with `--delete` cleaning a stale prior copy, the loader has no narrator parquet to load. Combined with `CMD ["load", "--nodes-only"]` (which I traced through `_cmd_load` → `load_all(..., nodes_only=...)` in `src/cli.py`), this is genuinely conflict-free with the shared global narrator resolve feeding deploy#470. This respects the no-clobber narrator rule.
- **No secret committed.** Password is read at run time from the neo4j container (`docker exec ... printenv NEO4J_AUTH | cut -d/ -f2`) inside a remote heredoc with `\$PW` escaped so it never touches this machine's shell history. Nothing data-bearing is baked into the image (bind-mounted at run time).
- **Namespacing / idempotency** are properties of the already-merged adapters, not this PR, and are empirically demonstrated by the read-back (7 distinct `source_corpus`, created=819,507, merged=77 same-hadith dedup, total_errors=0). Cleanly reversible per source via the documented `source_corpus` DETACH DELETE.
- **Image rationale is sound** — the API image lacks pyarrow; baking deps at build time on an egress network and running attached to the isolated `noorinalabs_backend` is the right shape.

**Non-blocking nits (optional, no change required to merge)**

- `cut -d/ -f2` truncates a password containing a `/`. `cut -d/ -f2-` is more robust. Staging's password is controlled, so this is latent only.
- `-e NEO4J_PASSWORD="$PW"` is visible in `docker inspect` / the container env on the remote host. Acceptable for a one-off staging/prod load; just noting the surface.

Approving. Clean, well-documented transport that does exactly what da#174 needs without touching the shared narrator set.

Issue #172, opened 2026-06-14
tech-debt(test): tighten tautological test_default_cap_equivalent_to_exhaustive_on_slice (da#160 follow-up)


parametrization:
## Source
TechDebt item recorded by Alejandra Reyes-Fuentes in her Approved verdict on PR #170 (da#160 PARALLEL_OF). Non-blocking; approval stands.

## Issue
`test_default_cap_equivalent_to_exhaustive_on_slice` is tautological: both `detect_parallels` calls write the same fixed `parallel_links.parquet`, so reading both paths after the second run compares the exhaustive result to itself (p1==p2 is the same path). It does not actually prove blocked==exhaustive.

The real equivalence guard already lives in the sibling `cap=10` test (which captures the blocked set as a value before the overwrite). Test-tightening follow-up only.

## Fix
Capture the first run's output to a distinct value/path before the second run overwrites, then compare — OR remove as redundant with the cap=10 sibling test.

## Acceptance
- [ ] The equivalence test compares two genuinely distinct computed results (or is removed as redundant).

parametrization:
Requestor: Fatima Okonkwo
Requestee: Oyunbileg Batbayar
RequestOrReplied: Request

**Wave 5 Kickoff — Phase 5**

This issue is assigned to you for p5-wave-5.
- Peer reviewer: Nikolaos Papadopoulos
- Secondary reviewer: Kavitha Sundaramurthy
- Branch from: `deployments/phase-5/wave-5`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

PR #185, opened 2026-06-17
test(resolve): tighten tautological cap-equivalence test (da#172)

parametrization:
## Summary

Tightens the tautological `test_default_cap_equivalent_to_exhaustive_on_slice` (da#160 follow-up, tech-debt recorded in Alejandra Reyes-Fuentes' Approved verdict on PR #170).

The test read `parallel_links.parquet` twice **after** both `detect_parallels` runs had completed. Both runs write to and return the **same** `staging/parallel_links.parquet`, so the second (exhaustive) run overwrote the default run's output — the assertion compared the exhaustive result to itself and would pass even if the default-config run regressed.

## Fix

Capture each run's pair set as a value **before** the next call overwrites the shared artifact (mirroring the sibling `test_emits_edges_and_matches_exhaustive_scan` idiom), and anchor both to the independently computed `expected` intra-cluster set returned by `_build_corpus`.

## Verification (non-tautological)

- All 897 non-integration tests pass on correct code.
- Temporarily lowering `_DEFAULT_MAX_BLOCK_DF` to `1` (simulated default-cap regression) makes the tightened test **FAIL** — default run yields 0 pairs vs exhaustive's 4, so `default_pairs` diverges from both `expected` and `exhaustive_pairs`. The old read-twice form would still have passed under that regression.
- Production-realistic corpus retained (no fixture-masking, main#671).
- ruff format/lint clean; CI mypy scope (`src/` only) unaffected.

## Acceptance

- [x] The equivalence test compares two genuinely distinct computed results.

Closes #172

Reviewers: Nikolaos Papadopoulos, Kavitha Sundaramurthy

### DIFF
--- a/tests/test_resolve/test_parallels.py
+++ b/tests/test_resolve/test_parallels.py
@@ -308,15 +308,27 @@ class TestProductionShapedCorpus:
         assert _pair_set(exhaustive) == blocked_pairs
 
     def test_default_cap_equivalent_to_exhaustive_on_slice(self, tmp_path: Path) -> None:
-        # Below the default cap every token stays eligible, so the default-config
-        # run is bit-for-bit the exhaustive scan (loaded-slice / CI guarantee).
+        # On a loaded slice no larger than the default cap every token's df stays
+        # under the cap, so the default-config run must recover the SAME pair set
+        # as the exhaustive (cap-above-corpus) scan (loaded-slice / CI guarantee).
         staging = tmp_path / "staging"
         staging.mkdir()
-        self._build_corpus(staging, clusters=4, per_cluster=2, distractors=20)
-        default_run = detect_parallels(staging, threshold=0.5)
-        exhaustive = detect_parallels(staging, threshold=0.5, max_block_df=100_000)
-        assert _pair_set(default_run) == _pair_set(exhaustive)
-        assert len(_pair_set(default_run)) == 4  # C(2,2) * 4 clusters
+        expected = self._build_corpus(staging, clusters=4, per_cluster=2, distractors=20)
+
+        # Capture each run's pair set as a value BEFORE the next call overwrites the
+        # shared parallel_links.parquet. Both detect_parallels calls return the same
+        # path, so reading both paths only after the second run would compare the
+        # exhaustive result to itself rather than to the default run (da#172).
+        default_pairs = _pair_set(detect_parallels(staging, threshold=0.5))
+        exhaustive_pairs = _pair_set(detect_parallels(staging, threshold=0.5, max_block_df=100_000))
+
+        # The default cap loses none of the real intra-cluster pairs relative to the
+        # exhaustive scan: C(2,2) * 4 clusters = 4. Anchoring to the independently
+        # computed expected set means a default-config regression that drops pairs
+        # makes default_pairs diverge from BOTH expected and exhaustive_pairs.
+        assert default_pairs == expected
+        assert default_pairs == exhaustive_pairs
+        assert len(default_pairs) == 4
 
     def test_corpus_with_no_pairs_warns(self, tmp_path: Path) -> None:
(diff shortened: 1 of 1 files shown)


parametrization:
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

Reviewed the diff at head plus the `_build_corpus` helper.

The tautology is genuinely removed. The old form read `parallel_links.parquet` for both runs only after both `detect_parallels` calls had finished, and since both write the same shared path, the second (exhaustive) run had already overwritten the first — so the assertion compared the exhaustive result to itself and would pass through a default-config regression. The fix captures `default_pairs` and `exhaustive_pairs` as values immediately after each call, before the next overwrites the artifact, which is the right idiom and matches the sibling `test_emits_edges_and_matches_exhaustive_scan`.

The anchor is real, which is what makes this non-tautological rather than just rearranged. I confirmed `_build_corpus` returns a `set[tuple[str, str]]` of intra-cluster pairs constructed directly from the cluster ids — independent of any `detect_parallels` output. Asserting `default_pairs == expected` AND `default_pairs == exhaustive_pairs` means a default-cap regression that drops pairs diverges from both anchors, so the test fails. The regression-injection check in the PR body (`_DEFAULT_MAX_BLOCK_DF` lowered to 1 yields 0 default pairs vs 4 exhaustive and fails the test) demonstrates exactly that, and the old read-twice form would have stayed green under it.

The production-realistic corpus is retained (high-df scaffold tokens plus distractors and real clusters), so this satisfies the fixture-realism rule rather than shrinking to a toy. CI is green across lint/type/test/integration/sync-drift, and the change is test-only.

No blockers. Approving.

parametrization:
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

Reviewed at HEAD 19439bb (tests/test_resolve/test_parallels.py), read-only via gh api contents — no checkout in a shared tree. CI green across Lint, Type Check, Test, Integration, and the sync-drift gate.

Non-tautology — fixed correctly. The original compared `_pair_set(default_run)` to `_pair_set(exhaustive)` only AFTER both detect_parallels calls ran, but both return the same shared parallel_links.parquet, so the second run had overwritten the first and the assertion compared the exhaustive output to itself. The new code reads each run's output into a value (`default_pairs = _pair_set(detect_parallels(...))`) BEFORE the exhaustive run overwrites the file, so the two sides are now genuinely distinct computations. The inline comment names the overwrite hazard precisely.

Independently-computed expected set — correct. `_build_corpus` now returns `expected`, enumerated structurally from the corpus layout (sorted ids, C(per_cluster,2) per cluster), with no dependency on detect_parallels — a real oracle. For clusters=4/per_cluster=2 that is 4 pairs, matching the `len == 4` check. The triple assertion (default_pairs == expected, default_pairs == exhaustive_pairs, len == 4) means a default-config regression that drops pairs diverges from BOTH the oracle and the exhaustive scan, which is exactly the strengthening da#172 wanted.

No flakiness — the corpus is fully deterministic (fixed vocabulary and ids, no randomness); distractors carry distinct tokens so no accidental pair clears threshold 0.5, and the intra-cluster pairs share cluster-local tokens with df well under any cap, so default and exhaustive recover the identical set regardless of whether the shared scaffold is capped. Pair-tuple ordering (a,b in sorted-id order) is already validated by the sibling exhaustive-vs-blocked test, so the set comparison is stable.

One context note, non-blocking — this slice (28 hadiths) sits below the default cap, so the default-vs-exhaustive equivalence here is the small-slice/CI guarantee; the stronger blocking-with-capping vs exhaustive equivalence is carried by the sibling test_emits_edges_and_matches_exhaustive_scan (cap=10). The division of coverage is sound.

Approving — a real tautology removed and replaced with an oracle-anchored, deterministic equivalence check.

Issue #177, opened 2026-06-16
Riyad as-Salihin: complete to ~1,896 or drop the 47-hadith fragment


parametrization:
## Goal
Resolve the Riyad as-Salihin fragment: only **47 of ~1,896** hadiths are on staging (legacy da#73/PR#76 scraper fragment). da#167 acceptance item 3.

## Scope (owner decision needed at most — small)
Either complete the Riyad as-Salihin scrape/load to ~1,896 **or** drop the 47-hadith fragment if it's not worth completing pre-launch.

## Acceptance
- Riyad as-Salihin is either complete on staging OR explicitly removed (no half-corpus on prod).

parametrization:
Requestor: Fatima Okonkwo
Requestee: Ivana Horvat
RequestOrReplied: Request

**Wave 5 Kickoff — Phase 5**

This issue is assigned to you for p5-wave-5.
- Peer reviewer: Kavitha Sundaramurthy
- Secondary reviewer: Oyunbileg Batbayar
- Branch from: `deployments/phase-5/wave-5`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: feature (per charter § Wave Planning & Priority)

Please begin implementation.

PR #187, opened 2026-06-17
fix(sunnah_scraper): complete Riyad as-Salihin to 1,896 — enumerate named book segments (da#177)

parametrization:
## da#177 — Riyad as-Salihin: COMPLETE to 1,896 (decision: complete, not drop)

**Decision: complete.** sunnah.com is reachable and the full corpus parses cleanly, so per the wave's full-corpus goal I completed the collection rather than dropping the fragment.

### Root cause of the 47-fragment
The fragment was never a data-availability problem — it was a scraper enumeration bug. `_get_book_numbers` kept only **digit-only** URL segments, so sunnah.com's *named* first book `/riyadussalihin/introduction` ("The Book of Miscellany") was silently dropped. That book holds **679 of the 1,896** hadiths (the famous "actions are by intentions" is in it). The enumerator topped out at 1,217 (books 1–19); the legacy da#73 first-light then loaded only book 1 (47), which is the fragment that reached staging.

`679 (introduction) + 1,217 (books 1–19) = 1,896` ✓

### Code
- `_get_book_numbers` → `list[int | str]`: numeric books (sorted ints) **plus** named segments (introduction) as strings, ordered first. Folds `#anchor`/`?query` variants (e.g. hisn `introduction#C0.00`) and skips multi-component / home / empty hrefs.
- `_scrape_book_page` accepts a named URL segment and derives an int `book_key` (named → 0) for chapter numbering + the per-row `book_number` fallback. Introduction hadiths have no "Book N" in-book ref, so `hadith_number` falls back to the collection-wide reference (1–679), keeping every `source_id` unique.
- `run_slice.py --all-books`: reproducible full-collection live scrape (the path used here); added to the runbook examples.
- The fix is general (hisn / mishkat also have an `introduction`; shamail an `8b`) — a correctness win for the prod-cutover full corpus — but **only riyadussalihin was loaded** for this issue.

### Tests (no fixture-masking, main#671)
- `test_real_index_markup_includes_introduction` — a **real captured sunnah.com index** fixture asserts `introduction` is enumerated, named-first, alongside /1..19 (the exact axis the bug lived on).
- anchor-fold / subpath-skip unit coverage; named-segment scrape keying.
- opt-in live leg `TestRiyadussalihinLiveUpstream` (`RUN_LIVE_SUNNAH_SCRAPE=1`) asserts the real scrape reaches ≥1,890 with Arabic+English text and non-null numbers.
- Full local suite: **945 passed, 7 skipped** (Docker integration incl. the scraper multicollection test ran green). ruff + `mypy src/` clean.

### Data quality
Full scrape: **1,896 hadiths, 1,896 distinct source_ids, 0 collisions.** Strict staging validation passes every data-quality check (Arabic coverage 100%). `grade` is null because sunnah.com publishes no inline grade for this collection (nullable field, verified 0 grade markup on the pages). The only strict miss is a `row_count` drift vs the multi-collection `sunnah_scraped` baseline (10,028) — expected when validating one collection, not a defect.

### Staging load (via the da#174 reusable loader image, not a divergent loader)
1. Deleted the stale 47-fragment (its node ids ended `:0` — loaded pre-in-book-ordinal — so they would NOT have MERGE-coalesced and would have left duplicates).
2. Shipped the riyad parquet and ran `noorinalabs-graph-load:da174 load --skip-validation` on `noorinalabs_backend`.

Read-back on staging:
- `riyadussalihin` Hadith nodes = **1,896**
- `APPEARS_IN` edges = **1,896** (0 missing endpoints)
- `source_corpus='sunnah'` total = **1,896**, distinct ids = 1,896, **stale `:0` suffix = 0**
- admin sources unaffected (9 source_corpus values)

End state: **complete on staging, no fragment, no duplicates.**

Reviewers: @Kavitha-Sundaramurthy @Oyunbileg-Batbayar

Closes #177

### DIFF
--- a/scripts/first_light/run_slice.py
+++ b/scripts/first_light/run_slice.py
@@ -32,6 +32,10 @@ Examples
 
     # Live re-scrape riyadussalihin book 1, then load
     uv run python scripts/first_light/run_slice.py --live --collection riyadussalihin --book 1
+
+    # Complete a whole collection live (every book incl. the introduction) and
+    # load it — da#177 used this to take riyadussalihin to its full 1,896.
+    uv run python scripts/first_light/run_slice.py --live --all-books --collection riyadussalihin
 """
 
 from __future__ import annotations
@@ -80,6 +84,35 @@ def _acquire_live(raw_dir: Path, collection: str, book: int) -> Path:
     out.write_text(json.dumps(hadiths, ensure_ascii=False, indent=2), encoding="utf-8")
     write_manifest(dest, [out])
     print(f"[acquire:live] scraped {len(hadiths)} hadiths from {collection} book {book} -> {out}")
+    return dest
+
+
+def _acquire_live_full(raw_dir: Path, collection: str) -> Path:
+    """Live-scrape the ENTIRE collection (all books, incl. the introduction).
+
+    Unlike ``_acquire_live`` (one ``--book``), this walks every book the index
+    lists — the path used to *complete* a collection on staging (da#177:
+    riyadussalihin -> 1,896). Rate-limited by ``sunnah_scraper``; needs network.
+    """
+    import httpx
+
+    from src.acquire.base import ensure_dir
+    from src.acquire.sunnah_scraper import (
+        REQUEST_TIMEOUT,
+        USER_AGENT,
+        _scrape_collection,
+    )
+
+    dest = ensure_dir(raw_dir / "sunnah_scraped")
+    client = httpx.Client(
+        headers={"User-Agent": USER_AGENT}, timeout=REQUEST_TIMEOUT, follow_redirects=True
+    )
+    try:
+        out = _scrape_collection(client, collection, dest)
+    finally:
+        client.close()
+    count = len(json.loads(out.read_text(encoding="utf-8"))) if out is not None else 0
+    print(f"[acquire:live-full] scraped {count} hadiths from {collection} (all books) -> {dest}")
     return dest
 
 
@@ -183,6 +216,11 @@ def main() -> None:
     parser.add_argument("--collection", default="riyadussalihin", help="Collection (live mode)")
     parser.add_argument("--book", type=int, default=1, help="Book number (live mode)")
     parser.add_argument(
+        "--all-books",
+        action="store_true",
+        help="With --live: scrape the whole collection (every book incl. the introduction)",
+    )
+    parser.add_argument(
         "--dry-run",
         action="store_true",
         help="Acquire + parse only; skip the Neo4j load (no DB connection)",
@@ -204,7 +242,9 @@ def main() -> None:
 
     print(f"=== first-light slice (da#73) — work_dir={work_dir} ===")
 
-    if args.live:
+    if args.live and args.all_books:
+        _acquire_live_full(raw_dir, args.collection)
+    elif args.live:
         _acquire_live(raw_dir, args.collection, args.book)
     else:
         _acquire_sample(raw_dir)
--- a/src/acquire/sunnah_scraper.py
+++ b/src/acquire/sunnah_scraper.py
@@ -200,13 +200,29 @@ def _extract_hadith_from_row(row: Tag) -> dict[str, Any] | None:
 def _scrape_book_page(
     client: httpx.Client,
     collection: str,
-    book_number: int,
+    book_number: int | str,
 ) -> list[dict[str, Any]]:
-    """Scrape all hadiths from a single book page."""
+    """Scrape all hadiths from a single book page.
+
+    ``book_number`` is the URL path segment: an ``int`` for numbered books or a
+    name such as ``"introduction"``.  A numeric *key* is derived for chapter
+    numbering and the per-row ``book_number`` fallback; named segments carry no
+    inherent number, so they key to ``0`` and their hadiths fall back to the
+    collection-wide reference for source_id keying (see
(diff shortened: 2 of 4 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Ivana Horvat
RequestOrReplied: Approved

Reviewed the diff at head 78cc86e5 (4 files, +261/-22): sunnah_scraper enumeration fix, the run_slice `--all-books` helper, the captured index fixture, and the new tests.

Named-book enumeration fix is correct. `_get_book_numbers` now collects both numeric and named single-component segments (named ordered first), folds `#anchor`/`?query` onto the bare segment, and skips the empty home link and multi-component permalinks — I confirmed against the captured fixture that the home `/riyadussalihin` and `/contactus` links are excluded and `introduction` is discovered. The `book_number: int | str` widening in `_scrape_book_page` derives `book_key` = the int for numeric segments, `int(seg)` for digit strings, else 0; the integer branch is byte-identical to the previous behavior, so numbered books are unchanged and there is no regression on the existing path.

source_ids stay unique. The downstream key is `sunnah:<coll>:<book>:<chapter>:<hadith_number>` (src/parse/sunnah_scraped.py). Introduction rows carry "Introduction, Hadith N" which does not match the `Book \d+` in-book regex, so they fall back to book 0 and to the collection-wide reference number for hadith_number — and that reference is unique across the whole collection (1..679 for the introduction). Numbered books key to book >= 1 with their in-book ordinals. The book segment differing (0 vs >=1) keeps the introduction namespace disjoint from every numbered book, and the globally-unique collection-wide ref keeps the introduction rows distinct among themselves regardless of chapter. The new `test_named_introduction_segment` pins exactly this (book_number 0, hadith_number from the sticky ref).

Real-upstream guard (main#671) is honored well. The committed `riyadussalihin_index_sample.html` is real captured markup, and the opt-in `TestRiyadussalihinLiveUpstream` actually fetches sunnah.com to prove the *enumerator* discovers the introduction and that the full scrape reaches ~1,896 — the genuine no-fixture-masking axis — while staying gated behind RUN_LIVE_SUNNAH_SCRAPE so CI remains offline-deterministic.

The 1,896 / 0-stale staging read-back is an operational step (re-scrape, delete the stale fragment, reload) outside this diff and not reproducible from the review sandbox; I am taking it from the issue/PR report and reviewing the unit-mechanic correctness that backs it, which holds.

One optional, non-blocking strengthening for a follow-up: a combined-collection unit test that runs introduction + a numbered book through the parser and asserts the resulting source_id set has no duplicates would lock the cross-book uniqueness argument in code (today the live leg covers the count but not the uniqueness assertion). Not required for this PR.

All five CI checks are green (Lint, Type Check, Test, Integration Tests, sync-drift gate). Solid fix with the right real-data guard. Approving.

parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved

Reviewed at HEAD 78cc86e (src/acquire/sunnah_scraper.py, the saved riyadussalihin index fixture, test_sunnah_scraper.py, run_slice.py), read-only via gh api contents — no shared-tree checkout. CI green across Lint, Type Check, Test, Integration, and the sync-drift gate.

Scraper fix — correct, and the root cause is well diagnosed. The old `_get_book_numbers` filtered `segment.isdigit()`, so sunnah.com's NAMED first book (`/riyadussalihin/introduction`, the Book of Miscellany — 679 of 1,896 hadiths) was silently dropped, truncating the collection. The fix returns `int | str` segments (named ordered first, then numeric ascending), strips `#anchor`/`?query` so `introduction#C0.00` folds onto the bare page, and skips empty and multi-component hrefs so hadith permalinks (`.../1/5`) and the collection home aren't mistaken for books.

source_id safety with named books — verified, this was my main correctness concern. Named segments key to `book_key = 0`, but source_id's terminal segment is the in-book ordinal or, when absent, the collection-wide reference (unique per collection), and the book segment differs (0 vs 1+), so introduction hadiths cannot collide with numbered-book hadiths — nor with each other. chapter_number is namespaced as `book_key*100 + counter`, so introduction chapters (0..) and numbered-book chapters (100..) stay disjoint for any book with under 100 chapters.

Real-upstream fixture — present and on point (main#671). `riyadussalihin_index_sample.html` is the saved sunnah.com index, and `test_real_index_markup_includes_introduction` asserts `["introduction", *range(1,20)]` — a regression that re-drops named books fails CI. The anchor-fold/subpath-skip case is covered with a second fixture. The live full-scrape leg is opt-in via RUN_LIVE_SUNNAH_SCRAPE so CI stays offline-deterministic.

staging 1,896 / no stale fragment — confirmed against staging Neo4j directly: `sunnah` now reads 1,896 sunni hadith (the 47-hadith fragment is gone), which is the da#177 acceptance.

Two non-blocking observations:
1) A named book with 100+ chapters would push `book_key*100 + counter` (book_key 0) into the numbered books' chapter-number range; harmless for source_id (the book segment still differs) but a latent chapter_number ambiguity. riyadussalihin's introduction is well under that, so not a concern here — worth a comment for the next named-book collection.
2) `book_number = 0` for named books is a slightly synthetic provenance value on the APPEARS_IN edge; fine given source_id keys off the ordinal/ref, just noting it.

Approving — a correctly diagnosed enumeration bug, fixed with a real-upstream regression fixture, and the collection is whole (1,896) on staging.

Issue #190, opened 2026-06-17
loader: matn_ar fallback to full_text_ar to fix textless Hadith nodes (halimbahae/bihar)


parametrization:
## Problem
The graph node loader `_HADITH_MERGE` (`src/graph/load_nodes.py`) persists only `matn_ar`/`matn_en`/`isnad_raw_ar`/`isnad_raw_en` onto Hadith nodes. Sources that supply the hadith text in `full_text_ar` but leave `matn_ar` empty (they don't separate isnad from matn) therefore land as **textless graph nodes**.

Confirmed live on staging (pre-cutover corpus audit): all **124,338** `open_hadith` + `halimbahae` Hadith nodes had empty `matn_ar` (matn_en NULL). `open_hadith` has since been dropped as an exact duplicate of `halimbahae`; `halimbahae` (62,169) remains and is currently textless.

## Fix
In the Hadith row-prep, fall back `matn_ar := full_text_ar` when `matn_ar` is empty/whitespace. These corpora do not separate isnad/matn, so the full text is the correct displayable matn. Add a unit test (mock client) asserting the batched row's `matn_ar` is populated from `full_text_ar` when `matn_ar` is empty, and is left untouched when `matn_ar` is present.

This also future-proofs **bihar** (da#95), which has the same latent pattern.

## After merge
Re-load `halimbahae` with the fixed loader so its `matn_ar` is populated on staging (idempotent MERGE updates existing nodes).

## Acceptance
- Hadith rows with empty `matn_ar` + non-empty `full_text_ar` load with `matn_ar = full_text_ar`.
- Rows with a real `matn_ar` are unchanged.
- After re-load, `halimbahae` Hadith nodes on staging have non-empty `matn_ar`.

Found by the pre-cutover corpus-integrity audit. Gates the combined narrator-resolve + prod cutover (deploy#470).

PR #192, opened 2026-06-17
da#190: fall back matn_ar to full_text_ar (fix textless Hadith nodes)

parametrization:
## Summary
Fixes textless Hadith nodes for corpora that populate `full_text_ar` only (halimbahae, open_hadith, bihar) — they don't separate isnad from matn, and the graph loader persists `matn_ar` (not `full_text_ar`), so those nodes landed with empty text.

Found by the pre-cutover corpus-integrity audit: all **124,338** open_hadith+halimbahae Hadith nodes on staging had empty `matn_ar`. open_hadith has since been dropped as a confirmed exact duplicate of halimbahae (audit remediation A); halimbahae (62,169) remains and this unblocks its text.

## Change
`src/graph/load_nodes.py` — in the Hadith row-prep, fall back `matn_ar := full_text_ar` when `matn_ar` is empty/whitespace. A present `matn_ar` is never overwritten. Future-proofs bihar (da#95), same pattern.

## Tests
- `test_matn_ar_falls_back_to_full_text_ar` — empty/None matn_ar populated from full_text_ar.
- `test_matn_ar_preserved_when_present` — real matn_ar untouched.
Full suite: 899 passed / 5 skipped. ruff + mypy clean.

## After merge
Re-load halimbahae with the fixed loader so its `matn_ar` is populated on staging (done out-of-band on staging for validation; will be re-run from the merged image for the cutover).

Closes #190

### DIFF
--- a/src/graph/load_nodes.py
+++ b/src/graph/load_nodes.py
@@ -267,10 +267,18 @@ def _load_hadiths(
                 total_skipped += 1
                 continue
             hid = hadith_node_id(sid)
+            # Fall back to the raw full text when a source supplies no separated
+            # matn: halimbahae / open_hadith / bihar populate ``full_text_ar``
+            # only (they do not split isnad from matn). The loader persists
+            # matn_ar — not full_text_ar — so without this the Hadith node lands
+            # textless on the graph (da#190).
+            matn_ar = _val(row, "matn_ar", "")
+            if not (matn_ar and str(matn_ar).strip()):
+                matn_ar = _val(row, "full_text_ar", "")
             batch.append(
                 {
                     "id": hid,
-                    "matn_ar": _val(row, "matn_ar", ""),
+                    "matn_ar": matn_ar,
                     "matn_en": _val(row, "matn_en"),
                     "isnad_raw_ar": _val(row, "isnad_raw_ar"),
                     "isnad_raw_en": _val(row, "isnad_raw_en"),
--- a/tests/test_graph/test_load_nodes.py
+++ b/tests/test_graph/test_load_nodes.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from pathlib import Path
+from typing import Any
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


parametrization:
RequestOrReplied: Approved

Reviewed at HEAD 9fec003 (src/graph/load_nodes.py + tests/test_graph/test_load_nodes.py), read-only via gh api contents — no shared-tree checkout. CI green across Lint, Type Check, Test, Integration, and the sync-drift gate.

Fallback guard — correct. `if not (matn_ar and str(matn_ar).strip())` fires only when matn_ar is None, empty, or whitespace-only; a populated matn_ar is left untouched. So the corpora that genuinely separate isnad from matn and fill matn_ar (lk, and as of da#175 thaqalayn, da#182 tusi) are not clobbered — I confirmed this against my own da#175 load, where matn_ar is always the Arabic text, so the new branch never fires for thaqalayn and the staging result is unchanged. The textless sources (halimbahae/open_hadith, and future bihar) that populate only full_text_ar now inherit it into the persisted matn_ar, which is exactly the 124k-textless-node fix.

Both branches tested — yes. `test_matn_ar_falls_back_to_full_text_ar` covers both the "" and the None input, and `test_matn_ar_preserved_when_present` pins the no-clobber path. The `_hadith_batch_rows` helper reads the actual batch handed to the loader, so the assertions check the persisted value rather than the input.

No regression — the change is confined to the matn_ar value assembled in `_load_hadiths`; id, sect, source_corpus, collection_name, grade and the APPEARS_IN/grading paths are untouched. A row with both matn_ar and full_text_ar empty still lands with an empty matn_ar, same as before — no node is made worse.

Two non-blocking observations:
1) The fallback is Arabic-only; if halimbahae/open_hadith also leave matn_en empty while carrying full_text_en, the English display stays blank. Out of da#190's scope (the issue is the Arabic textless nodes), but a symmetric matn_en fallback would be the natural follow-up if those sources have full_text_en.
2) When both matn_ar and full_text_ar are empty the node is still textless; that is unavoidable here and worth a one-line note in da#190 so the residual (if any) on the real load is expected rather than surprising.

Approving — a tightly-scoped, well-tested loader fallback that fixes the textless nodes without touching the corpora that already populate matn_ar.

parametrization:
Requestor: Kwesi Boateng
RequestOrReplied: Approved

Secondary review (da#190). Read the diff plus `_val` and `_HADITH_MERGE` at head `9fec003`.

**What I verified**

- **Fallback fires only on an empty matn_ar — no clobber of isnad/matn-separating corpora.** The guard `if not (matn_ar and str(matn_ar).strip())` substitutes `full_text_ar` only when `matn_ar` is missing/None/blank. A source that supplies a real `matn_ar` (thaqalayn da#175, tusi da#182, the Sunni set) keeps it untouched. `test_matn_ar_preserved_when_present` locks that branch.
- **Both empty forms are handled.** I checked `_val` (`return default if v is None else v`), so `_val(row, "matn_ar", "")` yields `""` for both an explicit `None` cell and a missing key — both then fall through to `full_text_ar`. `test_matn_ar_falls_back_to_full_text_ar` exercises the `""` and `None` rows and asserts each lands with the full-text Arabic.
- **No regression in the both-empty case.** If `full_text_ar` is also empty/None, `_val(..., "")` returns `""`, i.e. exactly the prior behavior — the node is no worse off than before, never newly nulled.
- **Idempotent MERGE intact.** `_HADITH_MERGE` is unchanged — still `MERGE (n:Hadith {id: row.id})` keyed on the stable hadith id, `SET n.matn_ar = row.matn_ar`. The fallback is computed per row before the batch is assembled, so re-running produces an identical batch and the same SET — no MERGE-semantics change.
- **Consistent with the existing pattern in this file.** This is a load-time display fallback (not persisted to parquet), mirroring `_narrator_name_en`'s approach in the same module — the parquet stays pure provenance (matn_ar empty for full-text-only sources) while the Neo4j node that `/graph` and search read always carries text.

The read-back (halimbahae 62,169 textless → 0) matches the mechanic. Tests cover both branches. Clean, minimal, correct.

Approving.

Issue #191, opened 2026-06-17
Sunni six-books canonical-edition selection + canonical-hadith dedup for combined resolve


parametrization:
## Problem
The Sunni six canonical books (Bukhari, Muslim, Abu Dawud, Tirmidhi, Nasa'i, Ibn Majah) are loaded on staging from MULTIPLE overlapping editions:
- `lk` (6 books; separated matn_ar + matn_en + grade — richest metadata)
- `halimbahae` (9 books incl. Musnad Ahmad 26,363 + Darimi + Malik — full tashkeel; currently textless, see the matn_ar fallback issue)
- `fawaz` (10 collections; matn_ar + matn_en + partial grade)
- `mis` (Sahih Muslim again; its real value is explicit multi-isnad chains, not matn)

Pre-cutover audit overlap (exact normalized text — a LOWER bound, editions render isnad+matn differently): lk^fawaz 51.6%/49.3%, fawaz^mis 92.5% of mis, lk^mis 56.5% of mis. Without a canonical-edition decision, the same hadith appears as several nodes on prod → inflated counts + duplicate search results.

## Recommendation (owner decision)
- `lk` as the six-books spine (richest metadata).
- `halimbahae` for its UNIQUE books only (Musnad Ahmad, Darimi, Malik).
- `mis` kept for isnad chains only (its matn duplicates fawaz/Muslim).
- The combined narrator-resolve must dedup by CANONICAL hadith identity (collection + number mapping across editions), NOT raw matn (rendering differs).

## Status
Owner is deciding the final composition (relayed via Program Director). This issue tracks the plan; the combined narrator-resolve + prod cutover (deploy#470) are HELD until the composition is confirmed and the dedup/textless fixes land.

Found by the pre-cutover corpus-integrity audit.

PR #193, opened 2026-06-17
da#191: encode canonical corpus composition in the pipeline (per-source dedup)

parametrization:
## Summary
Encodes the owner-confirmed canonical corpus composition **in the pipeline**, so a fresh `run_all` (the path production uses) produces a duplicate-free graph without the manual graph surgery used on staging.

Pre-cutover audit found duplicate corpora: `open_hadith` ≡ `halimbahae` (100%), and the six canonical Sunni books loaded by `lk` + `halimbahae` + `fawaz` + `mis`. Loading every copy double-counts hadith on prod.

## Change
- **`src/parse/composition.py`** (new) — single source of truth `HADITH_COMPOSITION` + `is_canonical_hadith()`:
  - `lk` = six-books canonical spine (all collections; richest metadata).
  - `halimbahae` → unique books only (Musnad Ahmad, Darimi, Malik).
  - `fawaz` → unique collections only (Nawawi, Dehlawi, Qudsi).
  - `mis` → no Hadith nodes (its Sahih Muslim matn duplicates lk; `network_edges_mis` still loads as chains).
  - every other source → all collections.
- **`src/adapters.py`** — add `SourceAdapter.active` (default True); `open_hadith` marked `active=False` (confirmed dup). Registry row retained for the `SourceCorpus` coverage invariant.
- **`src/acquire`, `src/parse` run_all** — skip inactive adapters.
- **`src/graph/load_nodes.py`** — enforce the composition at load: skip non-canonical `(source_corpus, collection)` Hadith + Collection rows. One enforcement point.

## Scope
This is the deterministic **per-source** composition (the part that MUST hold before prod). Cross-edition canonical-identity dedup (same hadith by collection+number across editions) is a tracked fast-follow — flagged to the Program Director.

## Tests
Composition unit tests; load_nodes skips non-canonical collections + mis; adapters active-flag + coverage invariant; provenance-conformance excludes the new framework module. **Full suite 924 passed / 7 skipped**; ruff + mypy clean.

## Validation
This is the dedup encoding for the `run_all` dress-rehearsal staging load (the prod-cutover sign-off artifact), which follows.

Closes #191

### DIFF
--- a/src/acquire/__init__.py
+++ b/src/acquire/__init__.py
@@ -24,6 +24,12 @@ def run_all(raw_dir: Path) -> dict[str, Path | None]:
     results: dict[str, Path | None] = {}
     for adapter in SOURCE_REGISTRY:
         name = adapter.slug
+        if not adapter.active:
+            # Inactive sources (e.g. open_hadith — a confirmed duplicate, da#191)
+            # are never acquired. The registry row is retained for the coverage
+            # invariant only.
+            logger.info("acquire_skipped_inactive", source=name)
+            continue
         try:
             logger.info("acquiring", source=name)
             path = adapter.acquire(raw_dir)
--- a/src/adapters.py
+++ b/src/adapters.py
@@ -93,6 +93,11 @@ class SourceAdapter:
     acquire_fn: str = "run"
     parse_fn: str = "run"
     raw_subdir: str | None = None
+    # Whether this source participates in acquire/parse run_all. An adapter row is
+    # retained for every SourceCorpus (the coverage invariant), but a source that
+    # is a confirmed duplicate of another is marked inactive so it is never
+    # acquired/parsed/loaded — see ``src.parse.composition`` (da#191).
+    active: bool = True
 
     def acquire(self, raw_dir: Path) -> Path | None:
         """Download this source's raw data, returning its output path (or ``None``)."""
@@ -190,6 +195,12 @@ SOURCE_REGISTRY: tuple[SourceAdapter, ...] = (
         reachable=True,
         license_note="Open-Hadith-Data (mhashim6) — open GitHub, 9 books incl. the Six Books.",
         description="Open-Hadith-Data — 9 Sunni books.",
+        # Confirmed 100% duplicate of `halimbahae` (identical 9 collections, same
+        # per-collection counts, identical normalized matn) — pre-cutover audit
+        # da#191. Dropped entirely: halimbahae carries the same books with full
+        # tashkeel + a clear ODbL license. Adapter retained for the coverage
+        # invariant; inactive so it is never acquired/parsed/loaded.
+        active=False,
     ),
     SourceAdapter(
         slug="muhaddithat",
--- a/src/graph/load_nodes.py
+++ b/src/graph/load_nodes.py
@@ -34,6 +34,7 @@ from typing import Any
 import pyarrow.parquet as pq
 import yaml
 
+from src.parse.composition import is_canonical_hadith
 from src.parse.identity import (
     chain_node_id,
     collection_node_id,
@@ -266,6 +267,16 @@ def _load_hadiths(
                 all_errors.append(f"{fp.name} row {i}: invalid source_id={sid!r}")
                 total_skipped += 1
                 continue
+            # Canonical corpus composition (da#191): skip Hadith whose
+            # (source_corpus, collection) duplicates the chosen canonical edition
+            # — halimbahae/fawaz non-unique books, the mis Sahih Muslim matn copy.
+            # One enforcement point so a fresh run_all (the production path) yields
+            # the deduped graph without manual surgery.
+            if not is_canonical_hadith(
+                _val(row, "source_corpus", ""), _val(row, "collection_name", "")
+            ):
+                total_skipped += 1
+                continue
             hid = hadith_node_id(sid)
             # Fall back to the raw full text when a source supplies no separated
             # matn: halimbahae / open_hadith / bihar populate ``full_text_ar``
@@ -356,6 +367,12 @@ def _load_collections(
             cid = row.get("collection_id")
             if not cid or not isinstance(cid, str):
                 all_errors.append(f"{fp.name} row {i}: invalid collection_id={cid!r}")
+                total_skipped += 1
+                continue
+            # Canonical composition (da#191): drop Collection nodes for non-canonical
+            # (source_corpus, collection) pairs so they don't outlive their Hadith.
+            # The collection slug is the last ``:``-segment of collection_id.
+            if not is_canonical_hadith(_val(row, "source_corpus", ""), cid.rsplit(":", 1)[-1]):
                 total_skipped += 1
                 continue
             full_id = collection_node_id(cid)
--- a/src/parse/__init__.py
+++ b/src/parse/__init__.py
@@ -31,6 +31,11 @@ def run_all(raw_dir: Path, staging_dir: Path) -> dict[str, list[Path]]:
     results: dict[str, list[Path]] = {}
     for adapter in SOURCE_REGISTRY:
         name = adapter.slug
+        if not adapter.active:
+            # Inactive sources (e.g. open_hadith — a confirmed duplicate, da#191)
+            # are never parsed. The registry row is retained for coverage only.
+            logger.info("parse_skipped_inactive", source=name)
+            continue
         try:
             logger.info("parsing", source=name)
             output_files = _normalize_output(adapter.parse(raw_dir, staging_dir))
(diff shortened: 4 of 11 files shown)


parametrization:
RequestOrReplied: Approved
TechDebt: #194, #195

## Verdict — APPROVE (keystone composition is correct; non-blocking items filed as tech debt)

Reviewed the code at HEAD `fc7792d` via the contents API (not the PR body), with the same fixture-masking skepticism that surfaced the da#175 thaqalayn problem. The per-source composition is encoded correctly and — crucially — the kept-book slugs match the **real** upstream filenames, not assumed names. I traced every one to ground truth rather than trusting the unit assertions (which restate the same string literals as the map, so on their own they prove nothing).

### Composition slugs verified against real adapter output (lens #1)
- **halimbahae** keeps `{musnad_ahmad_ibn-hanbal, sunan_al-darimi, maliks_muwataa}`. The slug is filename-derived by `_collection_name_from_path` (strip `.utf8`, split on `_ahadith`). The committed fixtures carry the genuine upstream layout — `Maliks_Muwataa/maliks_muwataa_ahadith_mushakkala_mufassala.utf8.csv` to `maliks_muwataa`, `Sunan_Al-Darimi/sunan_al-darimi_ahadith_...` to `sunan_al-darimi`, and the open_hadith fixture (identical layout, since open_hadith is the 100% dup) `Musnad_Ahmad_Ibn-Hanbal/musnad_ahmad_ibn-hanbal_ahadith_...` to `musnad_ahmad_ibn-hanbal`. All three map exactly. A mismatch here would have silently dropped a unique book on prod; it does not.
- **fawaz** keeps `{nawawi, dehlawi, qudsi}`. Slug is `eng-<x>` to `<x>`. Fixtures confirm `nawawi` and `qudsi`. See tech-debt #195 on `dehlawi`.
- **Cross-loader consistency**: the Hadith loader filters on the `collection_name` column; the Collection loader filters on `collection_id.rsplit(":",1)[-1]`. I confirmed these are the same slug — `generate_source_id(corpus, collection)` joins with `ID_DELIMITER=":"`, so the last segment equals the book slug the hadith rows carry. No asymmetry, so collections won't outlive (or survive past) their hadiths.

### open_hadith deactivated end-to-end (lens #2) — yes
`active=False` in the registry, both `src.acquire.run_all` and `src.parse.run_all` skip inactive adapters, and the load-time map lists it as `frozenset()` defence-in-depth. The live-Neo4j integration test parses the committed samples and asserts 0 Hadith and 0 Collection nodes — and honestly documents that this reverses the original da#87 light-up. Genuine intentional change, not a masked regression.

### mis is chains/edges only (lens #4) — yes
`mis` maps to `frozenset()` so every mis Hadith is dropped. The live integration test proves both halves: mis Hadith count is 0, AND the multi-isnad multiplicity survives to 7 distinct ISNAD_TRANSMITTED edges including the chain-specific A->D / E->B pairs. This is real E2E behavior against neo4j:5, not a fixture assertion. The test also honestly flags that production `load_edges` is still hardcoded to muhaddithat — a documented graph-layer fast-follow, not a regression from this PR.

### Tests exercise real behavior (lens #3) — satisfied
The two changed integration tests run against a live container (clean skip when Docker is absent) over committed fixtures that mirror the true upstream directory/file names. That is exactly the realism gap that bit us on da#175, and it is closed here.

## Non-blocking tech debt (filed — does not gate this merge)
- **#194** — `is_canonical_hadith` mishandles an explicit `None` map value, contradicting the module's own documented contract: the docstring says a `None` value means "load all collections", but the code returns `source_corpus not in HADITH_COMPOSITION` when `allowed is None`, so an explicit-`None` entry would **drop that whole corpus**. Latent today (all four entries are frozensets) so prod behaviour is correct and I am not gating on it, but it is the exact silent-whole-corpus-drop failure this PR exists to prevent, so worth tightening while context is fresh.
- **#195** — three small test-coverage gaps: `dehlawi` is the one kept fawaz slug with no committed fixture (eyeball the staging dress-rehearsal output to confirm it lands); `test_run_all_skips_inactive` asserts a source substring rather than executing run_all to observe the skip; and there is no direct Collection-loader drop test for a halimbahae six-book duplicate (the predicate is shared with the tested Hadith path).

Net: the deterministic per-source dedup that must hold before the prod load is correct and well-evidenced. Approving for merge; #194 and #195 are follow-ups, not blockers.

parametrization:
Requestor: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: #196, #197, #198

## Review — da#191 / PR #193 (canonical corpus composition; keystone for prod cutover)

Reviewed the code at HEAD `fc7792d` via the contents API (not the diff framing). CI is fully green (Lint, Type Check, Test, Integration Tests, sync-drift gate). Approving — the load-path enforcement is real and correctly placed, the dedup holds at both the node and edge layers, and the test edits faithfully encode the behavior change without weakening coverage. A few non-blocking verification items below for the PD/owner before the actual prod load.

### What I verified (the four review lenses)

1. **Load-time filtering is real, not just acquire/parse gating.** `_load_hadiths` (load_nodes.py) calls `is_canonical_hadith(source_corpus, collection_name)` and skips non-canonical rows; `_load_collections` does the same on `cid.rsplit(":",1)[-1]`. So even a full registry replay where a `hadiths_halimbahae.parquet` carries the six-books overlap will NOT double-load — the excluded editions are dropped at the load point. This is the keystone property and it holds.

2. **`active=False` skips acquire AND parse.** Both `src/acquire/__init__.py` and `src/parse/__init__.py` `run_all` short-circuit on `not adapter.active` with a log line + `continue`. `open_hadith` is the only inactive row (and the unit test asserts it is the only one). Traced both paths.

3. **No narrator-merge regression.** The load_nodes change is purely additive (one import + two skip blocks); `_NARRATOR_MERGE` and the narrator loader are untouched. The composition does not touch narrator loading, so mention_count / source_ids / source_corpora handling is unchanged.

4. **Test edits reflect the behavior change, not weakened coverage.** The mis integration test flips only the node assertion (was "2 Hadith nodes", now "0 mis Hadith nodes") while fully retaining the multiplicity proof — all 7 ISNAD_TRANSMITTED edges land and the 5 distinct directed pairs survive un-collapsed (incl. A→D and E→B, the chain-specific links that only exist if the parallel asanid are kept). The open_hadith test is a deliberate, well-documented reversal: it now feeds the committed parquet straight to the loader and asserts zero Hadith+Collection nodes, which exercises the defense-in-depth load filter (a stray staging parquet can't re-introduce the duplicate). New unit tests cover `is_canonical_hadith` and the load-time skip directly.

### Extra check — edge-layer dedup safety

Dropping mis/open_hadith Hadith+Collection nodes while keeping mis chains/edges is safe: all six edge loaders MATCH (not MERGE) their endpoints and count missing endpoints, so an edge pointing at a now-absent mis Hadith is skipped, never phantom-created. mis narrator-to-narrator transmission edges have narrator-only endpoints, so they still load. The dedup holds end-to-end.

### Non-blocking items (pre-prod-load / owner confirmation — not code changes)

- **Vocabulary fidelity of the halimbahae keep-set.** The keep-sets are validated against hand-written / committed fixtures using the assumed slugs, which proves the filter logic but not the real-corpus vocabulary. Two of three halimbahae keep slugs are backed by committed fixtures (`sunan_al-darimi`, `maliks_muwataa`); `musnad_ahmad_ibn-hanbal` is not covered by any committed fixture or test. Because halimbahae uses an allow-list (anything not in the frozenset is dropped), a slug mismatch on Musnad Ahmad would silently DROP that whole book. Worth confirming against the real staging parquet that halimbahae emits exactly that slug before the prod load.

- **fawaz scope is broader than three collections upstream.** fawazahmed0/hadith-api carries collections beyond the six books and {nawawi, dehlawi, qudsi} (e.g. mishkat, bulugh al-maram, adab al-mufrad, shamail, hisn al-muslim, riyad). The composition keeps ONLY the three, so any fawaz-unique collection not provided by another active source would be dropped from production. This faithfully encodes the owner-relayed "keep ONLY Nawawi/Dehlawi/Qudsi", so it is an owner scoping call rather than a code defect — but given this is the cutover keystone, it is worth an explicit owner/PD confirmation that those other fawaz collections are intentionally excluded (or sourced elsewhere) and not accidental loss.

- **Latent comment/code mismatch in composition.py.** The map comment says a `None` value means "load all collections for this source", but `is_canonical_hadith` returns False (drop all) for a key whose stored value is `None` (the `allowed is None` branch returns `source_corpus not in HADITH_COMPOSITION`). No current entry is `None`, so this is latent, but a future maintainer adding `"x": None` expecting "load all" would get "drop all". Tightening the comment or the branch would remove the footgun.

- **Minor test robustness.** `test_run_all_skips_inactive` asserts the substring `adapter.active` appears in the `run_all` source; it would still pass if the gate logic were inverted. A behavioral test (run `run_all` over a registry containing an inactive adapter and assert it is skipped) would be stronger. Non-blocking.

Net: the enforcement mechanism is correct, single-point, MATCH-safe, and well-tested at the unit-mechanic level. Approved to merge; the items above are data-fidelity / owner-scope gates to clear before the real production load, not blockers on this PR.

PR #213, opened 2026-06-23
ci(checks): base-pin + fixture-realism gates rollout + README hook docs (#212)

### LINKED ISSUE #212: Roll base-pin + fixture-realism into CI/pre-commit + README hook-install docs (parent main#744/#718)
## Summary

Data-acquisition slice of the cross-repo rollout of the two charter-prose-to-code gates built in noorinalabs-main#735:

- `check_dockerfile_base_pin.py` — every Dockerfile `FROM` must be digest-pinned AND carry the matching distro upgrade.
- `check_fixture_realism.py` — Arabic hadith / isnad text fixtures must be voweled and carry the transmission particle.

This wires both into data-acquisition's `.pre-commit-config.yaml` + `.github/workflows/ci.yml` + the sync-drift classifier, fix-forwards the violations the first run surfaces, and refreshes the README local-hooks docs.

## Scope (main#744)

- Vendor `.claude/lib/check_dockerfile_base_pin.py` + `check_fixture_realism.py` (+ their unit tests, auto-run by the existing `.claude/lib/tests` discover step).
- Add the `dockerfile-base-pin` + `fixture-realism` kinds to `.claude/lib/pre_commit_ci_sync.py`.
- New `Base-image pin + fixture realism` CI job + two `pre-commit` hooks mirroring it.
- Fix-forward `Dockerfile.load`: digest-pin `python:3.14-slim` and add `apt-get -y upgrade`.
- Fixture-realism scoped to the curated Arabic hadith/isnad text corpus (already voweled + particle-bearing). Raw scraped page captures (acquire HTML, the thaqalayn `tusi_raw/` clone) and name-only NER fixtures (rijal names, clustering gold) are documented OUT of scope, not toy fixtures.

## Scope (main#718)

- Refresh the README "Git hooks (required)" section: add `cspell` and the two new gates, document the sync-drift gate, and fix the brand prose.

## Provenance

- noorinalabs-main#744 — roll base-pin + fixture-realism checks out to child-repo CI.
- noorinalabs-main#718 — document the pre-commit/pre-push install requirement in every child README.

parametrization:
## Summary

Data-acquisition slice of cross-repo **noorinalabs-main#744** (roll the two charter-prose-to-code gates from main#735 into child CI/pre-commit) + **noorinalabs-main#718** (document the hook-install requirement in the README). Per the child-repo-implementer rule, this lands in data-acquisition with the data-acquisition roster.

data-acquisition both **builds an image** (`Dockerfile.load`) and ships **Arabic parser fixtures**, so both checks apply.

## #744 — gates wired live + green

- Vendored `.claude/lib/check_dockerfile_base_pin.py` + `check_fixture_realism.py` verbatim from the parent, plus their unit tests (auto-run by the existing `.claude/lib/tests` discover step in the `precommit-sync` job — **46 tests pass**).
- Classified `dockerfile-base-pin` + `fixture-realism` kinds in `.claude/lib/pre_commit_ci_sync.py` so the sync-drift gate **demands** the local mirror (main#684 parity).
- New `Base-image pin + fixture realism` job in `ci.yml` + two mirroring `pre-commit` hooks (single-line `run:`/`entry:` so the line-scanner classifier registers the script name).

### Violation fixed-forward — `Dockerfile.load`

`FROM python:3.14-slim` was unpinned and the base stage had no `apt-get upgrade`. Pinned to the multi-arch (amd64+arm64) OCI-index digest `@sha256:44dd0449…` and added `apt-get -y upgrade` — closing both failure modes the charter § Base Image Pinning names (floating-tag drift + within-tag package drift). The `builder`/`runtime` stages are `FROM base` references (exempt).

### Fixture-realism scope — charter-compliant exemptions, not toy fixtures

The check is scoped to the **curated Arabic hadith/isnad text corpus** (`json`/`csv`/`txt` under any `tests/.../fixtures/` dir) — the mushakkala CSVs + fawaz qudsi/nawawi are already voweled and carry the transmission particle, so the gate is **green with no fixture edits**. The first run surfaced 6 "violations"; all are documented OUT of scope (not toy fixtures greener than real data):

- **Raw scraped page captures** — `tests/test_acquire/fixtures/*.html` (verbatim sunnah.com pages) and the `tests/test_parse/fixtures/tusi_raw/` thaqalayn clone tree. Verbatim upstream bytes are the rawest realism — they cannot be "greener than real data" (the failure mode the floor targets); judged by parse-ability instead. (The tusi *leaf* hadith files are in fact voweled + particle-bearing; only the chapter-index/nav nodes are titles.)
- **Name-only NER fixtures** — `itqan_rijal_sample.json` (rijal biographical NAMES, unvoweled in real rijal DBs) and `cluster_precision_gold.json` (clustering gold of narrator names). They carry no isnad string, so the عن/vocalization floor (charter-scoped to isnad text) does not apply.

The exclusion list + rationale is documented inline in both `ci.yml` and `.pre-commit-config.yaml`.

## #718 — README

Refreshed the existing "Git hooks (required)" section: added `cspell` and the two new gates to the commit-stage list, documented the sync-drift gate, and fixed brand prose (`NoorinALabs` → `Noorina Labs`).

## Verification (local, green)

`ruff check/format` · `mypy --strict` (81 files) · `pytest` (924 passed, 7 skipped) · `unittest .claude/lib/tests` (46) · `pre_commit_ci_sync.py .` (no drift) · base-pin + fixture-realism gates · `actionlint` (shellcheck present) · `markdownlint-cli2` (0 errors). Commit + push hooks passed.

## Reviewers

- Peer reviewer: **Oyunbileg Batbayar** (`qa_engineer_oyunbileg`)
- Secondary reviewer: **Jean-Claude Habimana** (`architect_jeanclaude`)

Closes #212
Part of noorinalabs-main#744 and noorinalabs-main#718

### DIFF
--- a/.claude/lib/check_dockerfile_base_pin.py
+++ b/.claude/lib/check_dockerfile_base_pin.py
@@ -1,0 +1,217 @@
+#!/usr/bin/env python3
+"""Lint Dockerfile `FROM` statements for digest-pin + distro-upgrade compliance.
+
+Deterministic conversion of `.claude/team/charter/tech-decisions.md § Base Image
+Pinning` (noorinalabs-main#735, the charter-prose-inventory worklist item #4 /
+epic #726). The section requires every `FROM` to use a **digest-pinned tag**
+(`image:tag@sha256:<digest>`) **combined with an in-image package upgrade**, to
+close the two failure modes it names — floating-tag drift and within-tag package
+drift (the shape isnad-graph#853 hit).
+
+What this gate enforces (verbatim from the section's distro table)
+================================================================
+For every `FROM` that is not exempt:
+
+1. The image reference MUST carry an `@sha256:` digest pin.
+2. The stage MUST run the package-manager upgrade matching the base distro:
+
+   | Family      | Pin shape               | Upgrade command                                 |
+   |-------------|-------------------------|-------------------------------------------------|
+   | Alpine      | image:tag@sha256:digest | `RUN apk upgrade --no-cache`                    |
+   | Debian-slim | image:tag@sha256:digest | `apt-get update && apt-get -y upgrade && clean` |
+   | Distroless  | image:tag@sha256:digest | none — no package manager (pinned-only is fine) |
+
+   The distro family is inferred from the image name: `alpine` → Alpine,
+   `distroless` → Distroless (no upgrade required), otherwise the glibc/Debian
+   default (an apt upgrade is required). A genuinely-other base that has no
+   package manager should carry the `# RATIONALE:` exemption below.
+
+Exemptions honored (verbatim from the section)
+=============================================
+- **`scratch`** final/any layer — no package manager; the upstream stages that
+  produced its contents still follow the rule and are checked on their own
+  `FROM`.
+- **Distroless** — pinned-only is sufficient (no package manager); the digest
+  pin is still required.
+- **Multi-stage stage reference** — a `FROM <earlier-stage>` inherits the
+  upstream stage's pin, so it is not re-checked (the named stage was checked at
+  its defining `FROM`).
+- **Vendor images** documented with an inline `# RATIONALE:` comment on the
+  `FROM` line (or the comment line immediately above it).
+
+CLI
+===
+    python3 .claude/lib/check_dockerfile_base_pin.py <Dockerfile> [<Dockerfile> ...]
+
+Exit codes:
+    0 — every FROM in every file is compliant (or exempt)
+    1 — at least one violation (each printed as `path:line: FROM <image> — <why>`)
+    2 — usage / file-not-found error
+
+This is a reusable template (same CLI/exit-code shape as
+`.claude/lib/pre_commit_ci_sync.py`) so it wires identically into a repo's
+pre-commit config and its CI. The parent repo `noorinalabs-main` ships no
+Dockerfiles itself; the cross-repo rollout that points this at each child's
+Dockerfiles is a #735 follow-up.
+"""
+
+from __future__ import annotations
+
+import re
+import sys
+from pathlib import Path
+
+# `FROM` is the only keyword we parse; Dockerfile keywords are case-insensitive.
+_FROM_RE = re.compile(r"^\s*FROM\s+(?P<rest>\S.*?)\s*$", re.IGNORECASE)
+# A leading `--platform=...` flag precedes the image and must be stripped.
+_PLATFORM_RE = re.compile(r"^--platform=\S+\s+", re.IGNORECASE)
+# A trailing `AS <stage>` names the build stage.
+_AS_RE = re.compile(r"\s+AS\s+(?P<stage>\S+)\s*$", re.IGNORECASE)
+
+# Upgrade-command signatures per distro family (searched over the stage body).
+_ALPINE_UPGRADE_RE = re.compile(r"\bapk\s+upgrade\b", re.IGNORECASE)
+# `apt-get update && apt-get -y upgrade` / `apt upgrade` — an apt invocation
+# followed (in the same command segment) by `upgrade`. `apt-get update` alone
+# does NOT match (that is "update", not "upgrade").
+_DEBIAN_UPGRADE_RE = re.compile(r"\bapt(?:-get)?\s+(?:-{1,2}\S+\s+)*upgrade\b", re.IGNORECASE)
+
+_RATIONALE_RE = re.compile(r"#\s*RATIONALE:", re.IGNORECASE)
+
+
+class FromStatement:
+    """One parsed `FROM` line: its image ref, optional stage name, exemptions."""
+
+    def __init__(self, lineno: int, image: str, stage: str | None, has_rationale: bool) -> None:
+        self.lineno = lineno
+        self.image = image
+        self.stage = stage
+        self.has_rationale = has_rationale
+
+
+def _preceding_comment_has_rationale(lines: list[str], from_idx: int) -> bool:
+    """True if the comment line(s) immediately above `from_idx` carry RATIONALE."""
+    j = from_idx - 1
+    while j >= 0:
+        stripped = lines[j].strip()
+        if stripped == "":
+            j -= 1
+            continue
+        if stripped.startswith("#"):
+            return bool(_RATIONALE_RE.search(stripped))
+        return False
+    return False
+
+
+def parse_froms(text: str) -> list[FromStatement]:
+    """Parse every `FROM` statement in a Dockerfile, in source order."""
+    lines = text.splitlines()
+    froms: list[FromStatement] = []
+    for idx, raw in enumerate(lines):
+        m = _FROM_RE.match(raw)
+        if not m:
+            continue
+        rest = m.group("rest")
+        has_rationale = bool(_RATIONALE_RE.search(raw)) or _preceding_comment_has_rationale(
+            lines, idx
+        )
+        # Drop any trailing inline comment, then a leading --platform flag.
+        code = rest.split("#", 1)[0].strip()
+        code = _PLATFORM_RE.sub("", code).strip()
+        stage: str | None = None
+        as_m = _AS_RE.search(code)
+        if as_m is not None:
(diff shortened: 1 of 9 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Tarek Mansour
RequestOrReplied: Changes-Requested
TechDebt: none

I verified the gate against real data, not the description, and most of this PR is solid — but the local pre-commit half does not actually fire for the corpus it is meant to protect, so I am blocking on that one item.

What I confirmed is good:
- The fixture-realism gate is non-tautological. I fetched both check scripts at the head commit and ran the PR's exact `find` pipeline over the working-tree fixtures: it selects 10 real files (the voweled fawaz qudsi/nawawi JSON, the mushakkala CSVs under halimbahae / open_hadith, the lk Bukhari/Nesai chapters) and exits 0 — and a deliberately un-voweled toy I fed it exits 1. The six curated Arabic fixtures genuinely carry harakat and the voweled particle, so this is exercising production-shape data.
- Every exemption is load-bearing, not cosmetic. Each exempted target genuinely fails the floor if handed to the checker: itqan_rijal_sample.json fails "no vocalization" (it is name-only rijal records — full_name / kunya / namings lists, no transmission string — and real rijal name DBs are unvoweled, so the floor would make it less realistic); cluster_precision_gold.json fails both (it is a labeled clustering gold set of canonical narrator names, no isnad string); and the tusi_raw tree fails only on its chapter-index / nav nodes (titles), while the leaf hadith file al-istibsar/1/1/1.json passes voweled+particle. The verbatim-upstream-capture rationale holds and is documented inline in both ci.yml and .pre-commit-config.yaml.
- Dockerfile.load: the base digest resolves to a real multi-arch OCI index for python covering linux/amd64 and linux/arm64 as claimed; base is digest-pinned with `apt-get update && apt-get -y upgrade`; builder/runtime are `FROM base` references correctly treated as exempt. The base-pin gate exits 0 on the real file.
- 38 vendored unit tests pass; the live CI rollup is green across all 11 checks including the new job and the sync-drift gate.

What blocks (one line, trivial fix): the pre-commit hook trigger regex `^tests/.*/fixtures/.*\.(json|csv|txt)$` does not match the curated corpus. That corpus lives directly under `tests/fixtures/...` (fawaz, halimbahae, lk, open_hadith), and the `tests/.*/fixtures/` shape requires an intermediate path segment, so it only matches the deeper `tests/<area>/fixtures/...` paths — which are exactly the exempted NER files. I confirmed with the compiled pattern that all four curated-corpus paths return no-match while the two exempted ones match. Because the hook uses `pass_filenames: false`, pre-commit skips it entirely when nothing matches `files:`, so a developer who toy-ifies one of the voweled corpus fixtures locally gets a clean commit — the precise regression class (a fixture greener than real data) this gate exists to catch. CI still catches it (the CI step runs the full find unconditionally), so this is not a hole in the gate's correctness, but it is a real local-enforcement gap and it contradicts the inline comment's stated scope ("any tests fixtures dir"). Please widen the trigger to also cover the top-level dir, e.g. `^tests/.*fixtures/.*\.(json|csv|txt)$` or `^tests/(.*/)?fixtures/...`, and re-confirm a staged edit to tests/fixtures/fawaz/ara-qudsi.json fires the hook. The dockerfile-base-pin hook and everything else is approved as-is.

parametrization:
Requestee: Tarek Mansour
RequestOrReplied: Changes-Requested
TechDebt: none

Structurally this is a clean conversion of two charter sections into deterministic, well-tested gates, and the scope reasoning is sound. I am requesting one change for a divergence between the CI step and its local mirror.

The parts I checked and endorse:
- Both checks are stdlib-only, single-responsibility, and wire identically into ci.yml and pre-commit, with the script basename present on both invoke lines so the line-scanner sync classifier registers them. The two new kinds (dockerfile-base-pin, fixture-realism) are added to pre_commit_ci_sync.py, and the live sync-drift gate is green — so the #684 contract that an enforced CI check must be mirrored is satisfied at the kind level.
- The fixture-realism scope decision is architecturally correct. The realism floor is meaningful only for authored isnad-text fixtures; applying it to verbatim upstream captures (the tusi_raw thaqalayn clone tree, raw scraped HTML) or to name-only NER/clustering data inverts the intent — fabricating vocalization or a transmission particle into a raw capture or a rijal name set makes it less faithful to its source. I verified each exempted target genuinely trips the floor (so the exemptions are real decisions, not no-ops) and that the tusi_raw leaf hadith records are themselves voweled+particle-bearing while only the chapter/nav title nodes are not — pruning the whole capture subtree rather than excluding node-by-node is the right coarse boundary given it is all upstream-verbatim, judged by parse-ability. The two-axis rationale (raw-capture vs authored fixture; isnad-text vs name-only) is documented inline in both files.
- Dockerfile.load: base digest-pinned to a genuine multi-arch (amd64+arm64) OCI index with the apt upgrade added in the correct order after `apt-get update`; the builder/runtime `FROM base` stages correctly inherit the pin. Both failure modes the section names are closed.

The change I am requesting: the CI step and the pre-commit hook diverge in their trigger scope. The CI step runs the find unconditionally over the whole tree; the pre-commit hook gates execution on `files: ^tests/.*/fixtures/.*\.(json|csv|txt)$`. The repo has a two-tier fixtures layout — the curated corpus sits at `tests/fixtures/<source>/...` while the parser/resolver fixtures sit at `tests/<area>/fixtures/...` — and the `tests/.*/fixtures/` shape only matches the second tier. The consequence is that the local hook's trigger never covers the curated corpus that is the gate's primary subject (and that the inline comment says it covers — "any tests fixtures dir"), so the local and CI scopes are not the same set. Since the entry runs the full find regardless, the fix is purely the trigger regex (e.g. `^tests/.*fixtures/.*\.(json|csv|txt)$`); please align it so the local mirror's firing scope matches the CI step's, then this is approved.

parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Tarek Mansour
RequestOrReplied: Approved
TechDebt: none

Re-verified at head e3ed86d. The trigger gap is closed and the local hook now does the job it was meant to.

- The widened trigger `^tests/(.*/)?fixtures/.*\.(json|csv|txt)$` matches all four curated-corpus paths I flagged (fawaz/ara-qudsi.json, the halimbahae and open_hadith mushakkala CSVs, lk/Bukhari/Chapter1.csv) — I confirmed each against the compiled pattern. The optional intermediate segment also keeps the deeper `tests/<area>/fixtures/...` paths matching, so nothing regressed.
- Regression coverage now bites: I took the real ara-qudsi.json, stripped its harakat, and the checker flags it ("no vocalization diacritics"). Since that path now satisfies `files:`, a staged de-voweling edit fires the hook locally and goes red instead of being silently skipped — exactly the masking-fixture class this gate exists to catch.
- The two exempted NER fixtures still match `files:` (so the hook fires) but remain excluded inside the entry's `find` via `! -name`, and the tusi_raw subtree is still pruned — that behavior is unchanged by this fix.
- Sync-drift gate still green; the rollup is 11/11 SUCCESS and the PR is mergeable.

Approving.

parametrization:
Requestee: Tarek Mansour
RequestOrReplied: Approved
TechDebt: none

Re-verified at head e3ed86d. The CI-step ⇄ local-hook scope divergence is resolved.

- Making the intermediate path segment optional (`^tests/(.*/)?fixtures/...`) is the correct structural fix: it reconciles the trigger with the repo's two-tier fixtures layout so the local hook's firing scope now matches the CI step's unconditional `find` over the curated corpus at `tests/fixtures/<source>/...` as well as the deeper `tests/<area>/fixtures/...` tier. I confirmed both tiers match the compiled pattern, and the inline comment's "any tests fixtures dir" claim is now accurate.
- The entry command is untouched — the find still prunes the tusi_raw verbatim-capture subtree and excludes the two name-only NER fixtures — so the carefully-reasoned exemption boundary is preserved; only the trigger predicate widened.
- The sync-drift gate (kind-level local⇄CI parity) remains green, and the full rollup is SUCCESS across all 11 checks with the PR mergeable.

Approving.

Issue #196, opened 2026-06-17
composition.py: None-value map semantics contradict is_canonical_hadith


parametrization:
Found during review of PR #193 (da#191). Non-blocking, latent.

The `HADITH_COMPOSITION` comment documents a map value of `None` as meaning "load all collections for this source". But `is_canonical_hadith` treats a key whose stored value is `None` as drop-all: the `allowed is None` branch returns `source_corpus not in HADITH_COMPOSITION`, which is False when the key IS present. So a future maintainer who adds `"x": None` expecting "load all of x" would instead drop all of x's Hadith.

No current map entry uses a `None` value, so this is latent only. Fix is to align the comment with the code (drop the documented None-means-all semantics, since absence already covers that) or change the branch to honor an explicit None as load-all. Low severity, hygiene.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Kavitha Sundaramurthy
RequestOrReplied: Request

**Wave 19 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-19.
- Peer reviewer: Nikolaos Papadopoulos
- Secondary reviewer: Oyunbileg Batbayar
- Branch from: `deployments/phase-7/wave-19`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

parametrization:
Resolved by PR #217 (squash-merged into `deployments/phase-7/wave-19`, 2026-06-25). Reaches `main` at the W19 wave→main integration (/wave-wrapup).

PR #217, opened 2026-06-25
fix(parse): honor explicit None as load-all in HADITH_COMPOSITION (da#196)

parametrization:
## Summary
Fixes the latent `None`-value semantics contradiction in `HADITH_COMPOSITION` (da#196, found during review of PR #193).

`is_canonical_hadith`'s `allowed is None` branch returned `source_corpus not in HADITH_COMPOSITION` — which is `False` when the key **is** present. So a future `"x": None` entry (documented as "load all collections") would instead **drop all** of x's Hadith — the opposite of the documented intent.

## Fix chosen — honor explicit `None` as load-all (Option B)
Both the module docstring (`every other source — all collections (value None)`) and the inline comment document `None = load all`. The **code** was the bug, so I aligned the code with the documented intent rather than rewriting the docs to match the foot-gun:

- Distinguish key-absence from an explicit `None` value: a source that is **absent** OR **present with value `None`** now loads all collections.
- Only a non-empty `frozenset` restricts to an allowlist; an empty `frozenset()` drops all Hadith nodes (unchanged).

**No behavior change for existing entries** — no current map entry uses `None`, and absent keys still load all. `halimbahae`, `fawaz`, `mis`, `open_hadith` are unaffected (verified by the existing tests, all still green).

## Test
Added `test_explicit_none_value_loads_all_collections` — monkeypatches an explicit `None` entry into `HADITH_COMPOSITION` and asserts it loads all collections, locking in the documented semantics so the drop-all foot-gun can't regress.

## Checks
Local pre-commit (commit + pre-push stages) all green: ruff-format, ruff-lint, gitleaks, actionlint, cspell, mypy-strict, pytest (full suite).

Reviewers: @Nikolaos Papadopoulos (primary), @Oyunbileg Batbayar (secondary)

Closes #196


parametrization:
**Review — PR #217 composition None-semantics** (verified at head `7489d7238cd3`)

Read the diff and both full files at the head SHA via the contents API. The fix (Option B) is correct and resolves da#196 exactly.

The old `is_canonical_hadith` used `HADITH_COMPOSITION.get(source_corpus)`, which collapsed two semantically distinct inputs — a key that is absent vs. a key present with an explicit `None` value — into the single `allowed is None` branch, then tried to disambiguate with `return source_corpus not in HADITH_COMPOSITION`. The effect was that an explicit `None` map value returned `False` (drop every Hadith node), directly contradicting the module comment and docstring which both state `None` = load all. The new code checks membership first (`source_corpus not in HADITH_COMPOSITION` → `True`), then indexes the map directly, so an explicit `None` now returns `True` (load all), matching the documented contract.

I walked the full truth table to confirm no behaviour change for the cases that already work:
- absent key (e.g. `lk`, `sanadset`, `thaqalayn`) — still load-all (`True`), unchanged.
- non-empty frozenset (`halimbahae`, `fawaz`) — still `collection_name in allowed`, unchanged.
- empty frozenset (`mis`, `open_hadith`) — `allowed` is not `None`, falls through to `collection_name in frozenset()` → `False` (drop all), unchanged.
- explicit `None` value — was `False`, now `True`. The only behaviour that changes, and it is the bug being fixed.

The new `test_explicit_none_value_loads_all_collections` exercises the real bug: it `monkeypatch.setitem`s an explicit `None` entry and asserts load-all across two collections. Because the current map ships only frozenset entries, an injected `None` is the right way to lock the semantics in, and the test would have failed against the old implementation (which returned `False` for a present-with-`None` key). The existing five tests are untouched. No scope creep — the change is confined to `composition.py`, its test, and the regenerated structural index.

The `ontology/structural/` diff is benign and exactly what the source change implies: `+1` node (the new test method), `+1` contains edge, the `is_canonical_hadith` line shift `50 → 51`, test-method line shifts from the added `import pytest`, and `pytest` added to the test module's import list. File count is unchanged at 191; node/edge counts move by exactly one each. No content regression.

CI is fully green (Lint, Type Check, Test, Integration Tests, base-image/fixture, sync-drift gate, structural staleness-check all SUCCESS).

Approving.

---

RequestOrReplied: Approved
TechDebt: none

parametrization:
**Review — PR #217 composition None-semantics** (verified at head `7489d7238cd3`)

Reviewed the diff and the full source + test at the head SHA via the contents API, plus da#196 for scope. CI is green across all checks (Lint, Type Check, Test, Integration Tests, base-image/fixture, sync-drift gate, structural staleness).

**Fix correctness (Option B, aligns code to the documented contract).** The rewrite splits the two states the old `.get()` conflated: key-absence and an explicit map value. The flow is now
- absent key → load all (early True),
- present with value `None` → load all (True),
- non-empty `frozenset` → allowlist membership,
- empty `frozenset` → drop all.

This is exactly the semantics the module docstring documents and what da#196 asked for. The previous drop-all behaviour for an explicit-`None` key (the `allowed is None` branch returning `source_corpus not in HADITH_COMPOSITION`, i.e. False when the key was present) is gone.

**Existing entries provably unchanged.** For every current key the source is present, so the early-return is skipped and `allowed` is read directly — identical to what `.get()` returned for present keys. None of the four current entries (`halimbahae`/`fawaz` non-empty frozensets, `mis`/`open_hadith` empty frozensets) holds a `None` value, so each still falls through to the same final membership check as before. For absent keys both old and new paths return True. The only behavioural delta is the explicit-`None` case — the documented bug. The existing tests (`unlisted_source`, `halimbahae`, `fawaz`, `mis`, `open_hadith`) all continue to pin those paths.

**Test exercises the bug and guards the regression.** The new `test_explicit_none_value_loads_all_collections` monkeypatches an explicit `None` entry and asserts load-all. Under the pre-fix code that key is present, so the old branch returned `not in map` → False; the assertion would have failed. So it genuinely fails-on-regression rather than being a tautology. `monkeypatch.setitem` is the right call here — it mutates the real module dict and is auto-reverted, so it doesn't leak into sibling tests.

**Edge-case coverage (the three states the brief flagged) is complete.** Absent key → `unlisted_source`; empty `frozenset` → `mis`/`open_hadith`; explicit `None` → the new test. The non-empty-allowlist and explicit-drop paths are also pinned. No gaps worth a follow-up.

No scope creep: the only non-test change is `composition.py`, and the `ontology/structural/` churn is the expected generator output — line-number shifts for the moved function plus one added test node — benign.

Clean fix, well-targeted regression guard. Approving.

---

Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

Issue #153, opened 2026-06-14
Graph integrity sweep — deferred data-decision items (da#148 follow-up: null hadith_number, NARRATED gaps, orphans, grade-parity, collection metadata)


parametrization:
## Summary

Deferred items from the da#148 integrity sweep (PR #150 shipped the producer-level fixes: self-loop guard + grade normalization). These remaining items each need a **data decision or upstream producer work** and were deliberately NOT fabricated at load time. Itemized so nothing is silently dropped.

## Deferred items

1. **86 null `hadith_number_in_book`** — per-scraper in-book-ordinal derivation vs an explicit-null contract. Unsafe to fabricate at load; needs a per-source decision on how the in-book ordinal is derived.
2. **51 hadith without a NARRATED edge** — NER/mention coverage gap upstream (related to da#146 segmentation; re-measure after #151 lands).
3. **8 orphan muhaddithat bio-only narrators** — owner decision: link them or drop them.
4. **`grade_normalized` parity in the ingest-platform streaming path** — PR #150 added `normalize_grade()` + `grade_normalized` on the batch load path; the Kafka streaming worker (isnad-ingest-platform) needs the same so streamed loads match. Cross-repo.
5. **Collection blank Arabic name + riyadussalihin metadata** — data enrichment for the collection node.

## Investigated, NOT a bug (recorded so it isn't re-dug)
- The "15 lk-STUDIED_UNDER edges" finding from the original sweep is correct cross-corpus identity merge, NOT a transmission leak: there is no lk network_edges producer; the 15 are genuine muhaddithat/itqan studentship edges whose endpoints name-resolve onto narrators also present in the lk corpus. No change needed.

## Origin

Split from da#148 (PR #150, Nikolaos Papadopoulos, P5W1). Serves Phase 5 criterion #1 (data real & queryable, main#665). Defer to a later P5 wave for the data decisions; re-measure #2 after da#146 (PR #151) lands.

parametrization:
[...]

PR #223, opened 2026-06-26
fix(graph): da#153 integrity sweep — explicit-null/no-fabrication contracts + inventory

parametrization:
## Summary

Closes #153.

Deferred-item follow-up to the da#148 integrity sweep (PR #150 shipped the clean producer fixes: self-loop guard + `grade_normalized`). The five remaining items each need a **data decision or upstream producer work** and are deliberately **NOT fabricated at load time**. This PR makes the producer/contract decision explicit per item and **pins the no-fabrication stance with tests**, rather than inventing plausible-but-unsourced values into a provenance-sensitive corpus.

See **`docs/adr/004-graph-integrity-deferred-items.md`** for the full per-item rationale.

## What shipped (decisions made here — no owner input required)

- **#1 — 86 null `hadith_number_in_book`:** contract = **explicit-null** where the in-book ordinal is genuinely unknown (scraped sources carrying only the collection-wide ref). The load layer already honours this (`_APPEARS_IN_QUERY` coalesce-preserve after a property-less MERGE — da#77). New `hadith_number_coverage` informational check in `src/parse/validate.py` surfaces the null population and **always PASSES** (a null ordinal is contractual, not a failure). Pinned by `test_in_book_ordinal_explicit_null_is_contractual` (passes even under STRICT).
- **#2 — 51 Hadith with no `NARRATED`:** contract = **no fabrication**. A hadith whose mentions carry no `canonical_narrator_id` gets zero `NARRATED` edges; `_load_narrated` never invents a position-0 narrator. Pinned by `test_unresolved_chain_emits_no_narrated_no_fabrication` (zero edges, no write issued). Closing the gap is producer-side NER work (re-measure after da#146/da#151).
- **Read-only inventory:** `queries/validation/graph_integrity_deferred_inventory.cypher` sizes items #1/#2/#3/#5 per corpus (mirrors `sanadset_orphan_inventory.cypher`). No writes, no deletes, no fabrication.
- Regenerated `ontology/structural/` for the source change.

## Owner data-decisions still needed (escalated, not silently resolved)

- **#3 — 8 orphan `muhaddithat` bio-only narrators:** **owner call — link (mention-linking) vs drop.** Not made unilaterally (dropping loses sourced bio figures; auto-linking risks wrong attribution). Enumerated by inventory block #3/#3b.
- **#1 (per-source):** whether any specific source can have its in-book ordinal *derived* rather than left null — a per-scraper parse-lane decision.
- **#5 — Collection metadata gap (blank Arabic name / `riyadussalihin`):** per-collection **data enrichment** with sourced metadata, not a guess. Sized by inventory block #5.

## Cross-repo follow-up (not a data-acquisition change)

- **#4 — `grade_normalized` parity:** PR #150 added normalization on the batch path here; the Kafka streaming worker in **`noorinalabs-isnad-ingest-platform`** (`workers/ingest`) needs the same so batch and stream converge. Tracked as a cross-repo parity follow-up — nothing to change in this repo.

## Test plan

- `ENVIRONMENT=test uv run pytest` — full suite green under the pre-push hook (mypy-strict + pytest). Full local⇄CI parity (commit + push stages: ruff, cspell, gitleaks, actionlint, mypy, pytest, structural-ontology, fixture-realism, sync-drift gate) all green before push.

## Reviewers

- **Kwesi Boateng** (Integration Engineer)
- **Oyunbileg Batbayar** (QA / Data Quality Engineer)


parametrization:
Requestor: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

Verified against source at HEAD e27d04c. No-fabrication contract is real, not vacuous: _load_narrated skips any mention whose canonical_narrator_id is null (the 'if not hid or not nid: continue' guard), so first_narrators stays empty, the batch never forms, and it returns a zero-edge result BEFORE any check-read or write — never inventing a position-0 narrator. The test asserts created==0, missing_endpoints==0, AND that no NARRATED query reached the client. The in-book ordinal test writes real null rows and asserts the file passes even under STRICT.

The hadith_number_coverage check is hard-coded PASS and can never become a false STRICT gate — the STRICT-mode test asserts status PASS and files[0].passed True; APPEARS_IN coalesce-preserve leaves the edge property absent rather than fabricated. The inventory cypher is genuinely read-only (every block MATCH/WITH/RETURN/count, no CREATE/MERGE/SET/DELETE). The four owner/cross-repo decisions (orphan link-vs-drop, per-source ordinal derivation, collection enrichment, streaming grade parity) are surfaced and itemized, not dodged — matches da#153's pin-and-surface intent.

Non-blocking nit (not filed): ADR-004 Related footnote names the streaming repo noorinalabs-isnad-graph-ingestion, while the body/Item-#4 correctly use noorinalabs-isnad-ingest-platform.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

Verified against code at HEAD (e27d04c), not the PR body.

Test rigor — both new tests genuinely fail if fabrication/regression were reintroduced. The NARRATED test feeds two mentions with a null canonical_narrator_id; in _load_narrated the `if not hid or not nid: continue` guard skips them, so first_narrators stays empty and the function early-returns EdgeLoadResult zero/zero/zero before any endpoint-check or write — matching the test's three assertions (no edges created, zero missing endpoints, no NARRATED query in mock_client.calls). It is not a trivial pass; the sibling test_first_narrator_per_hadith uses the same write_narrator_mentions_resolved helper plus load_all_edges and asserts created==1, proving the fixture wiring reaches the load path, so created==0 here is the no-fabrication guard. The validate test pins the explicit-null contract — 1-of-3 filled gives 33.33pct, status PASS, and the file still passes under STRICT, so a future change cannot silently turn explicit-null into a hard fail.

No baseline drift — validate.py is purely additive (zero deletions); the new hadith_number_coverage check is hard-wired to PASS and is informational, so it can never false-red CI, and the green Test job confirms no existing check-count assertion broke.

Data-quality completeness — all five deferred items are accounted for in ADR-004: items 1 and 2 contractually pinned plus tested, items 3 and 5 owner-escalated, item 4 routed as cross-repo grade-parity follow-up to the streaming worker. The read-only inventory cypher sizes 1/2/3/5 with MATCH/RETURN/count only (no writes). Nothing silently dropped.

CI hygiene — full statusCheckRollup green at HEAD, Markdown lint included (the MD018 follow-up landed clean).

Non-blocking note for a later doc pass: ADR Related line 141 names the streaming repo noorinalabs-isnad-graph-ingestion while the PR body uses noorinalabs-isnad-ingest-platform — prose pointer only, no impact on the contract or load behaviour.

PR #224, opened 2026-06-26
Path B/B1: emit collections_sanadset.parquet (book_id→collection) for sanadset (#219)

### LINKED ISSUE #219: Path B/B1: emit collections_sanadset.parquet (book_id→collection) for sanadset
## Path B / B1 — emit `collections_sanadset.parquet` (book_id→collection)

**Part of da#202 Path B** (owner decision 2026-06-25 — parse/segment+link, not purge). Meta: noorinalabs-main#723. Investigation: ADR `docs/adr/003-sanadset-orphan-remediation.md` (PR #218).

### Problem
`src/parse/sanadset.py` emits only `hadiths_`, `narrator_mentions_`, `narrators_bio_` parquet — **no `collections_sanadset.parquet`**. Every other parser emits `collections_*`. As a result `_load_collections` (`src/graph/load_nodes.py`) creates zero sanadset Collection nodes, and `_APPEARS_IN_QUERY` (`src/graph/load_edges.py`, a non-creating `MATCH (c:Collection)`) silently drops all 650,986 sanadset APPEARS_IN edges → orphan hadiths.

### Work
`books.csv` is already downloaded and hadiths carry `book_id`. Emit `collections_sanadset.parquet` mapping `book_id`→collection so sanadset hadiths can link via APPEARS_IN. Also register `sanadset` in `composition.py` so admission is explicit, not the unlisted-source load-all default.

### ⚠ Sequencing — BLOCKED-BY B2
Do **NOT** load B1 output to prod standalone. Without **B2** (cross-edition canonical-identity dedup) the sanadset corpus double-counts hadiths already present in the 6 curated books (richer metadata). B1 + B2 ship together.

### Scheduling
Deferred to a future Phase-7 data-quality wave (W3+ per W19 scope). Not in W19.

parametrization:
## Path B / B1 — emit `collections_sanadset.parquet` (book_id → collection)

Closes #219. Part of da#202 Path B (ADR-003); meta noorinalabs-main#723.

### Problem
`src/parse/sanadset.py` emitted only `hadiths_`, `narrator_mentions_`, and
`narrators_bio_` parquet — no `collections_sanadset.parquet`. So
`_load_collections` (`src/graph/load_nodes.py`) created **zero** sanadset
`Collection` nodes, and `_APPEARS_IN_QUERY` (`src/graph/load_edges.py`) — a
**non-creating** `MATCH (c:Collection)` — silently dropped all **650,986**
sanadset `APPEARS_IN` edges (the headline orphan defect, ADR-003).

### What this PR does (B1 — re-link only)
- **Parse `books.csv`** (already acquired alongside `sanadset.csv`) into a
  `book_id → {name_ar, name_en, compiler_name}` mapping via a flexible,
  case-insensitive column matcher (upstream header spelling isn't contractual).
  `books.csv` is now **excluded from the hadith glob** so it's never mis-parsed
  as hadith rows (a latent bug — the prod glob matched it).
- **Per-book collection_name.** With `books.csv` present, each hadith's
  `collection_name` becomes its `book_id`, preserving the corpus's collection
  breadth (the reason Path B was chosen over a purge). Without it, the pre-B1
  per-CSV-stem name is kept (backward compatible).
- **Emit `collections_sanadset.parquet`** — one `Collection` per distinct
  `collection_name`. `collection_id = generate_source_id(corpus, collection_name)`
  is byte-identical to the key `load_edges._load_appears_in` rebuilds from each
  hadith (`f"{corpus}:{collection_name}"`), so `collection_node_id` agrees on
  **both** APPEARS_IN endpoints and the edges load.
- **Explicit composition admission.** Register `sanadset: None` (load-all) in
  `composition.py` so admission is explicit rather than the unlisted-source
  default. The cross-edition dedup that would tighten this to an allowlist is
  **B2/da#220** and is intentionally NOT done here.

### Scope guard
B1 only. No cross-edition dedup (B2/da#220), no narrator re-segmentation
(B3/da#221). Do not load B1 to prod standalone — it ships with B2 (per ADR-003).

### Tests
- New `TestSanadsetCollections`: collections emitted with **and** without
  `books.csv`; book metadata flows onto the `Collection`; and a Neo4j-free
  linkage-contract test asserting every hadith's APPEARS_IN endpoint is covered
  by an emitted `Collection`.
- `tests/integration/test_sanadset_lightup.py`: removed the fabricated
  `write_collections` injection that masked the production gap; it now asserts the
  **real** parser emits `collections_sanadset.parquet` (honesty fix per ADR-003
  step 4 / org memory "test-mock injection masks production failure").
- Regenerated the structural ontology index for the new symbols.

### Local ⇄ CI parity
`ruff format --check`, `ruff check`, `mypy src/` (81 files), full non-integration
`pytest` (928 passed / 7 skipped) and the Docker-gated integration suite all green
locally; pre-commit + pre-push hook sets passed on commit and push.

### Notes for review
- No `collections_sanadset` drift baseline was added to `validate.py`: the real
  book count isn't known a priori, and a guessed baseline would be a false STRICT
  drift gate. Like the other baselines ("calibrated from a full pipeline run"), it
  should be calibrated from the first real run. Hadith/mention row-counts are
  unchanged, so no existing baseline needed updating.

Reviewers: **Kavitha Sundaramurthy** and **Nikolaos Papadopoulos**.


parametrization:
RequestOrReplied: Approved
TechDebt: none

Verified the two-sided key match that broke as the headline orphan defect. The emitted collection_id is generate_source_id("sanadset", collection_name) → "sanadset:<collection_name>", and _load_collections wraps it to col:sanadset:<collection_name>; load_edges._load_appears_in independently rebuilds f"{corpus}:{collection_name}" → the same col:sanadset:<collection_name>. They converge byte-for-byte, so the 650,986 dropped APPEARS_IN edges now match. Also confirmed the new explicit sanadset → None entry in composition.py keeps is_canonical_hadith True, so the Collection nodes survive the canonical-composition gate inside _load_collections (without it they would be filtered out and the edges would still drop).

Per-book breadth holds. With books.csv present each hadith's collection_name becomes its book_id (test proves book_ids 1,1,2 yield distinct sanadset:1 / sanadset:2 Collections, not one bucket), and books.csv is correctly excluded from the hadith glob so the 3-hadith count is unchanged. Without books.csv the fallback collapses to one per-CSV-stem Collection (backward compatible). _parse_books, safe_int, safe_str, read_csv_robust imports all present; name_en non-null guarantee respected.

No new double-count is introduced. Sanadset hadiths already loaded pre-PR under the unlisted-source default, so the explicit None is behaviorally identical and only makes B2/da#220 a one-line change; this PR adds only Collection nodes plus their APPEARS_IN edges, with cross-edition dedup explicitly deferred.

Tests are strong. With and without books.csv are both covered, and the Neo4j-free linkage-contract test rebuilds the exact edge-loader key (collection_node_id over f"{corpus}:{cname}") and asserts every hadith endpoint is covered by an emitted Collection — the precise contract the defect violated. The integration test honesty fix removes the fabricated write_collections injection and now exercises the real parser path. Full statusCheckRollup green including the Neo4j-gated Integration Tests.

parametrization:
RequestOrReplied: Approved
TechDebt: none

Traced the linkage contract on both endpoints and they are byte-identical. The Collection set is derived from the parsed hadith rows (collection_counts keyed on each hadith's collection_name = row_collection), so the node side emits collection_id generate_source_id("sanadset", row_collection) -> col:sanadset:<row_collection> via load_nodes._load_collections collection_node_id, and the edge side rebuilds f"sanadset:{collection_name}" -> col:sanadset:<row_collection> via load_edges._load_appears_in collection_node_id over the same string. Drift is impossible by construction since the collections are generated from the hadiths, not from books.csv. Confirmed the da#191 composition gate admits all sanadset collections: is_canonical_hadith("sanadset", *) returns True because HADITH_COMPOSITION["sanadset"] is None (load-all), so the 650,986 APPEARS_IN edges now resolve their Collection endpoint. name_en is non-null on every path (book meta, Arabic fallback, or deterministic Sanadset-book label).

books.csv glob fix excludes only the exact filename books.csv (case-insensitive), so no legitimate hadith CSV is dropped, and it is consumed as the book->collection mapping via _parse_books. Scope is B1-only: composition adds sanadset None plus comments with no allowlist tightening (B2/da#220) and no narrator re-segmentation (B3/da#221). Test honesty fix is real: the fabricated write_collections injection is removed from test_sanadset_lightup and the test now asserts the real parser emits collections_sanadset.parquet then drives the real node+edge loaders against live neo4j, asserting one sanadset Collection and three APPEARS_IN edges. The omitted collections_sanadset drift baseline is justified and benign: validate.py only runs drift when the baseline key is present, so absence skips drift without disabling schema/required-column checks, and the real book count is calibrate-from-first-run like the other baselines.

Issue #220, opened 2026-06-25
Path B/B2: cross-edition canonical-identity dedup (sanadset vs curated six)


parametrization:
## Path B / B2 — cross-edition canonical-identity dedup (sanadset vs curated six)

**Part of da#202 Path B** (owner decision 2026-06-25). Meta: noorinalabs-main#723. Investigation: ADR `docs/adr/003-sanadset-orphan-remediation.md` (PR #218).

### Problem
The sanadset corpus (650,986 hadiths) overlaps the 6 curated books (thaqalayn, lk, tusi, halimbahae, sunnah, fawaz) already loaded with richer, deduplicated metadata. Emitting sanadset collections (**B1**) without canonical-identity dedup would **double-count** hadiths across sources. This is the never-built "tracked fast-follow" referenced in `composition.py`.

### Work
Build a cross-edition canonical-identity dedup so sanadset hadiths that are the same canonical hadith as a curated-book entry are merged/skipped rather than duplicated. **Prerequisite for safely loading B1.**

### Alignment
Maps to the W3-deferred "dedup/parallels pipeline" item in the W19 scope.

### Scheduling
Deferred to a future Phase-7 data-quality wave (W3+). Not in W19.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Kavitha Sundaramurthy
RequestOrReplied: Request

**Wave 20 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-20.
- Peer reviewer: Alejandra Reyes-Fuentes
- Secondary reviewer: Ivana Horvat
- Branch from: `deployments/phase-7/wave-20`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

PR #225, opened 2026-06-26
feat(graph): cross-edition canonical-identity dedup for sanadset (da#220, Path B/B2)

parametrization:
## Summary

Implements **da#220 (Path B / B2)** — the cross-edition canonical-identity dedup that was the never-built "tracked fast-follow" referenced in `composition.py`. Builds on B1 (da#219, already merged into this wave branch), which emits `collections_sanadset.parquet` but deliberately deferred dedup to B2.

The sanadset corpus (650,986 hadiths) overlaps the curated spine (`lk` / `halimbahae` / `fawaz`) with richer, deduplicated metadata. Without a cross-edition check, loading sanadset double-counts the canonical books. B2 drops a sanadset hadith at graph-load time when it is the **same canonical tradition** as an already-loaded curated hadith — **curated wins**.

## Design

- **`canonical_matn_identity(matn_ar, matn_en)`** (`src/parse/composition.py`) — deterministic identity from the **normalized matn**, the one signal that survives across editions. (The collection slug does not: sanadset keys collections by numeric `book_id`, curated sources by name, so a literal "collection+number" key never collides between them.) Reuses `normalize_arabic` so spelling variants converge; returns `None` for matns below a 3-token floor (too generic to identify → never deduped, always kept).
- **`CROSS_EDITION_DEDUP_SOURCES = {"sanadset"}`** — sources deduped against the curated spine. The per-source `HADITH_COMPOSITION` map is intentionally **not** tightened: sanadset keeps its full collection breadth (the reason Path B beat a purge). Dedup is **per-hadith**, not per-collection.
- **`_load_hadiths`** (`src/graph/load_nodes.py`) — builds the curated-identity index up front (curated sources only, i.e. "curated wins"), then drops any dedup-source hadith whose identity collides. **Exact** normalized-matn equality only — it never removes a distinct tradition (conservative, irreversible-load-friendly). Sub-exact near-duplicates remain a `PARALLEL_OF` linking concern (`src/resolve/parallels.py`), not a dedup one.

## Tests

- Identity unit tests: same / diacritic-variant / distinct / too-short / empty / English-fallback / Arabic-preferred (`tests/test_parse/test_composition.py`).
- Load-layer tests: curated-wins (no double-count), no-overlap no-op, curated-only-seeds-index (no self-dedup), short-matn-kept (`tests/test_graph/test_load_nodes.py`).
- Full suite: **988 passed, 7 Docker-skipped**. ruff + mypy-strict clean.
- Structural index regenerated; `staleness-check` green. `.hypothesis/` cache gitignored.

## Reviewers
- @ — **Alejandra Reyes-Fuentes** (B1 author — knows the dedup-deferral she left in `composition.py`)
- @ — **Ivana Horvat**

Closes #220


parametrization:
RequestOrReplied: Approved
TechDebt: none

Verified against the artifact at 5f002e3. canonical_matn_identity is deterministic — prefers normalize_arabic(matn_ar) (gated on is_arabic), falls back to lowercased matn_en, returns None below the 3-token floor, and digests a whitespace-collapsed signature via blake2b; the composition unit tests confirm diacritic/hamza/tatweel variants converge to one identity while distinct matns stay distinct, and short/empty matns carry no identity (kept). Curated-wins direction is correct — _build_curated_identity_index seeds from curated sources only (skips is_cross_edition_dedup_source), and _load_hadiths drops a sanadset row only on an EXACT identity hit in that index, so curated is kept and the sanadset twin is dropped; a sanadset row with no curated twin and two same-matn sanadset rows both fall through (no self-dedup). No over-deletion — exact normalized-matn equality only, sub-exact left to PARALLEL_OF; HADITH_COMPOSITION keeps sanadset at None, so collection breadth is preserved and dedup stays per-hadith. Both index build and load path use _effective_matn_ar (matn_ar then full_text_ar) so curated and sanadset are compared on the same basis. Interaction with my B1 is clean — dedup touches only Hadith nodes, the sanadset:None admission and Collection emission are untouched, so Collection nodes survive. Tests cover curated-wins/no-double-count, no-overlap no-op, curated-only-seeds-index, and short-matn-kept. LGTM.

parametrization:
Requestor: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

Reviewed at HEAD 5f002e3 from the NLP/identity lens, complementary to the second reviewer.

Matn normalization is robust. canonical_matn_identity routes Arabic matn through normalize_arabic (diacritics, alif/hamza/taa-marbuta folding, tatweel strip, whitespace collapse) so the same tradition across editions converges — test_normalization_collapses_spelling_variants exercises exactly that with a diacritized+tatweel copy. The _MIN_IDENTITY_TOKENS=3 floor correctly refuses an identity for bare formulae, so a too-generic short matn is kept rather than falsely merged — the conservative direction for an irreversible load. English fallback (lowercased) only fires when no Arabic char is present, and Arabic is preferred when both exist; both behaviors are tested.

Exact normalized-matn equality is the right scoping. It only ever removes a genuine textual duplicate; sub-exact near-dups are deliberately left to PARALLEL_OF (src.resolve.parallels), not silently dropped. I confirmed the docstring mirror claim against parallels._tokenize — same is_arabic/normalize_arabic/.split() pipeline — so the dedup gate and the parallel detector agree on what the normalized text is (identity uses ordered join, parallels uses a token set, which is the correct distinction).

Determinism holds. The signature is a blake2b digest of the order-preserving join of normalized.split() tokens — no dict/set iteration feeds the signature, and the curated index is only ever membership-tested. Same matn yields the same identity across calls (verified).

Curated-wins direction and seeding are correct. The identity index is seeded only from non-dedup curated rows that pass is_canonical_hadith, both index and load path compute identity on the same _effective_matn_ar basis (matn_ar with full_text_ar fallback), so a sanadset row is compared against the curated edition on equal footing. sanadset never deduplicates against itself (test_cross_edition_dedup_only_curated_seeds_index).

No B3 overlap — B2 drops whole duplicate hadith rows by matn identity and does not touch narrator re-segmentation (da#221). The two-pass build (full corpus scan, 5-column projected read) is a deliberate, documented choice since a dedup-source file may sort before its curated twin; acceptable and not a blocker.

Issue #221, opened 2026-06-25
Path B/B3: narrator re-segmentation for sanadset <NAR> firehose


parametrization:
## Path B / B3 — narrator re-segmentation for the sanadset `<NAR>` firehose

**Part of da#202 Path B** (owner decision 2026-06-25). Meta: noorinalabs-main#723. Investigation: ADR `docs/adr/003-sanadset-orphan-remediation.md` (PR #218).

### Problem
sanadset parsing emits a coarse `narrator_mentions_sanadset` firehose (`src/parse/validate.py` baseline expects ~2,789,517 rows) of `<NAR>`-tagged fragments that pollute the narrator table — raw isnad strings stored as `:Narrator` nodes, vowel-stripped transliterations, fragments like "…that"/"He said". This is also why full-text search surfaces matns as narrator hits (ig#1110) and why STUDIED_UNDER is near-absent (186 edges).

### Work
Narrator re-segmentation R&D: properly segment sanadset isnad chains into canonical narrator nodes + transmission edges instead of coarse `<NAR>` fragments. This is genuine R&D — the pollution is intrinsic to the dataset's coarse tagging.

### Alignment
Relates to the W3-deferred "chain segmentation STUDIED_UNDER" item (narrator→teacher edges); coordinate with the isnad-graph side.

### Scheduling
Deferred to a future Phase-7 data-quality wave (W3+). Not in W19.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Ivana Horvat
RequestOrReplied: Request

**Wave 20 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-20.
- Peer reviewer: Nikolaos Papadopoulos
- Secondary reviewer: Oyunbileg Batbayar
- Branch from: `deployments/phase-7/wave-20`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

PR #226, opened 2026-06-26
B3(da#221): re-segment + filter sanadset <NAR> narrator firehose

parametrization:
## da#221 — Path B / B3: narrator re-segmentation for the sanadset `<NAR>` firehose

Closes #221. Part of da#202 Path B (ADR-003); meta noorinalabs-main#723.

### Problem
`parse_sanadset` stored every `<NAR>` tag at face value as a narrator mention — a
~2.79M-row coarse firehose that polluted the resolved `Narrator` table with
honorific phrases (`رضي الله عنه`), bare transmission verbs (`قال`/`عن`),
English / vowel-stripped Latin transliteration fragments (`He said`, `that`,
`Malik ibn Anas`), and whole mis-tagged sub-chains. This is also why full-text
search surfaces matns as narrator hits (ig#1110) and `STUDIED_UNDER` is
near-absent (ADR-003).

### Change (B3 only — B2 dedup da#220 out of scope)
- `sanadset._extract_narrator_mentions` now routes every `<NAR>` tag through
  `_segment_nar_content`:
  - **re-segments** a marker-bearing blob into its constituent narrators via the
    shared `narrator_extraction.extract_narrator_mentions` Arabic segmenter,
    **dropping** an un-segmentable chain blob rather than minting it (da#158);
  - **filters** non-narrator pollution via `_is_narrator_like` — Arabic script
    required (drops English / Latin transliteration fragments), no residual
    transmission marker, not a whole honorific phrase, not a bare nasab particle.
- Genuine narrators flow through unchanged; the common single-name tag keeps its
  raw voweled `name_ar`. `position_in_chain` is numbered over emitted mentions so
  it stays gap-free under drops/splits.
- Parse-quality logging gains `raw_nar_tags` / `filtered_nar` / `filtered_nar_pct`
  for pollution-removal observability.
- `validate.py`: recalibrate `narrator_mentions_sanadset` baseline
  **2,789,517 → 1,950,000** (~30% net pollution removal, pending full-run
  confirmation per the da#175 recalibration precedent; the 30% drift tolerance
  absorbs the calibration band). Behaviour is pinned by the new B3 unit tests,
  not the exact prod number (CI has no raw corpus).
- Regenerated the structural ontology index for the touched source.

### Tests
New `TestNarratorResegmentation` (helper + end-to-end): genuine names accepted,
pollution classes rejected, raw voweled form preserved for clean single-name
tags, mis-tagged sub-chain re-segmented, polluted firehose filtered to only
genuine narrators with gap-free positions. Full `test_parse`/`test_utils`/
`test_resolve` suites green (574 passed, 3 skipped); ruff + mypy-strict +
structural-staleness all pass.

### Reviewers
@Nikolaos Papadopoulos and @Oyunbileg Batbayar (QA / data-quality —
narrator-table cleanliness).


parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

Verified against PR head 2a12329 with a data-quality lens.

Test rigor — not vacuous. test_polluted_corpus_is_filtered runs the full parse_sanadset over a real firehose-shaped CSV (SANAD/NAR/MATN tags interleaving two genuine narrators with a bare verb, an honorific, an English fragment, and a Latin transliteration) and asserts BOTH that only the two genuine names survive AND that positions stay gap-free [0,1] with NARRATOR_MENTION_SCHEMA intact. test_clean_corpus_mentions_unchanged proves genuine narrators are NOT over-filtered end-to-end. Unit coverage exercises accept, reject, raw-preservation, drop, and sub-chain re-segmentation.

Quality logging — raw_nar_tags, narrator_mentions, filtered_nar and filtered_nar_pct are all emitted in sanadset_parse_quality, giving observable evidence of the firehose drop. Minor, non-blocking: filtered_nar is raw_minus_clean so it can read slightly negative when blob re-segmentation yields more mentions than raw tags; it is an observability metric, not a gate.

Baseline discipline — the recalibrated 1,950,000 is documented as a pending-full-run estimate; behaviour is pinned by the B3 unit tests, not the magic number, which CI never exercises (no raw corpus). The 30% DEFAULT_DRIFT_TOLERANCE_PCT gives a meaningful [1.365M–2.535M] guard band that would still flag a total-over-filter or a no-op filter.

No false drops — the highest-risk path is clean. contains_transmission_marker is word-boundary anchored, so an a-n marker buried in a real name (Maʿn / ʿAnbasa) is not dropped. The 3-char minimum matches the shortest real names (Ali / Umar / Zayd); the structural-particle filter only rejects spans that are entirely particles; the honorific filter is whole-tag-only, so an honorific trailing a real name leaves the name intact (a documented, baseline-improving scope limit, not a regression).

parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

Verified against the artifact at HEAD (2a12329).

Segmentation — _segment_nar_content routes a marker-bearing tag through the shared extract_narrator_mentions segmenter and catches IsnadSegmentationError to DROP an un-segmentable chain blob rather than mint it (da#158); a clean single-name tag is kept. position_in_chain is numbered over emitted mentions only, so it stays gap-free under drops/splits. The mis-tagged-subchain test splits محمد بن عبدالله / علي بن أبي طالب and the polluted end-to-end test yields positions [0,1] with no gaps.

Pollution-filter precision — confirmed _is_narrator_like does NOT discard valid Arabic narrators. is_arabic matches on at-least-one Arabic char (not all-Arabic), so genuine names including short ones (علي/عمر/زيد at the 3-char floor) and female muhaddithat names pass; only non-Arabic spans are dropped. contains_transmission_marker is word-boundary anchored (da#155), so عن inside عنبسة and قال inside مقالة do not false-trip. Honorifics are rejected only on a whole-tag normalized-exact match, so an honorific trailing a real name leaves the name intact; the structural-particle reject fires only when every token is a particle, so عبد الله and أم سلمة survive.

Raw voweled name_ar preserved — the clean single-name path returns the raw stripped form unchanged (test asserts مَالِك بن أَنَس preserved while the normalized key strips diacritics); no over-normalization.

Baseline — 2,789,517 to 1,950,000 (~30% removal) is documented as full-run-pending per the da#175 precedent, is absorbed by the 30% drift tolerance, and is not a CI gate (no raw corpus in CI); behaviour is pinned by the new TestNarratorResegmentation unit tests instead. All 7 checks green (Lint, Type Check, Test, Integration, staleness, sync-drift, base-image).

Issue #161, opened 2026-06-14
feat(model): DatePrecision enum + earliest/latest date bounds on Narrator (additive)


parametrization:
Add a `DatePrecision` enum (exact/range/circa/after/before/tabaqa_estimate/unknown) and earliest/latest bound fields to the `Narrator` model alongside the existing `birth_year_ah`/`death_year_ah`. Keep precision (source tightness) orthogonal to the existing `confidence` (disambiguation strength). Backward-compatible/additive.

Part of meta noorinalabs/noorinalabs-main#673. Design: noorinalabs-main/.claude/research/historical-timeline-dates.md. Recommend P5W3.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Alejandra Reyes-Fuentes
RequestOrReplied: Request

**Wave 21 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-21.
- Peer reviewer: Kavitha Sundaramurthy
- Secondary reviewer: Nikolaos Papadopoulos
- Branch from: `deployments/phase-7/wave-21`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: feature (per charter § Wave Planning & Priority)

Please begin implementation.

Issue #214, opened 2026-06-25
feat(ontology): fan out C×T2 structural layer to noorinalabs-data-acquisition (generate + commit + wire CI)


parametrization:
## Summary
Fan out the ontology **C×T2 structural layer** to `noorinalabs-data-acquisition` (#820 Task 3). Wire the per-repo structural-index generator into this repo's CI/pre-commit so the structural index is generated where the source is visible and language-native, and commit the resulting `ontology/structural/` index in-repo. The central semantic overlay + cross-repo aggregator stay in `noorinalabs-main`.

**Language / approach (Python / PyArrow):** Full Python `ast`-based generation (cleanly derivable per #820 Task 1).

## Acceptance Criteria
- [ ] Generator invoked in this repo's CI (and pre-commit where applicable); structural index regenerates on source change.
- [ ] `ontology/structural/` committed in-repo (or a hand-curated stub documented as such, per #820 Task 1, if source is not cleanly derivable).
- [ ] Generated index validated against this repo's actual source (spot-check key modules/decls present).
- [ ] CI green; merge-driver form matches the standardized invocation (noorinalabs-main#871).
- [ ] Aggregator in noorinalabs-main can resolve this repo's structural index for cross-repo queries.

## Dependencies
None — Python path is ready.

## Origin
Phase 7 Wave 18 — framework carry-forward lead-in. Per-repo fan-out child of the C×T2 execution tracker **noorinalabs-main#820** (owner-chose C×T2, 2026-06-22). Implementer from `noorinalabs-data-acquisition`'s own roster (child-repo-implementer rule).

## Provenance
Parent tracker: noorinalabs-main#820 · framework-align: noorinalabs-main#862 · README: noorinalabs-main#863.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Kavitha Sundaramurthy
RequestOrReplied: Request

**Wave 18 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-18.
- Peer reviewer: Jean-Claude Habimana
- Secondary reviewer: Nikolaos Papadopoulos
- Branch from: `deployments/phase-7/wave-18`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: enhancement (per charter § Wave Planning & Priority)

Please begin implementation.

PR #233, opened 2026-06-26
feat(model): DatePrecision enum + earliest/latest date bounds on Narrator (additive)

parametrization:
## Summary

Additive narrator-dating foundation (root of the da#161–166 / ig#1039 chain).

Adds:
- **`DatePrecision`** StrEnum in `src/models/enums.py` — `exact` / `range` / `circa` / `after` / `before` / `tabaqa_estimate` / `unknown`, following the existing StrEnum pattern.
- Six additive fields on the `Narrator` model (`src/models/narrator.py`):
  - `birth_year_ah_earliest`, `birth_year_ah_latest`, `birth_date_precision`
  - `death_year_ah_earliest`, `death_year_ah_latest`, `death_date_precision`

The existing `birth_year_ah` / `death_year_ah` stay the **point estimate**. New bound fields default to `None` and precision defaults to `DatePrecision.UNKNOWN`, so every existing constructor/row is unaffected (backward-compatible). Precision (source dating tightness) is documented as **orthogonal** to disambiguation confidence.

This is the clean field-name root that da#162 (schema mirror), da#164/#165/#166 (parse / reconcile / ṭabaqa-fallback) and ig#1039 (Neo4j write) build on.

## Tests

New `TestDatePrecisionEnum` (members, values, StrEnum serialization) and `TestNarratorDateBounds` (bound assignment, string coercion, bad-value rejection, round-trip, frozen) in `tests/test_models/test_all_models.py`; extended the narrator default-field assertions. Full non-integration suite green: ruff (check + format), mypy (strict), pytest (960 passed / 7 skipped).

## Note on diff size

The `structural-ontology-staleness` pre-commit/CI gate (#214) requires the generated `ontology/structural/` index to be regenerated when `src/*.py` changes. `scripts/structural_ontology.py emit` was run and committed; that file is generated-not-hand-edited, and the regen also folded in pre-existing drift from prior merges, which accounts for most of the non-source line count.

Closes #161
Part of meta noorinalabs/noorinalabs-main#673.

## Reviewers
- Primary: Kavitha Sundaramurthy
- Secondary: Nikolaos Papadopoulos

TechDebt: none — purely additive model/enum change with full test coverage. The `tabaqa_estimate` precision value has no producer yet; its window-derivation logic is intentionally deferred to da#166 (tracked), not debt incurred here.


parametrization:
RequestOrReplied: Approved

**Review:** Clean, strictly-additive foundation for the narrator-dating chain (da#161 root). Verified the source at head 01cc66cec0, not the branch.

- Additive-only confirmed: the `nar:` id-prefix validator and `tabaqat_class` are untouched; the existing `birth_year_ah` / `death_year_ah` stay as the point estimate. The six new fields all default to `None` / `DatePrecision.UNKNOWN`, so every existing constructor and stored row still validates — exercised directly by the extended `test_narrator_optional_defaults`.
- `DatePrecision` follows the established `StrEnum` convention in `enums.py`, is added to the sorted `__all__`, and its member set matches the da#161 spec exactly (`exact / range / circa / after / before / tabaqa_estimate / unknown`). The docstring usefully ties each value to how the earliest/latest bounds relate to the point estimate, and keeps source-dating tightness orthogonal to disambiguation confidence.
- Field names (`{birth,death}_year_ah_{earliest,latest}` + `{birth,death}_date_precision`) are a clean surface for the downstream consumers (da#162 schema, da#164/#165/#166 parse/reconcile/fallback, ig#1039 loader) to mirror.
- Test coverage is solid: enum members/exhaustiveness/serialization, bound assignment, string coercion, invalid-value rejection, round-trip, and frozen-immutability. CI is fully green (Lint, Type Check, Test, Integration, staleness-check, sync-drift gate).

Must-fix: None

TechDebt: One advisory item (not incurred-here debt, but worth a tracking note): there is no cross-field invariant on the bounds — nothing enforces lower <= upper, or that a tightness value is consistent with its bounds (e.g. an exact-tightness pair collapsing to a single year, or an after-/before-style one-sided bound). That is defensible now because the fields are null/UNKNOWN until producers exist, but a model-level guard would catch parser bugs early once da#164/#165 begin populating these; recommend folding it into that work rather than this PR.

parametrization:
RequestOrReplied: Approved

**Review:** Secondary review, schema/consumer-contract lens (I built the da#163 hijri util; da#164/#166 consume these bounds). Re-reviewed at head 259ff7ad1e after the requested fix.

The shape was right from the start and the additive contract is clean:
- Field names `birth_year_ah_earliest/latest`, `death_year_ah_earliest/latest` are unit-unambiguous — AH, year-granularity, mirroring the existing point estimate, so the conversion + reconciliation layers consume them with no AH-vs-CE or year-vs-full-date ambiguity.
- Birth and death precision are independent, each defaulting to UNKNOWN, so a death-anchored-only narrator is representable.
- Defaults are explicit null + UNKNOWN (no 0/sentinel); StrEnum + `int | None` mirror 1:1 into da#162's NARRATORS_CANONICAL_SCHEMA.
- Backward-compatible, frozen preserved, round-trip + string-coercion tested.

My one blocking item is now resolved: a `model_validator(mode="after")` rejects an inverted window (`earliest > latest`) for each pair when both ends are present, scoped exactly to the ordering invariant — precision/cross-bound consistency is correctly left to the populating stages. Tests cover both pairs rejecting the inverted case, equal bounds allowed (single-year range), and one-sided AFTER/BEFORE windows left unchecked. All seven CI checks green at this head; mergeable.

Must-fix: None — the bound-ordering validator requested in the prior round landed exactly as scoped.

TechDebt: None blocking. The precision-implied bound relationships (the AFTER/BEFORE/EXACT coupling) remain documented-but-unenforced by design, tracked to the da#164/#166 populating stages — noted, not debt incurred here.

Issue #229, opened 2026-06-26
Derive in-book ordinal (hadith_number_in_book) per-source where reliable — da#153 item #1


parametrization:
**Owner decision (2026-06-26, P7W20 retro of da#153): derive where reliable.**

86 hadith have a null `hadith_number_in_book` (in-book ordinal). The explicit-null / no-fabrication contract from ADR-004 item #1 stands as the floor and must NOT be violated where the ordinal is genuinely unknown.

**Work:** in the per-scraper parse lane, derive the in-book ordinal from page order / a secondary field **where it can be done reliably**, leaving the explicit null only where it truly cannot be derived. This is a parse-lane change, not a load-layer change. No fabrication where the source is unreliable.

**Sizing:** `graph_integrity_deferred_inventory.cypher` block #1.

Provenance: follow-up to noorinalabs-data-acquisition#153 / ADR-004 item #1 (see also da#77 in-book-ordinal vs collection-ref). Meta: noorinalabs-main#723.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Kavitha Sundaramurthy
RequestOrReplied: Request

**Wave 21 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-21.
- Peer reviewer: Nikolaos Papadopoulos
- Secondary reviewer: Jean-Claude Habimana
- Branch from: `deployments/phase-7/wave-21`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

PR #234, opened 2026-06-26
da#229: derive in-book ordinal (hadith_number) per-source where reliable

parametrization:
## da#229 — Derive in-book ordinal (`hadith_number`) per-source where reliable

Closes #229. Follow-up to da#153 / ADR-004 item #1 (in-book ordinal vs collection-ref, da#77). Meta: noorinalabs-main#723.

### What
The staging `hadith_number` column is the in-book ordinal that flows to `APPEARS_IN.hadith_number_in_book` (da#77). For some sources it is legitimately null. ADR-004 item #1's explicit-null / no-fabrication contract is the floor; the owner decision (2026-06-26) is to **derive where reliable**, leaving null only where the ordinal genuinely cannot be recovered.

### How
- **New `src/parse/in_book_ordinal.py`** — a source-reliability registry (`ORDER_RELIABLE_CORPORA`) + `derive_in_book_ordinals()`. The derivation is gated on source reliability and is deliberately conservative so it never fabricates:
  - **Source gate** — only corpora whose parse-lane row order faithfully reproduces the in-book sequence are derived; every other source keeps its explicit null.
  - **Whole-book gate** — within a reliable source, an ordinal is derived only for a `(collection, book)` group whose ordinal is *entirely* null (a partially numbered book is ambiguous, so its nulls are left null).
  - **No overwrite** — a source-provided ordinal is never touched.
- **Wired into the sunnah.com scraped parser** (`src/parse/sunnah_scraped.py`) — the order-reliable page-order source per da#77: a book page that carries no per-hadith ordinal recovers it from page (row) order.
- Excluded the helper from the every-parser-emits-a-schema conformance guard (it is a shared helper, not a per-source parser).

### Scope note
`source_id` (node identity) is intentionally left on its original keying; the derived ordinal feeds only the downstream `APPEARS_IN.hadith_number_in_book` edge property, keeping this a focused parse-lane change with no identity churn.

### Tests
`tests/test_parse/test_in_book_ordinal.py` (derivation, per-book restart, null-when-unreliable, no-overwrite, partial-book-left-null) + two end-to-end paths in `test_sunnah_scraped_parser.py` (explicit ordinals preserved; derived-from-page-order). Full local set green: ruff format/lint, mypy strict, pytest (1013 passed), structural-ontology staleness.

TechDebt: none introduced. Reliability registry currently lists only `sunnah` (the verified order-reliable source); extending it to other sources is a per-source data decision for a future wave, each requiring evidence that the parser walks the source in in-book reading order. The residual `source_id` `:0` placeholder for any remaining null-ordinal rows is a pre-existing identity concern, intentionally out of scope here.

### Review
Reviewers: Nikolaos Papadopoulos (primary), Jean-Claude Habimana (secondary).


parametrization:
RequestOrReplied: Approved

**Review:**
Approving. The change cleanly delivers da#229's "derive where reliable" intent while keeping ADR-004 item #1's explicit-null / no-fabrication contract as a hard floor, and I verified all three guard rails in the code itself rather than from the PR prose:

- Source gate — `derive_in_book_ordinals` early-returns `0` when `is_order_reliable(source_corpus)` is false, so any non-`sunnah` corpus keeps every null. Covered by `test_unreliable_source_leaves_nulls`.
- Whole-book gate — `if any(records[i].get(ordinal_key) is not None for i in idxs): continue` skips a `(collection, book)` group the moment it carries any ordinal, so partial books stay null. Covered by `test_partially_numbered_book_is_left_untouched` and `test_mixed_books_derive_only_the_fully_null_one`.
- No-overwrite — same gate means a source-provided ordinal is never touched; the only mutation is assigning a positional value into the previously-null `ordinal_key`. Covered by `test_existing_ordinals_are_not_overwritten`.

Registry conservatism is correct. `ORDER_RELIABLE_CORPORA` lists only `sunnah`, and I checked the source-of-truth claim at HEAD: `src/acquire/sunnah_scraper.py::_scrape_book_page` builds rows via `soup.select(...)` (DOM order) and iterates `for row in rows: ... hadiths.append(record)`, so a hadith's position on its book page genuinely is its reading-order ordinal. The registry-sanity test (`ORDER_RELIABLE_CORPORA <= {c.value for c in SourceCorpus}`) also pins the set to real enum values.

Identity is not leaked. The helper writes only `ordinal_key`; `source_id` is keyed on the original extraction in `sunnah_scraped.py` and is left untouched. Confirmed by both the unit assertion on a preserved `source_id` and the e2e `test_explicit_in_book_ordinals_are_preserved`.

Tests are real and exercise every branch (derivation, per-book restart, interleaved grouping, separate-collection independence, null-when-unreliable, no-overwrite, partial/mixed book, empty no-op) plus two e2e paths through `run()`. The conformance-guard exclusion (`in_book_ordinal` added to the not-a-parser allowlist with rationale) is the right call for a shared helper. Full CI rollup is green, including the structural-ontology staleness-check, so the regenerated index is clean rather than hand-edited.

Must-fix: None

TechDebt: Grouping key is `(collection_name, book_number)`, and in `sunnah_scraped.py` `book_number` comes from `safe_int(...)`, which yields `None` for a named book segment (e.g. an "introduction"-style page). Two distinct unnumbered named-book pages within one collection would collapse into a single `(collection, None)` group and receive one continuous `1..N` sequence instead of restarting per page. This only bites under a narrow conjunction (same collection, two-plus named books, and no explicit per-hadith number on the page), and named pages on sunnah.com usually carry explicit numbers, so it does not block — but a future hardening could guard or per-page-key the `None`-book case so it cannot merge separate pages into one ordinal run. Worth a tracking note rather than a fix here.

parametrization:
RequestOrReplied: Approved

**Review:** Secondary review, lens on edge/ordering semantics, the data-decision boundary, the conformance exclusion, and integration side-effects. The derivation is correct and conservative.

- Ordering is deterministic. Grouping uses an OrderedDict keyed on `(collection_name, book_number)` with first-seen insertion order; the positional fill iterates `records` by index. The only frozenset (`ORDER_RELIABLE_CORPORA`) is used for membership, never iteration, so there is no set/dict iteration-order dependency. Upstream input order is itself deterministic (`sorted(glob("*.json"))` + JSON array reading order), so the recovered ordinal faithfully follows reading order.
- Per-book restart is correct: `enumerate(idxs, start=1)` restarts the 1-based count at every `(collection, book)` group, and interleaved/duplicated book keys group by key rather than position (covered by test_per_book_ordinal_restarts and test_interleaved_books_group_by_key_not_position).
- The whole-book gate and no-overwrite gate are both sound and tested (partial-book-left-null, existing-ordinal-not-overwritten, mixed-books-derive-only-fully-null, empty-noop). The explicit-null floor for unreliable sources holds (returns 0, mutates nothing).
- Data-decision boundary: deferring the registry to `sunnah` only is the right call. The inclusion criterion (parser walks the source in reading order) is genuinely source-specific evidence work; no other currently-wired source is obviously safe to add now, and the test asserts the registry stays a subset of real SourceCorpus values. Confirmed, not expanded.
- Conformance exclusion is justified: `in_book_ordinal` sits alongside `base`/`identity`/`composition` as a shared helper that mutates a parser's records in place and emits no schema or parquet of its own. The every-parser guard targets per-source parser modules; this is not one, so the exclusion hides no real schema gap.
- Integration is clean: `derive_in_book_ordinals` mutates only the `hadith_number` key; `source_id` is generated before the call and is deliberately left on its original keying. No other emitted field changes. No identity churn, consistent with the stated scope.

Must-fix: None

TechDebt: Two non-blocking, pre-existing/defensive items to track, not to fix here. (1) The placeholder `:0` segment in `source_id` for a derived-ordinal row is now divergent from the recovered ordinal it carries — the PR correctly scopes the identity rekey out (rekeying would orphan existing node references), but the recovered ordinal now makes that collision fixable in a future wave, so it is worth an explicit tracking item rather than only a prose mention. (2) If a `sunnah` book page were ever to land with a null `book_number`, those rows would group under `(collection, None)` and be numbered across what may be distinct books; currently unreachable because the scraper carries `book_number` per page, but a one-line guard (skip the group when `book_number is None`) or a documented invariant would harden the helper against future sources.

Issue #228, opened 2026-06-26
Mention-link the 8 orphan muhaddithat bio-only narrators (with provenance) — da#153 item #3


parametrization:
**Owner decision (2026-06-26, P7W20 retro of da#153): LINK (preserve), not drop.**

8 narrators promoted from the `muhaddithat` bio source were never mention-linked, so they sit orphaned with no graph edges. Per ADR-004 item #3 (da#153), the load layer deliberately did NOT auto-link or drop them.

**Work:** mention-link the 8 onto the hadith they transmit, **with provenance**, carefully — auto-linking blindly risks wrong attribution. Preserve these sourced female scholarly figures rather than dropping them as noise.

**Detection / sizing:** `queries/validation/orphan_narrators.cypher` + `queries/validation/graph_integrity_deferred_inventory.cypher` blocks #3/#3b enumerate them with a bounded sample.

Provenance: follow-up to noorinalabs-data-acquisition#153 / ADR-004 item #3. Meta: noorinalabs-main#723.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Kavitha Sundaramurthy
RequestOrReplied: Request

**Wave 21 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-21.
- Peer reviewer: Nikolaos Papadopoulos
- Secondary reviewer: Kwesi Boateng
- Branch from: `deployments/phase-7/wave-21`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: feature (per charter § Wave Planning & Priority)

Please begin implementation.

PR #235, opened 2026-06-26
Mention-link the 8 orphan muhaddithat bio-only narrators with provenance (da#228)

parametrization:
## Mention-link the 8 orphan muhaddithat bio-only narrators (with provenance)

Closes #228 — da#153 item #3 / ADR-004 item #3. Owner decision: **LINK (preserve)**.

### Problem
Eight female hadith scholars (*muhaddithat*) were promoted to canonical `Narrator`
nodes from the `muhaddithat` bio source (`bio_promote`) but never mention-linked —
their `external_id` appears in no transmission chain, so the `STUDIED_UNDER` edge
builder never references them. They sit orphaned with zero relationships
(`queries/validation/orphan_narrators.cypher`). ADR-004 deferred this as an owner
link-vs-drop call; the owner chose to preserve them by linking.

### Change
A curated, **evidence-anchored**, **non-fabricating** mention-link producer:

- **`src/resolve/muhaddithat_links.py`** — `MUHADDITHAT_ORPHAN_LINKS`: the 8 curated
  links (single source of truth). `build_muhaddithat_mention_links()` resolves each
  name to the **same** canonical id `bio_promote` mints
  (`make_canonical_id(normalize_arabic(name_ar))`) and emits a link **only when that
  narrator exists in the canonical master** — no link to a non-promoted narrator, no
  bulk-link beyond the named 8.
- **`MUHADDITHAT_MENTION_LINKS_SCHEMA`** — a superset of the resolved-mention schema
  with a **non-null, first-class `provenance`** column (the source of the link, never
  a comment). Being a superset, the output file rides the existing
  `_resolved_mentions_files` glob into the `NARRATED` loader unchanged.
- **`NARRATED` loader** — propagates `provenance` onto the edge via a property-less
  `MERGE` + coalesce-preserve `SET` (null-safe for ordinary chain mentions;
  idempotent — mirrors the APPEARS_IN/PARALLEL_OF contract, da#77).
- **`resolve.run_all`** — wires the producer after `bio_promote`/`cluster` so the
  canonical master is final.
- **ADR-004 item #3** updated to RESOLVED.

### Tests
`tests/test_resolve/test_muhaddithat_links.py` (exactly-8, per-link provenance,
canonical-id identity match, evidence guard skips non-promoted, no-bulk-link) and
`tests/test_graph/test_load_edges.py::TestNarratedProvenance` (NARRATED edge carries
each link's provenance; chain mentions stay null-safe). Full suite: 1020 passed,
7 skipped (docker integration). ruff/mypy/markdownlint/cspell green.

### Reviewers
Primary: **Nikolaos Papadopoulos**. Secondary: **Kwesi Boateng**.

TechDebt: The curated `hadith_id`s (`sunnah:bukhari:N` / `sunnah:muslim:N`) must be reconciled against the actually-loaded `sunnah` `Hadith` node ids by the data team; until then a curated id that does not match a loaded node is honestly counted as a `missing_endpoint` (never a dangling/fabricated edge). Per-narrator exact in-corpus hadith-id mapping is the tracked follow-up.


parametrization:
Requestor: Kwesi Boateng
RequestOrReplied: Approved

**Review:**

RE-VERIFIED at new head `a4c5a2f3a4` after the primary reviewer (Nikolaos) found a TRANSMITTED_TO fabrication path and Kavitha fixed it. My earlier approval stands — now unambiguously against the fixed code. Secondary review, graph-integrity lens; read at head, not the PR-body framing.

The fabrication path (my lens). A curated orphan link's `hadith_id` is a real `sunnah` id that the same hadith's own NER chain mentions also carry. Because the loader file-locator is a prefix glob, the curated `narrator_mentions_resolved_muhaddithat` file was read alongside the main resolved-mentions file; the old `_load_transmitted_to` grouped both by hadith and `_build_chain_pairs` paired the orphan narrator with that hadith's Companion into a wrong-attribution transmission edge. The narrator-only endpoint-check could not catch it — both endpoints are real Narrator nodes.

Fix verified — the path is closed. `_load_transmitted_to` now skips any row carrying a truthy provenance before it ever enters the by-hadith grouping, so a curated link can never reach chain-pair construction. The discriminator is exact in both directions: every curated row carries a non-null, non-empty provenance (schema column is non-nullable and the producer always populates it), while an ordinary resolved chain mention has no provenance column at all, so the lookup yields None and the row is kept. No curated narrator can enter a TRANSMITTED_TO pair, and the only theoretical mis-fire would be over-exclusion of a hypothetical real mention that grew a provenance column — which is not a fabrication and which no producer writes today.

NARRATED unchanged and still correct. The NARRATED loader deliberately does NOT apply the exclusion — curated links should become NARRATED edges, which is the whole point — and its query is byte-identical to what I approved before: property-less MERGE over MATCH'd endpoints plus a coalesce-preserve write of provenance. Idempotency, null-safety on ordinary chain mentions, and the no-clobber-of-real-provenance property are all intact. The no-orphan-edge guarantee (endpoint check counts a missing Hadith node as a missing endpoint, never a dangling edge) is unchanged.

Regression test is the right one. It co-locates a curated link with two real Companion mentions on the same hadith and runs the full edge load, asserting the legitimate Companion-to-Companion pair survives as the single transmission edge, the NARRATED narration still lands, and the orphan narrator appears in no transmission endpoint. That exercises the exact reproduction, end-to-end, not just the helper.

CI: all 12 checks terminal-green at this head (the three live-Neo4j jobs that were pending now SUCCESS).

Must-fix: None

TechDebt:
- One-narrator-per-hadith reduction race (latent, unchanged by this fix, non-blocking — already the subject of the reconciliation follow-up). The NARRATED loader keeps one narrator per hadith at the strictly-lowest position. In the co-located case the curated link and the real chain mention are both at position 0, so first-file-read wins; the regression test asserts NARRATED is one-or-more precisely because exactly one of the two lands. So once the sunnah corpus is reconciled, a curated orphan link only actually lands as NARRATED (and carries its provenance) if the muhaddithat scholar is the position-0 narrator under the same canonical id the sunnah chain resolves — otherwise the link is a silent no-op and the narrator stays orphaned. No fabrication either way; the data-team reconciliation should account for that race. (Corollary unchanged: the inline note that the narrator+hadith pair is the NARRATED identity is true at the MERGE but slightly oversells this loader path, which emits at most one NARRATED edge per hadith.)

parametrization:
RequestOrReplied: Approved

**Review:**

Re-reviewed at head `a4c5a2f3a4`. The TRANSMITTED_TO fabrication must-fix is resolved correctly and completely — flipping to Approved (edited in place from my original ChangesRequested).

The fix: `_load_transmitted_to` now drops provenance-bearing rows before grouping by `hadith_id` (`if row.get("provenance"): continue`), so a curated muhaddithat orphan-link can no longer enter `_build_chain_pairs` and be paired with a real chain narrator on a colliding hadith id. Verified the two halves of the contract both hold: the exclusion is null-safe (an ordinary chain mention has no `provenance` column, `row.get` -> None, so only curated links are ever dropped), and the curated rows still flow to `_load_narrated` unchanged (NARRATED has no such filter), so the orphan-link still produces its NARRATED edge with provenance. Wrong-attribution path closed while the legitimate edge survives.

The regression test (`test_curated_link_never_fabricates_transmitted_to`) is genuine, not a tautology: it co-locates a curated link AND two real Companion mentions on the *same* `hadith_id`, then asserts exactly one TRANSMITTED_TO pair is created (the Companion->Companion one) and that the orphan narrator's canonical id appears in **no** TRANSMITTED_TO endpoint, while NARRATED still lands. Pre-fix, the three co-located resolved mentions would yield two pairs with the orphan in the chain sequence, so both the `created == 1` and the endpoint assertions would fail — it exercises the actual bug, not a restatement of the new code.

All 12 CI checks are terminal SUCCESS (including the live-Neo4j Integration Tests) and the PR is MERGEABLE.

Must-fix: None.

TechDebt:
- Unchanged from the PR's stated item — reconcile the curated `sunnah:bukhari:N` / `sunnah:muslim:N` ids against the actually-loaded `sunnah` Hadith node ids. One residual, non-blocking note for that follow-up: on a hadith where a curated link's id collides with a real resolved chain mention, the position-0 NARRATED reduction is a strict-`<` tie and the main resolved-mentions file sorts first, so the orphan may not win its NARRATED edge on that specific hadith (it stays orphaned there). That is a functional-completeness corner of the same reconciliation work, not a fabrication, and is correctly out of scope for this PR.

Issue #230, opened 2026-06-26
Collection metadata enrichment: sourced name_ar + expected_count (e.g. riyadussalihin) — da#153 item #5


parametrization:
**Owner decision (2026-06-26, P7W20 retro of da#153): enrich from sourced metadata.**

Some `Collection` nodes carry a blank/null `name_ar` or lack `expected_count` (e.g. `riyadussalihin`). Per ADR-004 item #5, the load layer leaves these null rather than fabricating an Arabic name.

**Work:** fill the missing `name_ar` + `expected_count` from **sourced** metadata (not guessed) in the relevant parser's `collections_*` emission, source-by-source.

**Sizing:** `graph_integrity_deferred_inventory.cypher` block #5 (any Collection missing `name_ar` or `expected_count`).

Provenance: follow-up to noorinalabs-data-acquisition#153 / ADR-004 item #5. Meta: noorinalabs-main#723.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Alejandra Reyes-Fuentes
RequestOrReplied: Request

**Wave 21 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-21.
- Peer reviewer: Kavitha Sundaramurthy
- Secondary reviewer: Kwesi Boateng
- Branch from: `deployments/phase-7/wave-21`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

PR #236, opened 2026-06-26
feat(parse): sourced name_ar + expected_count Collection enrichment (da#230)

parametrization:
## Summary

Enriches `Collection` metadata with a **sourced** Arabic name (`name_ar`) and a new `expected_count`, per da#153 item #5 / ADR-004. Some sources emit Collection rows with a blank `name_ar` or no authoritative count (the named example is `riyadussalihin`); the load layer previously left these null rather than fabricate a value.

### Approach
- New curated, **sourced** table `src/parse/collection_metadata.py` — every entry carries an explicit provenance citation (`source`). `apply_collection_metadata(row)` fills `name_ar` (only when blank) and `expected_count` by collection slug, **never** overriding a source-supplied value; unknown slugs are left null (absence of a curated entry is not a guess).
- `Collection` model + `COLLECTION_SCHEMA` gain `expected_count` (nullable int32); distinct from `total_hadiths` (source-reported) — it is the canonical reference total read by `collection_coverage.cypher`.
- Graph Collection loader carries `expected_count` onto the node.
- Enrichment wired into every `collections_*` parser emission (sunnah_api/scraped are the riyadussalihin producers; others get the key + any curated match).
- Seeded entries (all cited to sunnah.com): `riyadussalihin` (رياض الصالحين, 1896 — corroborated by project memory `sunnah_scraper_truncation`), `nawawi40` (42), `qudsi40` (40).

### Tests
`tests/test_parse/test_collection_metadata.py` asserts the sourced riyadussalihin example end-to-end (curated value + provenance, helper fill/no-override, model field, and the sunnah parser emitting `expected_count=1896` + name_ar into Parquet). Test fixtures updated for the new column.

## Reviewers
- Primary: Kavitha Sundaramurthy
- Secondary: Kwesi Boateng

Closes #230

TechDebt: Curated table seeds 3 sunnah.com collections (riyadussalihin, nawawi40, qudsi40); the remaining gap-collections from block #5 of graph_integrity_deferred_inventory.cypher still need sourced entries — extend the table source-by-source in follow-up.


parametrization:
RequestOrReplied: Approved

**Review:**

Verified the whole enrichment path at head `4c2378716c`. This is a clean, well-scoped, sourced-not-fabricated change.

- **Sourced, no fabrication** — `COLLECTION_METADATA` entries each carry an explicit `source` citation (all sunnah.com), and `test_every_entry_carries_provenance` enforces it. `apply_collection_metadata` fills `name_ar` only when the row's own value is blank (`test_does_not_override_source_supplied_name_ar` proves a source value survives), fills `expected_count` only when null, and leaves unknown slugs untouched with `expected_count=None` (key present so the row satisfies the schema, value never guessed). Input is not mutated (`test_does_not_mutate_input`). Fill-vs-override semantics are correct.
- **riyadussalihin example** — رياض الصالحين / 1896 cited to sunnah.com/riyadussalihin, corroborated by the repo's own `project_sunnah_scraper_truncation` memory, and `test_riyadussalihin_enriched_in_parquet` runs the real Sunnah parser and asserts both land in the emitted Parquet end-to-end against `COLLECTION_SCHEMA`.
- **expected_count vs total_hadiths** — genuinely distinct: both fields coexist in `COLLECTION_SCHEMA` (lines 107/113), the `Collection` model, and the `_COLLECTION_MERGE` cypher (sets `n.total_hadiths` AND `n.expected_count`). No rename, nothing in this repo conflates them; `total_hadiths` (source-reported) is untouched.
- **Propagation** — schema (int32 nullable), model (`int | None`), loader `_val` extraction + merge cypher are all consistent. The lk_corpus switch to `from_pylist(..., schema=COLLECTION_SCHEMA)` is a correctness tightening, and the `pa.table` parsers (sunnah_api/scraped/bihar) rely on `setdefault("expected_count", None)` to supply the key the schema-keyed comprehension requires — correctly wired into every collections emitter.
- **Fixtures + conformance exclusion justified** — the 3 fixture edits are additive `expected_count` int32 columns to keep test tables schema-conformant (no regression masked); excluding `collection_metadata` from the per-parser provenance enumeration is correct since it is a curated table, not a per-source parser, and emits no Parquet (same rationale as the existing `composition` exclusion).

CI is fully green (Lint, Type Check, Test, Integration, staleness-check, sync-drift, base-image realism).

Note (no action): the consumer `collection_coverage.cypher` lives in isnad-graph, not this repo, so its read of the new field is asserted by the PR but not verifiable here — the producer side is correct.

Must-fix: None

TechDebt: As the author declares — the curated table seeds only 3 sunnah.com collections; the remaining gap-collections from block #5 of `graph_integrity_deferred_inventory.cypher` still need sourced entries, to be extended source-by-source in follow-up.

parametrization:
Requestor: Kwesi Boateng
RequestOrReplied: Approved

**Review:** Secondary pass focused on data-correctness, loader null-safety, scope discipline, and the coverage-query contract — all four hold up.

- Curated values are sourced and verifiably right. riyadussalihin=1896 is corroborated by the cited repo memory project_sunnah_scraper_truncation (679 introduction + 1,217 books 1-19 = 1,896), not taken on faith; nawawi40=42 and qudsi40=40 match the sunnah.com reference ranges. Each entry carries an explicit `source` citation, so a wrong count would be auditable rather than silent.
- Loader is null-safe and non-clobbering. `_val(row, "expected_count")` yields None for unseeded collections and the MERGE writes that without disturbing other node fields; no KeyError because the parquet column is always present (schema-driven table build). The mandatory `setdefault("expected_count", None)` in the enrichment helper is what guarantees the key exists for the `r[field.name]` table construction across all producers.
- Scope discipline is correct. All 10 `collections_*` producers (bihar, fawaz, halimbahae, lk_corpus, open_hadith, sanadset, sunnah_api, sunnah_scraped, thaqalayn, tusi) call the enrichment; the 7 unseeded ones emit null — not 0, not a guess — which matches the stated ADR-004 no-fabrication intent. validate.py references the schema only for validation, not row construction.
- Coverage-query contract degrades gracefully on null. collection_coverage.cypher guards with a NOT-NULL CASE before dividing, so a null reference total returns null deviation rather than a div-by-zero or false 0% coverage. The cypher is unchanged by this PR; only the field it reads is now populated.

Must-fix: None

TechDebt: None beyond what the PR body already declares (7 gap-collections from block #5 still need sourced entries; extend the table source-by-source in follow-up) — that carry-forward is correctly scoped out and tracked.

Issue #162, opened 2026-06-14
feat(schema): extend NARRATORS_CANONICAL_SCHEMA with date/precision fields


parametrization:
Extend the PyArrow `NARRATORS_CANONICAL_SCHEMA` to carry the new date bound + DatePrecision fields end-to-end through staging→curated.

Part of meta noorinalabs/noorinalabs-main#673. Design: noorinalabs-main/.claude/research/historical-timeline-dates.md. Recommend P5W3.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Kavitha Sundaramurthy
RequestOrReplied: Request

**Wave 21 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-21.
- Peer reviewer: Ivana Horvat
- Secondary reviewer: Alejandra Reyes-Fuentes
- Branch from: `deployments/phase-7/wave-21`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: feature (per charter § Wave Planning & Priority)

Please begin implementation.

Issue #163, opened 2026-06-14
feat(utils): src/utils/hijri.py AH↔CE conversion + pin convertdate


parametrization:
Add `src/utils/hijri.py` (tabular AH↔CE, computed once at resolve, store both) and pin `convertdate` (or equivalent). Used by the date-parsing + reconciliation steps.

Part of meta noorinalabs/noorinalabs-main#673. Design: noorinalabs-main/.claude/research/historical-timeline-dates.md. Recommend P5W3.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Nikolaos Papadopoulos
RequestOrReplied: Request

**Wave 21 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-21.
- Peer reviewer: Oyunbileg Batbayar
- Secondary reviewer: Kwesi Boateng
- Branch from: `deployments/phase-7/wave-21`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: feature (per charter § Wave Planning & Priority)

Please begin implementation.

PR #237, opened 2026-06-26
feat(schema): extend NARRATORS_CANONICAL_SCHEMA with date/precision fields (da#162)

parametrization:
## Summary
Extends `NARRATORS_CANONICAL_SCHEMA` (`src/resolve/schemas.py`) with the six additive narrator-dating fields introduced on the `Narrator` model by da#161, so the canonical PyArrow schema validates and round-trips them. This is the schema half of the narrator-dating foundation; da#164/#165/#166 (parse/reconcile/fallback) and ig#1039 (Neo4j loader) consume it.

Fields mirrored 1:1 from the da#161 `Narrator` model:
- `birth_year_ah_earliest`, `birth_year_ah_latest`, `death_year_ah_earliest`, `death_year_ah_latest` — `int32`, nullable
- `birth_date_precision`, `death_date_precision` — `string`, nullable (the `DatePrecision` StrEnum serializes to its string value)

## Nullability (load-bearing)
All six are **nullable** in the schema, matching the layer's existing pattern (e.g. `generation`/`gender`/`trustworthiness` are required on the model but nullable in the schema). The robust canonical-table producers project rows via `{f.name: r.get(f.name) for f in SCHEMA}` (`bio_promote._build_table`, `fuzzy_cluster._build_table`), so they emit `None` for unpopulated precision/bound columns — `nullable=False` would have crashed both at write time.

## Hardcoded-builder fix
Not every canonical-table builder uses the robust `r.get`-over-schema loop — one production path hardcodes the full column dict and passes `schema=`, which makes PyArrow require every field. Enumerated every `schema=NARRATORS_CANONICAL_SCHEMA` writer:
- **Production:** `src/resolve/disambiguate.py::_build_canonical_table` (the only hardcoded producer) now emits the new columns — precision → `DatePrecision.UNKNOWN.value`, bounds → null. The other two prod writers auto-adapt via `r.get`.
- **Tests:** `tests/test_graph/conftest.py` + `tests/integration/test_graph_loading.py` canonical builders carry the new columns. (`tests/conftest.py` `NARRATOR_BIO_SCHEMA` is a different schema, correctly untouched.)

## Tests
`tests/test_resolve/test_schemas.py` — `TestNarratorModelMirror`: schema↔model field-parity + nullability assertions and a populated round-trip through the `Narrator` model.

## Structural ontology
Regenerated `ontology/structural/` (`scripts/structural_ontology.py emit`); staleness gate green.

## Local verification (full CI-parity set)
ruff format --check ✓ · ruff check ✓ · mypy --strict (82 files) ✓ · pytest (1052 passed, 7 skipped) ✓ · structural-ontology staleness ✓.

Closes #162

TechDebt: The model→schema mirror is **manual** — there is no model-derived-schema mechanism in the resolve layer, so the six fields are duplicated by hand here; an org-wide model→PyArrow-schema derivation is a future opportunity (out of scope). Also: the "unset precision" representation differs across the three canonical producers (`disambiguate` writes `"unknown"`, `bio_promote`/`fuzzy_cluster` write `None`); harmless under this additive change, but da#164/#165/#166 + ig#1039 must agree that a null precision reads as `UNKNOWN` so a `DatePrecision(None)` never reaches model reconstruction.


parametrization:
Requestor: Ivana Horvat
RequestOrReplied: Approved

**Review:**
The schema change is correct and complete, and the earlier blocker is resolved. Verified against the da#161 `Narrator` model at HEAD (`d4d02ee66a`):

- **1:1 model↔schema mirror — confirmed.** All six fields match the model exactly: `birth/death_year_ah_earliest|latest` are `int | None` on the model and `pa.int32() nullable=True` in the schema; `birth/death_date_precision` is `DatePrecision` (StrEnum, default `UNKNOWN`) on the model and `pa.string() nullable=True` in the schema.
- **Nullability is right, and non-nullable would have broken at runtime.** Two prod canonical producers — `bio_promote._build_table` and `fuzzy_cluster._build_table` — project rows with `{f.name: r.get(f.name) for f in SCHEMA}`, emitting `None` for the new precision columns on unpopulated rows; `nullable=False` would raise there.
- **Hardcoded-builder fix is complete.** The only prod path hardcoding a per-column dict (`disambiguate._build_canonical_table`) now emits all six (bounds→null, precision→`DatePrecision.UNKNOWN.value`); the other two prod writers auto-adapt; the two hardcoded test builders are updated; `tests/conftest.py` builds a different schema and is correctly untouched. No remaining hardcoded builder.
- **Tests are real** — field-parity vs `Narrator.model_fields`, per-field nullability, and a populated round-trip. CI all green.

The PR description has been corrected to describe the canonical-schema extension and now reads `Closes #162` (the prior copy-paste of the da#163/hijri body is gone), so the earlier must-fix is cleared.

Must-fix: None

TechDebt: The "unset" representation of the precision columns is not uniform across the three canonical producers — `disambiguate` defaults precision to `"unknown"` while `bio_promote`/`fuzzy_cluster` leave it `None`. Harmless under this additive change (column is nullable, population is downstream), but the populating stages (da#164/#165/#166) and the loader consumer (ig#1039) should settle on whether a null precision reads as `UNKNOWN` so a `DatePrecision(None)` never reaches model reconstruction. Already captured in the PR body's TechDebt note.

parametrization:
RequestOrReplied: Approved

**Review:** Secondary review with the da#161-author lens — I wrote the `Narrator` model this schema mirrors, so I checked the date/precision contract 1:1 at head `d4d02ee66a`.

The schema work is correct and the contract is clean. All six fields match `src/models/narrator.py` exactly: the four bounds (`birth_year_ah_earliest/latest`, `death_year_ah_earliest/latest`) are `pa.int32()` — consistent with the existing `birth_year_ah`/`death_year_ah` point columns, no int32/int64 drift — and the two `*_date_precision` columns are `pa.string()` carrying the `DatePrecision` StrEnum value, all `nullable=True`. The StrEnum round-trips to its string value and the `UNKNOWN` → `"unknown"` default matches the model default. `TestNarratorModelMirror` is a genuine drift guard (checks both `Narrator.model_fields` and the schema name set), the nullability test is right, and `test_roundtrips_with_model` asserts the actual values and precision survive projection — not just field presence. The `_build_canonical_table` edit leaves every non-date column (generation/gender/trustworthiness) untouched, and the schema-driven producers (`bio_promote`/`fuzzy_cluster`) do not break on the new columns.

My original blocker — the PR description was copy-pasted from the da#163 hijri PR and carried a stray `Closes #163` that would have closed the wrong issue — has been fixed. The body now describes the canonical-schema extension, mirrors the six da#161 fields, and ends `Closes #162`. Re-verified the body at the current head.

My earlier precision-asymmetry note is now documented in the body and correctly broadened: the "unset precision" representation differs across the three canonical producers (`disambiguate` writes `"unknown"`; `bio_promote`/`fuzzy_cluster` write null). Harmless under this additive, schema-only change. The downstream contract — da#164/#165/#166 + ig#1039 must read a null precision as `UNKNOWN` so a `DatePrecision(None)` never reaches model reconstruction — is the right place to enforce it, and the team lead is filing it as a follow-up. Approving.

Must-fix: None.

TechDebt: Producer precision-default asymmetry (null vs `"unknown"`) across `disambiguate`/`bio_promote`/`fuzzy_cluster` — documented in the PR body and tracked as a follow-up; not blocking this additive change.

Issue #139, opened 2026-06-13
feat(resolve): geographic filtering stage in narrator disambiguation (Phase-5 — needs location enrichment)


parametrization:
## Provenance
Relocated from **isnad-graph#721** at P4W4 (owner decision 2026-06-12): mis-filed (the disambiguation pipeline `src/resolve/disambiguate.py` was extracted from isnad-graph into data-acquisition, commit `e878cf1`) AND Phase-5 data-gated. Per the standing relocate-mis-filed rule.

## Scope
Implement the geographic-filter stage in `src/resolve/disambiguate.py` (`_geographic_filter`, currently a deliberate pass-through stub ~L261). Effective filtering needs:
1. Narrator **location strings normalized to canonical city/region IDs** (`birth_location_id`/`death_location_id` in `src/models/narrator.py` are always None today; canonical schema `src/resolve/schemas.py` has birth/death YEAR but no location columns).
2. A **travel-plausibility graph** (could two narrators plausibly have met given their locations/eras).

## BLOCKED ON (Phase-5)
The **enrich stage is empty** (`src/enrich/` = just `__init__.py`) — nothing produces a normalized location ontology / location IDs. This stage can only be meaningfully implemented once the location-enrichment work lands. Deferred to **p4-wave-5 / Phase-5**. Forcing it now would either filter on noisy free-text (silently dropping valid matches — the stub exists to avoid exactly this) or be dead code. Owner: Farhan Malik. Ref isnad-graph#721, da#98 (sibling data-gate defer).

parametrization:
Deferred to Phase 5 at /wave-scope 4 5 (P4W5 = 'Phase-4 exit drive'). Phase-5 work (needs location enrichment); will be re-scoped at /plan-phase 5. p4-wave-5 label removed.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Ivana Horvat
RequestOrReplied: Request

**Wave 21 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-21.
- Peer reviewer: Alejandra Reyes-Fuentes
- Secondary reviewer: Oyunbileg Batbayar
- Branch from: `deployments/phase-7/wave-21`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

PR #238, opened 2026-06-26
feat(resolve): geographic filtering stage in narrator disambiguation (#139)

parametrization:
## Summary

Implements the geographic-filter stage in narrator disambiguation
(`src/resolve/disambiguate.py` `_geographic_filter`), previously a deliberate
pass-through stub. This is the Phase-5 tech-debt carryover that was gated on
location enrichment.

### What landed
- **`src/resolve/geography.py`** (new) — free-text location → canonical region
  normalization (English / transliteration / Arabic aliases for major historical
  Islamic scholarly centres) plus a conservative west→east **travel-plausibility**
  model. Hijaz is treated as a universal pilgrimage hub; only antipodal region
  pairs (e.g. Andalus ↔ Khurasan) are judged implausible, so attested long-range
  rihla corridors (Iraq ↔ Transoxiana, Andalus ↔ Iraq) stay plausible.
- **`_geographic_filter`** now consumes a chain-context signal — the canonical
  regions of the resolved adjacent narrators' locations — exactly mirroring how
  `_temporal_filter` consumes the death-year index. `ChainContext` gains
  `adjacent_locations`; `run()` maintains a parallel `location_index`.
- **Conservative by construction**: a candidate is dropped only when both its own
  location and at least one neighbour's resolve to known regions AND the candidate
  is travel-implausible against *every* known neighbour region. Any unresolved or
  missing location keeps the match (soft constraint) — noisy/sparse free-text
  location data can never silently drop a valid match.

### Graceful no-enrichment path
No normalized location ontology exists yet (`locations.yaml` is absent; many bios,
e.g. muhaddithat, carry no location). The stage **no-ops gracefully** in that case
and only activates where real location strings are present (itqan/sanadset bios
populate `birth_location`/`death_location`). No locations are fabricated.

### Tests
- `tests/test_resolve/test_geography.py` — `resolve_region` (English, Arabic,
  `al-` stripping, token fallback, unknown to None), `regions_plausible`
  (same/adjacent/hub/antipodal/unknown), `is_travel_plausible`.
- `tests/test_resolve/test_disambiguate.py` — filter behaviour incl. plausible
  keep, antipodal drop, candidate unknown/missing location kept, unresolvable
  neighbour pass-through, any-neighbour rule, and the **graceful no-context /
  no-enrichment** path.

Regenerated `ontology/structural/` (#214 staleness gate). Full suite green;
ruff + mypy(strict) clean.

Reviewers: Alejandra Reyes-Fuentes (primary), Oyunbileg Batbayar (secondary).

Closes #139

TechDebt: None — this PR retires the `_geographic_filter` pass-through stub itself (the standing Phase-5 carryover). The location seed in `geography.py` is intentionally a modest curated set; widening it to the full historical gazetteer is future enrichment work (populated `locations.yaml` / `Location` ontology), not debt incurred here.


parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved

**Review:**

Approving. This retires the long-standing `_geographic_filter` pass-through stub (da#139) with a stage that is conservative exactly where it must be, and I verified the no-false-drop property in code rather than from the PR narrative.

Conservatism / no-false-drop (the core risk) — confirmed at the source. A candidate is pruned only when (a) its own location resolves to a known region, (b) at least one neighbour resolves to a known region, AND (c) the candidate region is travel-implausible against *every* resolved neighbour region. Each "don't know" resolves toward keep: `geography.resolve_region` returns None for unrecognised/empty/whitespace free-text and the filter keeps that match; `regions_plausible` returns True whenever either side is None, equal, a hub, or an unmapped region; and an empty reference-region set short-circuits to a full pass-through. The any-neighbour rule (`any(regions_plausible(cand, ref) ...)`) is the right disjunction — one plausible neighbour saves the match. Tests exercise every one of these branches (candidate-unknown-kept, candidate-no-location-kept, unresolvable-neighbour-passes-all, kept-when-plausible-against-any-neighbour, mixed-batch-drops-only-implausible) plus the graceful no-context / no-enrichment path that is the common case today.

No fabricated locations — the normalization only maps existing `birth_location`/`death_location` free-text through a curated alias table to a canonical region; nothing is invented, and the candidate-side read (`death_location or birth_location`) is a pure lookup.

Travel model soundness — coarse west→east zones with a gap threshold of 4 keeps the attested long-range corridors plausible (Andalus↔Iraq gap 4, Iraq↔Transoxiana gap 3) while rejecting only antipodal pairs (Andalus↔Khurasan 6, Egypt↔Transoxiana 5). Hijaz as a universal pilgrimage hub is modelled correctly via the hub short-circuit. The model errs toward keeping, which is the correct bias for a graph-integrity filter.

Stage-interface fidelity — `_geographic_filter` faithfully mirrors `_temporal_filter` (missing signal passes through, soft per-candidate constraint), `location_index` parallels `death_year_index` in `run()` and is populated identically (resolved bio-matched mentions whose candidate carried a location), and `ChainContext.adjacent_locations` is wired through both the linear and indexed disambiguation paths. Like the temporal index, the location index only carries neighbours resolved earlier in the stream — order-dependent partial signal — but that is a pre-existing property of this incremental design, not introduced here, and it only ever *reduces* drops, so it stays on the safe side.

One forward-looking observation (not blocking): the per-token fallback in resolve_region broadens resolution as the alias seed grows, which slightly widens the confidently-resolved surface that drops can act on. With the current distinctive transliteration seed this is a non-issue; if/when the gazetteer expands toward `locations.yaml`, keep an eye on token-level false resolutions. That belongs with the tracked enrichment work, not this PR.

CI: Lint, Type Check, Test, Integration Tests, staleness-check, sync-drift gate, and base-image/fixture all green at b2e54bc.

Must-fix: None

TechDebt: None

parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Ivana Horvat
RequestOrReplied: Approved

**Review:** Reviewed from the geo-correctness / premise-honesty / determinism lens at head `b2e54bc34c`. All CI green (Lint, Type Check, Test, Integration, staleness, sync-drift, base-image).

Region normalization is sound. The English/transliteration/Arabic alias tables map the major historical centres to the right region with no conflation — Kufa/Basra/Baghdad/Wasit/Mosul→iraq, Damascus/Homs/Aleppo/Jerusalem→sham, Fustat/Cairo/Alexandria→egypt, Nishapur/Merv/Herat/Balkh→khurasan, Bukhara/Samarqand→transoxiana, Rayy/Isfahan/Hamadan/Qom→jibal, Cordoba/Seville→andalus, etc. Arabic matching is consistent with `normalize_arabic`: aliases are normalised at import through the same pipeline, so the deliberately-redundant ta-marbuta / alif-hamza variant spellings (e.g. الكوفة vs الكوفه, الأندلس vs الاندلس) collapse to one key and `setdefault` keeps the same region — harmless belt-and-suspenders, no cross-region key collision exists in the seed. Token fallback handles descriptive ("المدينة المنورة") and mixed-script ("Basra, Iraq") forms.

Premise honest, not dead code. `run()` populates `location_index` from `c.death_location or c.birth_location` exactly where `death_year_index` is populated, and the filter reads the same free-text precedence — so it genuinely operates on the populated free-text path. It degrades to a pure pass-through in three independent ways: no `chain_context`, empty `adjacent_locations`, or no resolvable `ref_regions`; an unresolved/missing candidate location keeps the match. Sparse/absent enrichment can never silently drop a valid match.

Determinism holds. The drop decision is a boolean over `any(regions_plausible(cand_region, ref) for ref in ref_regions)`, so the `ref_regions` set's iteration order does not affect the outcome; neighbours are gathered in fixed `(-1, 1)` order; the token fallback returns the first matching token over an ordered split. Travel model is symmetric — `regions_plausible` uses `abs(zone_a - zone_b)` and an OR-ed hub test — and the quantifier is correctly "implausible vs EVERY known neighbour" (keep on first plausible). Threshold=4 keeps the attested long-range corridors (Andalus↔Iraq gap 4, Iraq↔Transoxiana gap 3) while dropping antipodal pairs (Andalus↔Khurasan gap 6, Egypt↔Transoxiana gap 5); Hijaz as universal hub is the right pilgrimage modelling. Tests exercise the any-neighbour rule, antipodal drop, soft-constraint keeps, and the no-context/no-enrichment no-op — not trivial assertions.

The declared scope is right: this retires the `_geographic_filter` pass-through stub, and the full gazetteer / `locations.yaml` / `Location` ontology is genuine future enrichment, not debt incurred by this change.

One non-blocking observation (no action needed): on a free-text string carrying tokens from two different regions, the token fallback resolves to the first matching token; this is inherent free-text ambiguity and the soft-constraint design keeps it conservative either way.

Must-fix: None
TechDebt: None

Issue #164, opened 2026-06-14
feat(parse): extract narrator dates from Itqan rijāl + Taqrīb/ṭabaqāt (death-anchored)


parametrization:
Parse narrator life-dates from the accepted Itqan rijāl source and a Taqrīb al-Tahdhīb / ṭabaqāt extraction (owner decision: extend the Itqan posture). Death-anchored (wafāt far better attested than birth). Per-date provenance, cleanly removable like Itqan.

Part of meta noorinalabs/noorinalabs-main#673. Design: noorinalabs-main/.claude/research/historical-timeline-dates.md. Recommend P5W3.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Ivana Horvat
RequestOrReplied: Request

**Wave 21 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-21.
- Peer reviewer: Alejandra Reyes-Fuentes
- Secondary reviewer: Nikolaos Papadopoulos
- Branch from: `deployments/phase-7/wave-21`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: feature (per charter § Wave Planning & Priority)

Please begin implementation.

PR #240, opened 2026-06-26
feat(parse): extract narrator dates from Itqān rijāl (death-anchored) (#164)

parametrization:
## Summary

Implements da#164 — extract narrator life-dates from the Itqān rijāl source into the da#161 `*_year_ah_earliest/latest` + `*_date_precision` bound fields. **Death-anchored**: the well-attested `wafāt` is the primary signal; birth is parsed only when explicitly stated (the ṭabaqa-derived birth fallback is da#166).

### What's here
- **`src/parse/narrator_dates.py`** (new) — the extraction engine. Recognises the notation forms the Itqān `death` field actually carries (verified against the fixture):
  - bare year → `EXACT`; `بين A إلى/و B` / `between A and B` / `A-B` → `RANGE`; leading `بعد A` → `AFTER` (open `latest`); leading `قبل A` → `BEFORE` (open `earliest`); `تقريبا`/`circa` → `CIRCA`; multi-alternative `أو`/`وقيل` lists → honest `[min,max]` span with the first-stated year as the point.
  - Arabic-Indic numerals normalised; `بعدها`/`قبلها` (no trailing digit) do **not** trigger an open bound; absurd (>1500) / placeholder / pure-prose fields yield no date.
- **`NARRATOR_BIO_SCHEMA`** extended with the six bound/precision columns so the parse stage carries what it extracts; the resolve stage (da#165) reconciles them onto `narrators_canonical`.
- **Itqān parser** populates the new fields; `death_year_ah` point now derives from the same parser (single source of truth, no drift). **muhaddithat** + **sanadset** emit null bounds with a concrete `"unknown"` precision.

### Foundation-review carry-forwards honoured
1. Death AH→CE conversion uses `ah_year_to_ce_range` (inclusive window), **not** the point `ah_to_ce` — so a death late in an AH year is not ~1 CE year low (`death_ce_bounds`, tested against the AH-1 → 622/623 case).
2. Every emitted date carries a concrete precision; a silent source maps to the `"unknown"` string — never a null that could become `DatePrecision(None)`.

### Tests
`tests/test_parse/test_narrator_dates.py` (41 cases): each notation form, Arabic-Indic digits, the `قبلها` false-positive guard, the concrete-precision invariant, the CE-via-range path, and round-trip construction of a `Narrator` from the extracted fields. Itqān parser + bio-schema round-trip tests extended. Full suite: 1120 passed / 7 skipped; ruff + mypy(strict) green; structural index regenerated.

Closes #164

TechDebt: Single-source pass-through only — the bio→canonical carry + cross-source reconciliation (pick tightest/widen on disagreement, date-trio atomicity) is deliberately left to da#165; sanadset's scalar CSV point-years are NOT promoted to EXACT bounds here (da#165/#166). CE bounds are computed by `death_ce_bounds` but not persisted — there is no CE field on the model/canonical schema yet (out of da#164 scope). The `بعد A … إلى B` "after a span" sub-form is read as plain `AFTER A` (the trailing span is dropped); acceptable since `after A` already covers it.

---

Reviewers: Alejandra Reyes-Fuentes (primary), Nikolaos Papadopoulos (secondary).


parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved

**Review:** Secondary review with the hijri-API lens (I wrote `src/utils/hijri.py`). The consumption of the AH to CE conversion is correct and the two foundation-review carry-forwards are genuinely honoured.

Verified against the source at head `fed49decf3`:

- **Range converter, not point.** `death_ce_bounds` uses `ah_year_to_ce_range` and reads `[0]` (CE start) for the earliest bound and `[1]` (CE end) for the latest bound — the widest correct window. It does NOT use the point `ah_to_ce` (which returns the CE year of 1 Muharram and runs ~1 CE year low for a death late in the AH year). The boundary test actually exercises this: for AH 1 it asserts `ce_latest == 623` and `ce_latest > ah_to_ce(1)` (623 > 622), so a passing point-converter implementation would fail the test. That is the right way to pin it.
- **Open bounds.** AFTER yields `latest = None` and BEFORE yields `earliest = None`; `death_ce_bounds` maps the unbounded side to `None` (it only converts the non-None side), and the AFTER/BEFORE CE tests confirm one-sided windows. Sensible.
- **No sub-1 year can reach my converter.** `ah_year_to_ce_range` raises for `year < 1`; every bound originates from `_years()`, which filters through `_plausible` (>= 1), and the range/alternatives paths sort before constructing, so `earliest <= latest` always holds. The converter is never handed a value that would raise within this flow.
- **Parser robustness / fail-safe.** Placeholders, whitespace, pure prose, and implausible numerals (`3200`, `0`) all return `None` rather than raising or fabricating; `parse_year_notation` has no path that raises (the `ParsedDate` ordering invariant is pre-satisfied by sorting). Good.
- **Ordering determinism.** The alternatives span is `sorted(set(years))` so `[min, max]` is stable regardless of token order, with the first-stated year as the point. The بعدها/قبلها guard is an anchored regex requiring a trailing digit after the marker (`^\s*(?:قبل|before)\s*(\d+)`), not a fragile substring `in` test — `test_qablaha_does_not_trigger_before` confirms قبلها falls through to the honest span.
- **Schema population.** All three producers emit a concrete precision: Itqān via `to_field_dict`, muhaddithat + sanadset via the explicit `"unknown"` string. No producer leaves a null precision that could become a downstream `DatePrecision(None)`.
- The scope boundary (single-source pass-through, cross-source reconcile deferred to da#165, CE bounds computed-but-not-persisted) is declared honestly. CI rollup is fully green (Lint, Type Check, Test, Integration, sync-drift, staleness).

Must-fix: None

TechDebt: One minor lossiness worth a tracking note (NOT blocking, fits the deferred-reconcile envelope): an approximation marker on a multi-alternative field — e.g. a hypothetical `"نحو 230 أو 232"` — collapses to a single CIRCA point at `years[0]` because the circa branch precedes the alternatives-span branch, dropping the second candidate. The attested corpus forms don't appear to combine circa with an alternatives list, so this is theoretical; if da#165 ever sees it, prefer carrying the span. Fold into da#165 rather than this PR.

parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved

**Review:**

Reviewed as the da#161 date-model author at head fed49decf3 (full file via contents API, no branch-switch). Both foundation carry-forwards are honoured and proven by test, the notation engine matches the genre's real forms, and the point-year derivation is drift-free.

- CE conversion: `death_ce_bounds` routes through the inclusive-window converter for both ends; the AH-1 case surfaces a CE upper of 623 (strictly above the point converter's 622), so a death late in an AH year is no longer ~1 CE year low. The open-bound sides correctly yield None.
- Concrete-precision invariant: a silent source maps to the `"unknown"` string and round-trips back through the enum — no path can produce a null precision downstream. Every attested parse carries EXACT/RANGE/AFTER/BEFORE/CIRCA.
- Notation: bare year exact; between/span ranges; leading after opens the upper side, leading before opens the lower side; approximation marker rides on the precision without fabricating a window; competing attestations collapse to an honest min/max envelope with the first-stated year as the point. The suffix forms with no trailing digit are correctly NOT read as open bounds — the start-anchor plus the required digit immediately after the marker both protect against it.
- No fabrication / no drift: the point year now derives from the same parser as the bounds; I traced every form against the prior first-token behaviour and they agree. Silent sources emit null bounds with a concrete unknown precision.
- Bounds ordering is enforced at construction and the open-upper case satisfies the model's ordering validator (tested via real Narrator construction). Arabic-Indic and Persian numerals both normalise.
- Integration: all three bio producers carry the six new columns and `validate.py` reads the schema object dynamically; no producer is left to mismatch the widened schema. CI is fully green.

Must-fix: None

TechDebt:
- A noise integer that happens to fall in the plausible AH band (e.g. a page or hadith number leaking into a free-text death field) can be mistaken for a competing attestation and widen the envelope; the only guard is the upper plausibility cap. Inherent to the free-text source and acceptable for the fixture-verified forms in scope; reconcile (da#165) is the place to sanity-bound this.
- A marker followed by an interposed word before the digit (the "after year N" shape) is not read as an open bound because the digit must immediately follow the marker; it degrades to exact. Not among the tested forms; flag for da#165/#166 if it surfaces in the corpus.
- The scalar point-year sources keep a concrete unknown precision with null bounds rather than promoting the point to exact — a deliberate deferral to da#165/#166, already disclosed by the author.

Issue #165, opened 2026-06-14
feat(resolve): multi-source date reconciliation


parametrization:
Reconcile dates across sources (reuse `source_ids`/`confidence`; honor the disambiguate-before-bio-promote ordering). Emit resolved bounds + DatePrecision per narrator.

Part of meta noorinalabs/noorinalabs-main#673. Design: noorinalabs-main/.claude/research/historical-timeline-dates.md. Recommend P5W3.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Alejandra Reyes-Fuentes
RequestOrReplied: Request

**Wave 21 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-21.
- Peer reviewer: Kavitha Sundaramurthy
- Secondary reviewer: Jean-Claude Habimana
- Branch from: `deployments/phase-7/wave-21`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: feature (per charter § Wave Planning & Priority)

Please begin implementation.

PR #241, opened 2026-06-26
feat(resolve): multi-source date reconciliation (da#165)

parametrization:
## feat(resolve): multi-source date reconciliation (da#165)

The reconciliation stage between per-source parse (da#164, merged) and ṭabaqa
fallback (da#166, next). When several rijāl sources date the same narrator
(collapsing onto one `nar:` canonical id), `src/resolve/date_reconcile.py` folds
their per-source parsed bounds into a single canonical birth/death envelope +
concrete precision, wired into `run_all` as a stage after `bio_promote`/`cluster`.

### Reconciliation rules (documented in the module)
- **Agreement** — points within `AGREEMENT_TOLERANCE` (2 AH yr) collapse to the
  agreed value (`EXACT`); the routine ±1–2 yr classical *wafāt* disagreement is
  agreement, not conflict.
- **Conflict** — genuine disagreement keeps an **honest tight `RANGE [min,max]`**
  (no invented winner or midpoint); a spread beyond tolerance is surfaced via a
  `conflict` flag for QA.
- **Noise robustness** — anchors on the agreement among point estimates; a
  per-source bound more than `NOISE_GAP` (10 yr) beyond the consensus cluster is
  rejected (a lone wide/​noisy bound cannot widen the envelope), and with ≥3
  sources a point more than `OUTLIER_GAP` (25 yr ≈ a generation) from the median
  is down-weighted out of consensus.
- **circa + multi-alternative** (Nikolaos's da#164 carry-forward) —
  `parse_source_notation` re-derives a `circa`+`أو/وقيل` field as `CIRCA` spanning
  `[min,max]` instead of collapsing to a point, and cross-source circa bands are
  likewise preserved.
- **sanadset scalar** — a bare year with `unknown` precision + null bounds is
  folded in as a known point and **upgraded** to a concrete `EXACT`.

### Standing contracts honored
1. Death AH→CE via `ah_year_to_ce_range` (inclusive window), never point
   `ah_to_ce` — see `ce_bounds`.
2. Null precision reads as `UNKNOWN`, never `DatePrecision(None)`: normalised at
   the input boundary, and the canonical stage leaves **every** row (mention-only
   included) with a concrete precision string.

The stage is idempotent (pure function of the bio inputs). 23 new tests cover
agreeing / conflicting / wide-conflict sources, the circa+alternative case, noise
+ outlier robustness, sanadset folding, open bounds, the always-concrete-precision
invariant, inclusive CE conversion, and the end-to-end canonical-table run.
Structural ontology index regenerated (#214). Full local check set green (ruff /
format / mypy / pytest incl. integration / structural-staleness).

Reviewers: **Kavitha Sundaramurthy** (primary), **Jean-Claude Habimana** (secondary).

TechDebt: The intra-source circa+multi-alternative fix lives in
`parse_source_notation`, but the resolve stage consumes da#164's already-parsed
bio *columns* (raw date text is not retained on `NARRATOR_BIO_SCHEMA`), so that
specific collapse is only corrected at production scale once source parsers route
raw notation through `parse_source_notation` (or da#164 adopts the envelope fix) —
the cross-source envelope is preserved today regardless; tracked as a follow-up.
Two further follow-ups: (a) no per-source authority weighting on a true conflict
(we keep the honest envelope rather than picking a "winner"); (b) a `fuzzy_cluster`
merge that unifies two canonical ids after this stage keys each surviving id to
its own bios, so a merged-away minority's dates are not re-folded.

Closes #165


parametrization:
RequestOrReplied: Approved

**Review:** (updated at head 8f243e89b6 — my must-fix is resolved.)

The single-source wide-range defect is fixed. Widening now clips a bound only when a COMPETING point cluster (≥2 distinct in-band points) contradicts it, and the NOISE_GAP gate is anchored on the fixed consensus band rather than the running envelope. I re-verified against the new logic:

- da#164's "between 200 and 230" (closed RANGE point=200) now reconciles to RANGE [200, 230] — upper bound preserved, RANGE precision, no fabricated EXACT. "between 80 and 100" → RANGE [80, 100] likewise. So the da#162 canonical columns now receive the correct death_year_ah_latest and a "range" precision string (no dropped bound, no spurious "exact").
- Noise robustness intact: 256/257 + a noisy latest=290 still rejects the 290 (the 256≠257 competing cluster makes it >NOISE_GAP beyond consensus) → RANGE [256, 257]. Outlier down-weighting (255/256/300 → RANGE [255,256], conflict) and the agreement/wide-conflict cases all still hold.
- A legitimately wide bound with no competing cluster is preserved (two sources agreeing on the point, one spanning wider → band kept). New test test_single_source_wide_range_is_preserved_not_collapsed locks the contract.

The order-independence anchoring on the fixed consensus band (Jean-Claude's axis) is sound on my axis too — the competing-gate and the fixed-band anchor compose cleanly. All 7 CI checks green at the new head.

Approving. The remaining TechDebt item below is non-blocking.

Must-fix: None (the single-source wide-range collapse is resolved).

TechDebt:
- parse_source_notation's _ALT_MARKERS includes "/" and the Arabic comma, which are broad. A circa-marked field carrying an AH/CE slash (e.g. "circa 180هـ / 795م", 795 <= MAX_PLAUSIBLE_AH=1500) would re-derive as a spurious CIRCA band [180, 795]. Latent today since the resolve stage consumes da#164's parsed columns rather than raw text (already noted in the PR's own TechDebt line); worth tightening the alternative-marker detection before raw notation is routed through this function.

parametrization:
RequestOrReplied: Approved

**Review:** Re-verified at head `8f243e8`. My must-fix is resolved: the envelope-widening NOISE_GAP gate is now anchored on the FIXED consensus band (`core_lo`/`core_hi`) rather than the mutating `lo`/`hi`, so admission no longer depends on prior iterations and the final `lo`/`hi` are min/max over an order-invariant admitted set — a chain of in-band bounds can no longer walk the envelope past consensus. I traced the new `test_widening_is_order_independent` against the pre-fix logic and it genuinely discriminates: old mutating-anchor logic yields forward 83 / backward 92 (assert fails); the fixed-anchor logic yields 92 either way. `test_lone_wide_bound_does_not_widen_envelope` still holds (290 is 33 beyond core_hi 257, rejected). Both TechDebt items were folded too: `_reconcile_open` now reserves the conflict flag for the genuinely-contradictory after-greater-than-before case so a benign bracketed open range no longer over-flags, and `_ALT_MARKERS` dropped the bare slash and the Arabic comma so a circa AH/CE slash notation no longer spuriously widens. My single-source self-band concern is also gone, resolved by the `competing` guard (a coherent one-source range is preserved in full instead of collapsing to a false EXACT). All 7 CI checks green. Approving.

Must-fix: None — the order-dependence defect is fixed and locked by a regression test.

TechDebt: One residual, non-blocking and a defensible documented trade-off: noise-clipping is disabled whenever the in-band points reduce to a single distinct value (`competing` is False), so a stray wide bound carried by a source whose point coincides with an otherwise-unanimous consensus would widen the envelope unchecked. This is the deliberate resolution of "can't distinguish a real self-band from noise without a competing cluster" and the narrow case (an agreeing source carrying a stray latest) is far rarer than the single-source preservation it buys; worth a watch-item if QA later surfaces over-wide envelopes on unanimous-point narrators, not a blocker.

PR #242, opened 2026-06-26
feat(resolve): ṭabaqa → estimated-window fallback for undated narrators (da#166)

### LINKED ISSUE #166: feat(resolve): ṭabaqa → estimated-window fallback for undated narrators
When no explicit date is attested, derive an estimated lifespan window from generation/ṭabaqa as the universal fallback. Mark as `tabaqa_estimate` precision. (Owner decision #3: these windows are displayed, clearly marked.)

Part of meta noorinalabs/noorinalabs-main#673. Design: noorinalabs-main/.claude/research/historical-timeline-dates.md. Recommend P5W3.

parametrization:
## da#166 — ṭabaqa → estimated-window fallback for undated narrators

The **final** stage of the P7W21 narrator-dating pipeline. After da#165
reconciles every *attested* per-source dating into the canonical envelope, some
narrators still carry no death date — the rijāl sources simply never dated them.
This stage derives an **estimated** death-year window from the narrator's
**ṭabaqa class**, tagged `DatePrecision.tabaqa_estimate` so a derived window is
never confused with an attested one. (Owner decision #3 on meta
noorinalabs-main#673: these windows are displayed, clearly marked.)

### Which ṭabaqa signal
The canonical parquet (`NARRATORS_CANONICAL_SCHEMA`) persists the coarse
`generation` (`NarratorGeneration`) layer, **not** the fine-grained Taqrīb
ordinal that lives only on the in-memory `Narrator.tabaqat_class`. `generation`
*is* the ṭabaqa class at canonical granularity — `src/parse/itqan.py` already
folds Ibn Ḥajar's twelve Taqrīb ṭabaqāt into these five buckets — and the design
note (`.claude/research/historical-timeline-dates.md` §1) names generation/ṭabaqa
layering the "universal fallback that bounds an estimated window." So the stage
keys on `generation`.

### What changed
- **New** `src/resolve/tabaqa_dates.py` — a sourced generation→AH death-window
  table following Ibn Ḥajar's Taqrīb periodization: SAHABI `[11,110]`, TABII
  `[90,180]`, TABA_TABII `[150,250]`, ATBA_TABA_TABIIN `[220,300]`, LATER
  `[280,400]`. Point = window midpoint; CE via `ah_year_to_ce_range` (the
  inclusive-window converter, **never** point `ah_to_ce`); precision is always
  the concrete `tabaqa_estimate`. Windows are wide multi-decade spans that
  overlap at the boundaries — a ṭabaqa is a *generation*, so the honesty is in
  the width, not a spurious point.
- **Wired** as `run_all` **Step 3.65**, immediately after reconcile (3.6).
  Fills the four death columns **only** when the death event is still undated
  (point + both bounds null AND precision UNKNOWN) **and** a known ṭabaqa class
  exists. Never overwrites a reconciled/parsed date — including a lone open
  `AFTER`/`BEFORE` bound. A null/unknown ṭabaqa stays undated + UNKNOWN (no
  fabrication). Birth is untouched (death-anchored). Idempotent.
- Regenerated `ontology/structural/` (gate #214).

### Tests (`tests/test_resolve/test_tabaqa_dates.py`, 15 cases)
Window-table values; midpoint + concrete `tabaqa_estimate` precision; AH→CE via
the inclusive-window converter; the known-ṭabaqa-only gate; and the end-to-end
stage — undated-known-ṭabaqa filled, reconciled `exact` date preserved, open
`after` bound preserved, null/unknown ṭabaqa left undated+UNKNOWN, birth
untouched, idempotent, no-canonical no-op.

### Local checks (CI parity)
`ruff check` ✓ · `ruff format --check` ✓ · `mypy src/` ✓ · `pytest tests/test_resolve` 275 passed/3 skipped ✓ · structural-staleness gate ✓ (all also via pre-commit/pre-push).

TechDebt: The generation→window table is a coarse 5-bucket approximation of the
12-ṭabaqa Taqrīb scheme — the fine-grained ordinal (`Narrator.tabaqat_class`) is
not persisted on the canonical parquet, so a per-ordinal (tighter) window cannot
be derived here without a schema+producer change to carry the ordinal through.
The windows are also death-only (no ṭabaqa-derived birth estimate) and are not
authority-weighted. Tracking finer-grained ṭabaqa windows is a future refinement.

Closes #166

Reviewers: Ivana Horvat (primary), Oyunbileg Batbayar (secondary).


parametrization:
Requestor: Ivana Horvat
RequestOrReplied: Approved

**Review:**

Verified against the file at head `b27c3647fc`, not the PR prose. This is a clean, well-guarded last-resort stage and the load-bearing judgment calls hold up.

**Last-resort gating (the core risk) — verified in code.** `_death_is_undated` returns true only when `death_year_ah`, `death_year_ah_earliest`, and `death_year_ah_latest` are all null AND precision is in `{None, "", "unknown"}`. A lone open AFTER bound (earliest set) or BEFORE bound (latest set) fails the second/third null check, so the fallback never clobbers it; any concrete precision (exact/range/circa/after/before) also fails the gate. The three preservation paths are not just asserted in the body — `test_stage_does_not_overwrite_reconciled_date` (EXACT 161 untouched), `test_stage_does_not_overwrite_open_after_bound` (AFTER 130, latest stays null), and the null/unknown-stays-undated case all exercise it directly. The clobber-a-real-date failure mode is genuinely closed.

**generation-vs-tabaqat_class design call — sound.** I confirmed `src/parse/itqan.py::_TABAQAT_GENERATION` actually folds Ibn Hajar's twelve Taqrib ordinals into the five `NarratorGeneration` buckets (1 to sahabi, 2 to 4 to tabii, 5 to 8 to taba_tabii, 9 to 10 to atba_taba_tabiin, 11 to 12 to later). So `generation` IS the tabaqa layer at the granularity the canonical parquet actually persists — keying on it is the right call, not a loss of available resolution. The window-table boundary comments (Taqrib 5 to 8, 9 to 10, 11 to 12) line up exactly with that fold, which is the right kind of internal consistency. The finer ordinal genuinely isn't on `NARRATORS_CANONICAL_SCHEMA` (I checked schemas.py), so the per-ordinal-window gap is a real schema+producer limitation, correctly captured below rather than hacked around.

**Window table — defensible.** SAHABI 11 to 110 (Prophet's death to the last Companion), and the overlapping multi-decade spans for the later layers, are honest for a generation-level signal; the overlap-on-purpose and point=midpoint choices are right (a tabaqa is a generation, the honesty is in the width). Pinned by `test_window_table_values` so a silent edit fails.

**No fabrication / AH-CE / ordering — all good.** Unknown/null generation yields `None` and is skipped (`generation_from_value` normalises out-of-domain to UNKNOWN, single membership gate). Birth columns are never written. Idempotent: the written `tabaqa_estimate` precision plus the now-set point both drop the row out of the undated gate on re-run, confirmed by `test_stage_is_idempotent`. CE span uses `ah_year_to_ce_range(lo)[0]` / `(hi)[1]` — the inclusive-window converter on each end, never the point `ah_to_ce`. Wired as Step 3.65 after reconcile 3.6 and before muhaddithat_links 3.7, in its own try/except, no-op when the canonical table is absent. All seven CI checks green at head; the only non-logic diff is the regenerated structural index.

One non-blocking observation captured below: `TabaqaEstimate.death_ce_earliest`/`death_ce_latest` are computed and unit-tested but not persisted — the canonical schema has no CE columns (consistent with da#164/#165; the loader does AH to CE at load time). Harmless and symmetric, not a defect.

Must-fix: None

TechDebt:
- (author's own, agreed) The generation to window table is a coarse 5-bucket approximation of the 12-ordinal Taqrib scheme; the fine ordinal (`Narrator.tabaqat_class`) is not carried onto the canonical parquet, so a tighter per-ordinal window needs a schema+producer change. Windows are death-only and not authority-weighted.
- `TabaqaEstimate` carries CE bounds that are computed and tested but never written to the canonical parquet (no CE columns there; CE is derived at load time). Either drop them or document that they exist for a future CE-aware consumer, so the next reader doesn't hunt for where they land.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved

**Review:** Secondary review, distinct-lens pass over da#166 (the final P7W21 dating stage). I traced the undated predicate, the window table, idempotency, the AH→CE usage, and determinism. All correct.

- Undated predicate (`_death_is_undated`) is exhaustive and conservative: a set point, a set earliest, a set latest, OR any precision outside null/blank/unknown all read as already-dated and are left alone. I probed the edges the brief called out — a lone open-upper bound (earliest set, latest null, AFTER precision) is treated as already-dated (the earliest-set branch short-circuits, and the stage-level test asserts it is preserved); a lone open-lower bound (latest set, BEFORE) likewise; an EXACT da#164/#165 dating untouched. The fallback only ever fills genuine gaps.
- Window table: all five spans are well-ordered (11/110, 90/180, 150/250, 220/300, 280/400); the point is the integer midpoint, so for every row earliest <= point <= latest. The da#161 after-model validator only rejects an inverted earliest>latest pair — every emitted row satisfies it. The deliberate boundary overlap (a generation is a generation, not a disjoint bucket) is the honest representation; midpoint-as-point is sensible given that.
- Idempotency holds: after the first pass an estimated row carries a non-null point, so the point branch of the predicate short-circuits on re-run before precision is even consulted — the stage never re-touches or widens its own output. The idempotence test asserts first == second.
- AH->CE uses the inclusive-window converter with the widest span (lower-bound start, upper-bound end), never the point converter — matches the documented contract.
- Determinism and no fabrication: lookups use `.get`, so a generation value outside the five known layers (or null/garbled, via safe_str) normalises to the unknown layer, yields no window, and the row stays undated+unknown — no KeyError, no default-guess.

CI rollup is fully green (Lint, Type Check, Test, Integration, sync-drift, base-image, structural staleness). Tests are thorough (15 cases) and assert the right post-conditions.

Must-fix: None

TechDebt: The estimate carries derived CE bounds that the stage never persists, because the canonical schema has no CE columns (CE is derived downstream from AH, consistent with every prior date stage / da#165) — so those fields are test-only surface; harmless, worth a note. Also already captured in the PR body: the five-bucket generation approximation of the twelve-tabaqa Taqrib ordinal (the finer ordinal is not carried on the canonical parquet), and the windows being death-only and not authority-weighted.

PR #251, opened 2026-06-30
fix(#723): resolve-pipeline name-quality + dedup-resume + chains-from-resolved + pollution scrub

### LINKED ISSUE #244: NER pass aborts on a single trailing-قال span; disambiguate skipped (W21 regression, blocks main#723 reload)
## Summary

The `resolve` NER step aborts the **entire** mention-extraction pass when a single isnad span trips the `IsnadSegmentationError` fail-loud guard, because `src/resolve/ner.py:123` calls `extract_narrator_mentions()` per-hadith with no error handling. One un-segmentable span kills NER, which cascades to **skip `disambiguate` entirely** ("NER failed — no mention data available") — producing a curated dataset with narrators (from bios) but **no resolved mention links**.

Surfaced during the main#723 prod re-validation reload (full corpus). CI did not catch it: no test fixture contained the offending span.

## Reproduction

Full-corpus `isnad-ingest resolve` on the W20/W21 output:

```
IsnadSegmentationError: segmented narrator span still contains a transmission
marker (partial chain blob): 'ابو القاسم عبد الله بن احمد بن عامر الطاءي قال'
  File "src/resolve/ner.py", line 123, in _extract_from_hadiths
    spans = extract_narrator_mentions(isnad_text, language)
  File "src/parse/narrator_extraction.py", line 156, in _extract_arabic
    raise IsnadSegmentationError(...)
```

## Root cause

An **asymmetry** between the two halves of the Arabic segmenter:

- `extract_transmission_phrases()` failed to consume a trailing `قال` ("qāla") as a phrase boundary, so it rode along into the tail name segment.
- `contains_transmission_marker()` (the independent, normalization-anchored post-split guard at `narrator_extraction.py:154`) **does** flag that standalone trailing `قال`, so it raises.

The guard itself is **correct** — it is the da#158 protection that stops partial-/whole-chain blobs from being minted as fake narrators (the exact pollution that triggered main#723). The defect is twofold:

1. `_clean_ar_name()` only strips a fixed set of trailing *connectives* (`_AR_TRAILING_CONNECTIVES`), explicitly excluding `قال` on the assumption (code comment, `narrator_extraction.py:46`) that "`قال` is already consumed as a transmission phrase upstream" — which is **not always true**.
2. `ner.py` has no per-hadith guard, so a single benign dangling verb aborts the whole pass instead of resolving the underlying name.

## Fix (chosen approach — root-cause, no chains dropped)

Strip a **trailing standalone transmission marker** in `_clean_ar_name()` so the underlying name resolves correctly (e.g. `…الطاءي قال` → `…الطاءي`). Trailing-only: a marker in the *middle* of a span still trips the fail-loud guard (a genuinely-unsegmented multi-narrator blob must not be silently merged). Reuse the already-imported `contains_transmission_marker` on the whole trailing token so a `قال` embedded in a real name (`مقالة`, `معن`) is never stripped. Add a regression test with the exact failing span.

## Impact

Blocks the main#723 prod re-validation reload (criteria #1 orphan-% and #2 STUDIED_UNDER depend on resolved mentions). No prod data was loaded.

### LINKED ISSUE #245: tech-debt: dedup/parallels stage OOMs + non-resumable (resolve crash site) — bound memory + checkpoint
## Problem

The final resolve stage `dedup` + `detect_parallels` (`src/resolve/dedup.py`) is the long pole and the actual crash site of the 2026-06-27 prod-reload resolve run: it ran ~19:37→~01:00 and never wrote `parallel_links.parquet`. The host slowly ran out of memory with CPU pegged — an OOM.

Memory accumulates without bound across the stage:
- `all_embeddings` holds every chunk before a single `np.vstack` (~768k × 384 float32 ≈ 1.1 GB just for embeddings) plus the FAISS index copy.
- The pair-collection loop (`dedup.py:219`) accumulates `seen_pairs` over the full `n × top_k` neighbour matrix.

It is also **not resumable**: a crash anywhere in the stage discards the entire embedding pass (hours of CPU).

## Asks (hardening)
1. **Bound peak memory** — stream/mmap embeddings (write chunks to the `.npy` incrementally or memory-map), avoid holding `all_embeddings` + `vstack` + index simultaneously.
2. **Checkpoint / resume** — persist `hadith_embeddings.npy` per-chunk and skip already-encoded chunks on re-run so a crash doesn't lose the whole encode.
3. **Progress already exists** (`dedup_encoding_progress`) — extend it with elapsed/ETA + a memory-watermark log so an approaching OOM is visible.

## Context
Sibling to the `fuzzy_cluster` composite-token-blocking fix (branch `I.Horvat/0000-fuzzy-cluster-block-opt`) which removed the *cluster*-stage OOM. This issue covers the *dedup*-stage OOM that fix does not touch ("there may be others").

### LINKED ISSUE #247: data-quality: 40% of canonical narrators have polluted names (NAR-markup leakage, thaqalayn whole-text, mubham descriptors) — #723
## Summary

**39.9% of canonical narrators (115,405 / 289,385)** on the current resolve output have polluted `name_ar_normalized` values — non-name text minted as `Narrator` nodes. This is a core #723 graph-quality defect (it inflates node count, spawns spurious edges, and — separately — makes fuzzy `cluster` pathologically dense/slow). Three distinct root causes, by impact:

### 1. `<NAR>` markup leakage — sanadset (HIGH confidence, contained fix)
The staging file `narrator_mentions_sanadset.parquet` already carries **105,124** names with a literal, usually *unclosed/nested* `<NAR>` marker, e.g. `"<NAR> أَبُو عُبَيْدَةَ"`, `"<NAR> <NAR> عبد الرحمن"`, `"اسماعيل <IDF> يعني</IDF> ابن عليه"` (16,518 canonical names contain `<`; 4,856 contain `يعني`).
- **Why the existing `_is_narrator_like` filter misses it:** `is_arabic()` returns True for *mixed* Latin+Arabic (`is_arabic("abc محمد") == True`), so a name with an embedded `<NAR>` tag passes the Arabic-script gate and the tag is never stripped.
- **Fix:** strip residual `<…>` markup (`<NAR>`/`</NAR>`/`<IDF>`/`</IDF>`) from tag content in `src/parse/sanadset.py` before validation, AND harden `_is_narrator_like` to reject any residual angle-bracket markup. Recovers the real name (`أَبُو عُبَيْدَةَ`) rather than dropping it.

### 2. thaqalayn parser dumps whole hadith bodies into the name field (HIGH impact)
**48,710** thaqalayn canonical "names" are entire multi-hundred-word hadith texts (Arabic + English translation paragraphs). This is the known broken thaqalayn parser (memory `project_thaqalayn_parser_schema`: wrong field mapping). Needs the thaqalayn parser field-mapping fix; a token-count cap on emitted narrator names is a cheap backstop.

### 3. Collective / anonymous (mubham) descriptors minted as named narrators
`"جماعه من اصحاب"` (a group of companions), `"رجل من اصحاب النبي"` (a man from the Prophet's companions), `"رجال من قومه"`, `"من لا اتهم"`, honorific phrases `"صلى الله عليه وسلم"`. These are real isnad phenomena (unnamed narrators) but should not be distinct `Narrator` nodes. Needs `_NON_NARRATOR_PHRASES` / pattern expansion (precision-sensitive — must not drop real names).

## Acceptance
- `<NAR>`/markup-bearing canonical names → 0
- thaqalayn whole-text names eliminated
- collective/anonymous descriptors not minted as named narrators
- no regression in legitimate narrator count (precision guard: review a sample before/after)

parametrization:
## What

Integration branch resolving the #723 narrator/relational data-quality defects in the
resolve pipeline — the fixes already validated on staging (1.66M nodes, 100%-populated
chains, ~69% linkage, **zero name pollution**) and now reproduced + unit-covered here.

Bundles three sub-branches (merged) plus the integration-level fixes:

- **NER name-quality filter at source** (da#247) — strip NAR-markup, drop mubham/
  whole-text pollution and English non-name prose fragments minted as narrators
  (`أبيه` "his father" was the #1 "narrator" pre-fix).
- **Crash-resumable, memory-bounded dedup** (da#245) — fixes the OOM / non-resumable
  embedding stage at the resolve crash site.
- **NER pass survives un-split / trailing-`قال` isnad spans** (da#244) — skip the
  unsegmentable per-hadith span instead of aborting the whole pass.
- **Vectorized cdist scoring + common-name block cap** in `fuzzy_cluster` (kills the
  silent O(n²)/OOM pass), **chains built from resolved mentions** (the #723 hollow-chain
  root cause), and a **relational-pollution scrub** equivalent to NER-with-filter.

Also: prod graph-reload runbook (`docs/PROD_RELOAD_RUNBOOK_723.md`), and a `.gitignore`
hardening so the whole `data/` tree (incl. the ~1.5 GB curated snapshots) and
`.claude/worktrees/` are ignored.

## Why

#723: the prod isnad graph was polluted (≈40% of canonical narrators had markup/
relational-pronoun names) and its chains read hollow. These fixes correct the resolve
stage so the corrected graph is reproducible from curated artifacts.

## Test plan

- `uv run pytest tests/ -m "not integration"` — green (full pre-push gate passed:
  ruff, mypy-strict, pytest).
- New regression tests: `test_name_quality`, `test_dedup_resume`, `test_ner`,
  `test_load_nodes` — each fails on pre-fix code.
- Staging validated end-to-end; **stage↔prod graph parity confirmed** (identical node/
  rel/constraint/index counts; `أبيه` absent from both top-narrator lists).

## TechDebt

None introduced. (`fuzzy_cluster` common-name block cap is a tunable heuristic, tracked
in the resolve module docstring; not new debt.)

Closes #244
Closes #245
Closes #247
Part of noorinalabs/noorinalabs-main#723
Part of noorinalabs/noorinalabs-main#903


parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Ivana Horvat
Verdict: Approved
TechDebt: Elided mid-chain relational narrator (ابيه) lets _build_chain_pairs bridge a direct TRANSMITTED_TO across the dropped link — track modeling unnamed intermediates as anonymous nodes (P7 follow-up).

## Review — QA / data-quality lens

Reviewed every changed source file and test at HEAD `bc51f206` via the contents API. All 12 CI checks green, Test and Integration Tests included.

### What holds up
- The name-quality filter drops the relational-pronoun pollution (ابيه / جده / عنه plus the trailing-punctuation variants that shielded the bare token) alongside the markup, mubham-collective, and English-prose classes — while the precision guards keep real names. Multi-token kunya (ابي اسحاق), partitive-من lineages, and romanized al-/an- article forms all survive. Because the relational drop fires only on a single-token whole name, and fuzzy_cluster needs two shared significant tokens, these spans can never reach disambiguate or clustering.
- Regression tests are genuine, not tautological. The name-quality, dedup-resume, and ner suites exercise behaviour that does not exist on pre-fix code; the chains-from-resolved test pins the #723 hollow-chain defect directly (narrator_ids populated and position-ordered, read from curated rather than staging).
- Dedup resume is correct and idempotent. The sidecar meta (count, dim, model, ids_hash) refuses to reuse a memmap from a foreign corpus, the progress marker resumes from the right offset, and the FakeModel encode-count proves no re-encode on resume. The disk-backed memmap is the right memory-bound fix for the OOM.
- The fuzzy_cluster common-name block cap and the single-threaded-cdist-on-one-shared-pool rework are a sound runtime fix; the cap is a documented recall/runtime tradeoff in the conservative (keep-apart) direction.

### One data-quality nuance to track, non-blocking
- When a relational-pronoun mention is dropped from the middle of a chain, load_edges._build_chain_pairs pairs the two surrounding narrators consecutively, minting a TRANSMITTED_TO edge that bridges the real but unnamed intermediate. This is strictly better than the 65k-mention mega-node it replaces, but it does fabricate a localized direct link. The clean remedy is modeling unnamed intermediate narrators as anonymous nodes instead of eliding them — captured in the TechDebt line above.
- Minor and cosmetic only — Chain.narrator_ids keeps adjacent duplicates (chain_length counts them) while TRANSMITTED_TO drops self-loops. Not a correctness bug.

Approving. The change resolves the #723 pollution and the NER/dedup OOM at source, the tests are real regression guards, and CI is fully green.

parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Request
Verdict: Changes Requested
TechDebt: None new introduced. The fuzzy_cluster common-name block cap (250) stays a tunable recall/runtime knob already tracked in the resolve-module docstring (da#118 harness), and the da#138 token-order false-merge harness remains the standing cluster-precision debt — neither is added by this PR.

## Review (data-engineering lens — chains-from-resolved, dedup resume, cluster perf, graph integrity)

This is strong, well-reasoned work and the bulk of it is approve-quality. I found one chain-integrity gap that sits squarely in the #723 subject matter and is a one-line fix, so I am asking for that change before merge.

### Must-fix

- The new Chain node loader (`_load_chains` in `src/graph/load_nodes.py`) folds the curated muhaddithat orphan-link rows into `Chain.narrator_ids`. The file glob there is the prefix `narrator_mentions_resolved` over `curated_dir`, so it also reads `narrator_mentions_resolved_muhaddithat.parquet` (da#228). Those orphan-link rows carry `position_in_chain=0`, a `provenance` value, a real `canonical_narrator_id`, and a real `hadith_id` (e.g. `sunnah:bukhari:2`). The loop appends every `(pos, nid)` with no `provenance` filter, so for each of the ~8 hard-coded muhaddithat links the orphan narrator is injected into that hadith's chain sequence and its `chain_length` is inflated by one — a non-isnad narrator inside a chain, which is exactly the kind of chain pollution #723 set out to remove. The sibling edge loader already guards this precise collision in `_load_transmitted_to` (`if row.get("provenance"): continue`, with a long comment explaining the same-id-grammar collision); the chain loader needs the identical guard. Suggested fix is to skip `provenance`-bearing rows in that `_load_chains` loop, mirroring the TRANSMITTED_TO path, and add a small regression test feeding a resolved-mentions set plus one orphan-link row for the same hadith and asserting the orphan is absent from `narrator_ids`. NARRATED and TRANSMITTED_TO endpoints are unaffected (those paths are correctly guarded), so this is a denormalized-property correctness gap, not a graph-edge break — narrow blast radius, but it does contradict the chains-from-resolved intent.

### Verified and solid (no action)

- Chains-from-resolved root-cause fix is correct. Reading the canonical-id-bearing `narrator_mentions_resolved*` from `curated_dir` (not the raw staging mentions, which lack `canonical_narrator_id`) is the right cure for the hollow `chain_length==0` defect, and `is_complete = len(narrator_ids) > 0` plus position-ordered `narrator_ids` matches the intent. The regression test asserts the populated, position-ordered chain and would fail on the pre-fix staging-read path.
- Dedup crash-resume state machine is correct and idempotent. The memmap + sidecar-meta + progress-marker design resumes from the last committed offset, refuses reuse on a corpus/model meta mismatch, and re-encoding a partially written batch overwrites deterministically. The progress marker is written after the chunk is assigned, so an OOM-killed process (the actual target failure, OS page cache intact) resumes safely. Tests cover full-encode, no-reencode resume, mid-offset resume, meta-mismatch full re-encode, and batch-size invariance — good coverage that fails on the pre-resume code.
- Fuzzy-cluster performance rework is sound. Composite-token-pair blocking makes the `>=2` shared-token guard hold by construction, so skipping that re-check inside the cdist/scalar block paths is safe; the per-block single-threaded `cdist` on one shared pool, completion-order drain, no `seen` set, and bounded in-flight futures together address both the O(n^2)/OOM and the thread-pool pathology described in the docstring. float64 keeps the scalar/cdist threshold cut identical. The block cap at 250 is a documented recall/runtime tradeoff, not a correctness issue.
- The identity.py double-corpus-collapse log sampling is a correct hot-path fix for the main#139 condition (8.45M warnings at the stage reload), and the per-edge SET-after-MERGE coalesce-preserve contracts in load_edges remain intact.

CI is fully green (13/13 checks, including Test, Integration Tests, Type Check, Lint, and the sync-drift gate) at the reviewed head.

Once the `_load_chains` provenance guard lands (plus its regression test), this is an approve from me.

parametrization:
@reviewer — the `_load_chains` provenance guard + regression test landed in 7bb5a41 (Kavitha's must-fix). The orphan-link rows (provenance-bearing) are now skipped in chain assembly, mirroring `_load_transmitted_to`; `test_chains_exclude_provenance_orphan_links` asserts the orphan narrator is absent from `narrator_ids` and fails on pre-fix code. Ready for re-review.

parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved
Verdict: Approved
TechDebt: None new introduced. The fuzzy_cluster common-name block cap (250) stays a tunable recall/runtime knob already tracked in the resolve-module docstring (da#118 harness), and the da#138 token-order false-merge harness remains the standing cluster-precision debt — neither is added by this PR.

## Re-review at HEAD 7bb5a41 — must-fix resolved, flipping to approve

Re-reviewed the new commit at head 7bb5a41 (read via the contents API, no checkout). My earlier chain-integrity must-fix is fully addressed:

- `_load_chains` in src/graph/load_nodes.py now skips `provenance`-bearing rows (`if row.get("provenance"): continue`) before folding mentions into `narrator_ids`, mirroring the `_load_transmitted_to` guard in load_edges.py, with a comment that explains the orphan-link glob collision. An ordinary chain mention has no `provenance` column (row.get returns None), so only the curated muhaddithat orphan-links are dropped — the right scope.
- The regression test `TestLoadChains::test_chains_exclude_provenance_orphan_links` is non-vacuous and exercises the REAL producer: it calls `build_muhaddithat_mention_links(curated_dir)` with the default `canonical_path=None`, so the existence guard is skipped and the actual `sunnah:bukhari:2` orphan-link row is emitted alongside a two-mention isnad chain for the same hadith. It asserts the orphan narrator is absent and `narrator_ids == ["nar:real-0", "nar:real-1"]` with `chain_length == 2` — pre-fix code would fold the orphan in and report length 3, so the test genuinely fails on the old path and passes on the new one. Using the real producer rather than a hand-rolled stub is the stronger choice.

All earlier approve-quality findings still hold (chains-from-resolved root-cause fix, dedup crash-resume state machine, fuzzy_cluster cdist/blocking rework, identity.py log sampling). CI is green at 7bb5a41 (Test, Integration Tests, Type Check, Lint, Spellcheck, actionlint, and the Pre-commit/CI sync-drift gate all SUCCESS).

No remaining concerns from my lens. Approved.

Issue #239, opened 2026-06-26
tech-debt: default precision columns to UNKNOWN in bio_promote/fuzzy_cluster canonical builders


parametrization:
## Problem
The `NARRATORS_CANONICAL_SCHEMA` date/precision extension (da#162, P7W21) added two nullable precision columns (`birth_date_precision`, `death_date_precision`) carrying the `DatePrecision` StrEnum value, defaulting to `"unknown"` (`DatePrecision.UNKNOWN.value`).

The `disambiguate._build_canonical_table` producer correctly defaults unset precision to `"unknown"`. But the other two schema-driven producers (`bio_promote._build_table`, `fuzzy_cluster._build_table`) project rows via `{f.name: r.get(f.name) for f in SCHEMA}` — their rows carry no precision key, so bio-only / fuzzy-clustered narrators get **null** precision instead of `"unknown"`.

This diverges from the da#161 `Narrator` model's non-Optional `UNKNOWN`-default invariant. The schema permits null so nothing breaks today, but a `DatePrecision(None)` could reach model reconstruction in a downstream consumer (da#164/#165/#166, ig#1039).

## Fix
Symmetric one-line default: emit `DatePrecision.UNKNOWN.value` for the two precision columns in `bio_promote._build_table` (and `fuzzy_cluster._build_table` if it materializes precision), matching `disambiguate`.

## Provenance
Flagged independently by both reviewers (Ivana Horvat, Alejandra Reyes-Fuentes) on PR #237 (da#162); recommended as a follow-up rather than expanding that PR. Consumers in da#164/#165/#166 + ig#1039 must agree that null precision reads as `UNKNOWN`.

PR #255, opened 2026-07-01
fix(canonical): default precision columns to UNKNOWN in bio_promote/fuzzy_cluster (da#239)

parametrization:
## Summary

Defaults the `birth_date_precision` / `death_date_precision` columns to `DatePrecision.UNKNOWN.value` (`"unknown"`) in the two schema-driven canonical builders that previously left them null.

`disambiguate._build_canonical_table` already defaults unset precision to `"unknown"`, but `bio_promote._build_table` and `fuzzy_cluster._build_table` project rows generically (`{f.name: r.get(f.name) for f in NARRATORS_CANONICAL_SCHEMA}`), so bio-only and fuzzy-clustered narrators — which carry no precision key — shipped **null** precision. That diverged from the da#161 `Narrator` non-Optional `UNKNOWN`-default invariant and risked a `DatePrecision(None)` reaching model reconstruction in downstream consumers (da#164/#165/#166, ig#1039).

## Change

- `bio_promote._build_table`: after the generic projection, override the two precision columns with `r.get(col) or DatePrecision.UNKNOWN.value`.
- `fuzzy_cluster._build_table`: same symmetric override.
- Added `DatePrecision` import to both modules. No other columns changed; matches the existing `disambiguate` idiom and vocabulary exactly (no new enum value invented).
- Regenerated the committed structural ontology index (new `src.models.enums` import edges + line numbers).

## Tests

- `test_bio_promote.py::test_unset_precision_defaults_to_unknown` — promotes a bio with no precision set, asserts both precision columns read `"unknown"` (not null).
- `test_fuzzy_cluster.py::test_unset_precision_defaults_to_unknown` — clusters records with no precision set, asserts both columns read `"unknown"` on the output.
- Both tests **FAIL pre-fix** (`assert None == 'unknown'`) and pass post-fix. Full suite: 1322 passed, 7 skipped. ruff + ruff-format + mypy-strict clean.

## Provenance

Flagged independently by both reviewers (Ivana Horvat, Alejandra Reyes-Fuentes) on PR #237 (da#162), recommended as a follow-up.

## Reviewers

- **Dilara Erdogan** (peer)
- **Ivana Horvat** (secondary)

Closes #239


parametrization:
**Requestor:** Dilara Erdogan
**Requestee:** Kavitha Sundaramurthy

**RequestOrReplied:** Approved

Reviewed at HEAD `6507276` (artifact via `gh api contents`, no checkout).

The fix is correct and minimal. Both `bio_promote._build_table` and `fuzzy_cluster._build_table` override the two precision columns after the generic schema projection with `r.get(col) or DatePrecision.UNKNOWN.value` — byte-identical to the established `disambiguate._build_canonical_table` idiom (`disambiguate.py:743/753`), so no new or variant sentinel is invented: `DatePrecision.UNKNOWN.value == "unknown"`, the same value the da#161 `Narrator` non-Optional invariant uses.

Key risk (unconditional-set clobbering a real precision) is **not present**. The `... or ...` form is default-if-falsy, and every valid `DatePrecision` member is a non-empty truthy string (`exact`, `range`, `circa`, `after`, `before`, `tabaqa_estimate`, `unknown`). A narrator with an attested precision therefore round-trips unchanged; only genuinely-absent (None) or empty-string precision — neither of which is a valid domain value — is coerced to `"unknown"`. The override reads from the source row and is scoped to exactly the two precision columns; no other column is touched. Symmetric across both builders.

Both new tests genuinely fail pre-fix: they assert the columns equal `"unknown"`, whereas the pre-fix generic projection leaves them None (`None == "unknown"`). `test_fuzzy_cluster` correctly exercises the no-merge path (two distinct narrators) so the default is verified on plain projection, not just a merge. CI green across all 7 checks (Lint, Type Check, Test, Integration, base-image, sync-drift, staleness).

Non-blocking observation (not a defect): neither new test adds a positive-preservation case (a row that DOES carry an attested precision, asserting it survives). The `or` idiom makes that behaviour self-evident and it mirrors the already-covered `disambiguate` path, so I do not require it — noting only for future symmetry.

TechDebt: none

parametrization:
**Requestor:** Ivana Horvat
**Requestee:** Kavitha Sundaramurthy

**RequestOrReplied:** Approved

Secondary review — schema/downstream-consistency + sibling-builder-coverage lens (distinct from the peer's clobber/enum-value check).

**Sibling-builder coverage is complete.** Four canonical builders share the generic `{f.name: r.get(f.name) for f in NARRATORS_CANONICAL_SCHEMA}` projection. Two are fixed here (`bio_promote._build`, `fuzzy_cluster._build_table`); `disambiguate._build_canonical_table` is the reference idiom. The remaining two emit these columns but were correctly left untouched because neither carries the raw NULL gap: `date_reconcile.reconcile_canonical_dates` already backfills concrete precision on every row inline (`date_reconcile.py:510-511`) before its projection, and `tabaqa_dates.apply_tabaqa_fallback` runs after reconciliation and only writes `tabaqa_estimate` for undated-death rows, passing already-concrete precision through. I grepped every `date_precision` emitter and `NARRATORS_CANONICAL_SCHEMA` consumer in `src/` — no other builder shares the gap. So the two-builder scope of this PR is exactly the unaddressed set.

**Schema + downstream consumer consistency is correct.** The columns are `pa.string(), nullable=True` in `resolve/schemas.py`, so writing the plain string `"unknown"` (= `DatePrecision.UNKNOWN.value`) is dtype-consistent and invents no new vocabulary. "unknown" — not NULL — is what consumers expect: the da#161 `Narrator` model (`models/narrator.py:54,60`) is non-Optional `DatePrecision`, so a null column reconstructs as `DatePrecision(None)` and raises at model build in the graph-load/enrich path (ig#1039). "unknown" everywhere is genuinely safer than NULL here. The `or`-coalescing matches disambiguate exactly and only rewrites falsy values; real precision strings ("exact", "circa", "tabaqa_estimate") are never falsy, so no populated precision is lost downstream. Regression tests assert both the `== "unknown"` and `is not None` legs. CI fully green (Lint, Type Check, Test, Integration, sync-drift, staleness).

TechDebt: none

Issue #246, opened 2026-06-28
enhancement: parallelize dedup embedding encode (~768k texts, serial single-process long pole)


parametrization:
## Problem

`dedup.run_dedup` (`src/resolve/dedup.py`) encodes ~768k hadith matn texts through a multilingual sentence-transformer on CPU, single-process, in a serial `for start in range(0, len(texts), batch_size)` loop. This is the dominant wall-clock cost of the whole resolve pass (multiple hours) and is embarrassingly parallel.

## Asks (parallelize / decompose)
1. **Chunk + fan out the encode** across processes (multiprocessing pool over text chunks) or a GPU batch path when available — collect chunk embeddings into the final matrix in order.
2. **Decouple encode from search** — the FAISS build/search is a separate, vectorized phase; chunked encode output should feed it without holding everything in RAM (coordinate with the memory-bounding work in the sibling hardening issue).
3. Keep output **deterministic** (stable ordering of `parallel_links.parquet`) regardless of chunk completion order.

## Context
Sibling to the `fuzzy_cluster` composite-token-blocking fix (branch `I.Horvat/0000-fuzzy-cluster-block-opt`, 10–50× on the cluster stage). The cluster fix made *that* stage near-instant; `dedup` encode is now the remaining long pole for the prod data-reload (main#723) resolve runs.

PR #256, opened 2026-07-01
perf(dedup): parallelize embedding encode (da#246)

parametrization:
## Summary

Parallelizes the dedup embedding encode (da#246) — the single-process, serial
`for start in range(0, len(texts), batch_size)` loop in `_encode_with_resume`
that ran ~768k hadith matn texts through the multilingual sentence-transformer
one batch at a time. This was the dominant wall-clock cost of the resolve pass
(multiple hours) and the remaining long pole for the prod data-reload (main#723)
now that the `fuzzy_cluster` blocking fix made clustering near-instant.

## Approach

- **Fan-out.** The remaining rows are split into `workers × 4` chunks; each
  worker process builds its own model (via a `model_provider` — sentence-
  transformers are not fork/pickle-safe to share) and encodes a chunk, and the
  parent writes each returned block to its **fixed** memmap row-range
  `emb[cs:ce]`.
- **Determinism.** Placement is by row index, not completion order, so the
  parallel output is **byte-identical** to the serial output. Dedup thresholds,
  FAISS build/search, and pair ordering (`parallel_links.parquet`) are
  untouched — the change is scoped strictly to the encode step.
- **Configurable + oversubscription guard.** Worker count comes from
  `DEDUP_ENCODE_WORKERS` (or the `encode_workers` arg): `0`/unset = auto (scale
  to the box, capped at 8); any value is clamped to the physical core count;
  never `< 1`. Per-worker BLAS/torch threads are pinned to 1 so `workers × cores`
  threads don't thrash the box.
- **Serial fallback** preserved: `workers <= 1`, no `model_provider`, or too few
  remaining rows to amortize process overhead all take the original in-process
  path.
- **da#245 resume preserved.** The progress marker advances over the contiguous
  completed prefix; a crash mid-fan-out resumes correctly and any non-contiguous
  written block is idempotently re-encoded. Peak RAM stays ~one chunk per
  in-flight worker.
- Default start method is `spawn` so a torch-initialized parent doesn't fork
  into a deadlock.

## Tests

New `tests/test_resolve/test_dedup_parallel.py` (deterministic fake model, no ML
deps, `fork` for speed):
- parallel output **exactly equals** serial output + parent model unused (proves
  the workers ran) + stable placement;
- ordering stable across multiple out-of-order chunk completions;
- parallel path honours da#245 resume (skips completed prefix);
- serial fallbacks (`workers=1`, small corpus, no provider);
- worker-count resolution: auto-cap, explicit request clamped to cores, never
  exceeds item count, single-core → serial.

`ruff` + `mypy --strict` + full `pytest` green locally and via pre-push hooks;
structural ontology index regenerated for the `src/` change.

## Reviewers
- **Alejandra Reyes-Fuentes** (peer)
- **Jean-Claude Habimana** (secondary)

Closes #246


parametrization:
**Requestor:** Alejandra Reyes-Fuentes
**Requestee:** Nikolaos Papadopoulos

**RequestOrReplied:** Approved

Reviewed at HEAD `3efae7598c306ab62e5e97e5fc5728a013d29746` via `gh api contents` (no checkout). Focus axis: parallel-vs-serial determinism / row-alignment.

**Determinism holds — and the design is safer than "each worker writes its own range" implies.** Workers only *encode and return* their block; the parent process serializes every `emb[cs:ce] = block` write in the `as_completed` loop. There is therefore no concurrent memmap write at all — the write-race hazard is eliminated by construction, not merely guarded. Placement is by fixed row index `cs`, so out-of-order worker completion cannot scramble ordering. Chunk bounds (`s=start; while s<n: e=min(s+chunk_size,n); s=e`) partition `[start,n)` disjointly, contiguously, no gaps/overlap. FAISS build/search, `_classify_pair` thresholds, and the canonical `hid_a >= hid_b` pair-ordering are untouched — the change is scoped strictly to the encode step.

**Resume (da#245) survives chunking.** The marker advances only over the contiguous completed prefix (`while contiguous in completed: contiguous = completed[contiguous]`), and `emb.flush()` precedes `prog_path.write_text` so a mid-fanout crash leaves the marker lagging (never ahead of) durable data; non-contiguous written blocks are idempotently re-encoded on restart. The resume test drives the true parallel path (remaining=14 >= 2*batch_size=8) and confirms preserved-prefix + correctly-placed remainder.

**Oversubscription guard is correct.** `_resolve_encode_workers` caps auto at `_AUTO_WORKER_CAP`, clamps explicit requests to `os.cpu_count()`, never exceeds item count, never returns < 1; per-worker BLAS/torch threads pinned to 1 via `initargs=(..., 1)`; `spawn` is the default start method (torch fork-safety), tests override to fork for speed. Serial fallback preserved for `workers<=1`, missing provider, and sub-threshold corpora — each asserted via the parent-model encode counter.

**Tests genuinely exercise the row-misalignment failure mode.** `_texts(i)="x"*(i+1)` + `FakeModel` encoding to `[len(text)]*dim` means `emb[i,0]` must equal `i+1`; a block landing in the wrong range would fail both `np.array_equal(serial,parallel)` and the per-row placement assertion. `parent_model.encoded == 0` proves the workers, not the parent, ran. CI 7/7 green (Lint, Type Check, Test, Integration, base-image, sync-drift gate, structural staleness).

Two minor, non-blocking observations:
1. Byte-identity of the float *values* under the real SentenceTransformer rests on `model.encode` being batch-invariant — the parallel per-chunk sub-batch boundaries do not align with the serial boundaries. This is the exact tolerance the serial path already accepts (its own batch grouping shifts with the resume point), so it is consistent and acceptable; the fake-model tests prove placement/ordering, not real-model batch-invariance. Worth an eyeball on the first real prod run's cluster counts vs the last serial run.
2. `test_parallel_resume_skips_completed_prefix` proves the final state is correct but does not strictly prove the prefix was *not* re-encoded (the prefill values coincide with what a re-encode would produce). Minor test-strength nit; the code path (start read from progress marker) is correct.

Neither warrants changes. Nice, tight, correctness-preserving parallelization.

TechDebt: none

parametrization:
**Requestor:** Jean-Claude Habimana
**Requestee:** Nikolaos Papadopoulos

**RequestOrReplied:** Approved

Secondary review, resource-safety + operability lens (peer covers byte-equality/determinism). This runs in the high-stakes wave-end prod reload (main#723), so I focused on the OOM / worker-death / silent-partial failure modes.

**Worker death — no silent zero-block (the key concern).** In `_encode_range_parallel`, `fut.result()` re-raises any worker exception, and a killed worker (OOM-killer SIGKILL) surfaces as `BrokenProcessPool`. Either way the exception propagates out of the `as_completed` loop, cleanly shuts down the pool via the `with` block, and aborts `run_dedup` with a diagnosable traceback. The parent never ships a zero-filled block for a dead worker. On restart the resume marker holds only the *contiguous* completed prefix, so a crash mid-fan-out resumes correctly and non-contiguous written blocks are idempotently re-encoded (deterministic values). This is the correct, loud-fail behavior. Verified.

**memmap flush/sync ordering is crash-safe.** `emb.flush()` is called *before* `prog_path.write_text(contiguous)` on every chunk completion, so the progress marker can never claim rows that aren't yet msync'd — no torn-write hazard where a resume trusts data that never hit disk. The marker is conservative (contiguous prefix only) while flush() durably syncs the whole mapping. Correct.

**Memory is bounded and documented.** Auto worker count is capped at 8 (`_AUTO_WORKER_CAP`), per-worker BLAS/torch threads pinned to 1, peak RAM ~one chunk per in-flight worker, and RSS watermark logged per chunk so an approaching OOM shows in the logs rather than as a silent kill. Operator override via `DEDUP_ENCODE_WORKERS`. Config surface is clean: `dedup_encode_workers: int = 0` on root `Settings` (reads the env var, default 0=auto), documented in both `config.py` and the `run_dedup` docstring, clamped to physical cores, never <1, never > item count. da#245 resume interaction is preserved and unit-covered.

Non-blocking caveats (folded into TechDebt, none rise to a change request):
1. The highest-stakes operational path — a worker raising / `BrokenProcessPool` propagating and NOT silently zero-filling — has no regression test. The behavior is correct by construction (`fut.result()` re-raise is standard), but for the wave-end reload a test asserting loud-fail-with-resumable-marker would lock it in.
2. Auto-scale is `min(cpu, 8)` — CPU-based, not RAM-aware. On a high-core / low-RAM box, auto=8 × ~0.5 GB models (plus the parent's model, which stays resident but unused during fan-out ≈ 9 models total) could over-commit RAM. It is documented and operator-overridable, so acceptable, but the auto heuristic is a core count, not a memory budget.
3. `emb.flush()` msyncs the full ~1 GB mapping on every one of the ~32 chunk completions; correct but a minor per-flush cost on slow disks.

Default behavior note (not a defect): the default is now auto-parallel, not serial — an operator wanting the old serial path must set `DEDUP_ENCODE_WORKERS=1`. This is the intended perf change and is documented; the determinism + loud-fail + bounded-memory guards justify parallel-by-default. CI is fully green (7/7 checks SUCCESS at head).

TechDebt: add a worker-death/BrokenProcessPool regression test (loud-fail + resumable marker, no silent zero-fill); consider a RAM-aware cap or documented per-worker RAM budget for the auto heuristic — track as a da#246 follow-up.

Issue #258, opened 2026-07-01
NER residual: Arabic matn-prose + compound/chain-connective narrator spans at token cap


parametrization:
## Summary

The da#253 fix (`clean_narrator_name`) fully eliminated the **English** matn-as-narrator pollution class (the owner-reported "Thawban:The Messenger of Allah (ﷺ) sacrificed during a journey…" example — verified 0 survivors on the corrected artifacts + reloaded stg/prod graphs). A **smaller, distinct residual class remains**: Arabic-script matn-prose and compound/chain-connective narrator spans that sit right at the 30-token cap.

## Evidence (corrected narrators_canonical.parquet, 219,849 rows)

A prose-shape heuristic (>=8 tokens, >=3 Arabic stop/function words, not ibn-chain-dominated) flags **22,830 name_ar_normalized values (10.4%)**, but weighted by mention_count the visible footprint is tiny:

- mention_count >= 5 (UI-visible / popular-narrators): **7**
- mention_count 2-4: 224
- mention_count <= 1 (tail noise): 22,599

Three sub-classes are conflated in that bucket:

1. **Genuine Arabic matn body** — e.g. `كان علي بن ابي طالب بالكوفه في الجامع اذ قام اليه رجل من اهل الشام…` ("Ali ibn Abi Talib was in Kufa in the mosque when a man from Sham stood up…"). True pollution.
2. **Compound co-narrator joins** — e.g. `سعد بن عبد الله و عبد الله بن جعفر الحميري جميعا` ("Sa'd ibn Abdullah AND Abdullah ibn Ja'far al-Himyari, together"). Two real narrators joined by `و … جميعا`.
3. **Chain-connective fragments** — e.g. `سالم بن عبد الله ان عبد الله بن عمر…` ("Salim ibn Abdullah THAT Abdullah ibn Umar…"). A real name + isnad connective `ان` + the next narrator.

## Why this was NOT hot-fixed in the reload

A crude post-pass drop would destroy the real names embedded in classes 2 and 3 (losing legitimate narrators + their mention edges). The correct fix is at the NER/resolve stage, mirroring how da#253 handled the English colon-join:

- Split compound co-narrators on `و … جميعا` into separate canonical narrators.
- Truncate chain-connective spans at `ان` / `عن` when the tail begins a new isnad link, keeping the leading name.
[...]

PR #260, opened 2026-07-02
fix(name-quality): Arabic isnad/matn residual — truncate connectives, drop matn, detect compounds (da#258)

parametrization:
Refs #258

Direct Arabic-script sibling of da#253 (English `<name>:<matn>` colon-join fix). Record-level validation of the #723 stage/prod reload found **7,451 narrator nodes still carrying isnad text** in `name_ar` (96.6% the mention≤1 tail). This cleans that residual at the NER/name-quality stage — WITHOUT destroying real names — so the next regenerate→reload drops it.

## The three residual classes

All defeat the 30-token cap because real `ibn`-lineage nasab names legitimately reach ~30 tokens.

| Class | Example | Treatment |
|-------|---------|-----------|
| **4 — genuine matn body** | `كان علي بن ابي طالب بالكوفه في الجامع اذ قام اليه رجل …` | **DROP** (boundary verb leads the span) |
| **5 — chain-connective fragment** | `ابو عبد الرحمن عن ابي هريره` · `سالم بن عبد الله ان عبد الله بن عمر` · `بندار هو محمد بن بشار` | **TRUNCATE** at the connective, keep the real lead |
| **6 — compound co-narrator join** | `سعد بن عبد الله و عبد الله بن جعفر الحميري جميعا` · `يحيى بن ايوب وقتيبه وابن حجر قالوا` | **SPLIT** (detection primitive here; row-emission = follow-up) — never dropped |

## Design

Classes 4+5 unify into a single **truncate-at-first-isnad-boundary** rule in `clean_narrator_name`: cut the span at the first matn-verb / isnad-connective (`عن`/`ان`) / relative-pronoun (`الذي`/`التي`) / editorial-gloss (`هو`/`قيل`) token and keep the preceding real name; drop only when that boundary **leads** the span (the narrator sits after an elided-grammar verb and is unrecoverable). This recovers more real names than a plain drop — `عاءشه قالت كان` → `عاءشه`, `بندار هو محمد بن بشار` → `بندار` (= Ibn Bashshar). A matn-density backstop (≥2 particle words) catches sentences whose first boundary is a preposition the cut can't anchor on. **None of the boundary tokens is ever a name component**, so every real nasab in the da#247 test set is preserved unchanged.

Class 6 gets `split_compound_narrators()` — a detection+split primitive (splits on standalone and proclitic `و`, strips the trailing `جميعا`/`قالوا` marker; the proclitic split fires **only inside a confirmed compound**, so a lone `و`-initial real name — وكيع "Waki'", وهب "Wahb" — is never mutated). `clean_narrator_name` deliberately does **not** drop an un-split compound (no matn token fires on a name-only `و`-join), so no co-narrator is lost while the split wiring is pending.

### Class-6 split scoping decision (in-PR detection, deferred row-emission)

The full 1→N split requires emitting one mention row per member, which touches `src/resolve/ner.py`'s **two** one-span→one-row call sites — the raw/normalized display alignment and the co-narrator chain-position semantics are a distinct **resolve-stage** change that deserves its own review focused on graph-load implications. The genuinely-clean multi-narrator compounds are also small and all mention≤2. So this PR ships the **fully unit-tested split primitive** + **no-regression guarantee** (compounds preserved, not deleted), and defers the ner.py wiring to a scoped follow-up. Filing recommended: *"da#258 follow-up: wire `split_compound_narrators` into `src/resolve/ner.py` to emit one mention row per co-narrator member."*

## Record-level before/after evidence

Over the live curated artifact `data/curated/narrators_canonical.parquet` (**219,849 nodes / 3,108,384 mentions**), running the shipped `clean_narrator_name`, weighted by `mention_count`:

| Effect | Nodes | Mentions |
|--------|------:|---------:|
| class 5 truncated → **real lead** (recovered, not lost) | 33,659 | 31,466 |
| class 1/4 matn **dropped** (genuine pollution) | 3,901 | 4,157 |
| **total names changed** | **37,560** | **35,623** |
| class 6 compounds **preserved** (split = follow-up) | 7,837 | 8,070 |

Top drops are all matn (`ثم`, `كان`, `قالت عاءشه`, `ان رسول الله صلى الله عليه واله`); **no real name is destroyed** (max-mention drop = `ثم`, 84).

## Tests

65 new unit tests (three residual classes + the compound-split primitive) assert: matn dropped, connectives truncated to the lead, compounds preserved/split, long real nasab names kept unchanged (no false positives), and **no regression** on the da#253 English + da#247 mubham/markup classes. Full suite: **1377 passed, 7 skipped**; mypy clean (90 files); ruff clean; structural index regenerated.

## Closure gate

A full **regenerate → stg reload → live record-level verify** is the owner/orchestrator-run closure gate (heavy compute), so this is `Refs #258`, not `Closes` — the PR delivers unit-mechanic + artifact-level correctness.

Intended reviewers: **Jean-Claude Habimana** + **Alejandra Reyes-Fuentes** (persona reviewers — orchestrator drives verdicts).

TechDebt: None

https://claude.ai/code/session_01SvQH5QhX7z6dAVM1rtTAXJ


parametrization:
RequestOrReplied: Approved

**Architect review — class-6 detection-only scoping is sound; no-regression drop-guarantee verified end-to-end**

I traced the full `clean_narrator_name` at HEAD (c8be168) and confirm the central architectural claim: no code path drops a name-only compound while the resolve-stage wiring is pending. The conjunction `و` and the co-narrator markers `جميعا`/`قالوا` are correctly EXCLUDED from `_ISNAD_BOUNDARY`, so step 3c never cuts a `X و Y جميعا` join; steps 4-8 do not fire on Arabic name tokens; and the step-9 density backstop counts zero for a `جميعا`-ending compound (`جميعا` is not in `_MATN_DENSITY`). The one nuance worth naming for the record: a compound ending in `قالوا` is actually preserved by the 3c truncation path (`قالوا` is a boundary token, so it is trimmed as a trailing marker leaving the co-narrator body intact) rather than by the density-1 argument in the step-9 comment — both paths land on "preserved, not dropped", so the invariant holds; the step-9 comment's `قالوا`-density-1 rationale is slightly redundant given 3c already handles it, but that is documentation nuance, not a defect.

On the deferral: shipping `split_compound_narrators` as a fully-tested detection/split primitive while deferring the 1→N row emission is the right call. That emission touches `src/resolve/ner.py`'s one-span→one-row contract (raw/normalized display alignment + co-narrator chain-position semantics) — a genuinely distinct resolve/graph-load concern that warrants its own focused review, not a rider on a parse-stage cleanup. I verified the tracking issue da#261 exists and is OPEN, so the deferred work is not a latent untracked gap; preserving the compound as a recoverable single node is strictly better than dropping (no co-narrator or mention edge is lost), and the follow-up can split it losslessly later.

The truncate-at-first-boundary design (classes 4+5) is ordered correctly — 3c runs before the token cap so a recovered lead is re-measured, not the raw sentence — and the split primitive's public API is clean (exported in `__all__`, returns a list, singleton for non-compounds, `[]` for empty, proclitic-`و` split gated on a confirmed compound so lone `و`-initial names like وكيع/وهب are untouched). All 7 CI checks are green; 65 new tests cover the three classes plus the da#253/da#247 non-regression set.

Must-fix: None
TechDebt: None

parametrization:
RequestOrReplied: Approved

**Data-engineer review — record-level correctness sound; no real name is destroyed; one inert-guard note (non-blocking tech debt)**

I read the code at HEAD (c8be168) and checked the two things that matter most for the live curated artifact — false-positive risk against real nasab names, and internal consistency of the before/after evidence. Both hold.

False-positive safety is the strong part of this PR. The boundary/matn sets are matched **whole-token** (`tok in _ISNAD_BOUNDARY`), which is exactly what saves the flagged homographs — `معن` (Ma'n), `ابان` (Aban), `سنان` (Sinan), `عثمان` all survive because they are single tokens distinct from `عن`/`ان`, and there is no legitimate name whose token is bare `عن`, `ان`, `هو`, `الذي`, or `والله`. Every member of `_ISNAD_BOUNDARY` is a transmission/matn verb, isnad subordinator, relative pronoun, temporal particle, or `هو`/`قيل` gloss — none is ever a name component — so the truncation cannot cut inside a real lineage. The `test_long_real_nasab_preserved` cases (including `عباد بن زياد من ولد المغيره بن شعبه`, which carries mid-lineage `من`) confirm the long-name precision guarantee, and the da#253/da#247 non-regression tests still pass on the English + mubham/markup classes. The compound path is also safe: `و` and `جميعا` are deliberately excluded from the boundary set, so `clean_narrator_name` preserves an un-split `X و Y جميعا` (no boundary token fires, density 0) rather than dropping co-narrators — verified by `test_compound_not_dropped`. Note `split_compound_narrators` is not yet wired into `ner.py`, so its own split behavior is test-only and not on the live data path this wave — that further shrinks the false-positive surface.

The before/after numbers are internally consistent when weighted by mention_count: 33,659 + 3,901 = 37,560 nodes and 31,466 + 4,157 = 35,623 mentions both reconcile to the stated totals, class-5 truncations skew node-heavy (<1 mention/node, the mention≤1 tail) as claimed, and the max-mention drop being `ثم` (84) is consistent with "no real name destroyed." I could not re-run the 219,849-node parquet here (that is the owner-run closure gate, correctly `Refs #258` not `Closes`), but the shipped logic plausibly produces these effects and the drops are carried by the leading-boundary rule (class-4 matn bodies start with `كان`/`قالت` → dropped) plus the existing token-cap/prose guards.

One honest observation that I want on record but that does NOT block: the step-9 `_MATN_DENSITY` backstop is effectively dead code. `_MATN_DENSITY` is a subset of `_ISNAD_BOUNDARY`, and step-3c already truncates at (or before) the first boundary token, so by the time step 9 runs the surviving tokens contain zero boundary tokens and therefore zero density tokens — the `sum(...) >= 2` test is always `0 >= 2`. It can never fire. This is harmless to record correctness because it only ever fails to drop (never wrongly drops a real name), and the measured 3,901 drops already reflect the shipped function's real behavior, so the evidence stands. But the docstring/comment present it as an active guard for "a bare-preposition run the truncation cannot anchor on," which the current subset relationship makes impossible. Worth a follow-up to either populate `_MATN_DENSITY` with genuine non-boundary particles (so it actually catches the residual it describes) or drop the backstop and its prose.

Must-fix: None
TechDebt: step-9 _MATN_DENSITY backstop is inert (subset of _ISNAD_BOUNDARY, always evaluates 0>=2) — populate with non-boundary particles or remove; harmless (false-negative-only), so not a merge blocker.

Issue #259, opened 2026-07-01
Loader chain_integrity validation hangs (>40min) on full graph — forces --skip-validation


parametrization:
## Symptom

The graph loader's built-in `chain_integrity` validation query (`src/graph/validate.py`, run at the end of `isnad load`) is **pathologically slow on a full-size graph** — it ran >40 min without completing on the #723 prod reload (stopped manually), and to load both stg and prod cleanly during the wave-22 window the loader had to be invoked with `load --skip-validation`.

## Impact

- Full production loads currently **cannot run their own post-load integrity validation** — they complete the data load (nodes + edges, 0 errors) but the validation step never returns, so it is routinely skipped. Loads are validated *out-of-band* via direct cypher instead, which is manual and not part of the loader contract.
- A skipped validation means a genuinely-broken load (hollow chains, dangling endpoints) could ship unflagged if the operator forgets the out-of-band check.

## Likely root cause

The chain-integrity / cycle-detection query does variable-length traversal that degrades super-linearly in the presence of transmission cycles (see da#248 — TRANSMITTED_TO cycles from single-token narrators). With ~576k chains and ~2.5M TRANSMITTED_TO edges the traversal blows up. Two orthogonal fixes:

1. **Data:** remove the spurious transmission cycles (da#248) so the traversal terminates quickly.
2. **Loader robustness (this issue):** the validation query needs a bounded/timeout-guarded form (e.g. `apoc.cypher.runTimeboxed`, a hop cap, or a batched per-source check) so a slow/cyclic graph downgrades validation to a warning instead of hanging the loader indefinitely. `--skip-validation` should be an explicit opt-out, not the de-facto default because the default hangs.

## Acceptance

- `isnad load` (full graph, no `--skip-validation`) completes its integrity validation within a bounded time (minutes, not >40) OR emits a clear timeout warning and exits non-fatally.
- The out-of-band cypher validation used in the wave-22 window is folded into (or replaced by) the loader's own validation.

## Context

Surfaced during the wave-22 / main#723 corrected-corpus reload to stg + prod (2026-07-01); both loads used `--skip-validation` to avoid the hang. Related: da#248 (the cycle data that likely triggers the blow-up).

PR #263, opened 2026-07-02
graph: bound post-load validation so isnad load never hangs (da#259)

parametrization:
TechDebt: da#249 (validation harness runs multi-statement .cypher as one query) and da#248 (super-linear cycle traversal root cause) remain out of scope and are tracked separately; this PR is the loader-robustness half.

Refs #259

## Problem

On the full graph (~576k chains, ~2.5M TRANSMITTED_TO edges) the `chain_integrity` variable-length cycle traversal (`(:Narrator)-[:TRANSMITTED_TO*1..20]->(:Narrator) WHERE n=m`) degrades super-linearly and never returns (>40 min). Both stg and prod loads (main#723) therefore ran with `--skip-validation`, which turned an explicit safety opt-out into the de-facto default. The load itself succeeds; only the post-load validation hangs.

## Bounding mechanism chosen — and why

Defence-in-depth, because "never hang the loader" has to hold even if one layer misbehaves:

1. **Server-side statement timeout** — the per-query budget is handed to Neo4j via the driver's `unit_of_work(timeout=...)`, so the database itself aborts the runaway transaction and releases its resources. This is the idiomatic Neo4j bound and is preferred to fire first.
2. **Client-side wall-clock backstop** — the query runs on a daemon worker thread joined for the budget. If it is still alive we abandon it (the daemon dies with the process; the server-side timeout reclaims the query server-side) and downgrade. This guarantees the loader process moves on even if a server ignores the statement timeout.

Both paths converge on the same outcome: the check is recorded as a non-fatal `WARN`, never a hang and never a hard fail.

Why not the alternatives: a **hop cap** on the variable-length pattern changes detection semantics (silently misses longer cycles) and is really a fix for the da#248 root cause, not loader robustness; a **per-source batched check** is a larger rewrite that still needs a bound per batch. A timeout keeps the existing query semantics intact while making the loader safe.

## Behaviour changes

- `run_validation(..., timeout_seconds=300.0)` — per-query budget; overrun downgrades to `WARN`.
- `ValidationResult` gains a `warning` flag plus `status` (PASS/FAIL/WARN) and `is_fatal` properties. Only `is_fatal` (a real failure, not a downgrade) trips `validation_passed`, so warnings never fail the load.
- `--skip-validation` is now documented as a genuine explicit opt-out, not a hang workaround. New `--validation-timeout` flag tunes the budget.
- `Neo4jClient.execute_read` accepts an optional server-side `timeout`.

## Acceptance mapping

- **(1)** A full-graph load without `--skip-validation` no longer hangs — a slow/cyclic query emits a clear timeout warning and the load exits non-fatally.
- **(2)** `--skip-validation` is an explicit opt-out with clear help text and semantics.
- **(3)** The chain_integrity check that had to be run out-of-band in the wave-22 window is now safe to run inline in the loader's own path (that hang was the reason it was pushed out of band). The multi-statement `.cypher` handling (da#249) is orthogonal and stays out of scope.

## Tests

`tests/test_graph/test_validate.py` (all green locally, full suite 1385 passed / 7 skipped):
- fast graph validates normally and the budget is handed through as a server-side timeout;
- a wall-clock-slow path downgrades to a non-fatal warning and `run_validation` returns quickly (asserts no hang);
- a server-side `TransactionTimedOut` path downgrades to a non-fatal warning;
- a genuine (non-timeout) query error stays a hard failure;
- a warning result keeps `validation_passed` true (the aggregation `load_all` uses).

The closure gate is a real owner-run full-graph load, so this is `Refs #259`, not `Closes`.


parametrization:
RequestOrReplied: Approved

**Data-engineer review — bounded validation delivers "never hang the loader"; two layers, aggregation, and timeout classification all check out**

Traced the mechanism at HEAD 0dbf02d.

Two layers. The server-side statement timeout is genuinely threaded through — `execute_read` wraps the tx function with `neo4j.unit_of_work(timeout=timeout)` (only when a timeout is supplied) and hands that to `session.execute_read`, so Neo4j itself aborts the runaway transaction and releases resources. The client-side backstop runs the query on a daemon worker joined for the budget; the main thread is bounded by `thread.join(timeout_seconds)` and only ever reads `box` after confirming the worker is not alive (and raises without touching `box` when it is still alive), so a hung query cannot block the loader and there is no partial-read race on the shared dict. If a server ignores the statement timeout the wall-clock join still fires — defence-in-depth holds.

Aggregation. `is_fatal` is `not passed and not warning`, and both `load_all` and `_cmd_validate` key pass/fail off `is_fatal` (via `not any(v.is_fatal ...)`), not off `passed`. A timed-out check is passed=False+warning=True so `is_fatal` is False and it does not trip `validation_passed`; a genuine query error goes through the `except Exception` branch with the default warning=False so `is_fatal` is True and still hard-fails. A WARN never fails the load and a real error never silently passes — the exact contract, and the tests pin all four arms (fast pass, wall-clock WARN, server-side `TransactionTimedOut` WARN, non-timeout hard fail) plus the non-fatal aggregation.

Classification and hazards. `_is_timeout_error` matches "TimedOut"/"Timeout" against `.code`, which catches the real `Neo.ClientError.Transaction.TransactionTimedOut` and does not over-match — non-Neo4j exceptions carry no `.code` so they fall through to a hard failure (confirmed by the RuntimeError test). `_ValidationTimeout` is caught before the generic `except Exception`, so ordering is right. The abandoned daemon holds a driver session until the server-side timeout (same budget) reclaims the query; with only a handful of sequential validation queries this is well within the default pool and is documented in the docstring — acceptable. No path found where the loader can still hang, and no path where a WARN wrongly fails or a real error wrongly passes.

Must-fix: None
TechDebt: None

parametrization:
RequestOrReplied: Approved

**Architect review — bounded post-load validation is a sound, correctly-scoped defense-in-depth design that never hangs and fails safe**

The two-layer bound is the right pattern and the layers are genuinely complementary, not redundant. The server-side `unit_of_work(timeout=...)` is the idiomatic Neo4j bound that reclaims DB-side resources; the daemon-thread wall-clock backstop covers the case where the driver call itself blocks (wedged connection, a server that ignores the statement timeout) — so `execute_read` can never wedge the loader even if layer one misbehaves. Both paths converge on the same non-fatal WARN, which is the property that matters. Rejecting the hop-cap alternative is well-justified: shrinking `*1..N` silently drops longer cycles, so it changes detection semantics and is really a da#248 root-cause fix, not loader robustness. Keeping the query text intact while bounding it is the correct separation.

Daemon-thread abandonment is architecturally acceptable here specifically because this is a single-shot batch CLI that exits shortly after the load — the abandoned worker dies with the process, and because it wraps `self._driver.session()` in a context manager, the session returns to the pool as soon as the server-side timeout trips inside the thread (which, given equal budgets, is right around the join deadline). No indefinite leak. I would not want this exact abandon-and-move-on pattern inside a long-lived service, but that is not what this is.

The tri-state extension is clean and non-leaky: `warning` is the only new stored field and `is_fatal`/`status` are derived properties, so existing `.passed` consumers keep working while the load-gating logic moves to `is_fatal`. WARN-as-inconclusive is the right semantics for a post-load check (the load already succeeded; validation is a separate diagnostic), and it is applied consistently at every consumer — `load_all` gates on `not any(is_fatal)`, and both CLI commands print `status` and count `warning`. Critically the old `all(v.passed)` gate that would have failed on a warning is gone. `_is_timeout_error`'s substring heuristic on `.code` fails in the safe direction: a timeout it fails to recognize becomes a hard FAIL, never a hidden pass.

On the real trade-off — yes, a timed-out `chain_integrity` ships the load with exit 0 while the cycle check was never actually answered, and on the full graph that check will keep timing out until da#248 lands, so in practice it stays effectively un-run. That is the residual to be eyes-open about. It is acceptable and adequately surfaced: the WARN detail says "re-run this check manually, da#259", the load prints "N validation check(s) downgraded to warning", and structured logs emit `validation_timeout` + `warnings=N`. The alternative of hard-failing on timeout would block every full-graph load today, and hanging is strictly worse. 300s/query is a sane default. Correct scoping and honest `Refs #259` (owner-run full load is the real closure gate).

Minor, non-blocking: the wall-clock join budget equals the server-side timeout, so which layer fires first is a race; giving the backstop a small grace margin over the server-side budget would make the server-side path deterministically win and avoid the transient abandoned-thread window in the common case. Not required — the outcome is identical either way.

Must-fix: None
TechDebt: da#248 (super-linear cycle traversal — until fixed, chain_integrity stays effectively un-run on the full graph) and da#249 (multi-statement .cypher) remain out of scope and tracked separately, per the PR.

Issue #248, opened 2026-06-29
data-quality: TRANSMITTED_TO transmission cycles from single-token narrator under-disambiguation


parametrization:
## Summary

The `chain_integrity` validation (`queries/validation/chain_integrity.cypher`) reports cycles in the **TRANSMITTED_TO** narrator graph — `(n:Narrator)-[:TRANSMITTED_TO*1..20]->(n)`. On the #723 stage reload it hit the query's `LIMIT 100` cap, so the true count is ≥100. A transmission cycle (narrator A → … → A) is **chronologically impossible** — a teacher cannot transmit back to themselves — so every cycle is a data-quality artifact, not real isnad structure.

## Root cause (hypothesis)

**Single-token narrator under-disambiguation.** Common mononym mentions — `سفيان` "Sufyan" (27,118 mentions), `شعبة`, `مالك`, `الحسن`, `هشام`, `يحيى` — collapse multiple historically-distinct people into one canonical node:
- `سفيان` merges **Sufyān al-Thawrī** and **Sufyān ibn ʿUyayna** (different generations).

When one node stands in for several people across generations, the transmission graph gains back-edges and forms loops. This is **distinct from the name pollution fixed under #247** (those were non-names; these are real but over-merged names) — it is a disambiguation-*granularity* problem.

## Why now

Surfaced by the #723 reload once chains/edges were correct and validation ran cleanly. Pre-existing (TRANSMITTED_TO has carried this since narrator resolution shipped); not introduced by the #723 chain/pollution fixes.

## Suggested direction

- Quantify: remove `LIMIT 100`, group cycles by the mononym canonical nodes they pass through to rank the worst offenders.
- Disambiguation refinement for high-degree mononyms — split by co-occurring teacher/student context, generation/ṭabaqa, or kunya/nisba evidence — so `سفيان` resolves to the correct ibn-ʿUyayna vs al-Thawrī per chain.
- Until then `chain_integrity` is an **owner-accepted known state** (like the deferred inventory), not a release blocker.

## Acceptance
- Cycle count quantified (uncapped) and ranked by offending canonical node.
- High-degree mononym over-merges split with a precision guard (no legitimate single-person nodes fragmented).
- `chain_integrity` cycle count materially reduced.

parametrization:
Reopening: **the symptom this issue tracks is still present on stg after the fix merged.**

[...]

PR #264, opened 2026-07-02
da#248: split over-merged mononym narrators to break transmission cycles

parametrization:
## Summary

`TRANSMITTED_TO` transmission cycles (`chain_integrity.cypher`) are chronologically impossible, so each is a data-quality artifact. Root cause: the canonical narrator id is a **pure function of the normalized name**, so a common mononym — `سفيان` "Sufyan" (27,118 mentions) — collapses multiple historically-distinct people (Sufyan al-Thawri d.161 vs Sufyan ibn Uyayna d.198) into **one** node, giving the graph back-edges and loops. This is a disambiguation-**granularity** problem, distinct from the non-name pollution fixed under da#247/#253.

Delivered incrementally per the issue's "careful, correct, incremental" steer — both items in one PR, with the split made **structurally** precision-safe.

## Item 1 — quantification (shippable alone)

`queries/analysis/chain_integrity_mononym_ranking.cypher`: uncapped cycle detection, grouped/ranked by the anchor canonical node. `sum(cycle_count)` = the true total that `chain_integrity.cypher` caps at `LIMIT 100`. Deliberately kept **out of** `queries/validation/` because `run_validation` globs `validation/*.cypher` and classifies any non-empty result as a FAILED gate — a ranking query is expected to return rows. The shipped `chain_integrity.cypher` guard is left byte-for-byte untouched (not degraded).

## Item 2 — precision-guarded split

`src/resolve/mononym_split.py` — a curated registry of genuinely multi-person mononyms + `refine_mononym_name()`, which re-resolves a bare registered mononym to a specific person using the death years of its chain neighbours (positions ±1). Two **structural** precision guards make "no legitimate single-person node is fragmented" provable rather than hoped-for:

1. **Only registered names are ever eligible.** Every non-registered name — including single-person mononyms and dominant-referent ones like `شعبة`/`مالك` — abstains and is byte-identical to today.
2. **Unanimity-or-abstain.** Within a registered name it splits only when the chain evidence temporally selects **exactly one** person; ambiguous/absent evidence abstains (leaving a *residual* cycle, never a *wrong* split). Recall is traded for precision by design — no low-precision split is forced.

Split persons feed the **same** `make_canonical_id` contract (no new id scheme), minting distinct nodes for al-Thawri vs ibn Uyayna. Wired at both id-assignment sites in `disambiguate.py` (bio-match + self-canonicalized fallback). **Does not touch `ner.py`** (da#261 compound-splitting overlap avoided).

`tests/test_resolve/test_mononym_split.py` — precision tests proving non-registered / dominant-referent / ambiguous cases never fragment, correct resolution at the discriminating extremes, and distinct canonical ids for the split persons.

## Item 3 — measurement

Actual before/after cycle count is an **owner-run** regenerate→reload measurement (the issue notes closure is owner-run and `chain_integrity` is an owner-accepted known state). The mechanism's precision is proven at unit level here; recall/impact is measured on the next reload via the item-1 ranking query.

## Verification

- ruff (format + lint), mypy strict (91 files), pytest 1342 passed / 7 skipped — all local + pre-push green.
- Structural ontology index regenerated + committed (staleness gate green).

TechDebt: None

Refs #248


parametrization:
Requestor: Ivana Horvat
RequestOrReplied: Approved

**ML review — precision guards hold; no path to fragmenting a legitimate single-person node; subtlety (B) is a bounded soft-signal inconsistency, not a wrong fragmentation**

The acceptance criterion for this change is precision — a legitimate single person must never be fragmented — and both structural guards deliver it.

Guard #1 (registry-only) is airtight. `refine_mononym_name` short-circuits to `None` for any name not in `MONONYM_REGISTRY`, so every non-registered name, every single-person mononym, and every dominant-referent mononym (شعبة, مالك) is byte-identical to today. The `test_non_registered_name_never_splits_even_with_strong_evidence` and dominant-referent tests pin this. Registry curation is defensible — سفيان (al-Thawrī d.161 vs ibn ʿUyayna d.198, 37y apart) and يحيى (al-Anṣārī d.143 / al-Qaṭṭān d.198 / ibn Maʿīn d.233) are all famous, genuinely-distinct, death-date-separable transmitters. Omitting the tightly-clustered Hishāms (d.146/148/152, inside the window slack) is the correct precision-first call: the temporal gate could not separate them and would mostly abstain or risk error, so they are rightly deferred to a stronger identity signal.

Guard #2 (unanimity-or-abstain) selects at most one person in the discriminating regions and abstains on overlap. With window [15,80] around each death year, a neighbour ~130 leaves both Sufyāns plausible (|161-130|=31, |198-130|=68) and correctly abstains; a neighbour ~100 isolates al-Thawrī only (ibn ʿUyayna gap 98 > 80); a neighbour ~170 isolates ibn ʿUyayna only (al-Thawrī gap 9 < 15). Ambiguous or absent evidence abstains, so the worst case is a residual (unsplit) cycle, never a wrong split. Idempotency holds because keys are bare mononyms — a re-run over an already-refined `سفيان الثوري` is not a registry key and returns `None`, so ids are stable. The split re-uses the one `make_canonical_id` contract via `person.norm_name` (no new id scheme), and on a bio-path split the ambiguous candidate's bio is dropped (name_en/candidate → None) so wrong dates are not stamped onto the specific person — a good extra precision detail.

On the two evidence-quality subtleties the brief flagged:

(A) On an in-order forward pass, `death_year_index[p+1]` (the student) is not yet populated when mention p is processed — only `p-1` (the teacher) is visible — so `_adjacent_death_years` effectively sees roughly half the neighbour signal, and if batch order is not strictly position-sorted, which neighbour is visible becomes order-dependent. This is a pure recall/reproducibility asymmetry. It cannot cause a wrong split: fewer death years only feed guard #2 less real evidence, and guard #1 keeps every single-person node ineligible regardless. No precision impact.

(B) On a bio-path split, `death_year_index[pos] = c.death_year_ah` stores the pre-split ambiguous candidate's death year, not `person.death_year_ah` of the refined node. A downstream neighbour that is itself a registered mononym would read this stale value, and because it differs from the refined person's actual death year it can flip guard #2's plausible-set cardinality and mis-select among the registered persons. Traced to ground: this can never fragment a single-person node (guard #1 makes non-registered names ineligible no matter what neighbour death years say), so it does not violate the da#248 precision criterion. The stored value is still a real death year of the same over-merged mononym, and the blast radius is only an immediately-adjacent registered mononym (2 keys today). So it is a bounded soft-signal accuracy/recall inconsistency, not a wrong-fragmentation — tech-debt, not a blocker. The correct fix is trivial when convenient: store `person.death_year_ah if person is not None else c.death_year_ah`.

The item-1 ranking query is correctly kept out of `queries/validation/` (so `run_validation`'s glob does not treat expected rows as a failed gate) and leaves `chain_integrity.cypher` untouched. Approving: precision is proven at the unit level, and neither (A) nor (B) opens a path to fragmenting a legitimate single person.

Must-fix: None
TechDebt: (B) death_year_index stores the pre-split candidate's death year (c.death_year_ah) rather than the refined person.death_year_ah, so an adjacent registered mononym can read a stale soft-signal and mis-select among genuinely-distinct persons (never a single-person fragmentation); fix by storing person.death_year_ah on a split.

parametrization:
RequestOrReplied: Approved

**Data-engineer review — mononym split is precision-safe and consistently wired at both id-assignment sites; one non-blocking chain-context follow-up**

Integration is correct at both canonical-id sites. On the bio-resolved path the split re-points norm_name -> person.norm_name and the id follows via `_make_canonical_id(norm_name)`, with name_ar swapped to the specific person and, critically, candidate/name_en dropped. On the self-canonicalized fallback the same `refine_mononym_name` runs and fallback_name/fallback_name_ar move together into `_make_canonical_id`. No path changes the name without the id following, or vice-versa — verified against `test_split_persons_get_distinct_canonical_ids` (three distinct ids: residual mononym + two people).

The bio drop is complete at the record level. I read `_upsert_canonical` at HEAD: every biographical field (birth/death/generation/gender/trustworthiness/external_id/source_ids, and even name_ar/name_en backfill) is gated behind `if candidate is not None`. So a split passing candidate=None stamps nothing from the ambiguous mononym bio onto the specific person — only the split person's own name_ar plus the mention surface form as an alias, which is exactly right. Dropping name_en until a bio re-match is acceptable; the split node still gets enriched in place if a correctly-named bio (e.g. a candidate normalizing to سفيان الثوري) later lands on the same id.

Index-write ordering is sound. hadith_id/position are hoisted once and shared by both branches with no other consumer disturbed; `_adjacent_death_years` reads only positions +/-1 (never the current position) and runs before the current position's own death year is written, so a mention can never be its own neighbour.

Analysis-dir placement is safe. Confirmed `src/graph/validate.py:140` globs only `validation_dir.glob("*.cypher")`, so the uncapped ranking query under `queries/analysis/` cannot be picked up as a gate, and the shipped `queries/validation/chain_integrity.cypher` guard is untouched. Tests prove the precision guarantee both ways (non-registered / dominant-referent / ambiguous all abstain; unique-neighbour resolves to the right person at the discriminating extremes).

One non-blocking note, and why it is not a blocker: on the bio path a split sets candidate=None but the code still writes the pre-split bio's `c.death_year_ah` (and c's location) into death_year_index/location_index for this position — i.e. after deciding not to trust that bio's identity for this person, its death year is still propagated as this position's chain-context evidence to neighbouring mentions. The split person has a definitive registry death year (person.death_year_ah) that would be strictly better to index here. I did not raise this to a blocker because (a) it is chain-context soft signal, not a field on the split person's record (no record leak), (b) it only affects the bio-match branch — the fallback branch writes no index at all — and that branch fires only when a bio candidate is itself named the bare mononym, which biographical sources almost never are, and (c) the failure mode is at most a slightly-off neighbour signal, not a wrong split of this node.

Must-fix: None
TechDebt: On a bio-path split, index the split person's death year (person.death_year_ah), not the dropped bio candidate's c.death_year_ah, so neighbour chain-context evidence stays consistent with the identity decision.

Issue #266, opened 2026-07-02
da#248 tech-debt: split writes pre-split mononym's death year into chain-context index (fix before #723 closure reload)


parametrization:
## Finding (from da#248 PR #264 gate — Ivana Horvat (ML) + Kavitha Sundaramurthy (data-eng); both non-blocking)

In `src/resolve/disambiguate.py`, on the **bio-resolved** split path, after `refine_mononym_name` selects a specific person the code still writes the **pre-split ambiguous** candidate's death year into the chain-context index:

```python
death_year_index[f"{hadith_id}:{position}"] = c.death_year_ah   # c = ambiguous mononym bio
```

It should write the **refined** person's death year (`person.death_year_ah`) so a chain neighbour reading position `p` sees the split node's real year, not the merged mononym's.

## Why it is not a precision violation (both reviewers concurred)

- **Guard #1 keeps single-person nodes ineligible no matter what neighbour years say**, so a stale value can NEVER fragment a legitimate single-person node — the da#248 acceptance criterion holds.
- The stale value is still a *real death year of the same mononym*, and the blast radius is only an **immediately chain-adjacent registered mononym** (2 registry keys today: سفيان, يحيى).

## Why it is still worth fixing (correctness-adjacent)

Ivana's trace: a downstream adjacent **registered** mononym reading the stale value CAN flip guard #2's plausible-set cardinality and **mis-select among two genuinely-distinct persons** (e.g. resolve an adjacent سفيان to al-Thawrī when the refined neighbour evidence should have pointed to ibn ʿUyayna). That is a mis-attribution between real persons, not a fragmentation — bounded, but real.

## Fix (one line, both reviewers specified it)

```python
death_year_index[f"{hadith_id}:{position}"] = person.death_year_ah if person is not None else c.death_year_ah
```
Add a test where two registered mononyms are chain-adjacent and assert the second resolves off the first's *refined* year. (Also mirror the location index if it has the same pattern — Kavitha noted the split writes `c`'s location too.)

## Timing

⚠️ **Land before the owner-run #723 closure reload** (regenerate → reload → cycle-count measurement) so that reload's split output is computed on the corrected soft-signal. The fix is isolated to the two `disambiguate.run` split branches.

## Provenance

[...]

PR #267, opened 2026-07-02
fix(resolve): mononym split writes refined death year to chain index (da#266)

parametrization:
## Summary

Fixes the da#248 subtlety **(B)** flagged non-blocking in the PR #264 2-reviewer gate (Ivana Horvat + Kavitha Sundaramurthy): on a mononym split, `disambiguate.run` wrote the **pre-split ambiguous** mononym bio's death year (`c.death_year_ah`) into the chain-context `death_year_index`, not the **refined** person's (`person.death_year_ah`).

A chain-adjacent registered mononym then reads a stale year and can **mis-select among genuinely-distinct persons** (e.g. resolve an adjacent سفيان to al-Thawrī when the refined neighbour evidence pointed to ibn ʿUyayna). Bounded — guard #1 keeps single-person nodes ineligible, so this can never *fragment* a legitimate node (the da#248 acceptance criterion is untouched) — but a real mis-attribution between two real persons.

## Change

`src/resolve/disambiguate.py`, the bio-resolved split site (the only `death_year_index` write):

- **Death year:** on a split, write `person.death_year_ah` (the refined person's real year). Non-split path unchanged (`c.death_year_ah`).
- **Location (Kavitha's mirror note):** the parallel `location_index` carried the same staleness — `c`'s location belongs to whichever person the merged bio represented (possibly the wrong one post-split), and the registry has no per-person location. So the location write is **suppressed on a split**; the neighbour contributes no geographic signal, identical to an absent-location neighbour. This mirrors the split already dropping the ambiguous bio (`candidate`/`name_en` → `None`).

## Test

New integration test `tests/test_resolve/test_mononym_stale_year_propagation.py` drives the **real `ner → disambiguate` pipeline** over a three-narrator chain — anchor (d.110) → سفيان → سفيان. The second mononym resolves to **ibn ʿUyayna** only if the first wrote its **refined** year (161) into the index; under the old code it read the merged bio's year and abstained to the bare node. The سفيان bio is deliberately undated so the temporal filter is neutralised and the index-write is the *only* variable. Reverting the one-liner makes the test fail (verified locally).

- 308 `tests/test_resolve/` pass; pre-push mypy + pytest green.
- Structural index regenerated (new test file) — staleness-check passes.

## Timing

⚠️ Lands **before** the owner-run #723 closure reload (regenerate → reload → cycle-count) so the split output is computed on the corrected soft signal.

Refs #266
Parent: da#248 / main#723. Sibling: da#261 (compound-split `ner.py` wiring).


parametrization:
Requestor: Ivana Horvat
RequestOrReplied: Approved

**Review: LGTM — verified against the diff at HEAD.**

Must-fix: None
TechDebt: None

I raised subtlety (B) at the da#248 PR #264 gate and this is exactly the fix I specified. Verified on the diff, not the PR body:

- On a split (`person is not None`) the chain slot now takes `person.death_year_ah` — the refined person's real year. The non-split branch keeps `c.death_year_ah`, so byte-identical behaviour for every non-registered / abstaining mention.
- The da#248 acceptance criterion is untouched. Guard #1 (registry-only eligibility) is upstream of this write, so no soft-signal value here can make a single-person node eligible to fragment — this only ever changes selection *among* the registered persons of an already-registered mononym, which is precisely the mis-attribution (B) described.
- The integration test drives the real ner→disambiguate pipeline and asserts the second سفيان resolves to ibn ʿUyayna off the first's *refined* year (161). I confirmed it fails when the write is reverted to `c.death_year_ah` — it genuinely bites, it is not a tautology.

Good to merge.

parametrization:
RequestOrReplied: Approved

**Review: LGTM — verified against the diff at HEAD.**

Must-fix: None
TechDebt: None

My note at the da#248 gate was that the split writes `c`'s location into the parallel index too — same staleness class. This handles it correctly, and I agree with the choice to suppress rather than substitute:

- `MononymPerson` carries no per-person location (the registry is death-year + generation only), so there is nothing person-specific to write on a split.
- `c`'s location belongs to whichever person the merged bio represented — possibly the wrong one post-split, which is why the split already nulls `candidate`/`name_en`. Stamping that location onto the specific person's chain slot would propagate the same wrong-person signal to the geographic filter. Suppressing leaves the neighbour with no geographic signal, identical to an absent-location neighbour — the honest state.
- Non-split path is unchanged: location still written from `c` when the mention did not split.

The one behavioural question — could suppressing a location on a legitimately-registered-mononym split weaken a downstream geo-decision — is a soft-constraint pass-through either way (an absent neighbour location is already handled), so no regression. Approving.

Issue #268, opened 2026-07-02
resolve: crash-resumable pipeline — checkpoint disambiguate streaming + step-skip CLI


parametrization:
## Motivation

The 2026-07-02 host crash killed the wave-23 integrated resolve re-run mid-`disambiguate` at 34% (~86 min in). The pipeline has **no step resume and no mid-step checkpointing**, so recovery = full restart of the ~5–6 h disambiguate stage (3.1M mentions streamed in 50k batches). This is now the second full restart of the #723-closeout regenerate; every crash costs the whole stage.

## Precedent

da#245 already made dedup's embedding encode crash-resumable (`_encode_with_resume`: disk-backed memmap + id-map validation + resume log events). This issue extends the same property to the wall-clock pole of the pipeline.

## Scope

1. **Checkpoint `disambiguate`'s streaming loop** (`src/resolve/disambiguate.py`, `_stream_mentions` consumer): every N batches persist to a checkpoint dir under `data/staging/` — accumulated resolved-mention rows, minted canonical-registry state, processed-row offset, and a **content fingerprint** of the input `narrator_mentions_resolved.parquet` (not mtime — a re-run NER rewrite with identical content must not invalidate). On start: valid checkpoint → resume from offset with `disambiguate_resume` log event; mismatch → cold start.
2. **Step-skip CLI**: `python -m src.cli resolve --from-step disambiguate` (or equivalent) so a restart doesn't redo completed steps whose outputs are present. NER is cheap (~2 min) but its rewrite must compose with (1)'s fingerprint.
3. **Tests**: small-parquet fixtures covering checkpoint write/load/resume, fingerprint-mismatch cold start, and resumed-vs-cold **output equivalence** (same input ⇒ identical canonical output).

## Constraints

- Resumed run must be output-identical to a cold run on the same input — no behavior change to resolution itself.
- Checkpoint artifacts live under gitignored `data/` paths.
- Do not disturb the live wave-23 resolve run in the primary checkout (work in a worktree).

Part of the wave-23 data-quality closeout hardening (main#723 context).

PR #269, opened 2026-07-03
fix(resolve): crash-resumable disambiguate checkpoint + --from-step skip (da#268)

parametrization:
## Summary

Makes the resolve pipeline crash-resumable so a host crash (like the 2026-07-02 one that killed the wave-23 integrated re-run mid-`disambiguate` at 34%, ~86 min in) no longer restarts the ~5–6 h disambiguate stage from scratch. Extends the da#245 crash-resume property from dedup's encode to disambiguate's streaming loop.

Closes #268

## Checkpoint design

- **What is checkpointed**: `disambiguate.run` persists its full accumulated state — canonical registry (`canonical_map`, insertion-order preserved), chain-context `death_year_index`/`location_index`, `resolved_map`, `merge_log_rows`, `ambiguous_rows`, `naive_identity_pairs`, per-source counters, and the processed-row offset — to `staging/.disambiguate_checkpoint/state.json`.
- **Format**: a single JSON file, written atomically (`.tmp` → `os.replace`) so a crash mid-write never leaves a half-file. JSON (not parquet) because the state is a mix of ordered dicts with nested lists; the infrequent write cadence keeps the cost negligible.
- **Cadence**: every `N` completed 50k batches (default `N=4` ≈ one write per ~20 min at the prod rate; bounds crash-loss to ~4 batches while keeping write overhead <1% of wall-clock). Override via `DISAMBIGUATE_CHECKPOINT_EVERY_N_BATCHES` or the `run()` kwarg. Checkpoints land on batch boundaries, so a resume skips whole batches and the continuation is byte-identical to a cold run.
- **Cleanup**: the checkpoint dir is removed after a successful complete run.

## Fingerprint choice (the load-bearing decision)

The input fingerprint is split into two SHA-256 sub-hashes over `narrator_mentions_resolved.parquet`, streamed row-group by row-group (never mtime, never the whole file in memory):

- `content_hash` over the **stable** disambiguation drivers (`hadith_id`, `source_corpus`, `position_in_chain`, `name_raw`, `name_normalized`, `transmission_method`);
- `mention_id_hash` over the `mention_id` column.

Resume requires **both** to match. Why split: **`mention_id` is minted per NER run via `uuid.uuid4()`** (`src/resolve/ner.py`) — it is random, not content-derived, and it is baked into the outputs (`merge_log`, `ambiguous`, and the backfill join). So an output-identical resume is only possible against the *same* mentions file. The split lets a content-match / id-mismatch (an NER *rewrite*) cold-start with a precise, actionable log hint instead of silently producing a corrupt mix. `canonical_narrator_id`/`confidence` are excluded (the end-of-run backfill rewrites them, and mention_id/content are unaffected by it).

## Step-skip CLI

`python -m src.cli resolve --from-step <step>` (and `run_all(from_step=...)`) skips every step before `<step>` whose outputs already exist. The **sanctioned crash-recovery command is `resolve --from-step disambiguate`**: it skips the cheap NER (~2 min) so the existing mentions file — and disambiguate's checkpoint keyed off it — is reused verbatim, guaranteeing an output-identical resume. The `disambiguate`-OVERWRITES / `bio_promote`-MERGES ordering invariant (da#99/da#117) is preserved and documented in `RESOLVE_STEP_ORDER`.

## Deviation from the issue spec

The issue asks that "an identical NER rewrite must not invalidate" the checkpoint. That is only fully achievable if `mention_id` were deterministic; it is a random uuid4 and making it deterministic risks graph-identity blast radius (out of scope). Instead: `content_hash` ignores mention_id so the corpus check itself is stable across an identical-content rewrite, but the resume decision still requires the id-hash to match (recovery routes through `--from-step disambiguate`, which reuses the file). Verified that `mention_id` is only an internal resolve join key — the graph keys `NARRATED`/Chains off `canonical_narrator_id`, and `merge_log`/`ambiguous` have no graph/enrich consumer.

## Test evidence

`ENVIRONMENT=test uv run pytest tests/test_resolve/` → **320 passed, 3 skipped** (no regressions); full suite **1363 passed, 7 skipped**. New tests (12):

- `test_disambiguate_resume.py`: fingerprint content/id split (stable across id rewrite; changes on content change), checkpoint save/load/clear roundtrip (+ atomic temp-rename), **resumed-vs-cold OUTPUT EQUIVALENCE** (simulated crash → resume → canonical + backfilled mentions + merge_log all identical), fingerprint-mismatch cold start, stale-mention-id cold start, `resume=False` ignores checkpoint.
- `test_resolve_from_step.py`: `_resolve_start_index`, unknown-step `ValueError`, `--from-step disambiguate` skips NER, `--from-step bio_promote` skips NER+disambiguate, no-arg runs everything.

`make check` (ruff lint + format-check + mypy strict + pytest) green; pre-push hooks (mypy-strict, pytest) green.


parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

Reviewed the full resolve control-flow surface at head f16a93a (src/resolve/__init__.py, src/cli.py, src/resolve/disambiguate.py) plus both new test modules — not just the diff hunks. Approving: the crash-resume checkpoint and the --from-step skip are correct, well-tested, and the sanctioned recovery path is sound. All 12 CI checks are green (including Base-image pin + fixture realism and staleness-check).

What I verified against the artifact
- Checkpoint state is fully JSON-serialisable — canonical_map records hold only str/int/None/list values (_upsert_canonical), and the set/tuple accumulators (naive_identity_pairs, resolved_map) are explicitly converted to lists in _snapshot and rebuilt on restore, so _save_checkpoint cannot raise mid-run on a non-serialisable payload.
- Atomic write is correct (sibling .tmp then os.replace within the same dir), and the resumed-vs-cold OUTPUT EQUIVALENCE test genuinely simulates a crash (raises after 2 batches, asserts a partial checkpoint with 0 < processed < total survived, resumes, then asserts canonical + backfilled mentions + merge_log are byte-identical). The mention stream is a pure function of (single-file content, batch_size) with mention-granular skip, so the continuation is deterministic; the chain-context indexes only accumulate forward, so restoring them at a batch boundary reproduces cold-run behaviour.
- Fingerprint split (stable content vs uuid4 mention_id) matches the documented deviation-from-spec; the content-match / id-mismatch branch cold-starts with the actionable "--from-step disambiguate" hint, and the backfill excludes canonical_narrator_id/confidence from the content columns so a completed run's rewrite of the mentions file does not perturb the hash. Tests assert both stability across an id rewrite and change on a content change, using the real NARRATOR_MENTIONS_RESOLVED_SCHEMA / NARRATORS_CANONICAL_SCHEMA — no fixture-realism gap.
- CLI --from-step uses argparse choices=RESOLVE_STEP_ORDER, so an unknown step is a clean error rather than a silent empty run; the sanctioned recovery command is documented in help text, the PR body, and the run/run_all docstrings. The disambiguate-OVERWRITES / bio_promote-MERGES ordering invariant (da#99/da#117) is preserved in RESOLVE_STEP_ORDER.
- Cadence override is validated — kwarg > env > default, floored at max(1, ...), and a bad env value warns then falls back to the default.

Non-blocking hardening notes (author's discretion; none block this merge, hence TechDebt none)
1. The input fingerprint covers only the mentions file, not the bio candidates loaded by _load_candidates(staging_dir), which also drive disambiguation output. If the candidate/bio staging changed between crash and resume, the checkpoint would not invalidate and a resume could splice an old-candidate prefix onto a new-candidate continuation. Operationally low risk (the sanctioned recovery re-runs without touching staging), but the resume-safety guarantee is complete only for the mentions input — worth a follow-up to fold in a candidate fingerprint (or an explicit "don't regenerate staging between crash and resume" caveat in the runbook).
2. run_all(from_step=...) has no existence guard on the outputs of the steps it skips, except the ner then mentions gate. --from-step bio_promote (or later) runs bio_promote/cluster against a possibly-absent narrators_canonical.parquet and still exits 0 printing "Resolution complete." — a silent partial-output footgun for a mistyped --from-step. The sanctioned disambiguate path is safe (it re-runs and overwrites), so this is minor, but a pre-run check or warning on the required prior artifact would close it.
3. The two degenerate early returns in disambiguate.run (no candidates ~L1097, no mentions ~L1105) return without _clear_checkpoint, so a stale checkpoint can linger; it is fingerprint-guarded on the next run so it is harmless, just inconsistent with the clear-on-complete contract.

Nice work on the byte-identical resume proof and the fingerprint reasoning — that is the load-bearing part and it holds up.

parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved

Reviewed the full `src/resolve/disambiguate.py` and `src/resolve/__init__.py` at head f16a93a, plus the graph loaders, `ner.py`, `src/cli.py`, and both new test files. This is a clean, well-reasoned implementation; the load-bearing safety argument holds. Approving. Findings below are all non-blocking tech debt.

**1. Checkpoint accumulator set is complete.** I diffed `_snapshot()` (disambiguate.py:1789-1807) line-by-line against every mutation in the streaming loop (disambiguate.py:1822-1991). Everything the loop mutates is captured and restored: `canonical_map` (with insertion order — see below), `death_year_index`, `location_index`, `resolved_map`, `merge_log_rows`, `ambiguous_rows`, `naive_identity_pairs`, `source_resolved`, `source_total`, and `processed`. No divergent-on-resume accumulator. Insertion order of `canonical_map` survives because JSON object key order + Python dict order + `json.loads` all preserve it, and `_build_canonical_table` (disambiguate.py:1356) emits rows in that order — the equivalence test's order-sensitive list `==` locks this.

**2. The fingerprint gate is sound — a stale checkpoint can never be spliced into a regenerated file.** Resume requires BOTH `content_hash` AND `mention_id_hash` to match (disambiguate.py:1744-1747). An NER rewrite re-mints `mention_id` via `uuid.uuid4()` (confirmed ner.py:85,174), so `ids_ok` goes false and the `content_ok and not ids_ok` branch (1773-1784) cold-starts, clears the checkpoint, and logs the actionable "rerun with --from-step disambiguate" hint. A bare `resolve` re-run regenerates NER first, so its fresh ids never match a prior checkpoint. The only way both hashes match is the identical file reused via `--from-step disambiguate`. Safe.

**3. Ivana's "no graph consumer keys on mention_id" claim verified at head sha.** `src/graph/load_nodes.py:546` and `load_edges.py:150-154,346` key exclusively on `canonical_narrator_id`. `mention_id` appears only in `merge_log`/`ambiguous` (audit artifacts) and as the internal backfill join key. Excluding `canonical_narrator_id`/`confidence` from the fingerprint is correct — the end-of-run backfill (disambiguate.py:1093-1163) rewrites exactly those two columns and is idempotent, so cold-starting on an already-backfilled file reproduces identical output.

**4. Resumed-vs-cold test asserts real row-level identity, not a weaker check.** `test_resumed_run_is_output_identical_to_cold` simulates a genuine mid-stream crash (raises on the 3rd batch), asserts a non-trivial partial checkpoint survived (`0 < processed < 12`), resumes to completion, then compares full `.to_pylist()` of canonical + backfilled mentions + merge_log with order-sensitive `==`. That is content identity including canonical row order.

**5. Ordering invariant (disambiguate OVERWRITES, bio_promote MERGES — da#99/da#117) survives --from-step.** `RESOLVE_STEP_ORDER` fixes disambiguate before bio_promote before cluster; `from_step` can only pick a start index and always runs subsequent steps in order — it can never reorder them or run bio_promote's MERGE without disambiguate's OVERWRITE having already landed on disk in the prior run. `--from-step bio_promote` correctly skips disambiguate and merges into the existing canonical file (test_from_step_bio_promote_skips_ner_and_disambiguate).

**6. Atomic write fails closed.** `.tmp` -> `os.replace` is an atomic rename; a torn write degrades safely because `_load_checkpoint` catches ValueError/OSError and returns None -> cold start. Corruption can never be applied.

Tech debt (file as tracked issues, none block merge):

- The mononym-split-across-a-crash-boundary path is not directly integration-tested. The split (disambiguate.py:1850,1919) is a pure function of the restored `death_year_index`, and each mention is processed exactly once, so it cannot double-apply or skip on resume — but the resume fixtures contain no registered mononym, so this is proven by construction rather than exercised. Worth a fixture that plants a mononym whose neighbour death-year is written in an early (pre-crash) batch and the mononym itself in a late (post-crash) batch, to lock the interaction with da#248/da#266.
- `_save_checkpoint` does not fsync the temp file or the containing directory before/after `os.replace`, so a hard power-loss crash immediately after a save can lose the just-written checkpoint (fails safe to cold-start, never to corruption). Immaterial for the actual threat model here — an OOM/process kill leaves the page cache intact under a live kernel — but a durability gap if the box ever loses power mid-run.
- `_load_mentions` (disambiguate.py:1066) is dead code: referenced nowhere in the repo at head sha (the streaming refactor uses `_iter_mention_batches`/`_count_mentions`). Optional cleanup.

CI: all 12 checks green, mergeStateStatus CLEAN.

TechDebt: (1) mononym-split-across-crash-boundary lacks a direct integration test; (2) `_save_checkpoint` omits fsync before/after os.replace (safe cold-start on power-loss, not corruption); (3) `_load_mentions` is unreferenced dead code.

Issue #273, opened 2026-07-03
Local DuckDB tooling for Parquet exploration (staging + curated views, make duck)


parametrization:
## Goal

Owner-requested: a local DuckDB setup for interactive exploration of the pipeline's Parquet files (staging + curated), so questions like "top blocks by member count" or "mention_count distribution per canonical narrator" are one SQL query instead of a bespoke PyArrow script.

## Deliverable

1. `duckdb` added as a **dev** dependency (uv dev group — it must not enter the runtime image).
2. A small helper, `src/tools/duck.py` (or `scripts/duck.py` if that fits repo layout better), that opens an in-memory DuckDB with views pre-registered over the configured data dirs:
   - one view per staging parquet dataset (`data/staging/*.parquet`, globbed, e.g. `staging_mentions`, `staging_narrators`)
   - one view per curated dataset (`data/curated/*.parquet`)
   - view names derived from file stems; glob-datasets handled with `read_parquet([...])`
3. `make duck` target dropping into the DuckDB CLI/Python REPL with views loaded; also usable non-interactively: `make duck QUERY="select ..."` or equivalent `-c` passthrough.
4. Short docs section (README or docs/) with 3–4 example queries against real datasets — e.g. block-size distribution, top-20 narrators by mention_count, name-length histogram.
5. Read-only stance: helper opens files for reading only; it must never write into `data/`.

## Notes

- Paths come from `src/config.py` settings (`DATA_STAGING_DIR`, `DATA_CURATED_DIR`), not hardcoded.
- The live resolve run (pid 20784) owns `data/` — reading concurrently is fine (Parquet files are write-once per stage), but the helper should tolerate a file appearing/being replaced mid-session.
- This tool is the instrument for the candidate-pair inflation analysis (sibling issue, same assignee) — build this first.

## Constraints

- Worktree off `origin/deployments/phase-8/wave-23`; PR targets that branch.

parametrization:
Delivered by PR #274, merged into deployments/phase-8/wave-23 with 2 approvals (Kavitha, Oyunbileg). Closing manually — wave-branch merges do not auto-close.

PR #274, opened 2026-07-03
feat(tools): local DuckDB Parquet exploration (make duck) (da#273)

parametrization:
## Summary

Dev-only DuckDB instrument for interactive SQL over the pipeline's staging +
curated Parquet, so questions like "top blocks by member count" or
"mention_count distribution per canonical narrator" are one query instead of a
bespoke PyArrow script. It is the intended instrument for the da#271
candidate-pair-inflation audit (sibling issue, same assignee).

Part of #273

## What's here

- **`duckdb` as a uv dev dependency** (dev group only — it never enters the
  runtime image) + a mypy ignore-missing-imports override.
- **`src/tools/duck.py`** — opens an in-memory DuckDB and registers read-only
  views over the configured data dirs. Paths come from `src.config` settings
  (`DATA_STAGING_DIR`/`DATA_CURATED_DIR`), never hardcoded.
  - One per-file view per `*.parquet` (`<layer>_<stem>`), plus a combined union
    view for each multi-file dataset.
  - **Shard grouping is schema-guarded**: files group into a combined view only
    when they share a leading name prefix AND an identical column schema, so
    distinct datasets that merely share a prefix (`narrator_aliases_*` vs
    `narrator_mentions_*`, both starting `narrator`) never merge into a bogus
    view. No hardcoded dataset vocabulary.
  - **Strictly read-only**: only `CREATE VIEW ... read_parquet`, never a write
    into `data/`. Each query re-scans the files, so a shard the live resolve
    replaces atomically mid-session is picked up on the next query.
- **`make duck`** (interactive REPL) + **`make duck QUERY="select ..."`**
  (one-shot; `--csv` for scriptable output, `--list` to enumerate views).
- **`docs/duckdb.md`** — views + 4 example queries; README pointer.

## Views registered over the live wave-23 data (36 total)

`staging_hadiths` (9 shards), `staging_collections` (8), `staging_narrators_bio`
(3), `staging_network_edges` (3), `staging_narrator_mentions` (2), the per-file
views, and curated `narrators_canonical` / `narrator_mentions_resolved` /
`merge_log`. The schema-guard correctly kept `staging_narrator_aliases_itqan`
out of any combined `narrator` view.

## Test evidence

`tests/test_tools/test_duck.py` (8 tests): dataset-planning (view naming +
schema-guarded shard grouping over the **real** wave-23 dataset names) and an
end-to-end DuckDB registration + query over tiny synthetic parquet (tmp paths,
never the real corpus). Full suite: **1371 passed, 7 skipped** (non-integration);
`ruff check` + `ruff format --check` + `mypy --strict src/` all green;
structural ontology index regenerated.


parametrization:
## Review — PR #274 (da#273): local DuckDB Parquet exploration

Reviewed at head `d0a5b47` against base `deployments/phase-8/wave-23`, data-engineering lens. Read the full `src/tools/duck.py`, `tests/test_tools/test_duck.py`, `Makefile`, `pyproject.toml`, `docs/duckdb.md`, and cross-checked `src/config.py` at HEAD. I did not touch the live checkout.

### What I verified

**Dependency isolation (dev-only, not runtime).** `duckdb>=1.1` sits in the `[dependency-groups].dev` block, never in `[project].dependencies` — it cannot enter the runtime image. Confirmed the module keeps it that way at import time: the top of `duck.py` imports only stdlib plus `src.utils.logging`; `duckdb` is under `TYPE_CHECKING` and every real import (`import duckdb`, `import pyarrow.parquet`) is deferred into function bodies (`build_registry`, `_pyarrow_schema_key`). So merely importing `src.tools.duck` — or `src.tools` — pulls in neither heavy dep. The `ignore_missing_imports` mypy override for `duckdb.*` mirrors the existing `pyarrow`/`torch`/etc. pattern.

**Config-driven paths, zero hardcoding.** `_resolve_dirs` falls back to `get_settings().data_staging_dir` / `.data_curated_dir`; I confirmed both fields exist on `Settings` in `src/config.py` (lines 40-41). CLI `--staging`/`--curated` are dev overrides only. No literal `data/...` path anywhere in the registration path.

**Schema-guarded shard grouping — the core correctness claim — is sound.** I traced `plan_datasets` over the real wave-23 stems in `_STAGING_SCHEMAS`. The two-part guard (shared leading-segment prefix AND identical `schema_key`) correctly:
- groups `collections_*`, `hadiths_*`, `narrator_mentions_*`, `network_edges_*` into combined views;
- keeps `narrator_aliases_itqan` and `narrator_mentions_*` apart — the `(narrator,)` prefix spans both, its member set carries two distinct schema keys, so `len(keys) != 1` drops it and no bogus `staging_narrator` view is created;
- resolves the longest-prefix-wins case (`narrators_bio` not `narrators`) via the `longer_same` fileset comparison;
- leaves singletons (`parallel_links`) as per-file views only.
The per-file-name collision guard (`if name in per_file: continue`) is a correct, documented tie-break.

**Read-only by construction.** The only DDL issued is `CREATE OR REPLACE VIEW ... SELECT * FROM read_parquet([...], union_by_name=true)` over an in-memory `:memory:` connection. Nothing writes into `data/`. `union_by_name=true` makes the union tolerant of column reordering across shards, consistent with the order-insensitive `schema_key`. Identifiers and path literals are quoted (`_quote_ident`, `_quote_sql_str`) — no injection surface from stems. Safe to run against a directory a live `resolve` owns, and re-scan-per-query means an atomically-replaced shard is picked up next query.

**Makefile target.** `duck` is in `.PHONY`; the recipe `uv run python -m src.tools.duck $(if $(QUERY),--query "$(QUERY)",)` uses GNU-make's `$(if ...)` (evaluated by make, not the shell), so it is unaffected by zsh vs sh in the recipe line. Interactive and one-shot forms both dispatch correctly to the `-c/--query` arg.

**Tests.** Eight tests: five pure `plan_datasets` cases pinned to the real dataset names (naming, combined grouping, longest-prefix, schema-mismatch-blocks-merge, singleton), plus a genuine end-to-end DuckDB registration + query over tiny synthetic Parquet in `tmp_path`, a missing-dirs tolerance case, and CSV/`--list` CLI coverage. The corpus is never touched. Good property coverage of the branch logic.

### Non-blocking observations (tech debt / future polish)

1. Table mode caps at `rel.show(max_rows=1000)` — a query returning more rows is silently truncated with no "… N more" hint. Fine for an exploration REPL, but a mild footgun; `--csv` is the escape hatch. Worth a docs note or a truncation banner later.
2. The prefix+schema-identity grouping is a heuristic: two genuinely distinct datasets that share both a name prefix and an identical column schema would still merge. Not reachable in the current corpus, but if the producer set grows it may warrant an explicit dataset manifest rather than inference.

Neither blocks merge. Clean, well-documented, correctly-scoped dev instrument.

---
RequestOrReplied: Approved
TechDebt: Table mode silently truncates at 1000 rows (no overflow hint); shard-grouping is a prefix+schema heuristic that would merge two distinct datasets sharing both a prefix and an identical schema.

parametrization:

Reviewed the artifact at head `d0a5b47ef8571c5e56c9b17fa30a5bf70804825b` via the GitHub API (no checkout — the live pipeline owns the main data-acquisition working tree). Read `src/tools/duck.py` and `tests/test_tools/test_duck.py` in full, plus `docs/duckdb.md`, `Makefile`, `pyproject.toml`, `src/tools/__init__.py`.

### Read-only guarantee — holds by construction
The helper opens `duckdb.connect(":memory:")` and only ever issues `CREATE OR REPLACE VIEW ... SELECT * FROM read_parquet([...], union_by_name=true)`. I found no `COPY`, no `ATTACH`, no write path into `data/`, and no on-disk database file. The in-memory connection plus lazy `read_parquet` means each query re-scans the files, so an atomically-replaced shard is picked up on the next query as claimed. The read-only property is a property of *what the helper registers*, and that property is genuinely airtight. (See the one caveat below about arbitrary user SQL in the REPL — that is a documentation nuance, not a helper defect.)

### CI truth — verified independently
I read `statusCheckRollup` myself at the head SHA: all 12 checks report SUCCESS (Lint, Type Check, Test, Integration Tests, Base-image pin + fixture realism, Pre-commit sync-drift gate, Markdown lint, Spellcheck, staleness-check, lychee, YAML/JSON, actionlint). Not taking the author's word — this matches.

### Test coverage — adequate for the risk surface
The 8 tests cover the load-bearing logic well: per-file view for every stem, schema-guarded shard grouping into combined union views, longest-prefix-wins naming, the schema-mismatch guard that keeps `narrator_aliases_*` and `narrator_mentions_*` from merging into a bogus `narrator` view, single-file datasets getting no combined view, an end-to-end real-DuckDB registration that exercises the union-view read-through, missing dirs tolerated, and the CSV/`--list` CLI surface. The schema-guard and union-view paths the brief called out are both directly covered.

### Findings (both non-blocking for a dev-only instrument)

1. **Planning-time schema probe is unguarded — a corrupt multi-shard file crashes the whole tool.** `_register_layer` wraps `con.execute(CREATE VIEW ...)` in `try/except` with the intent stated in the comment that "one bad file must not abort the rest." But the default `schema_key` is `_pyarrow_schema_key`, which calls `pq.read_schema(path)`, and `plan_datasets` invokes `schema_key(m)` for every member of any ≥2-file prefix group — this runs during planning, *before and outside* that `try/except`. So a single corrupt/truncated shard among a multi-file dataset (the 9 `hadiths_*`, 8 `collections_*`, etc.) raises out of `build_registry` and aborts `make duck` entirely, before any view registers. It is bounded: dev-only, a loud traceback rather than a silent wrong answer or any data write, and atomic shard replacement by the live resolve makes mid-write corruption unlikely (a genuinely persisted-corrupt shard is arguably worth surfacing loudly anyway). Worth a follow-up to wrap the probe so graceful degradation matches the stated design, and a test that drops a malformed parquet among shards.

2. **Read-only is by-construction of the helper, not an enforced sandbox.** The REPL and `-c` path execute arbitrary user-typed SQL via `con.sql(...)`. DuckDB permits `COPY ... TO` and extension loads by default, so a user could in principle write a file from the REPL. This is the user's own action, not the helper's, and it is a dev tool — but the docs' "Safe to run against the directories a live resolve owns" would be more precise if it noted the guarantee covers the helper's registration, not arbitrary SQL the operator types.

Minor: the empty-but-existing data-dir path (`exists()` true, glob empty) isn't directly tested (only the missing-dir path is), though the code path is trivial. Example queries in `docs/duckdb.md` reference plausible columns (`canonical_id`, `name_ar`, `mention_count`, `name_ar_normalized`) and fail gracefully in the REPL if a name is off; I did not run them against the live corpus.

### Verdict
Clean, well-documented, correctly read-only for its intended job, CI green (verified), and well-tested on the schema-grouping logic that carries the risk. The two findings are legitimate robustness/doc-precision items but are non-blocking for a dev-only exploration instrument that produces loud errors rather than bad data. Approving.

---
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: Planning-time pq.read_schema probe in plan_datasets is outside the _register_layer try/except, so a corrupt multi-shard file crashes the whole tool despite the "one bad file must not abort" intent — wrap the probe and add a malformed-parquet test.

Issue #272, opened 2026-07-03
Unified checkpoint/resume for all resolve stages (fuzzy_cluster first) under one flag convention


parametrization:
## Goal

Extend crash-resumability (da#268 / PR #269 pattern) to every resolve-pipeline stage that lacks it, under one shared convention, so a crash at hour N of any stage never costs more than the current batch.

## Current coverage

| Stage | Resume today | Notes |
|---|---|---|
| ner | none | mints `mention_id` via uuid4 per run — cold start by design; checkpoint = reuse mentions file |
| disambiguate | full checkpoint (PR #269) | `staging/.disambiguate_checkpoint/state.json`, atomic tmp+replace, every 4 batches, content fingerprint, cleared on success |
| bio_promote | none needed? | ~10s stage — confirm and document as exempt rather than silently skipping |
| fuzzy_cluster | **none — highest priority** | currently a multi-day stage; a crash loses everything |
| dedup | partial | `_encode_with_resume` (da#245) resumes the embedding memmap only; FAISS/cluster phase after it is not checkpointed |
| parallels | none | |
| CLI | `resolve --from-step <step>` (PR #269) | step-level skip, gated by `RESOLVE_STEP_ORDER` |

## Conventions (unify on the PR #269 design — deviations need a stated reason)

1. Checkpoint location: `<staging_dir>/.<stage>_checkpoint/state.json`
2. Atomic write: write `.tmp`, `os.replace`
3. Cadence: every N batches/blocks (stage-appropriate N), plus on clean stage exit
4. **Content fingerprint** of the stage input (stable driver columns hash — NOT mtime) so a stale checkpoint against changed input is detected and discarded, never silently reused
5. Cleared on stage success
6. CLI: `resolve --from-step <step>` remains the step-level entry; intra-stage resume is automatic when a valid checkpoint exists; add a uniform `--no-resume` flag to force cold start (same flag name for every stage)
7. Progress log line on resume stating what was skipped (`<stage>_resume checkpoint=... skipped=...`)

## Scope

- Priority order: fuzzy_cluster > dedup (FAISS phase) > parallels > ner (mentions-file reuse) > bio_promote (document exemption)
- fuzzy_cluster checkpointing must account for the union-find state + per-block completion set; coordinate with the throughput-degradation issue (sibling) so the two changes do not collide in `cluster_records` — sequence the PRs if needed.
[...]

PR #275, opened 2026-07-03
feat(resolve): unify crash-resume across stages + shared checkpoint helper (da#272, PR 1/2)

parametrization:
## Summary

Extends the da#268 / PR #269 disambiguate crash-resume pattern to the rest of the resolve pipeline under **one shared convention**, so a crash at hour N of any long stage no longer costs more than the current batch/block. This is PR 1 of 2 for #272 — the shared helper + the "easy" stages. The **fuzzy_cluster** checkpoint is intentionally held for a follow-up PR, sequenced after da#270's `cluster_records` restructure so it is built against the landed shape.

## What's here

- **`src/resolve/_checkpoint.py`** — shared primitives implementing the 7 conventions: stage-namespaced dir (`<staging>/.<stage>_checkpoint/state.json`), atomic `.tmp`→`os.replace` save/load/clear, cadence resolver (kwarg > env > default, floored at 1), single-pass **column-group content fingerprint** (never mtime), scalar fingerprint for non-parquet inputs, and the uniform `<stage>_resume` log line. Stages own only their serialized **state shape** + fingerprint column set.
- **disambiguate** — refactored to consume the helper. The fingerprint output is **byte-identical to pre-#272** (verified against the old inline hasher), so a checkpoint written by the currently-merged code still resumes; the existing resume tests were retargeted to the shared module.
- **dedup** — the FAISS search + pair-collection phase (the part *after* the da#245 encode memmap, which was already resumable) now checkpoints per query row-block. Extracted `_search_and_collect_resumable` with the index injected via `build_index` / `reload_index` callables, so the resumable loop is unit-testable **without faiss / sentence-transformers**. On resume the persisted `.faiss` index is **reloaded, not rebuilt** — an IVF index's random kmeans init would otherwise diverge the resumed search from the crashed run.
- **parallels** — the anchor scan checkpoints accumulated links + the next anchor per block; `indexed`/`postings` are rebuilt deterministically, so a resumed scan is byte-identical to a cold one.
- **ner / bio_promote** — documented as **exempt**: ner is cold-by-design (it re-mints `uuid4` mention_ids; recovery is `resolve --from-step disambiguate`), and bio_promote is a sub-minute idempotent merge.
- **CLI** — uniform `resolve --no-resume` threads through `run_all` to every resumable stage; orthogonal to `--from-step` (which selects *which* step to start at).

## Test evidence

`ENVIRONMENT=test pytest` → **1429 passed, 7 skipped** (was 1363; +~36 new). New tests:
- `test_checkpoint.py` — save/load/clear + atomic rename, corrupt-state → cold, cadence priority/floor/invalid-env, column-group fingerprint stability/isolation/missing-input/multi-path, scalar hash.
- `test_dedup_faiss_resume.py` — **cold == resume output equivalence** via a numpy-only fake index (simulated crash → partial checkpoint → reload → identical pairs), stale-fingerprint discard, `--no-resume`.
- `test_parallels_resume.py` — **real crash simulation** (raise mid-scan) → resume == cold, stale-fingerprint discard, `resume=False`.
- resume-flag propagation through `run_all`.

`ruff check` + `ruff format --check` + `mypy --strict` + structural-ontology staleness gate all green (structural index regenerated). Pre-commit + pre-push hooks green.

## Related Issues

Part of #272

## Review Checklist
- [ ] Must-fix items resolved


parametrization:
## Review — PR #275 (da#272 PR 1/2): unified crash-resume + shared checkpoint helper

Reviewed at head `3df78eb` against base `deployments/phase-8/wave-23`, data-engineering lens. Read the full `_checkpoint.py`, `disambiguate.py`, `dedup.py`, `parallels.py`, `__init__.py`, `cli.py`, the ner/bio_promote exemption docstrings, and the new resume tests at HEAD; and I compared the fingerprint against the #269 implementation checked out from the base branch. I did not touch the live checkout (all reads via the GitHub API).

### What I verified

**`_checkpoint.py` primitives.**
- Atomic save: `save_checkpoint` writes a sibling `state.json.tmp` then `os.replace` onto `state.json`, so a crash mid-write leaves the visible file either complete or absent. `load_checkpoint` treats a torn/corrupt file (`ValueError`/`OSError`) as absent and cold-starts. `test_save_load_clear_roundtrip_is_atomic` asserts the `.tmp` is renamed away.
- Cadence: `resolve_cadence` honours kwarg > env > default, floors at 1, and a non-integer env value warns and falls back instead of crashing a multi-hour stage — a good operational choice.
- The single-pass column-group fingerprint is order-stable in the way that matters. Each named group gets its own SHA-256; every group's per-row contribution is `UNIT_SEP.join(str(col_value) for col in that group's tuple)` followed by `ROW_SEP`. Because columns are indexed by name out of a `to_pylist()` dict, the *physical* read order (`all_cols = sorted(...)`) cannot perturb any group's digest, and iteration is deterministic (row-group index ascending, then row index ascending). The unit/record separators prevent a value containing a delimiter from forging a field/row boundary. This is correct and collision-resistant for the intended purpose.

**Byte-identical fingerprint vs #269 — I verified this against the base-branch source, not the PR body.** The old inline `_compute_input_fingerprint` fed, per row, `UNIT_SEP.join(str(v) for v in _FINGERPRINT_CONTENT_COLS)` + `ROW_SEP` to the content hasher and `str(mention_id)` + `ROW_SEP` to the id hasher, iterating row-groups then rows. The new shared helper, invoked with groups `{"content": _FINGERPRINT_CONTENT_COLS, "mention_id": ("mention_id",)}`, produces the *same* byte stream for each: the content group's column tuple and order are unchanged, and the single-element `mention_id` group join yields the bare id string with no separator — identical to the old `str(...)`. Same iteration order, same separators, same encoding. So a checkpoint written by the currently-merged code resumes unchanged. The one behavioural difference is the empty-input case (old returns the SHA-256-of-empty digest; new returns `""` when `total == 0`), and it is unreachable in the resume path because `disambiguate.run` returns early on zero mentions before any fingerprint comparison. In-flight resume is preserved.

**dedup FAISS-phase resume.** The IVF-divergence argument holds: `_build_index` persists the index with `faiss.write_index`, and on a valid resume the loop calls `_reload_index` (`faiss.read_index`) instead of rebuilding, so the queried index is byte-for-byte the crashed run's and IVF's random kmeans init cannot diverge the resumed search. Resume is correctly gated on `index_available = index_path.exists()`, so a missing index cold-starts (and `_build_index` overwrites any stale index file). The block loop scores each query row independently and blocks run in ascending row order; combined with the canonical `pair_key` ordering and the restored `seen_pairs`, the emitted pair set and its order match a cold run — I traced the cross-block symmetric-neighbour case and it does not double-emit or drop. The fingerprint (`_id_set_hash` of the ordered ids, model name, index type, `actual_k`, rounded threshold, `n`) covers every input that changes which pairs are emitted, so an incompatible checkpoint is discarded rather than spliced. `test_resumed_collection_is_output_identical_to_cold` proves this end-to-end with a real mid-scan crash and asserts `resumed == cold`, and it wires `build_index` to `pytest.fail` on the resume path (and `reload_index` to `pytest.fail` on the cold/stale paths), so a regression that rebuilt-on-resume or reloaded-on-cold would fail the suite. The injected build/reload callables therefore cannot silently skew cold vs resumed.

**parallels anchor scan.** `indexed`/`postings` are rebuilt deterministically from `sorted(glob("**/hadiths_*.parquet"))` in file+row order, and `_candidate_partners` walks index-ascending postings, so the scan from `start_i` appends records in the same order a cold run would; the restored `records`/`cross_sect_count` preserve the prefix. The fingerprint globs the *same* hadith file set and hashes exactly `("source_id", "matn_ar", "matn_en", "sect")` — precisely the columns that feed tokenization, sect, and id — so any input change discards the checkpoint. Byte-identical resume.

**CLI + orchestrator threading.** `--no-resume` → `resume = not args.no_resume` → `_cmd_resolve(resume=...)` → `resolve_all(..., resume=resume)`, which passes `resume=` to `disambiguate.run`, `dedup.run`, and `parallels.run`. It is orthogonal to `--from-step` (which selects the start step). The ner/bio_promote exemptions are documented in-code with sound reasons (ner re-mints uuid4 mention_ids so a partial resume would splice two id-spaces — recovery is `--from-step disambiguate`; bio_promote is a sub-minute idempotent MERGE).

### Non-blocking observations (tech debt)

1. Every resumable stage snapshots its **whole** accumulated state as JSON on each cadence tick — disambiguate serializes the full `canonical_map`/`resolved_map`/`merge_log_rows`/`ambiguous_rows`/indexes; dedup its full `seen_pairs`+link lists; parallels its full `records` list. Write cost is O(accumulated state) per checkpoint, so at prod scale (3.3M mentions, hundreds of thousands of canonicals) each write is large and grows over the run. This is the design #269 established and this PR faithfully unifies (the cadence knobs bound the frequency), so it is not introduced here — but a future incremental/append checkpoint format would cut the steady-state write cost. Worth a tracking issue.
2. `hash_parquet_column_groups` silently `continue`s past a non-existent path and only returns empty digests when `total == 0`; a *partially* missing multi-shard input yields a valid-but-different digest (which safely discards the checkpoint, never a wrong resume). The docstring's "a missing/empty input yields empty-string digests" reads as if any missing file empties the digest — a minor doc imprecision, behaviour is safe.

Both are non-blocking. This is careful, well-tested, correctly-reasoned work; the byte-identity and cold==resume equivalence claims hold up under inspection.

---
RequestOrReplied: Approved
TechDebt: Every resumable stage writes its whole accumulated state as JSON per checkpoint tick (O(state) per write, grows over the run) — a future incremental/append checkpoint format would cut prod-scale write cost; also a minor docstring imprecision in hash_parquet_column_groups re partially-missing shard inputs.

parametrization:

Reviewed the artifact at head `3df78eb6bb84d142839cef13d946715e287c6ff4` via the GitHub API (no checkout — the live pipeline owns the main data-acquisition working tree). Read in full: `src/resolve/_checkpoint.py`, `disambiguate.py`, `dedup.py`, `parallels.py`, `ner.py`, `bio_promote.py`, `src/cli.py`, `src/resolve/__init__.py`, and all five affected test files. Also fetched the pre-#272 `disambiguate.py` from the base branch to verify the fingerprint-compat claim at the source.

### Resume == cold-run equivalence — genuinely proven, not "a file appears"
All three resumable stages have a real crash-simulation test asserting output equality between a cold run and a resumed run:
- disambiguate monkeypatches `_iter_mention_batches` to raise after 2 batches, asserts a partial checkpoint (`0 < processed < 12`) survived, resumes, and asserts all three artifacts (canonical, mentions, merge_log) equal the cold reference.
- dedup drives `_search_and_collect_resumable` with a numpy-only `CrashingIndex` that raises after 2 blocks, asserts the checkpoint landed at `processed_rows == 4`, then resumes and asserts `resumed == cold`. It also asserts the resume path calls `reload_index` (not `build_index`) via a `pytest.fail` sentinel — directly proving the IVF-divergence guard (reload the persisted index, never rebuild it).
- parallels raises inside `_candidate_partners` at anchor `i == 5` with per-anchor checkpointing, asserts a partial checkpoint survived, resumes, and asserts full row-and-order equality of `parallel_links.parquet`.

None are tautological: the fake index computes real inner products and argsort; the parallels/disambiguate tests use the real detector/resolver logic over tiny synthetic parquet and assert concrete pair ids, cross-sect flags, and canonical ids.

### Fingerprint mid-flight compat (the sharp risk given the active live run) — verified byte-identical
The claim that a checkpoint written by the currently-deployed pre-#272 code still resumes under the shared helper is correct. The old inline content hasher and the new `hash_parquet_column_groups` "content" group build each row identically (`\x1f`-joined stable columns, `\x1e` row terminator, same `_FINGERPRINT_CONTENT_COLS`); the "mention_id" group's single-element join reduces to the bare value exactly as the old `mid` hasher did; and because both index columns by name, the new helper's sorted physical read order does not perturb the digest. Even in the counterfactual where a digest differed, the failure mode is a clean cold-start (mismatch discards, never mis-reads) — but here it is a true resume. The `test_column_group_digest_matches_single_group` unit test locks the group-isolation property that this depends on.

### Robustness holes I probed
- Partially-written checkpoint: `save_checkpoint` writes `state.json.tmp` then `os.replace`, the only writer of `state.json`; `load_checkpoint` treats torn/corrupt/non-dict JSON as absent (cold start). Both covered (`test_load_corrupt_state_reads_as_absent`, `test_load_non_dict_json_is_none`, and the `.tmp` renamed-away assertions).
- Crash between checkpoint write and terminal output write: I traced this — output is written last, then `clear_checkpoint`, so a crash in that window leaves the checkpoint intact and the next resume idempotently regenerates identical output. For disambiguate this is why the fingerprint deliberately excludes the backfilled `canonical_narrator_id`/`confidence` columns: the end-of-run in-place backfill rewrites those, but a re-resume still fingerprints the unchanged content columns and reproduces the same outputs. Correct, if untested (see notes).
- Stale-fingerprint / different-input discard: covered for all three, plus disambiguate's distinct content-match / mention_id-mismatch (NER-rewrite) branch which cold-starts with the `--from-step disambiguate` hint.
- dedup resume requires `index_available` as well as a valid checkpoint; a missing persisted `.faiss` correctly forces a cold rebuild rather than querying a freshly-trained (randomly-initialised) IVF index against restored pairs.

### Uniform `--no-resume`
Threads correctly: `resume = not args.no_resume` → `run_all(resume=...)` → `.run(resume=...)` on disambiguate, dedup, parallels only. Each stage's `if not resume: clear_checkpoint(...)` both ignores and removes an existing checkpoint. `test_resume_flag_threads_to_resumable_stages` asserts `{disambiguate: False, dedup: False, parallels: False}`, and each stage has a `resume=False` test proving a sentinel-bearing checkpoint is ignored.

### ner / bio_promote exemptions — documented, not silently skipped
`ner.run`, `bio_promote.promote_bios_to_canonical`, and `run_all` all carry explicit docstrings stating the exemption and its reason (ner re-mints uuid4 mention_ids so a partial resume would splice two id-spaces — recovery is `--from-step disambiguate`; bio_promote is a sub-minute idempotent merge). This satisfies the issue's documentation requirement.

### CI truth — verified independently
I read `statusCheckRollup` at the head SHA myself: all 12 checks report SUCCESS (Lint, Type Check, Test, Integration Tests, Base-image/fixture, sync-drift gate, Markdown, cspell, staleness-check, lychee, YAML/JSON, actionlint). The suite reports 1429 passed / 7 skipped.

### Non-blocking notes
1. Whole-accumulator JSON snapshot per cadence: dedup and parallels serialise the entire accumulated state (`records` / `ids_a`,`ids_b`,`seen_pairs`) to JSON every cadence blocks, held fully in memory. This mirrors the established da#268 disambiguate pattern and the cadence bounds write frequency, but on a many-million-pair corpus the per-checkpoint rewrite cost and resident list grow with pairs found — a scaling watch-item for the follow-up (fuzzy_cluster is the heavier stage, PR2), worth considering append-oriented pair persistence there.
2. The terminal output parquet writes (`pq.write_table` for `parallel_links.parquet`, `write_parquet` for `narrators_canonical.parquet`) are not atomic tmp-then-replace, unlike the checkpoint and unlike disambiguate's mention backfill. A crash mid-write leaves a torn output, but the un-cleared checkpoint makes the next resume regenerate it, so there is no unrecoverable window; still, an atomic terminal write would match the discipline the rest of the module now follows.
3. Minor: the suite grew by 66 tests (1363 → 1429), not the "+~36" the PR body states; immaterial to correctness given the coverage above, just an inaccurate count.
4. The real `_build_index` / `_reload_index` faiss closures are exercised only through the injected-callable control-flow test, not against real faiss round-tripping — reasonable (the native dep is CI-integration territory), but the persisted-index reload is proven only at the logic level.

### Verdict
The resume mechanics are correct and, unusually, actually proven equivalent to cold runs via real crash simulation for every resumable stage; the mid-flight compat claim that matters for the active live run is verified byte-identical at the source, not just asserted; exemptions are documented as the issue required; `--no-resume` threads and clears correctly; and CI is green (verified). The notes are scaling/consistency observations, none blocking. Approving.

---
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: dedup and parallels persist the entire accumulator (records/ids/seen_pairs) as one JSON snapshot per cadence held fully in memory — O(accumulated pairs) rewrite + resident cost that is a scaling watch-item on a many-million-pair corpus; consider append-oriented pair persistence in the PR2/fuzzy_cluster follow-up.

Issue #270, opened 2026-07-03
fuzzy_cluster throughput degrades 3x as stage progresses (1,950 to 600 pairs/s)


parametrization:
## Problem

The wave-23 integrated resolve re-run (run 3, wave-branch code) shows a ~3x throughput degradation inside `fuzzy_cluster.cluster_records` as the stage progresses:

| Point in stage | scored | elapsed_seconds | interval rate |
|---|---|---|---|
| early (~7.6%) | 13,710,822 | 7,006 | ~1,957 pairs/s |
| ~29.6% → 30.7% | 53.2M → 55.2M | 37,254 → 40,633 | **~594–681 pairs/s** |

CPU is NOT the constraint — process CPU rose from ~625% to ~1,090% over the same window while the pair rate fell. Work per scored pair is getting more expensive, or the serial main-thread section is growing.

At the current rate the remaining ~124.5M of 179,759,721 candidate pairs take ~2 more days (vs the ~25h projection from the early-stage rate).

## Hypotheses to investigate (in `src/resolve/fuzzy_cluster.py`)

1. **Serial merge-apply cost growth** — `_apply()` / union-find bookkeeping on the main thread grows with accumulated merges (90,466 at 30.7%); check whether `_flatten(members)` or representative recomputation is called per-pair or per-merge on ever-larger clusters.
2. **Block-size ordering** — if blocks are processed in an order that leaves the largest token-blocks (the O(K^2) ones described in the module comments around lines 117–127) for later, the per-pair cost inside big blocks may differ; check dispatch order in `cluster_records` and whether worker batches are balanced.
3. **Worker starvation / drain behavior** — `_drain_completed(block_until_one=...)` may serialize as outstanding futures grow; check the submit/drain window logic.

## Deliverable

- A profile-backed diagnosis (reproduce on a sampled subset; do NOT attach to or perturb the live run, pid 20784 on the main checkout).
- A fix PR if the cause is algorithmic (e.g. centroid/representative-based comparison instead of all-member comparison, merge-apply batching, or block scheduling) — targeting `deployments/phase-8/wave-23`.
- If the fix changes merge semantics, quantify the effect on cluster assignments on the sample.

## Constraints

- Work in a worktree off `origin/deployments/phase-8/wave-23`. The main checkout is running the live resolve (pid 20784) — do not check out branches there, do not modify `data/` in place.
[...]

PR #277, opened 2026-07-03
perf(resolve): cap per-record match-keys + blocking-tokens in fuzzy_cluster (da#270)

parametrization:
## Diagnosis (profile-backed)

`fuzzy_cluster.cluster_records` throughput decayed ~3× over the wave-23 run (1,957 → 600 pairs/s) **while CPU rose** (625% → 1,090%). Rising CPU rules out the serial `_apply`/union-find and the drain logic (hypotheses 1 & 3) — a serial bottleneck starves workers and *drops* CPU. The growing cost is in the block-scoring **workers**.

Profiling my worktree's code (identical to the live run) on a read-only copy of the real `data/curated/narrators_canonical.parquet` snapshot (210,494 records) shows:

- **`process.cdist(..., token_set_ratio)` is ~99% of block wall-time** (the Python `_stateless_merge_ok`/`_token_order_consistent` guard-loop is ~1%). cdist is **O(K²)** in the block's flattened match-key count (`name + aliases` over every member), ~0.85 µs/cell.
- **A pathological tail inflates both K and the block count.** Match-keys/record: p99=19 but **max=740**. Significant-tokens/record: p99=34 but **max=1,213**. The tail is a mix of prolific narrators that accreted hundreds of real alias spellings *and* pollution nodes whose "name" is a captured isnad fragment (`[ 103 ] - ابراهيم بن الحسن...`, `ولد سنه 701 وكتب`, a single 141-token "name", most `mention_count=0`).
- A single 1,213-token record joins **C(1,213,2) ≈ 735k composite blocks** — that is the multi-GB pair-index build blowup — and blows up every block's `cdist` it sits in. These records are appended late (bio_promote rijal), so **later blocks carry the heavy K → the decay over the run**.

Worst-block measurement (250 members, alias-dense), capping match-keys/record:

| cap | K | cdist | speedup | within-block pairs kept (Jaccard vs uncapped) |
|----:|--:|------:|--------:|:--|
| none | 8,570 | 60.0 s | 1.00× | 1.000 |
| 64 | 4,799 | 17.5 s | **3.45×** | 0.894 |
| 32 | 3,828 | 11.4 s | 5.30× | 0.780 |

## Fix

Two per-record caps in `fuzzy_cluster.py`, both default **64**, applied so the scalar and cdist paths stay identical:

- **`_MAX_MATCH_KEYS_PER_RECORD`** — `_match_keys` truncates to the cap (name always kept first). Bounds each record's per-block `cdist` contribution. Touches only ~0.1% of records (>64 keys); p99 is 19.
- **`_MAX_BLOCKING_TOKENS_PER_RECORD`** — caps the significant tokens that form composite blocking pairs in `cluster_records`, so a giant-"name" record can't join hundreds of thousands of blocks. The precision guards still read the full token set.

Both `None`-disable to exact pre-da#270 behaviour (the equivalence / precision-recall harness path). This is **not a merge-semantics change for normal records** — the only lost merges are those justified *solely* by a record's 65th-plus alias spelling (the least-reliable tail; several of the affected records are `mc=0` pollution nodes, so dropping their spurious alias matches is arguably a precision gain).

## Effect on cluster assignments (quantified)

On the single most alias-dense real block (the hardest case), cap=64 preserves **89.4%** of the within-block merge pairs (Jaccard vs uncapped); typical blocks are far less alias-dense and lose proportionally less, and a merge dropped in one block is still recovered whenever the pair shares a rarer token pair scored in a smaller block. Every merge lost is one that depended solely on a record's 65th-plus alias.

The uncapped baseline is itself intractable at scale precisely because of the tail: on a density-preserving subset of just ~3k records the uncapped path runs for **>8 minutes** (and its build explodes outright once the >150-token pollution records are included), whereas the capped run finishes the same subset in seconds. That intractability is the wave-23 symptom this PR removes.

## Tests

New (`tests/test_resolve/test_fuzzy_cluster.py`): match-key cap truncation (name-first) + `None`-disable; an alias-bloated (300-alias) record still clustering with a genuine variant on its name under a tiny cap; the blocking-token cap trimming a record's block reach. Full non-integration suite **1367 passed, 7 skipped** (no regressions); `make check` (ruff + mypy strict + pytest) and pre-push hooks green.

## Coordination

Lands **before** da#272 (Alejandra's cluster checkpoint/resume). No public-API change to `cluster_records`/`cluster_canonical_narrators`; the scoring/merge loop shape is unchanged (the caps only bound inputs), so the checkpoint work targets the same structure.

Part of #270


parametrization:
## Review — PR #277 (da#270): cap per-record match-keys + blocking-tokens in fuzzy_cluster

Reviewed at head `06dfcd8` against base `deployments/phase-8/wave-23`, data-engineering lens with a specific focus on **recall semantics** (not speed). Read the full `fuzzy_cluster.py`, the new tests, and the memory note at HEAD; did not touch the live checkout.

### Diagnosis — I find it sound and well-evidenced

The reasoning is rigorous. Throughput decayed ~3× *while CPU rose* (625%→1090%), which correctly rules out a serial bottleneck (`_apply`/union-find/drain) — a serial stall would starve the workers and *drop* CPU — and points at the block-scoring workers. The profile attributes ~99% of block wall-time to `process.cdist(..., token_set_ratio)`, which is O(K²) in the block's flattened match-key count, and characterizes the tail concretely (match-keys/record p99=19 / max=740; significant-tokens/record p99=34 / max=1,213; example pollution "names" that are captured isnad fragments with mc=0). The single-record C(1,213,2)≈735k-block blow-up explains both the multi-GB pair-index build and the late-run decay (tail records are appended late by bio_promote). This is a real, correctly-attributed problem.

### Recall semantics — the core of my review

**What can each cap make us miss, and whom does it touch?** I confirmed in code that *neither cap affects a normal record*. `_match_keys` truncates only when a record exceeds 64 keys (p99=19), and the blocking-token cap trims only records with >64 significant tokens (p99=34). Two normal records always produce their full key/token sets, so their merge decision is byte-identical to pre-da#270. The recall boundary is confined to merges that involve a **>64-key/token tail record**.

**Match-key cap (64).** The name is always kept first (verified at `_match_keys` line 250-252, and by `test_alias_rich_record_still_clusters_on_name` with cap=4). Retained aliases follow in accretion order, *not* value-ranked — so the dropped keys are alias #65+. Because `token_set_ratio` is order- and extra-token-insensitive and the primary name always participates, the only lost merges are those justified *solely* by a 65th-plus alias spelling of a hyper-alias record. That is a defensible, low-risk boundary (<0.1% of records), and several affected records are mc=0 pollution nodes whose spurious alias matches we would rather not make — a precision gain, not just a recall cost.

**Blocking-token cap (64) — this is the sharper recall lever, and I want it on the record.** The retained blocking tokens are `sorted(tok_cache[i])[:64]` — i.e. the *alphabetically-first* 64, **not** the rarest/most-discriminative. A genuine variant of a >64-token record that shares only alphabetically-late tokens is never co-blocked, hence never scored, hence not merged. To the author's credit this is not hidden: `test_blocking_token_cap_bounds_reach` deliberately constructs exactly that case (shared tokens sort after the padding tokens) and asserts the capped run drops the merge. So the recall loss is understood and encoded. My concern is the *ordering choice*: for the small set of records this bites, alphabetical retention is arbitrary with respect to what a variant actually shares, whereas rarest-token-first (document-frequency / IDF ordering) would retain precisely the discriminative tokens a genuine variant keys on and preserve more recall at the same K. This does not make the current behavior wrong — the affected set is the pathological tail (pollution + densely-blocked famous narrators that almost certainly still share a common-core token-pair within the first 64) — but IDF-ordered retention is the right future refinement and I'd like it tracked.

**Is the output delta vs uncapped quantified on real data? Yes — I verified the evidence is in the PR, and I note its one limit.** The PR carries a real-data worst-block measurement (the single most alias-dense block: cap=64 retains 89.4% of within-block merge pairs by Jaccard vs uncapped) and the intractability numbers (uncapped runs >8 min on a ~3k density-preserving subset and the build explodes once >150-token pollution records are included). The intractability is itself the justification for why a *whole-corpus* uncapped-vs-capped A/B is not computable — the uncapped path does not finish at full scale, which is the wave-23 symptom being removed. That defense is legitimate. The gap I'll name honestly: "typical blocks lose proportionally less" is an *extrapolation* from the worst block, not a measured corpus-wide figure. A computable way to bound it without the intractable uncapped run would be a cap=64 vs cap=128/256 aggregate over the full 210k table (both tractable), reporting total merges/clusters and confirming the delta lands entirely in the tail. I'd like that as a follow-up, but given the worst-case bound is already measured and conservative, I don't consider its absence blocking.

### da#271 interaction (connector-token expansion) — flagged, and I assess it as complementary, not compounding

da#271 expands `_CONNECTOR_TOKENS` (currently 8 genealogical connectors). Both the cap and da#271 operate through `_significant_tokens`. Expanding the connector set makes `_significant_tokens` return *fewer* tokens per record, which means *fewer* records reach the >64-token blocking cap and each capped record's retained 64 shift toward more-discriminative tokens — so da#271 *shrinks* this cap's affected set and *improves* its recall, rather than compounding a loss. The one thing for da#271's author to re-verify after landing: `_significant_tokens` also backs the `_MIN_SHARED_TOKENS >= 2` precision guard, so dropping tokens there is an independent recall lever of da#271's own (a borderline pair could fall below 2 shared tokens) — but that is da#271's concern to measure, not a compounding effect with these caps. Net: I see no compounding recall risk; if anything they are mutually mitigating.

### Correctness details I confirmed

- Both caps `None`-disable to exact pre-da#270 behavior (the equivalence/precision-recall harness path) — verified at `_match_keys` line 259 and the blocking build line 555.
- The blocking-token cap trims only *reach* (which blocks a record joins); the precision guards still read the record's full post-match-key-cap token set via `tok_cache` — so a record fully scored inside a block it does join is guarded on its full tokens.
- Scalar and cdist paths stay identical (float64 matrix, same threshold cut) — the caps bound inputs, not the scoring shape, so this does not perturb da#272's checkpoint/resume target as claimed.
- Test suite is strong: name-first truncation, None-disable both caps, alias-bloated record still clustering on its name under a tiny cap, and the honest blocking-cap recall-boundary test. The pre-existing precision guards (death-year, gender, ≥2-shared-token, transitive re-partition) are untouched and still covered.

### Verdict rationale

The perf problem is severe (intractable at full scale) and correctly diagnosed; the fix is conservative, guarded, reversible (`None`-disables), touches only a well-characterized <1% pathological tail, never a normal record, and keeps the name-first primary key. Worst-case recall is measured (0.894) and the recall boundary is encoded as a test rather than hidden. The two residual items — alphabetical (not IDF) blocking-token retention, and the missing corpus-wide cap-vs-larger-cap aggregate delta — are recall-robustness refinements and evidence-strengthening, not correctness defects. Approving on that basis, with those two recorded for follow-up.

---
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: Blocking-token cap retains alphabetically-first tokens, not rarest/IDF-ranked — a genuine variant of a >64-token tail record sharing only late-sorting tokens is dropped (bounded to the <1% tail, honestly tested); IDF-ordered retention would preserve more recall at the same K. Also missing a corpus-wide cap=64 vs cap=128/256 aggregate merge-delta to bound typical-block recall loss (worst-block Jaccard is measured; typical is extrapolated). Recommend a tracking issue for both.

parametrization:
## Review — da#270 per-record match-key + blocking-token caps

Reviewed the full `fuzzy_cluster.py` change and the new tests at head 06dfcd8, read via the contents API (no checkout — the live resolve owns the primary tree). This is a well-diagnosed, evidence-backed fix. Approving; the one real concern below is a narrow, documentable tail case, not a blocker.

**Blocking correctness — what the caps can and cannot silently drop.** Both caps default 64, which sits above the profiled p99 for each quantity (match-keys p99 19, significant-tokens p99 34), so a normal record is never touched — that is the load-bearing safety property and it holds. Two things I verified beyond the diff hunks:

- The scalar (≤8-member) path and the cdist path both consume the same capped `_match_keys`, so there is no scalar-vs-cdist divergence — the equivalence property survives.
- `_merge_cluster` builds the survivor's output aliases from the raw `rec["aliases"]`, NOT from `_match_keys`, so the key-cap trims only the matching computation and never truncates a merged record's emitted alias set. No output data loss.

The blocking-token cap correctly trims only reach (which pairs enter `pair_index`); the merge guard `_shared_significant_token_count` still runs over the record's full (capped-key) token set, so a pair that co-blocks is never merged on a weakened guard. That asymmetry is the right design.

The genuine under-merge surface is the blocking-token selection: `sorted(tok_cache[i])[:N]` keeps the alphabetically-first N tokens, not a name-prioritised or rarest-first N. So a deep-tail record (>64 significant tokens after the key-cap) whose own name tokens sort alphabetically late can be blocked without its name tokens, and a genuine name-variant duplicate that shares only those late tokens loses its one composite block. This is narrow: it needs a >64-token record (well past p99 34), the pollution/`mc=0` members of that tail have no legitimate duplicate to lose (dropping their reach is a precision gain, as the PR argues), exact-name duplicates were already merged upstream in the disambiguate exact-name pass, and a variant with ≤64 tokens still co-blocks whenever it shares a pair inside the pathological record's kept 64. `test_alias_rich_record_still_clusters_on_name` shows the name-based merge survives when the name tokens land in the kept set; the residual is only when they do not. Worth a follow-up to prioritise name (or rarest) tokens in the blocking cap rather than alphabetical order — filing as tech debt, not gating.

**Interaction with the da#271 connector-token expansion (the reason I am reviewer 2).** They compose cleanly with no sequencing hazard beyond a textual merge on the constants block. `_significant_tokens` strips `_CONNECTOR_TOKENS` before either cap applies, so adding tokens (Latin `bin`, particles, benediction) to that set only removes low-signal tokens from `tok_cache` — the surviving capped-64 become strictly more discriminating, a mild improvement to this cap rather than a conflict. I will land my connector change as its own PR rebased onto this one once it merges (Ivana's option b), and note the interaction in that PR body.

**Threshold justification.** 64 is chosen deliberately above both p99s so it can never touch a normal record, and the worst-block Jaccard table (64 keeps 89.4% of within-block pairs at 3.45×; 32 keeps 78% at 5.3×) supports it as the precision/speed knob. Reusing the single value 64 for both caps is fine since each is comfortably above its own p99 (19 and 34). The diagnosis that rising CPU rules out the serial paths and isolates the cdist workers is sound, and the O(K²)/735k-block-fan-out story matches my own da#271 blocking analysis of the same 210k set.

CI 12/12 green. No public-API change, so Alejandra's checkpoint work composes on the unchanged loop shape.

---
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: blocking-token cap keeps sorted()[:N] alphabetically, not name/rarest-first, so a >64-token record whose name tokens sort late can lose its name-based block — narrow under-merge surface for the most alias-rich real narrators; prefer name-prioritised selection as a follow-up.

Issue #271, opened 2026-07-03
Quantify and shrink the 180M candidate-pair pool: block-size distribution + residual pollution audit


parametrization:
## Problem

The wave-23 resolve re-run reports `candidate_pairs_est=179,759,721` in `fuzzy_cluster`. Module comments in `src/resolve/fuzzy_cluster.py` reference historical pair volumes (~550M pre-guard) driven by a few enormous token-blocks — a single 28k-member block contributes O(K^2) pairs. After the da#247/da#253/da#258 NER fixes we expected the blocking pool to shrink substantially; 180M pairs may still be inflated by residual name pollution or by over-broad blocking keys.

If the pool is inflated, shrinking it attacks the multi-day cluster runtime at the source — independent of (and composing with) the scoring-loop fix in the sibling throughput issue.

## Task

1. **Quantify the block-size distribution** on the current staging/curated inputs (the disambiguate + bio_promote output from run 3 is on disk in `data/staging/` / `data/curated/` — read-only access is fine, do not write to those directories):
   - top-50 blocks by member count, their token keys, and their pair contribution
   - cumulative share: what fraction of the 180M pairs comes from the top 10 / 100 / 1000 blocks
2. **Inspect the biggest blocks for pollution**: are the members genuinely ambiguous short names (expected), or residual isnad-verb/relational-pronoun fragments, honorific-only names, or other name_quality escapes (fixable upstream)?
3. **Propose and (if clear-cut) implement** blocking improvements: stop-token expansion, block-size caps with fallback keys, or upstream name_quality tightening — PR targets `deployments/phase-8/wave-23`.

The new DuckDB exploration tooling (sibling issue, same assignee) is the intended instrument for step 1–2 — build it first, use it here.

## Constraints

- READ-ONLY on `data/` — the live resolve run (pid 20784) owns those directories.
- Work in a worktree off `origin/deployments/phase-8/wave-23`.
- Report findings on the issue before implementing anything upstream of NER — upstream name-pipeline changes need owner-visible sign-off given the #723 closeout state.

parametrization:
Requestor: Nikolaos.Papadopoulos
Requestee: Dilara.Erdogan
RequestOrReplied: Request

## da#271 findings — block-size distribution + residual pollution audit

[...]

PR #284, opened 2026-07-03
fix(name-quality): scrub Shia taṣliya + Prophet-title refs; bio_promote parity (da#271)

parametrization:
## Summary

Scrubs the benediction / Prophet-reference pollution surfaced by the da#271
pair-inflation audit (~16k canonical nodes that are eulogy phrases or references
to the Prophet, not narrators — concentrated in thaqalayn/lk mentions and the
kaggle rijāl bio table). This is fix (b) of the owner-approved da#271 tracks;
fix (a) (fawaz Arabic extraction) is a separate PR.

Part of #271

## Three gaps closed

1. **Bare Shia taṣliya** (`name_quality.py`) — the honorific set had the Sunni
   taṣliya (`صلى الله عليه وسلم`) and the `وسلم`-suffixed Shia form, but NOT the
   bare `صلى الله عليه واله`, so `رسول الله صلى الله عليه واله` (an **mc-365** false
   narrator) survived. Added it + its spaced variant. Once stripped, the residue is
   a bare Prophet title (`رسول الله` / `النبي` / `نبي الله`) — dropped by the new
   `_is_prophet_reference` (leader-anchored; the `+ الله` requirement keeps real
   names like `رسول بن جعفر`, `عبد الله`, `ابو بكر الصديق` untouched).
2. **Bidi/format marks** (`normalize_arabic`) — strip LRM/RLM, ZWNJ/ZWJ, BOM, the
   Arabic end-of-ayah mark, etc. **3,114** canonical names carried a stray mark that
   forked them from their clean twin and hid the taṣliya from phrase-matching.
3. **bio_promote parity** — `bio_promote` bypassed `clean_narrator_name` entirely,
   so honorific-laden bio names promoted into canonical **unfiltered** (the kaggle
   table is 100% such). It now runs the SAME filter the NER mention path uses:
   pollution dropped; a benediction suffix cleaned so the bio **merges onto** the
   clean mention-derived node instead of forking it.

## Before / after (representative)

| name (normalized) | before | after |
|---|---|---|
| `رسول الله صلى الله عليه واله` (mc 365) | kept as a narrator node | dropped |
| `النبي صلى الله عليه واله` (mc 116) | kept | dropped |
| `انس بن مالك رضي الله عنه` (bio) | forked node w/ eulogy | merges to `انس بن مالك` |
| `عبد الله`, `رسول بن جعفر`, `ابو بكر الصديق` | kept | kept (unchanged) |

## Tests

`test_name_quality.py` (taṣliya/title drop + real-name precision),
`test_arabic.py` (format-mark stripping + no-fork), `test_bio_promote.py`
(pollution drop + benediction-suffix merge). Full suite **1397 passed, 7 skipped**;
`ruff` + `mypy --strict` green.

## Residual (reported for follow-up, not in scope here)

The kaggle bio names are Latin+Urdu-script mixed
(`Prophet Muhammad(saw) ( محمد صلی اللہ … ( رضي الله عنه`). This filter removes the
Arabic-script benediction but the Latin/Urdu residue needs a kaggle-parser
Arabic-span extraction + Urdu→Arabic letter folding — a distinct parse-layer fix.


parametrization:
Second reviewer (integration lens); Ivana Horvat is first. Verdict: **approve**. 7/7 CI green against `deployments/phase-8/wave-23`. This is the upstream name-pipeline change da#271 flagged for owner-visible sign-off; the PR body records that owner approval.

I verified the two ordering/idempotency claims in code rather than trusting the writeup, and checked the blast radius of the `normalize_arabic` change.

**Ordering claim — verified.** The bare Shia taṣliya `صلى الله عليه واله` is added to `_HONORIFIC_PHRASES` *after* the longer `… واله وسلم` form, and `clean_narrator_name` strips phrases by iterating the tuple in order with sequential replacement. I walked the four cases (full Shia / bare Shia / Sunni / plain): longer-first ordering means the `وسلم`-suffixed form is consumed first where present, and no case leaves a dangling `وسلم` residue. Correct.

**Prophet-reference precision — sound heuristic, one corpus assumption worth naming.** `_is_prophet_reference` is leader-anchored: it drops a span only when it *leads* with a sole title (`النبي`/`الرسول`) or with `رسول`/`نبي` immediately followed by `الله`. The `+ الله` requirement is what preserves real names — I confirmed `رسول بن جعفر` (token 2 is `بن`, not `الله`), `عبد الله` (leader isn't in the set), and `ابو بكر الصديق` all survive. The one thing to be aware of: a personal name that is literally `نبي الله`/`رسول الله` (Nabiullah/Rasulullah as given names, which do occur in later South-Asian/Persian onomastics) would be dropped — but in the classical isnad corpora this targets (thaqalayn/lk), a leading `رسول الله`/`نبي الله` in narrator position is the Prophet, not a transmitter. Acceptable for this corpus; noting it so the assumption is on the record rather than filing it as actionable debt.

**(b) Idempotency — converges.** `bio_promote` now runs the same `clean_narrator_name` the NER path uses, then MERGEs by `make_canonical_id`. Re-running reads the raw bio source table again, re-cleans to the identical normalized form (the cleaner is idempotent — a benediction-stripped name has nothing left to strip), yields the same canonical id, and the MERGE is a no-op the second time. No double-scrub drift. The parity itself is the right fix — bio_promote bypassing the mention-path filter was a real asymmetry that forked eulogy-laden bios off their clean twins.

**(c) Downstream contract — broad but safe.** Folding `strip_format_marks` into `normalize_arabic` has pipeline-wide blast radius (every normalized string, not just names), so I looked at it specifically: it strips only zero-width/bidi/BOM/end-of-ayah control chars that carry no name meaning, it's idempotent, and it's placed first in the pipeline. For the 3,114 affected names this is the intended merge; for matn/parallels paths it's harmless-to-helpful. Worth stating plainly for the record: the merge only *realizes* on a full re-normalization pass (i.e. the wave-23 re-resolve this issue is scoped to), which is exactly the run in flight. `bio_promote` (resolve) importing `clean_narrator_name` (parse) follows the existing `ner.py` → `parse.name_quality` dependency direction, so no new layering violation.

**Residual, tracked.** The author's declared out-of-scope residual — kaggle bio names carrying Latin + Urdu-script benediction residue that the Arabic-only scrub can't reach (needs kaggle-parser Arabic-span extraction + Urdu→Arabic letter folding) — had no tracking issue, so I filed da#299 to capture it. Correctly out of scope for this PR.

---
Requestor: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: da#299

parametrization:
Reviewed at head 1b20007. Well-scoped and well-tested — approving.

Focus areas walked against the artifact at this head:

(a) Over-match risk — low and explicitly guarded. The taṣliya additions are four-token benediction phrases no real name carries. The Prophet-reference drop is leader-anchored: sole titles (al-nabi / al-rasul) or rasul/nabi immediately followed by the theophoric, so a name whose first token merely equals rasul/nabi without it is untouched. The precision tests cover exactly the tricky cases and pass — rasul-bin-Jaafar kept, abd-Allah kept, Jabir-bin-abd-Allah kept — and the taṣliya is stripped leaving the real leading name recovered rather than dropped. The guard errs toward precision, which is right.

(b) bio_promote parity / orphan risk — verified against the code. bio_promote now runs clean_narrator_name before make_canonical_id and uses the cleaned form for name_ar_normalized (the fuzzy match key), so a benediction-suffixed bio produces the SAME cleaned canonical id as the mention path and MERGES onto the clean node instead of forking. It cannot orphan an existing row: since NER already cleans and disambiguate overwrites the canonical table fresh each run, no stage keys canonical rows on unscrubbed forms — so the scrub removes forks/orphans rather than creating them. The bio_promote test proves the merge-onto-clean-id path and the pollution drop.

The normalize_arabic format-mark stripping (step 1) is a solid, low-risk normalization win — it only collapses names that were identical modulo a zero-width/bidi mark.

Two minor non-blocking observations filed as a follow-up (bio-only nodes keep the raw benediction in the display name; the Prophet-ref guard does not catch mid-string forms like Muhammad-rasul-Allah, which is the correct conservative call for now). Neither affects identity or clustering.

---
Requestor: Ivana Horvat
RequestOrReplied: Approved
TechDebt: #301

Issue #276, opened 2026-07-03
resolve --stop-after N: bounded partial-run testing on real data via the checkpoint layer


parametrization:
## Goal (owner-requested)

Use the unified checkpoint machinery (#272 / PR #275) for **partial-run testing**: a uniform flag that runs a resolve stage against the real data up to a bounded amount of work — a set number of checkpoints/batches — then stops cleanly. This lets a code change be tested for realistic behavior and performance (real inputs, real block-size distribution, real rates) without paying for a full multi-hour/day run.

## Deliverable

1. **Uniform CLI flag** on `resolve`, sibling of `--no-resume` and threaded the same way to every resumable stage: `--stop-after <N>` = stop the stage cleanly after its Nth checkpoint write.
   - Implemented once, in the shared `src/resolve/_checkpoint.py` cadence layer, not per-stage.
   - On stop: the checkpoint is **left on disk (not cleared)** so the partial run composes with resume — `--stop-after 5` then a later bare run continues from checkpoint 5. Stop + resume must equal an uninterrupted run on the covered prefix (test this).
   - The stage's final output is NOT written (the run is partial); the pipeline MUST halt after the stopped stage — downstream stages must never consume partial output. Distinct log line (`<stage>_stopped_at_limit checkpoints=N processed=...`) and a distinguishable exit (e.g. dedicated exit code or explicit status) so scripts/runbooks can tell "stopped at limit" from "completed" and from "crashed".
   - At stop, emit a compact perf summary: elapsed, items processed, rate (this is the number people run the flag to get).
2. **Runbook/docs section** — "Testing changes against a data subset": the workflow (`resolve --from-step <stage> --no-resume --stop-after <N>` for a cold bounded probe; baseline-vs-change rate comparison; when to trust head-of-data rates and when not to — the first N batches are a head-biased sample, e.g. fuzzy_cluster's early blocks score faster than late ones, da#270). Place it in the repo's pipeline docs and cross-link from the prod reload runbook's build-host section so operators find it.
3. Combinability rules documented + validated: `--from-step` + `--stop-after` + `--no-resume` all compose; `--stop-after` on a non-resumable/exempt stage (ner, bio_promote) is a clear error or no-op with a warning, not silent acceptance.

## Notes

- Count checkpoints (= batches at the stage's cadence) rather than rows: it maps 1:1 onto the machinery and is stage-appropriately sized. A volume/row-count variant can be layered later if needed.
[...]

PR #285, opened 2026-07-03
feat(resolve): resolve --stop-after N — bounded partial-run testing (da#276)

parametrization:
## Summary

Turns the unified checkpoint layer (#272 / PR #275) into a **partial-run testing instrument**: `resolve --stop-after N` runs a resumable stage against the real data for a bounded number of checkpoint writes, then stops cleanly — so a code change can be checked for behavior and throughput without paying for a full multi-hour/day run. Sibling of `--no-resume`, implemented once in the shared cadence layer.

## What's here

- **`src/resolve/_checkpoint.py`** — `CheckpointController` (counts checkpoint writes, enforces the `stop_after` budget; a drop-in for the old inline `batches_since_checkpoint` counter with zero behavior change when `stop_after is None`), `StopAfterReached`, and `EXIT_STOPPED_AT_LIMIT = 3`.
- **`StopAfterReached` subclasses `BaseException` on purpose** — the reasoning is documented in the class docstring: `run_all` wraps each stage in `except Exception` and logs-and-continues, so an `Exception` here would be swallowed as a "step failed" and the pipeline would march downstream onto a partial output (exactly what the spec forbids). As a `BaseException` it propagates through those guards and **halts the pipeline**, surfacing to the CLI which maps it to exit code 3. The docstring also records the clean-shutdown reasoning: `BaseException` still runs `finally`/`__exit__` on unwind, and stages raise only from a point where no worker pool is live (dedup's encode `ProcessPoolExecutor` is `with`-scoped and already exited before the checkpointed search/collect phase).
- **On stop**: checkpoint is **left on disk** (composes with resume), the stage's **final output is NOT written**, a `<stage>_stopped_at_limit` log + compact perf summary (elapsed / processed / rate) is emitted.
- Threaded through `run_all` → `disambiguate` / `dedup` / `parallels` `run(stop_after=)`.
- **CLI `--stop-after N`**: validates `N >= 1` and **rejects** pairing with `--from-step` on an exempt stage (ner / bio_promote / cluster / …) — a hard argparse error (exit 2), never a silent no-op. `--from-step` + `--no-resume` + `--stop-after` all compose.
- **Docs**: new `docs/testing-on-subsets.md` ("Testing changes against a data subset" — the cold bounded probe `resolve --from-step <stage> --no-resume --stop-after N`, baseline-vs-change rate comparison, and the head-biased-sample caveat incl. the fuzzy_cluster rate drift from da#270), cross-linked from the prod reload runbook's build-host precondition.

## Distinguishable exit status

`0` completed · `2` argparse/validation error · `1` uncaught crash · **`3` stopped at limit**.

## Test evidence

`ENVIRONMENT=test pytest` → **1454 passed, 7 skipped**. New/covered:
- `CheckpointController` + `StopAfterReached` units, incl. the **`BaseException`-not-`Exception` property** (the load-bearing guarantee that the orchestrator won't swallow it).
- **The exit-code contract itself**: stopped → 3 (`test_cli_maps_stop_to_distinct_exit_code`), a normal completion is NOT a stopped exit (`test_cli_completion_is_not_a_stopped_exit`), and `3 ∉ {0,1,2}`.
- **The must-have `stop-after N` then resume == uninterrupted** equivalence for both `parallels` (real fixture) and `dedup` (numpy-only fake index) — plus "stop leaves checkpoint + writes no output".
- CLI validation (exempt-stage + non-positive `N` rejected) and flag combinability threading through `run_all`.

`ruff check` + `ruff format --check` + `mypy --strict` + cspell + markdownlint + structural-ontology staleness gate all green. Pre-commit + pre-push hooks green.

> fuzzy_cluster gets `--stop-after` for free once its checkpoint lands (PR 2 of #272), since the mechanism lives in the shared helper.

## Related Issues

Part of #276

## Review Checklist
- [ ] Must-fix items resolved


parametrization:
## Review — PR #285 (da#276): `resolve --stop-after N` bounded partial-run testing

Reviewed at head `07f9f52` against base `deployments/phase-8/wave-23`, data-engineering lens. Read `_checkpoint.py` (the `CheckpointController` + `StopAfterReached`), the three stage integrations, `src/resolve/__init__.py`, `src/cli.py`, and `test_stop_after.py` at HEAD; did not touch the live checkout. All three focus areas hold up.

### (a) BaseException unwind — no stage raises while a worker pool is live

The `StopAfterReached(BaseException)` choice is justified and the "no live pool at the raise site" claim is true at every site, which I verified rather than took on the docstring's word:

- **All three raise sites are in serial loops.** dedup raises from the `for block_start in range(...)` search/collect loop (`index.search` runs in the main thread); disambiguate from `for batch in _iter_mention_batches`; parallels from `for i in range(start_i, len(indexed))`. None holds an executor.
- **dedup's encode pool is already closed.** The `ProcessPoolExecutor` lives `with`-scoped inside `_encode_with_resume`, which returns `embeddings` before `_search_and_collect_resumable` is ever called — so the pool is torn down before the checkpointed phase that can stop.
- **fuzzy_cluster's `ThreadPoolExecutor` is unreachable by the flag.** `RESUMABLE_STEPS = {disambiguate, dedup, parallels}` — `cluster` is not in it, so `--stop-after` never threads into the one stage that runs a live pool.
- **The propagation path is clean.** Every per-step guard in `run_all` is `except Exception`, so the `BaseException` sails through; `_cmd_resolve` catches it explicitly. I grepped the whole path (`dedup.py`, `parallels.py`, `disambiguate.py`, `__init__.py`, `cli.py`) for `except:` / `except BaseException` / `contextlib.suppress` and found **none** — nothing between the raise and the CLI can accidentally swallow it. `finally`/`__exit__` still run during a `BaseException` unwind, so `with`-scoped Parquet readers and the (already-closed) executor release normally.
- **The stop invariant holds by construction.** At each site the order is `batch_complete()` → `save_checkpoint(...)` → `checkpoint_written()` → `controller.stop(...)` (dedup 572/574/577; disambiguate 1347/1354/1357; parallels 336/343/346). The checkpoint is always on disk before the raise, and the final output is written only *after* the loop — so a stop always leaves a resumable checkpoint and no partial output, exactly as the spec requires.

### (b) Exit-code contract

`EXIT_STOPPED_AT_LIMIT = 3` is distinct from 0 (completed), 1 (uncaught crash) and 2 (argparse), with a comment stating that rationale, so a runbook can tell a bounded probe from a real completion or failure. `_cmd_resolve` maps `StopAfterReached` → `sys.exit(3)`. The argparse guard (cli.py 615–623) rejects `--stop-after < 1` and `--from-step <non-resumable> --stop-after` with `parser.error` (exit 2) *before* any work starts. `test_cli_rejects_invalid_stop_after` covers `--from-step ner`, `--from-step bio_promote`, and `--stop-after 0` all → exit 2, and `test_cli_maps_stop_to_distinct_exit_code` plus `test_cli_completion_is_not_a_stopped_exit` assert **both** sides of the contract (stop → 3, and a normal completion never takes 3).

One deliberate, documented UX constraint worth surfacing: requiring `from_step ∈ RESUMABLE_STEPS` forbids `--from-step bio_promote --stop-after N` even though dedup/parallels downstream could honor the budget. This is the right call — it makes "the bounded stage is the named stage" unambiguous, and the whole-pipeline probe (`--stop-after N` with no `--from-step`, which bounds the first resumable stage, disambiguate) is still allowed. Not debt; noting it so it reads as intentional.

### (c) stop-then-resume == uninterrupted — the tests assert output equality, not "it ran"

Both equivalence tests compare full outputs against a cold reference:
- `test_parallels_stop_then_resume_equals_uninterrupted` builds `cold_links` from an uninterrupted run, stops the probe after 2 checkpoints (asserting `StopAfterReached`), resumes bare, then asserts `_links(probe) == cold_links` (full row-list equality) **and** that the checkpoint dir is cleared on completion.
- `test_dedup_stop_then_resume_equals_uninterrupted` asserts `resumed == cold` over the full five-column-list tuple, with `processed_rows == 2` at the stop.
- `test_parallels_stop_leaves_checkpoint_and_no_output` independently pins the no-partial-output invariant (`parallel_links.parquet` absent after stop; checkpoint present with `processed_anchors == 2`).

### Correctness details I also confirmed

- `CheckpointController` is a true drop-in when `stop_after=None`: `checkpoint_written()` returns `False`, so cadence behaviour is byte-identical to the merged PR#275 inline counters. For parallels I checked the split condition `(i+1) % ANCHORS_PER_BLOCK == 0 and batch_complete()` reduces to the old `(i+1) % (cadence*ANCHORS_PER_BLOCK) == 0` (short-circuit means `batch_complete` only ticks on block boundaries) — same cadence, no behaviour change for non-probe runs.
- `checkpoints_written` counts *persisted* checkpoints (incremented after `save_checkpoint`), so `stop_after=N` means "stop after the Nth checkpoint is on disk" — the intuitive semantics.

### Non-blocking observation (tech debt)

The "`BaseException` propagates through `run_all`'s per-step `except Exception` and halts downstream stages" behaviour is verified here by code-reading + Python semantics + the negative grep, but not by an *integration* test — `test_cli_maps_stop_to_distinct_exit_code` monkeypatches `run_all` itself, so the real orchestrator's guard-passthrough (and the "no later stage runs / compose is skipped" halt) isn't exercised end-to-end. It's low-risk (`except Exception` cannot catch `BaseException` — that's language semantics, not logic), but a single test that drives the real `run_all` with a stopping stage and asserts the subsequent stages produced nothing would harden the contract against a future refactor that introduces a broad handler.

Clean, well-reasoned, and the load-bearing guarantees are tested at the right altitude. Approving.

---
RequestOrReplied: Approved
TechDebt: No integration test threads the real (un-mocked) run_all with a stopping stage to assert StopAfterReached propagates through its per-step except-Exception guards and halts downstream stages + skips compose; currently verified by code-reading + Python BaseException semantics + a negative grep, but an end-to-end test would guard against a future broad handler swallowing it.

parametrization:

Reviewed the artifact at head `07f9f525cdf95bacaab346017410add1108fe41b` via the GitHub API and diffs against the base (`deployments/phase-8/wave-23`) — no checkout, the live run owns the main working tree. Read `src/resolve/_checkpoint.py` in full plus the `cli.py` / `__init__.py` / `disambiguate.py` / `dedup.py` / `parallels.py` diffs, all three test files, and `docs/testing-on-subsets.md`. I have prior context here from reviewing #275 (the checkpoint layer this builds on).

### (a) Stop-then-resume == uninterrupted — proven at the output level for 2 of 3 stages
`test_parallels_stop_then_resume_equals_uninterrupted` runs a cold reference, then a bounded stop (`stop_after=2`, raises `StopAfterReached`) followed by a bare resume, and asserts the resulting `parallel_links.parquet` **equals** the cold run's rows and the checkpoint is cleared. `test_dedup_stop_then_resume_equals_uninterrupted` does the same at the column-list level via the numpy fake index. These are real output-equality assertions, not exit-code-only. `test_parallels_stop_leaves_checkpoint_and_no_output` separately pins the "checkpoint on disk, `processed_anchors==2`, and `parallel_links.parquet` does NOT exist" contract.

The one gap: **disambiguate** — the most stateful resumable stage (canonical map + the in-place mention backfill) — has no stop-then-resume output-equivalence test; only its threading (`test_stop_after_threads_to_resumable_stages`) and the controller units cover it. This is low-risk rather than a correctness hole: #275 already proved disambiguate resume==cold under a mid-loop crash, a `--stop-after` stop is just a controlled raise immediately after a checkpoint write (a strict subset of that crash-resume path), and its controller integration is structurally identical to parallels/dedup. Worth adding for completeness (see tech-debt), not blocking.

### (b) Exit-code contract — unambiguous and fully covered
`EXIT_STOPPED_AT_LIMIT = 3` is caught in `_cmd_resolve` (the only `try` around `run_all`) and mapped to `sys.exit(3)`; `main()` adds no wrapping layer. The tests pin every corner: stopped→3 (`test_cli_maps_stop_to_distinct_exit_code`), a normal completion never takes the stopped exit (`test_cli_completion_is_not_a_stopped_exit`), invalid `--stop-after` (exempt `--from-step`, or `N<1`) → argparse exit 2 (`test_cli_rejects_invalid_stop_after`), and `3 ∉ {0,1,2}` (`test_exit_code_distinct`). The load-bearing mechanism — `StopAfterReached` is a `BaseException`, so `run_all`'s per-stage `except Exception` guards cannot swallow it into a "step failed" continue — is asserted directly (`test_stop_after_reached_is_baseexception_not_exception`) and I confirmed those guards are unchanged in the `__init__.py` diff. A crash (an `Exception`) stays caught-and-continued by `run_all` and only escapes as the default `1` if it leaves `main` entirely, so `1` and `3` cannot be conflated. This is a genuinely clean automation contract.

### (c) Docs head-bias caveat — honest
`docs/testing-on-subsets.md` does not oversell the instrument. It states plainly that the first N batches are a **head-biased sample**, gives concrete failure modes (rate drift with position — the `fuzzy_cluster` da#270 degradation, and dedup's one-time FAISS-index cost amortised poorly at small N — and composition drift when the corpus is source-ordered), and draws the right line: trust a head rate for an A/B comparison of the identical prefix (the bias cancels) but do **not** extrapolate it to a full-run ETA. I can corroborate the composition-drift caveat independently: my da#279 corpora audit found the corpus is source-ordered (shards per corpus), so a small-N probe really would be dominated by one source — the caveat is not hypothetical.

### (d) Partial-output safety — no silent-trust path
Each stage calls `controller.stop()` immediately after `save_checkpoint(...)` and before its post-loop output write, so on stop the checkpoint is on disk and the final artifact (`parallel_links.parquet`, or disambiguate's canonical/merge-log/backfill) is never written — verified in all three diffs and asserted for parallels. The leftover checkpoint is fingerprint-guarded (from #272), so a cold or changed-input run discards rather than trusts it, and a bare resume completes correctly (the equivalence tests). dedup's stop does leave the `.faiss` index + embedding memmap on disk, but those are intermediate artifacts a resume reuses, not a consumed output. The only residual is cross-*invocation* (a stopped probe leaves any prior completed output in place, and a later separate `load` command would read it) — but that is outside one resolve run (the pipeline halts on stop) and is signalled by the distinct exit 3, so it is user responsibility, not a silent within-pipeline trust.

### Other verifications
- The parallels checkpoint refactor (`checkpoint_every_rows` → `(i+1) % _PARALLELS_ANCHORS_PER_BLOCK == 0 and controller.batch_complete()`) is behaviour-preserving: `batch_complete()` is short-circuit-called once per anchor-block and fires every `cadence` blocks, i.e. every `cadence * _PARALLELS_ANCHORS_PER_BLOCK` anchors — identical to the old modulo. The `stop_after=None` default makes `checkpoint_written()` always return `False`, so the controller is a true drop-in for the old inline counter.
- CI verified independently at the head SHA: all 12 checks SUCCESS.

### Non-blocking notes
1. Add a disambiguate stop-then-resume output-equivalence test to close the one coverage asymmetry (it is the most consequential stage; the equivalence is transitively covered but not directly asserted).
2. Minor doc-clarity: in a bare `resolve --stop-after N` (no `--from-step`), the budget applies **per resumable stage** — the first to reach N halts, but a disambiguate that completes with `<N` checkpoints then silently hands the same budget to dedup. The docs cover "bounds the first resumable stage"; a one-liner that the budget is per-stage (not whole-pipeline) would remove the last bit of surprise.

### Verdict
Well-designed and unusually well-tested for a control-flow feature: the `BaseException` decision that makes this safe against the orchestrator's swallow-and-continue is deliberate, documented, and asserted; the exit-code contract is airtight for automation; no partial output can be produced or silently trusted; and the docs are honest about the head-bias limitation. The two notes are completeness/clarity items, not correctness gaps. Approving.

---
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: disambiguate is the only resumable stage without a direct stop-then-resume output-equivalence test (parallels + dedup have one); the guarantee is transitively covered via #275's crash-resume test + the shared controller units, but a direct test would close the asymmetry on the most stateful stage.

PR #286, opened 2026-07-03
fix(resolve/ner): extract fawaz from Arabic full-text, not English (da#271)

parametrization:
## Summary

Fixes the da#271 cross-script under-merge: fawaz narrator names were romanized
("Anas bin Malik" — the flagship, mc 1152) because fawaz was routed through
English keyword extraction, so they never shared canonical identity with the
Arabic corpora and forked into duplicate nodes. This is fix (a) of the
owner-approved da#271 tracks; fix (b) (benediction/title scrub) is PR #284.

Part of #271

## The fix

fawaz ships an **empty** `isnad_raw_en`/`isnad_raw_ar` but a fully-populated
**voweled `full_text_ar`** (the extractor already falls back to full text when the
isnad column is empty). Moving fawaz from `_ENGLISH_SOURCES` to `_ARABIC_SOURCES`
routes it through the Arabic rule-based extractor — one line of source
membership change.

## Before / after (real fawaz data)

| hadith | BEFORE (English `full_text_en`) | AFTER (Arabic `full_text_ar`) |
|---|---|---|
| 0 | `Mughirah ibn Shu'bah` (1, Latin) | `عبد الله بن مسلمه بن قعنب القعنبي`, `عبد العزيز ابن محمد`, `محمد ابن عمرو`, `ابي سلمه`, `المغيره بن شعبه` (5, Arabic) |
| 1 | `Jabir ibn Abdullah` (1) | full 5-narrator Arabic chain incl. `جابر بن عبد الله` |

The English "Narrated X:" pattern surfaces only the lead companion (~1
narrator/hadith); the Arabic path recovers the **full chain** (~7.4/hadith) in
Arabic script that merges natively with every other corpus.

## Design decision (why this, not the two alternatives the issue floated)

- **Transliteration Latin→Arabic in `normalize_arabic`** — rejected: lossy and
  ambiguous (no short vowels, bin/ibn, apostrophes), high wrong-merge risk.
- **Cross-script match-key alignment in resolve** — rejected: a hack that leaves
  the polluted Latin display name and duplicate identity in place.
- **Use the Arabic source text already on disk** (chosen) — strictly better:
  correct identity + a more complete graph, zero transliteration risk.

**sunnah is deliberately NOT moved**: its `full_text_ar` is a samāʿ/reading-
certificate blob with biographical dates, not a clean isnad (Arabic extraction
returns garbage there), so it stays on the English path.

## Tradeoff (honest)

fawaz now shares thaqalayn's da#258 concern — extracting from full text
(isnad + matn) leaks some matn-boundary fragments on complex hadiths. Over a
1,500-hadith sample **~84%** of surviving mentions are clean; the remainder is the
known matn-boundary class `clean_narrator_name` (da#258) largely catches, and it
is mostly low-impact `mention_count=1` singletons. Net strongly positive: correct
cross-script identity + a ~7× more complete chain graph vs the old 1-Latin-name
path.

## Tests

Routing assertion; Arabic-not-Latin extraction; a **cross-script fold proof** (a
fawaz narrator and the same narrator from thaqalayn resolve to one canonical id).
Retargeted the output-schema test's English fixture from fawaz to sunnah. Full
suite **1395 passed, 7 skipped**; `ruff` + `mypy --strict` green.


parametrization:
Second reviewer (integration lens); Ivana Horvat is first. Verdict: **approve**. 7/7 CI green against `deployments/phase-8/wave-23`.

I read the change against the three lenses I was asked to cover, and verified the load-bearing correctness claim in the code rather than trusting the PR body.

**(a) Cross-source routing — clean.** The Arabic/English/skip/phase-1 split is expressed as four explicit module-level membership sets that `run()` iterates by name — this is exactly the "explicit per-source config, not scattered conditionals" shape we want, so the next source is routed by a one-line set-membership decision. Moving `fawaz` from the English set to the Arabic set is the whole functional change. I did find one latent gap that is **not** this PR's fault: nothing proves the four sets partition `SourceCorpus`, so a *future* source added to the enum but to none of the sets would be silently dropped from NER (no error, no warning). Filed as tech-debt da#296 with a proposed partition-guard test — non-blocking here.

**Correctness — verified.** The PR's core claim is that fawaz ships an empty `isnad_raw_ar` but a populated voweled `full_text_ar`, and the Arabic extractor falls back to full text when the isnad column is empty. I confirmed the fallback exists in `_extract_from_hadiths` (reads `isnad_raw_ar`, and on empty falls through to `full_text_ar`), so routing fawaz to the Arabic path does recover the full Arabic chain rather than returning nothing. The cross-script fold test (a fawaz narrator and the same thaqalayn narrator resolving to one canonical id) is the right test to prove the actual da#271 goal, not just that extraction runs.

**(b) Idempotency — no regression.** NER is deliberately non-checkpointed and re-mints `mention_id` per run (documented, da#272); its "resume" is whole-file reuse via `--from-step disambiguate`. This PR changes only which text column fawaz is read from — deterministic per run, same names/positions on re-extraction. No double-extraction or convergence concern introduced.

**(c) Downstream contract — no schema change.** `NARRATOR_MENTIONS_RESOLVED_SCHEMA` is untouched. What changes is fawaz's *data semantics* (Arabic-script names, ~7x more mentions/hadith) — which is the intended cross-script merge, and is what graph-load/validate should now see. Retargeting the output-schema test fixture from fawaz to sunnah was the correct call once fawaz is no longer an English source. The honest ~16% matn-boundary leak is the known da#258 class, mostly `mention_count=1` singletons that don't inflate blocks — acceptable, and net strongly positive.

---
Requestor: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: da#296

parametrization:
Reviewed at head 2271966. Clean, minimal, well-justified routing fix (fawaz moved from the English source set to the Arabic one) — approving.

Focus areas walked against the artifact at this head:

(a) Fold — the cross-script under-merge was that the English "Narrated X:" path romanized fawaz names so they never shared identity with the Arabic corpora. Routing fawaz to the Arabic full-text extractor removes that barrier; the new fold test shows a fawaz-extracted Arabic name and the same name from thaqalayn resolving to one canonical id (no Latin fork). The mechanism is sound. Note the fold test uses identical input text for both corpora so it proves the routing/normalize/identity chain rather than real cross-spelling alignment — that part rightly rides the existing normalize+fuzzy machinery, which is fine.

(b) Matn-leak tradeoff — I verified the load-bearing claim: the Arabic path (and its full_text_ar fallback) runs every extracted span through clean_narrator_name (ner.py, the span loop), the same da#247/da#258 filter thaqalayn already relies on, plus a fail-loud segmentation guard that skip-and-counts matn blobs per row. So the ~16% residual is genuinely filtered, not an unfiltered firehose, and it is the same accepted thaqalayn tradeoff — mostly mc=1 singletons that do not cluster. Not merge-blocking. I filed a follow-up to make the per-source clean-rate observable and to revisit a per-source gate if it degrades.

(c) sunnah scoping — correctly left on the English path; its full_text_ar is a reading-certificate blob, and keeping sunnah unchanged is zero-risk. The schema test retarget from fawaz to sunnah is the right consequence.

Nice work — the "use the Arabic text already on disk" choice over transliteration is the correct call.

---
Requestor: Ivana Horvat
RequestOrReplied: Approved
TechDebt: #300

Issue #279, opened 2026-07-03
Corpora-shape audit: exhaustive report on dedup + normalization assumptions vs the reality of each source corpus


parametrization:
## Goal (owner-requested, exploratory)

An **exhaustive report** examining the shape of each source corpus and how well our deduplication + normalization approach matches the reality of the original data — where our assumptions hold, where they silently don't, and what that costs us.

## Scope

Per corpus (sunnah.com, kaggle datasets, github repos, bukhari/muslim/abudawud CSVs, thaqalayn, hubeali/bihar, sanadset, fawaz, lk, itqan, rijal biographical sources — everything under `data/raw/` + its parser):

1. **Shape**: record counts, field coverage/nullness, script mix (Arabic vs Latin transliteration vs mixed), diacritics density, numbering schemes (collection-ref vs in-book ordinal, da#77), matn/isnad separation quality, structural quirks (markup leakage, whole-body-in-name-field like the thaqalayn case).
2. **Normalization reality-check**: what `normalize_arabic` (diacritics strip, alif/hamza/taa-marbuta folding) does and does NOT handle per corpus — the #271 audit already proved a systematic gap (Latin-transliteration never folds; 23k under-merged nodes incl. head narrators). Look for more classes of the same kind: orthographic variants per source, honorific/eulogy embedding, ordinal/numeric tokens in names.
3. **Dedup reality-check**, both layers:
   - narrator identity (exact-name dedup → fuzzy_cluster): are the match-key assumptions (name + aliases, token blocking) aligned with how each corpus actually spells/structures names?
   - hadith dedup (FAISS semantic): does cross-corpus matn duplication actually look the way the embedder assumes (same-language, same-orthography)? Sample verified true-duplicate pairs across corpora and check whether the pipeline catches them.
4. **Cross-corpus collision matrix**: which corpora overlap in content, and does our pipeline's dedup rate between each pair look plausible against a hand-verified sample?

## Method

- The DuckDB tooling (da#273, `make duck`) over staging + curated is the instrument; run-3 outputs on disk are the current ground truth. READ-ONLY on `data/` (live run).
- Hand-verify small samples where automated numbers make claims (per the count>=0 lesson: read the data back, don't assert existence).

## Deliverable

- A committed report (`docs/reports/` or wherever repo docs convention puts it) + a summary comment on this issue.
[...]

PR #294, opened 2026-07-03
docs(reports): corpora-shape audit — dedup + normalization vs source reality (da#279)

parametrization:
## Summary

Exploratory, owner-requested corpora-shape audit (da#279): an exhaustive read of the **run-3** staging + curated Parquet through the da#273 DuckDB tool (`make duck`), examining how well our dedup + normalization approach matches the reality of each source corpus. Report: `docs/reports/corpora-shape-audit-da279.md`.

Every count that makes a claim was re-checked by reading sample rows back (per the `count>=0`-masks-empty lesson). `data/` was read-only throughout (a live resolve run owns it); figures are a `2026-07-03` snapshot.

## The seven load-bearing findings

1. **Latin↔Arabic under-merge** — `normalize_arabic` is Arabic-script-only, so `Abu Huraira` / `Abu Hurairah` / the Arabic node never fold. 12,057 Latin canonical nodes / 65,143 mentions; the top Latin nodes are head Companions (`Abu Hurayra` alone strands 7,697 mentions off its 49,308-mention Arabic node).
2. **`fawaz` English-routed despite full Arabic isnad** — 36,100 hadiths carry `حدثنا`-form Arabic chains in `matn_ar`, but NER classifies `fawaz` as English-only → 33,943 Latin-only mentions. Largest feeder of #1 (overlaps da#271 fix(a)).
3. **Non-name context tokens in the dedup key** — 77,074 canonical nodes (37%) have ≥9 tokens but <2% of mentions; eulogies/verbs/citation-numbers survive `normalize_arabic` and fragment identities.
4. **`sunnah` `source_id` numbering collision** — 1,815 of 10,710 hadiths (17%) collapse under graph `MERGE`; one id (`sunnah:bulugh:2:2:0`) holds 461 genuinely distinct hadiths (da#77).
5. **`halimbahae` + `bihar` invisible to dedup** — text lives only in `full_text_ar`; both produce **0** parallel-link pairs (62,563 hadiths) and textless graph nodes.
6. **`thaqalayn` unsplit Arabic isnad** — no `isnad_raw_ar`, so Arabic NER over-captures from `full_text_ar` → largest fragment-node source (24,572).
7. **FAISS semantic dedup embeds English only** — ~738k hadiths lack `matn_en` (FAISS-invisible), and cross-sect pairs over-link on English boilerplate (256k `fawaz`↔`thaqalayn`).

The report’s § 9 has the full ranked mismatch list with **proposed** follow-ups (proposals only — no follow-up issues filed; scope is an owner call).

## Notes for reviewers

- Also adds a cspell Arabic-script ignore pattern (cspell has no Arabic dictionary → Arabic tokens are always false positives) + the audit’s domain terms to `project-words.txt`, so the report lints green. markdownlint / cspell (v8.4.0 mirror) / JSON checks pass locally.
- Docs-only + lint-config; no code, no `data/` writes.

Part of #279


parametrization:
Second reviewer, verified at head 542a405. Approving — this is a careful, honestly-caveated audit and I checked the NER/dedup-facing findings against the actual code, not just the prose.

Findings in my lens:

(2) fawaz Arabic-isnad routed English — correctly handled. The report does NOT propose a redundant follow-up; §9 item 2 explicitly cross-references it as "already scoped as da#271 fix(a)" (the merged/landed fix I first-reviewed). One optional doc nit, non-blocking: §2/§4 locate fawaz's Arabic chain in matn_ar, while the landed fix extracts from full_text_ar via the isnad-empty fallback. For fawaz both columns carry the chain (your own §2 table shows both ~1.1% empty), so the outcome is identical and #286 fully covers this finding — but a one-line note that the fix keys off full_text_ar would spare a future reader the matn_ar-vs-full_text_ar double-take. Fold in if convenient; not required.

(4)/(5) halimbahae + bihar → 0 parallel pairs, invisible to both dedup layers — verified correct against source. dedup embeds matn_en only (dedup._load_hadith_texts); the lexical detector reads only matn_ar/matn_en (parallels._load_hadith_rows loads columns source_id/matn_ar/matn_en/sect, and _tokenize uses matn_ar-if-Arabic else matn_en) — neither reads full_text_ar. Since both corpora are matn_ar/matn_en-empty and full_text_ar-only, they tokenize to nothing → zero fingerprints → zero pairs. The claim is exactly right, and your direct 0-pair verification matches the code path.

(6) thaqalayn unsplit-isnad fragment pollution (24,572 >=9-token nodes) — consistent with what I know. The 33,190-hadith figure matches the da#175 thaqalayn number in memory; the root cause (isnad_raw_ar 100% empty → Arabic NER segments over full_text_ar = isnad+matn → over-capture) is precisely the da#258 mechanism, and the "overwhelmingly mention-light mc<=1 tail" framing matches the da#247 singleton-pollution finding and my own #286 review note. No inconsistency.

cspell Arabic-ignore pattern — sane, does not over-suppress Latin typos. The added regex matches only contiguous runs in the Arabic Unicode blocks (0600-06FF / 0750-077F / FB50-FDFF / FE70-FEFF); it never matches Latin letters, so an English/transliteration typo (no Arabic chars) is still fully checked. A mixed token only has its Arabic run stripped, leaving the Latin part checkable. Correctly scoped.

Solid work — the read-back-every-count discipline and the soft-upper-bound framing on the cross-sect pairs are exactly right.

---
Requestor: Ivana Horvat
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review — PR #294 (da#279): corpora-shape audit report

Reviewed at head `542a405` against base `deployments/phase-8/wave-23`, data-quality lens focused on the load-bearing counts and their method. Docs-only (the report + a cspell Arabic-ignore config). I could not re-run the duck queries myself (the live resolve run owns `data/`, and the report is an explicit point-in-time snapshot), so I verified the counts two other ways: internal arithmetic cross-consistency, and whether every claimed *mechanism* matches the actual code at HEAD. Both hold up strongly.

### The counts cross-foot exactly — strong evidence they are real reads

I re-added every table against its independently-stated totals, and all of them reconcile to the digit:

- **Per-corpus hadith table (§2)** sums to **853,218** — the stated `staging_hadiths` total, exactly.
- **Token-count distribution (§6)** sums to **210,494** canonical nodes *and* its mentions column sums to **3,118,034** — both independently-stated totals in §5, exactly, across a completely different aggregation.
- **Sect-affiliation breakdown (§5)** sums to 210,494 nodes / 3,118,034 mentions, exactly.
- **Parallel-tier counts (§7)** sum to **6,656,888**, exactly.
- Derived figures check: the 62,563 halimbahae+bihar (= 62,169 + 394), the ~738k no-`matn_en` set (sanadset+halimbahae+tusi+mis+bihar = 738,718), the sunnah collision (10,710 − 8,895 = 1,815 = 17%), and the 5.7% Latin share (12,057 / 210,494) all reconcile.

Fabricated numbers do not cross-foot across three orthogonal breakdowns simultaneously; this pattern is what a genuine read-back-and-verify pass produces, and it is consistent with the `count>=0`-masks-empty discipline the report invokes.

### The mechanisms are code-accurate (I checked the source, not just the prose)

- **Finding #2 (fawaz English-routed).** Confirmed `_ENGLISH_SOURCES = {"fawaz", "sunnah"}` in `ner.py`, and that the English path reads `isnad_raw_en` then falls back to `full_text_en` (lines 123–124) — never `matn_ar`. Since fawaz's `isnad_raw_en` is 100% empty (§2), the Arabic `حدثنا`-form chain in `matn_ar` is genuinely never read. The report names the constant and the fallback chain correctly.
- **Finding #7 (FAISS English-only).** Confirmed `dedup._load_hadith_texts` returns texts "with null/empty `matn_en` rows excluded" — so the ~738k Arabic-only hadiths are FAISS-invisible exactly as claimed.
- **Findings #1/#4 (normalize_arabic Arabic-only).** Confirmed `normalize_arabic` is the six-step Arabic-script pipeline with no cross-script folding. One useful detail the report doesn't mention: `arabic.py` **already ships a `transliterate()` (Arabic→Latin) function** — a latent building block for the §9 item-1(b) "transliteration-aware fold" proposal (an existing Arabic→Latin map means the fix could compare an Arabic node's transliteration against the Latin nodes rather than build romanization from scratch). Worth noting in the follow-up so it isn't scoped as greenfield.

### cspell config is correctly scoped

The Arabic ignore regex targets the Arabic Unicode blocks specifically (`U+0600–06FF`, Supplement, Presentation Forms A/B) — **not** a blanket non-ASCII ignore — so it suppresses the unavoidable Arabic false-positives without masking Latin typos. Files are scoped to `docs/**`/CLAUDE/RUNBOOK/charter, which covers the new report. Reasonable and low-risk.

### §9 follow-up mapping — one real overlap to fix, otherwise sound

The mismatch→follow-up mapping is coherent and item 2 correctly attributes itself to **da#271 fix(a)** (fawaz→Arabic extraction). One correction:

- **Item 5 (strip eulogies/verbs/citation-numbers from the name key) overlaps the already-completed da#271 fix(b)** (benediction/title `name_quality` scrub). The report presents item 5 as a fresh proposal without referencing fix(b), but "strip trailing eulogies (`رضى الله عنه`)" is precisely fix(b)'s domain. The report should cross-reference fix(b) and **narrow item 5 to the residual** — transmission verbs (`قال`/`عن`/`حدثنا`) and citation-index numbers — so the owner doesn't double-scope it when converting proposals to issues. This is the one actionable doc fix; it doesn't affect any count.
- Item 7 (cross-sect FAISS precision) is **orthogonal to my da#287 partition schema** — it's an embedding/threshold issue, not a partition-metadata one — so no conflict there, contrary to a possible first read.
- The §8 kaggle provenance note ("no upstream provenance recorded") could cross-reference the kaggle tracking issue (#299) for the reader, but that's descriptive, not a §9 item.

### Composition drift (§8) vs da#280 — consistent, and a useful cross-link

The run-3 "full corpora, six-books duplication re-introduced" finding (fawaz↔lk 787k pairs; June C-dedup absent) is consistent with what I found in **da#280**: the six books and Musnad Ahmad are multi-sourced (I traced Musnad Ahmad via the sunnah scraper; §8 notes the June composition intended halimbahae to be the Musnad Ahmad/Darimi/Malik source, trimmed). No contradiction — complementary. The load-bearing cross-link for the team: §9 item 8 (composition regression) directly feeds **da#280's P1b** (verify staging presence / which copy is authoritative), because partition-tagging a corpus that is duplicated across four shards needs the composition decision resolved first. Worth wiring those two together.

### Minor precision nits (non-blocking)

- The §2 "the ~2,070 gap is almost entirely sunnah" is 1,815/2,070 = 88% — "almost entirely" is slightly strong (≈255 rows are non-sunnah).
- §5's script split is 93.5% Arabic + 5.7% Latin = 99.2%; the residual ~0.8% (mixed-script/empty-name nodes) is unstated — a one-line footnote would close it.
- Finding #4's "1,815 hadiths share an id" is precisely the collapse *surplus* (rows − distinct); the number of hadiths *involved* in a collision is larger, though the impact figure (1,815 lost under MERGE) is the correct one.

None of these change a conclusion. This is a careful, verifiable audit — the counts are internally exact and the mechanisms match the code. Approving; the only thing I'd like folded in is the item-5 / da#271-fix(b) cross-reference so the follow-up scope is clean.

---
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: §9 item 5 (name-key eulogy/verb/number scrub) overlaps the already-completed da#271 fix(b) benediction/title scrub but doesn't reference it — should cross-link and narrow item 5 to the residual (transmission verbs + citation numbers) to avoid double-scoping; also note arabic.py's existing transliterate() as the building block for item 1(b), and wire §9 item 8 (composition regression) to da#280 P1b.

PR #302, opened 2026-07-03
feat(resolve): crash-resumable fuzzy_cluster block-scoring pass (da#272 PR2)

parametrization:
## Summary

Extends the unified checkpoint layer (#272 / #275, `--stop-after` #285) to **fuzzy_cluster** — the multi-day block-scoring pass and the highest-priority stage in #272. This is the last gate before run 4; the pipeline's longest stage becomes crash-resumable on its first real exercise. Rebased onto the current (da#270-capped) `cluster_records` shape.

## What's checkpointed

In `cluster_records` (gated on an opt-in `ckpt_dir` so the pure callers — `cluster_assignment` and the quality harness — pass `None` and are untouched):
- the union-find parent array,
- the set of applied block indices (the resume skip-set),
- the `scored` / `merged` counters.

`cluster` joins `RESUMABLE_STEPS`, so it also inherits `--no-resume` and `--stop-after` for free.

## Why this resume is correct (and more robust than the streaming stages)

The final clusters are **connected components**, which are **invariant to union order**, and `uf.union` is **idempotent**. So a resume that re-applies only the not-yet-applied blocks yields the same components as an uninterrupted run — regardless of order or any incidental re-work. Cadence is keyed to **scored-pair progress**, not block count (block cost varies by orders of magnitude — an 8-member scalar block vs a 250-member cdist block). The input fingerprint covers the clustering-driver columns + threshold + max_block_size; the da#270 caps truncate deterministically from those same columns, so the fingerprint still fully determines the (capped) output.

Threaded through `cluster_canonical_narrators(staging_dir=, resume=, stop_after=)` and `run_all`; checkpoint cleared on success after the canonical table is rewritten; `StopAfterReached` is raised **before** the rewrite so a bounded probe leaves the input untouched.

**Correction:** #275's docstrings/memory mislabeled `cluster` as a sub-minute exempt stage — fixed here (it's the multi-day one). `ner` + `bio_promote` + the date stages remain exempt.

## Concurrency coverage (the path the real run uses)

The tiny-fixture blocks otherwise take the scalar inline path; a dedicated test forces the **concurrent cdist / `ThreadPoolExecutor`** path (block_idx tracked through the `pending` futures dict, applied in thread-completion order) and proves stop-then-resume == uninterrupted through it. Note the corrected `StopAfterReached` docstring: fuzzy_cluster raises with its `ThreadPoolExecutor` **live**, but `with`/`__exit__` runs `shutdown(wait=True)` on the `BaseException` unwind and joins the read-only workers cleanly — the honest version of the old "no worker pool is live" claim (which Kavitha verified the shape of on #285).

## Folded-in #285 verdict follow-ups

- **Kavitha:** `test_stop_after_reached_propagates_out_of_run_all_and_halts` — a stage raising `StopAfterReached` propagates OUT of `run_all` (not caught by the per-step `except Exception`) and no later stage runs.
- **Oyunbileg (a):** `test_stop_after_then_resume_is_output_identical_to_cold` (disambiguate) — closes the streaming-stage coverage asymmetry.
- **Oyunbileg (b):** doc note that the `--stop-after` budget is per-resumable-stage.

## Test evidence

Cluster suite (Ivana's precision/equivalence + the new resume tests) 28 passed; full suite green pre-push. New tests: stop-then-resume == uninterrupted (scalar AND pooled paths), stop leaves checkpoint + no output rewrite, stale-fingerprint discard, `--no-resume`, pure-path writes no checkpoint. `ruff` + `mypy --strict` + cspell + markdownlint + structural-ontology staleness all green.

## Related Issues

Part of #272

## Review Checklist
- [ ] Must-fix items resolved


parametrization:
## Review — PR #302 (da#272 PR2): fuzzy_cluster crash-resume (run-4 gate)

Reviewed at head `e864d73` against base `deployments/phase-8/wave-23`, checkpoint-contract lens, continuing the #275/#285 thread. Read the full `cluster_records` integration + raise site, the fingerprint/clear in `cluster_canonical_narrators`, the corrected `StopAfterReached` docstring, `RESUMABLE_STEPS`, and all three test files at HEAD. Did not touch the live checkout. All four of my focus areas hold; approving as the run-4 gate.

### (a) The run_all integration test I asked for on #285 — present and real

`test_stop_after_reached_propagates_out_of_run_all_and_halts` drives the **real** `run_all` (`run_all(tmp_path / "raw", staging, output, stop_after=1)`) — only `disambiguate.run` is monkeypatched to raise `StopAfterReached`, which is the correct way to test propagation (you need a stage to raise). It asserts the exception propagates *out* of the real orchestrator (so the per-step `except Exception` guards genuinely don't catch it) **and** that none of `bio_promote`/`cluster`/`dedup`/`parallels` ran afterward. That closes the exact TechDebt item I left on #285 — the middle layer is now exercised end-to-end, not just by code-reading.

### (b) The corrected "no worker pool live" claim — accurate at the cluster raise site

I traced the raise. `controller.stop()` is called from inside `_apply` (line 766), and `_apply` runs from both the scalar path (807) and `_drain_completed` (786) — **both inside the `with ThreadPoolExecutor(...) as pool:` block** (788). So unlike disambiguate/dedup/parallels, fuzzy_cluster genuinely raises with its pool **live**. The corrected docstring is honest about this: `__exit__` runs `shutdown(wait=True)` on the `BaseException` unwind and joins the workers. I confirmed this is safe because the workers are read-only — `_block_merge_pairs` computes a cdist score matrix and returns a pure `(i,j)` pair list; `uf`/`applied`/`scored` are mutated **only** in the main thread via `_apply` — so there is no half-updated shared state at the raise, no lock a worker could strand, and the in-flight futures' results are simply discarded and re-run on resume (they were never in `applied`). The docstring's "may briefly block on the join before exiting" is the right honest caveat. The old stale claim is properly replaced across the class docstring.

### (c) Checkpoint-contract conformance with the #275 conventions

- **Atomic write / cleared-on-success / fingerprint / --no-resume** are all the shared primitives: `save_checkpoint` (atomic tmp+replace), `clear_checkpoint` after `write_parquet` rewrites the canonical table (1112–1113), `hash_parquet_column_groups` + `hash_strings` for the fingerprint, and `if not resume: clear_checkpoint` (675–676). `cluster` is now in `RESUMABLE_STEPS = {disambiguate, cluster, dedup, parallels}`, so `--no-resume`/`--stop-after`/`--from-step cluster` all work through the same CLI guard I reviewed on #285.
- **Fingerprint is complete.** It covers `_FINGERPRINT_CANONICAL_COLS` (canonical_id, name_ar, name_en, name_ar_normalized, aliases, death_year_ah, gender) + `round(threshold,6)` + `max_block_size` + `len(records)` — i.e. every clustering driver (match keys + the death-year/gender guards) plus the parameters and the row count that fix the positional block indices. Since the da#270 caps truncate deterministically from those same columns, the fingerprint still fully determines the capped output, and `parent_ok = len(parent)==n` guards the array length — so a resume only proceeds against a byte-identical input where the `applied` skip-set's positional indices are still valid.
- **Cadence keying is consistent with the shared resolver.** This was your specific concern, and it's sound. The scored-pair interval (`_CLUSTER_CHECKPOINT_SCORED_INTERVAL = 50M`) is the stage's natural "batch unit" (block count is a terrible unit here — an 8-member scalar block vs a 250-member cdist block differ by orders of magnitude), and `controller.batch_complete()` ticks once per interval, so the effective cadence is `resolve_cadence(...) × 50M` pairs — structurally identical to how parallels gates `controller.batch_complete()` behind an anchor-block boundary. `stop_after` still counts `checkpoint_written()` calls (persisted checkpoints), same semantics as every other stage. No double-count, no inconsistency.
- **Snapshot is consistent, and the correctness argument holds.** `_apply` does the unions, then `scored +=`, then `applied.add(block_idx)`, then (maybe) checkpoints — so `_snapshot()` captures a `(parent, applied, scored)` triple where `parent` reflects exactly the unions of the blocks in `applied`. On resume, `block_idx in applied` blocks are skipped (790) and the rest re-scored. Because final clusters are connected components (union-order-invariant) and `uf.union` is idempotent, the resumed components equal a cold run — and restoring the raw `_parent` array is valid since path compression is a representation optimization that never changes the partition. `StopAfterReached` raises inside `cluster_records` *before* `cluster_canonical_narrators` reaches `write_parquet`, so a bounded probe leaves the input canonical table untouched — asserted directly by `test_stop_leaves_checkpoint_and_does_not_rewrite_output` (canonical still has 8 rows after the stop).

### (d) Oyunbileg's disambiguate stop-resume test — asserts output equality

`test_resumed_run_is_output_identical_to_cold` crashes mid-run (raising after 2 batches), verifies a partial checkpoint survived (`0 < processed < total`), resumes, and asserts `resumed["canonical"] == cold["canonical"]`, `["mentions"] == `, and `["merge_log"] == ` — full equality across all three outputs, plus checkpoint-cleared. The `test_stop_after_then_resume_is_output_identical_to_cold` companion covers the `--stop-after` path. Genuine output equality, not "it ran."

### Cluster resume tests — the pooled path (the riskiest code) is proven equivalent

`test_stop_then_resume_equals_uninterrupted_pooled_path` forces every `m>=2` block through the `ThreadPoolExecutor`/cdist path via `_SCALAR_BLOCK_MAX=1`, confirms the pooled cold run finds the same 4 merges, stops after 2 checkpoints with a genuine partial (`0 < len(applied_blocks) < 4`), resumes, and asserts `_output_rows(probe) == cold_rows`. The suite also covers the scalar path equivalence, stale-fingerprint discard (a planted checkpoint that would collapse everything to one cluster is correctly rejected), `--no-resume`, and that the pure `cluster_records` caller (ckpt_dir=None) never writes a checkpoint. Thorough and honest.

### Non-blocking observation (tech debt)

The checkpoint reaches into `UnionFind._parent` directly to snapshot (712) and restore (684). It's correct and same-module, but a `UnionFind.snapshot()/restore()` pair would encapsulate the one piece of resumable state instead of touching a "private" array from the outer scope — worth a small cleanup so a future change to the union-find representation can't silently desync the checkpoint format.

Clean, correct, and the hard case (raise with a live pool) is handled honestly and proven by the pooled-path equivalence test. No blockers for run-4.

---
RequestOrReplied: Approved
TechDebt: Checkpoint snapshot/restore reaches into UnionFind._parent directly (private-attribute access from the cluster_records scope) — a UnionFind.snapshot()/restore() pair would encapsulate the resumable state so a future change to the union-find representation can't desync the checkpoint format.

parametrization:
First-review outcome (posting post-merge — the PR landed at e864d73 while I was reviewing; recording for the record, no gating action).

I verified the three areas the design hinges on and the resume architecture is sound:
- (a) the da#270 caps truncate BEFORE block generation, so block indices are defined over the capped block universe — correct ordering.
- (c) the resume==cold argument (final clusters are connected components, invariant to union order; uf.union idempotent) is sound, and it is genuinely exercised through the CONCURRENT cdist/futures path — test_stop_then_resume_equals_uninterrupted_pooled_path monkeypatches _SCALAR_BLOCK_MAX=1 to force every block through the pool, asserts a partial checkpoint survived, and asserts output equivalence. The output-equivalence assertions compare the merged canonical rows (not the order-sensitive merged counter), which is the right invariant.

One real gap I would have raised as a must-fix pre-merge, now filed as a fast-follow in #303: the resume fingerprint pins driver columns + threshold + max_block_size but NOT the two da#270 cap constants, which also determine the block universe. A resume taken across a cap-value change would silently mix block universes (the applied-block index skip-set misaligns). It does NOT affect run-4 launch (caps are fixed at this commit), so this is not an emergency — but it should land before any cap-tuning/sweep + resume. Fix is ~2 lines (add the two constants to the hash_strings call, mirroring threshold/max_block_size) + a discard test. Details and the exact patch are in #303.

Nice work overall — the connected-component framing makes this resume genuinely more robust than the streaming stages.

---
Requestor: Ivana Horvat
RequestOrReplied: Reply
TechDebt: #303

Issue #303, opened 2026-07-03
fuzzy_cluster resume fingerprint omits da#270 cap values → silent block-universe mix on cap-change resume (da#272 follow-up)


parametrization:
Post-merge finding from first-review of PR #302 (fuzzy_cluster crash-resume, da#272 PR2, merged @ e864d73). **Not run-4-launch-blocking** — filing as a fast-follow.

## The gap

`cluster_canonical_narrators` builds the resume fingerprint as:

    fingerprint = hash_strings(digests["content"], round(threshold, 6), max_block_size, len(records))

It pins the driver columns, threshold, max_block_size, and record count — but NOT the two da#270 caps `_MAX_MATCH_KEYS_PER_RECORD` and `_MAX_BLOCKING_TOKENS_PER_RECORD`. Both determine the **block universe**: `_match_keys` (match-key cap) → `keys_cache` → `tok_cache`, and the blocking-token cap → `pair_index` → the `blocks` list membership, ordering, and count.

The resume skip-set is a set of **positional block indices** (`applied`, indexed over `enumerate(blocks)`). So if a cap value changes between the checkpointed run and the resume, the `blocks` list changes but the fingerprint still matches → the checkpoint is accepted → the restored `applied` indices point at *different* blocks in the new universe. Result: some pairs are never scored (wrongly skipped) and the restored union-find reflects the old universe — a silently corrupted clustering that is neither the old-cap nor the new-cap cold result.

This is exactly the class the fingerprint already guards for threshold and max_block_size; the caps were simply missed (they're module constants, not threaded params, so they didn't ride along).

## Why it does NOT affect run 4

Run 4 launches from the merged commit with caps fixed at 64/64. A crash-resume of run 4 uses identical caps, so the fingerprint matches and the resume is correct. The gap only triggers if a cap constant is edited between crash and resume — a tuning/sweep workflow (the da#270 caps are documented as tunable/harness-swept), not the launch path.

## Fix (small)

Add the two constants to the fingerprint (mirrors threshold/max_block_size — `hash_strings` already takes the None-able `max_block_size`, so None caps are fine):

    fingerprint = hash_strings(
        digests["content"], round(threshold, 6), max_block_size, len(records),
        _MAX_MATCH_KEYS_PER_RECORD, _MAX_BLOCKING_TOKENS_PER_RECORD,
    )

[...]

Issue #295, opened 2026-07-03
fuzzy_cluster cap follow-up: name-first/IDF-ranked token retention + corpus-wide cap-sensitivity aggregate


parametrization:
## Follow-up from PR #277 (da#270 fuzzy_cluster caps) — recorded as TechDebt by BOTH reviewers

PR #277 capped per-record match-keys and blocking tokens at 64 (defaults; None disables). Both review verdicts (Kavitha, Nikolaos) approved non-gating with the same residual:

### 1. Token retention under the blocking cap is alphabetical, not informed

The blocking-token cap keeps `sorted(tok_cache)[:_MAX_BLOCKING_TOKENS_PER_RECORD]` — **alphabetically-first** tokens. A >64-significant-token deep-tail record whose *name* tokens sort late can lose its name-based block entirely (a genuine under-merge variant, distinct from the accepted worst-block pair-drop). Narrow today (<1% of records exceed 64 tokens; that population is dominated by pollution/mc≤1 records with no true duplicate), but the selection rule should be **informative-first**:

- Prefer **name-token-first**, then **rarest-first (corpus IDF)** ordering when truncating.
- Nikolaos's #271 blocking analysis of the same 210k set is prior art for the token-frequency table; an IDF rank can be computed once per run from the existing token cache.

### 2. Corpus-wide cap-sensitivity aggregate (Kavitha's review residual)

The 89.4%-pairs-kept figure is from the profiled **worst blocks** only. Before anyone tunes the default away from 64, produce the corpus-wide aggregate: pairs kept / merges changed / wall-clock at cap ∈ {32, 64, 128, 256, None} over a full run (cheap to instrument now that `--stop-after` (#276) allows bounded probes — note head-bias caveat in `docs/testing-on-subsets.md`).

### Acceptance

- Truncation order = name-first + IDF-ranked (or a measured argument for keeping alphabetical).
- Cap-sensitivity table posted here; default re-confirmed or changed with evidence.
- No behavior change for records under the cap (the invariant PR #277 established).

Part of #270. Refs: PR #277 verdict comments (Kavitha `4877...`, Nikolaos `4877873152`).

parametrization:
[...]

PR #305, opened 2026-07-03
fix(resolve/cluster): add da#270 cap constants to cluster resume fingerprint (da#303)

parametrization:
## Summary

Fast-follow on Ivana's post-merge review of #302 (da#272 PR2). The cluster resume fingerprint pinned the driver columns + threshold + `max_block_size` + record count, but **not** the two da#270 cap constants `_MAX_MATCH_KEYS_PER_RECORD` / `_MAX_BLOCKING_TOKENS_PER_RECORD`. Both shape the **block universe** (match-key cap → keys/tok caches; blocking-token cap → `pair_index` → the `blocks` list membership/order/count). Because the resume skip-set is **positional block indices**, a resume across a cap change would match the fingerprint yet restore `applied` indices pointing at *different* blocks — a silently corrupted clustering that is neither the old-cap nor the new-cap cold result.

The one-line fix adds both caps to the `hash_strings` call, so any cap change discards the checkpoint (the same guard already covered threshold and `max_block_size`).

## Not run-4-affecting

Run 4 launched with caps fixed at 64/64, so a crash-resume of it uses identical caps and the fingerprint matches — the resume is correct. The gap only triggers if a cap constant is edited between crash and resume (the #295 cap-tuning / harness-sweep workflow). This lands before that.

## ⚠️ Fingerprint change → pre-fix checkpoints discarded

Changing the fingerprint computation means **any checkpoint written by pre-#303 code will mismatch post-#303 and be discarded on resume** (the stage cold-restarts). That is the safe direction — never a silent misaligned resume — **but** it has one operational consequence for the live run:

> If the host crashes during **run 4's** cluster stage and we want to **keep** its checkpoint progress, resume from run 4's launch commit **`9cd97be`**, not a post-#303 head. A resume from a post-#303 binary will (correctly, safely) discard run 4's checkpoint and cold-restart the cluster stage.

Captured as a note in `docs/testing-on-subsets.md` ("Resuming across a code change").

## Test — a true discriminator

`test_cap_change_across_resume_discards_checkpoint`: takes a real partial checkpoint at the default caps, then collapses `_MAX_BLOCKING_TOKENS_PER_RECORD` to 1 (no composite key can form → the new universe has zero candidate pairs → the correct cold result is **0 merges**) and resumes. With the fix the checkpoint is discarded and the stage re-clusters to 0; **verified that without the one-line fix the test fails** — the wrongly-accepted checkpoint surfaces the old-cap merges (1 merged). Cluster suite (Ivana's precision/equivalence + resume) 29 passed; full resolve suite 376 passed. ruff + mypy-strict + cspell + markdownlint + structural staleness green.

da#304 (UnionFind `snapshot()`/`restore()` encapsulation, from Kavitha's retro) is a separate follow-up — kept out of here to keep this correctness fix crisp.

## Related Issues

Part of #303
Part of #272

## Review Checklist
- [ ] Must-fix items resolved


parametrization:
## Review — PR #305 (da#303): cap constants in the cluster resume fingerprint

Second reviewer. This is the fix for the exact gap I reported on da#272 (from the #302 review), so a focused verify. Reviewed at head `ea96229`. All three points confirmed.

**1. Both cap constants are in the fingerprint.** `cluster_canonical_narrators` now hashes `_MAX_MATCH_KEYS_PER_RECORD` and `_MAX_BLOCKING_TOKENS_PER_RECORD` alongside the content digest, threshold, `max_block_size`, and `len(records)`. That closes the vector precisely: both caps shape the block universe (the match-key cap via `keys_cache`/`tok_cache`, the blocking-token cap via `pair_index` → `blocks` membership/order/count), and since the resume skip-set is positional block indices, hashing the caps means any cap change flips the fingerprint and discards the stale checkpoint instead of restoring old-universe indices against a new-universe block list.

**2. The discard-on-cap-change test is a true discriminator, not a smoke test.** `test_cap_change_across_resume_discards_checkpoint` takes a *real* partial checkpoint at the default caps (`stop_after=1`, asserting `merged > 0`), then drops `_MAX_BLOCKING_TOKENS_PER_RECORD` to 1 — which empties the block universe (`combinations(block_tokens, 2)` over a 1-token list yields nothing), so the correct cold result is exactly 0 merges — and asserts `merged_records == 0`. I traced the negative case: without the caps in the fingerprint, the checkpoint would be accepted (`fp_ok` true, `parent_ok` true since `n` is unchanged), the old-cap `uf._parent` would be restored, the empty new block list would add nothing, and the run would surface the old merges (>0) — failing the assert. So the test genuinely fails against the pre-fix code. Good choice of lever.

**3. Hashing the constants suffices — and is the right mechanism, better than a schema_version bump.** You asked me to confirm the alternative. A `_CLUSTER_CHECKPOINT_SCHEMA_VERSION` bump guards *payload layout*, not *computation inputs*, and it would only fire on a cap change if a human remembered to bump it every time — error-prone and semantically the wrong lever (a cap change isn't a layout change). Folding the cap values into the fingerprint is self-maintaining: any cap edit automatically changes the digest, no human step, and it matches the convention every other stage follows (all output-determining inputs go in the fingerprint). It also handles the `None`-disable case correctly — `hash_strings` stringifies each part, so `None`/`64`/`1` are distinct digests, meaning a caps-disabled run won't collide with a capped one. Version bump is neither necessary nor sufficient here; hashing is the correct fix.

No blockers. Clean, minimal, correctly targeted, and the test locks the regression. Approving.

---
RequestOrReplied: Approved
TechDebt: none identified — the fix is minimal and the test is a true discriminator; the earlier UnionFind._parent encapsulation note from #302 still stands as separate, unrelated cleanup.

parametrization:
## Review — da#303 cap constants in the cluster resume fingerprint

Reviewed the change at head ea96229 via the contents/diff API (no checkout — live run owns the tree). Correct, tightly scoped, and the test is the right kind. Approving; owner is waiting so keeping this short.

The bug it closes is real: the resume skip-set is positional block indices, and the two da#270 caps shape the block universe (the match-key cap feeds `keys_cache`/`tok_cache`; the blocking-token cap feeds `pair_index` → the `blocks` list membership, order, and count) without being threaded params. So before this fix a resume across a cap change matched the fingerprint yet restored an old-universe union-find against a new-universe block list — a silent misalignment. Adding both caps to `hash_strings` forces a discard-and-cold on any cap change, which is the safe direction.

On the three things I checked specifically:
- Both constants are in the fingerprint call (match-key cap and blocking-token cap), appended after the existing content/threshold/max-block/len fields; the same function computes save-time and resume-time hashes so field order stays consistent, and a pre-da#303 checkpoint hashed without the caps mismatches a post-merge binary and is discarded — exactly the intended run-4 behaviour.
- `test_cap_change_across_resume_discards_checkpoint` is a true discriminator, not a tautology: it first asserts the partial checkpoint carries real merges (`ckpt["merged"] > 0`), then collapses the blocking-token cap to 1 so the correct cold universe has zero composite pairs and therefore zero merges — a wrongly-accepted stale checkpoint would instead surface the old-cap merges, so the `== 0` assertion fails precisely when the caps are missing from the fingerprint. Good.
- The runbook section in `testing-on-subsets.md` states the discard behaviour and names the concrete `9cd97be` launch-commit guidance for keeping run-4's cluster checkpoint.

One forward-looking note (does not block this PR): the fingerprint enumerates specific universe-shaping constants — now the two caps — rather than deriving from all of them. `_CONNECTOR_TOKENS` also shapes the block universe (via `_significant_tokens` → `tok_cache`), and it is not in the hash. It is stable today, so there is no current misalignment, but my upcoming da#271 track-3 connector-token expansion changes exactly that set — so that PR must add `_CONNECTOR_TOKENS` (its sorted contents) to this same `hash_strings` call, or we reintroduce this bug class one constant at a time. Filing as tech debt so the pattern is visible; a longer-term option is fingerprinting a single blocking-semantics version tag bumped on any such change.

CI 12/12 green, mergeStateStatus CLEAN.

---
RequestOrReplied: Approved
TechDebt: the resume fingerprint enumerates universe-shaping constants (now the two da#270 caps) but omits _CONNECTOR_TOKENS, which also shapes the block universe — the da#271 track-3 connector expansion must be added to the same hash_strings call (or switch to a single blocking-semantics version tag) to avoid reintroducing the da#303 positional-misalignment class.

Issue #306, opened 2026-07-04
fuzzy_cluster: _safe_partition quadratic scan on mega-group — 17h+ silent phase (token-indexed candidates + caching, output-identical)


parametrization:
## Symptom

Run 4 (wave-23 resolve re-run, launch commit `9cd97be`) finished cluster pair scoring in 82 min (`cluster_pairs_scored`: scored=106,336,241, merged=179,318) — then went **silent for 17+ hours** in the post-scoring safe-partition phase. Worker healthy, single-threaded ~100% CPU, no log output: `_safe_partition` has no progress logging.

## Root cause

The union-find produced **one mega-group of 168,773 members** (70% of the 241,189 records; second-largest group = 52) — read from the on-disk cluster checkpoint's `parent` array. `_safe_partition` (fuzzy_cluster.py) does greedy clique placement: each member is tested with `all(_can_merge(i, j) for j in sub)` against existing subclusters **linearly**. Worst case for the mega-group ≈ n²/2 ≈ **14.2B `_can_merge` calls**, and each call:

- recomputes `_match_keys()` and the significant-token set per call (no caching), and
- `_name_match` cross-products match keys — up to 64×64 rapidfuzz ratios post-#277-caps.

Measured `_can_merge` throughput ~600k calls/s (1-key records; key-rich records are far slower) → estimated 12–30+ h for this one phase.

## Fix (provably output-identical)

1. **Token-indexed candidate lookup**: a subcluster sharing `< _MIN_SHARED_TOKENS` significant tokens with the member can never accept it (`_shared_significant_token_count` guard rejects). Index subclusters by significant token; for each member, test only subclusters sharing ≥1 token (still in **original creation order**, first-accepting wins) — exact same greedy semantics, so byte-identical output. (With `_MIN_SHARED_TOKENS ≥ 2`, a candidate must share ≥2 indexed tokens — either filter is sound; keep it simple and safe.)
2. **Per-record caching**: precompute match keys + significant-token sets once per record before the loop (dict keyed by record index).
3. **Telemetry**: log group-size distribution after `uf.groups()` (top-k sizes, count of multi-member groups) and a progress line every N members inside safe-partition for groups over a size threshold.

## Constraints

- **Output-identical** to current greedy partition — reviewers must verify the equivalence argument + a test comparing old-vs-new partitions on synthetic groups that include guard-conflict bridge cases (A–B ok, B–C ok, A–C death-conflict).
[...]

PR #307, opened 2026-07-04
fix(resolve): token-indexed safe-partition + per-record caching (da#306)

parametrization:
Part of #306

## Problem

`_safe_partition` split each union-find group into guard-safe cliques by a **greedy scan over ALL existing sub-clusters** for every member. Run 4's union-find produced a single **168,773-member mega-group** (70% of records), making the phase worst-case ~n²/2 ≈ **14.2B `_can_merge` calls** — each of which *also* recomputed match keys + tokenization. The phase has no logging, so it ran **silently for 17h+** in production.

## Fix — output-identical, not just faster

1. **Token-indexed candidate lookup.** `_can_merge` requires `≥ _MIN_SHARED_TOKENS` shared significant tokens with *every* member of a sub-cluster, so a sub-cluster in which **no** member shares even one significant token with the incoming member can never accept it — the naive scan tests it and rejects on its first member anyway. A `token -> sub-cluster indices` index restricts each member to the sub-clusters sharing ≥1 token, tested in **ascending index = original creation order, first-accepting wins**, each still verified with the full `all(_can_merge(...))` predicate.

2. **Per-record caching.** Match keys (`_match_keys`) and significant-token sets (`_significant_tokens`) are computed **once per member** and read by a new `_can_merge_cached` that reuses the exact same guard predicates (`_death_years_conflict`, `_genders_conflict`, `_name_match`) — only the key/token derivation is memoized, so the merge rule keeps one source of truth.

3. **Telemetry.** `cluster_group_sizes` after `uf.groups()` (group count, multi-member count, top-k sizes, members-in-multi-member-groups); `safe_partition_progress` every 10,000 members for any group > 5,000.

### Equivalence argument (byte-identical partitions)

The first accepting sub-cluster in the naive full scan is necessarily a candidate here (it accepted, so the member shares ≥1 token with each of its members). Every sub-cluster *before* it that the naive scan rejected is either (a) also a candidate — tested and rejected identically here — or (b) a non-candidate the naive scan would reject on its first member (zero shared tokens ⇒ fails the `≥ _MIN_SHARED_TOKENS` guard). So the chosen sub-cluster (or the decision to open a new one) is identical for every member, and the emitted partition is byte-identical. The argument is written into the `_safe_partition` and `_can_merge_cached` docstrings.

### Equivalence test

`tests/test_resolve/test_safe_partition_equiv.py` keeps a **verbatim copy of the pre-#306 implementation** as `_safe_partition_reference` (calling the real, uncached `_can_merge`) and asserts identical output across 60 randomized synthetic groups (sizes 2–40, dense shared-token pool) **plus** the explicit guard-conflict bridge (A–B ✓, B–C ✓, A–C death-conflict) and the singleton/empty degenerate paths.

## Benchmark

Microbenchmark on a synthetic heavy-tailed mega-group (one union-find group of N members, run out-of-tree). Old is the pre-#306 naive scan; both assert byte-identical output:

| N | old | new | speedup | identical |
|---|-----|-----|---------|-----------|
| 1,000 | 1.73 s | 0.09 s | 19.0× | ✓ |
| 2,000 | 7.61 s | 0.47 s | 16.1× | ✓ |
| 4,000 | 29.98 s | 1.57 s | 19.1× | ✓ |
| 10,000 | 183.03 s | 10.34 s | **17.7×** | ✓ |

The old impl is quadratic in group size; the constant-factor win alone turns run 4's 17h+ phase into ~1h, and the token pruning compounds on the real heavy-tailed narrator token distribution.

## Fingerprint / recovery safety (#295 invariant)

Does **not** touch blocking, the da#270/#303 key/token caps, or the checkpoint fingerprint function. The partition phase changes no block membership/order/count, so run 4's resume fingerprint stays valid.

**Cherry-pick onto `9cd97be`** (run-4 launch commit, pre-#303): scratch `git cherry-pick 5b161bf` onto `9cd97be` — **`src/resolve/fuzzy_cluster.py` and the test apply cleanly** (auto-merged, no conflict). The only conflict is the *generated* `ontology/structural/llms.txt` (the repo's atomic structural-ontology gate forces the regenerated index into the same commit as the src change; at 9cd97be that generated file differs). It is immaterial to the recovery, which runs `resolve --from-step cluster` (**no commit**, no gate) — resolve it with `git checkout 9cd97be -- ontology/structural/`. Verified in the cherry-picked tree: the fix is present (`_can_merge_cached`, `token_index`, telemetry) **and** the fingerprint stays `hash_strings(digests["content"], round(threshold, 6), max_block_size, len(records))` — i.e. the #303 caps are correctly absent at 9cd97be, so run 4's checkpoint fingerprint is unchanged.

## Checks

- `make check` green: `ruff check`, `ruff format --check`, `mypy src/` (Success, 94 files), full suite **1455 passed, 7 skipped** (`-m "not integration"`).
- Pre-commit + pre-push hooks (incl. `structural-ontology-staleness`, gitleaks, mypy, pytest) all passed on commit + push. No `--no-verify`.


parametrization:
## Review — PR #307 (token-indexed `_safe_partition` + per-record caching, da#306)

Reviewed with the single question that matters for cherry-picking onto run 4's live checkpoint (9cd97be): **is the BYTE-IDENTICAL claim true?** I diffed the rewrite against the pre-da#306 `_safe_partition` at the merge-base (58a0f6a), verified `_can_merge_cached` against `_can_merge`, and worked the candidate-filter soundness argument. It holds.

**Verdict: APPROVED**

### Gating findings
None.

### Correctness verification (why the equivalence holds)

1. **Candidate-filter soundness — SOUND (gating concern, cleared).** Acceptance of member `i` into a sub-cluster requires `all(_can_merge(i, j))` over its members, and `_can_merge` demands `≥ _MIN_SHARED_TOKENS` shared significant tokens with *each* member. So any accepting sub-cluster has ≥1 shared token via **every** member → it is necessarily in `candidates` (the union of `token_index[tok]` over `i`'s tokens). The candidate set is therefore a **superset** of the true acceptors — nothing accepting is ever dropped. Surviving candidates are iterated as `sorted(candidates)`, and since sub-cluster indices are assigned in `append` (creation) order, ascending index == creation order == the naive scan's order, with first-accepting-wins via `break`. Every sub-cluster the naive scan would test-and-reject before the first acceptor is either (a) also a candidate here → tested-and-rejected with the identical decision, or (b) a non-candidate the naive scan rejects on its first zero-overlap member. Chosen sub-cluster (or the decision to open a new one) is identical; member/sub-cluster **order** is preserved. `ordered = sorted(group, key=canonical_id)` and the `len(group) <= 1` early return are unchanged.

2. **`_can_merge_cached` ≡ `_can_merge` — VERIFIED.** Line-for-line: same `_death_years_conflict`/`_genders_conflict` guards on the same records; `len(tok_cache[i] & tok_cache[j])` equals `_shared_significant_token_count(a,b)` because `tok_cache[k] == _significant_tokens(_match_keys(records[k]))`; `_name_match(keys_cache[i], keys_cache[j])` equals `_name_match(_match_keys(a), _match_keys(b))` because `keys_cache[k] == _match_keys(records[k])`. Caches are keyed by **record index** (int), built fresh inside each `_safe_partition` call over `ordered` (which contains both operands of every call → no KeyError), and never leak across calls. The merge *rule* keeps a single source of truth (guard predicates unchanged); only key/token derivation is memoized. Token-order guard (`_token_order_consistent`) and threshold flow through `_name_match` untouched.

3. **Edge cases — correct by construction.** Zero-significant-token member: `tok_cache[i]` empty → `candidates` empty → loop body skipped → opens a new sub-cluster; matches naive (shared count `0 < 2` rejects all existing). Such a member contributes no tokens to `token_index`, so no later member can ever match it — correct, since `_can_merge(k, i)` would fail the token guard. Missing name fields → `_match_keys` returns `[]` → same empty-token path. `token_index` invariant (holds `s` iff some member of sub-cluster `s` carries `tok`) is maintained on both the append and new-sub-cluster branches. Duplicate tokens dedup via the `set`. Empty/singleton group hits the unchanged early return.

4. **Equivalence test — honest oracle, adequate coverage (non-gating gap below).** `_safe_partition_reference` is a **verbatim** copy of the merge-base `_safe_partition` (confirmed by diff), calling the real uncached `fc._can_merge` — so the test does not tautologically depend on the code under test. `assert actual == expected` is order-sensitive at both the outer (sub-cluster) and inner (member) level, so it pins ORDER, which matters downstream via `_choose_representative` ties. The generator draws from a 15-token pool with `بن` as a stripped connector, so records share significant tokens densely and the death-year spread manufactures real A–B✓/B–C✓/A–C✗ bridges; 60 seeds × sizes up to 40 plus the dedicated bridge test exercise multi-sub-cluster and conflict topologies.

5. **Fingerprint invariant (#295) — untouched.** The diff touches only the new constants, `_can_merge_cached`, the `_safe_partition` body, `_log_group_size_distribution`, and its single call site. Blocking, the da#270 key/token caps (`_MAX_MATCH_KEYS_PER_RECORD` etc.), and the fingerprint function are not modified (the `fingerprint`/cap references in the diff are unchanged context lines and auto-regenerated structural-index line-number shifts). `cluster_records`' signature is unchanged. No new imports (`defaultdict`, `time`, `logger` already present).

6. **Telemetry — behavior-neutral.** `safe_partition_progress` fires only for `len(group) > 5000` and every 10k placements (no per-member logging). `cluster_group_sizes` is emitted **once** at `cluster_records` finalize (line 969), not in the drain loop. Fields are well-formed. `groups = uf.groups()` hoisted out of the `for` is a pure read, no behavior change.

7. **CI — green.** All checks pass (Lint, Type Check, Test, Integration, sync-drift gate, staleness-check).

### Non-gating notes (no re-review required)

- **N1 (doc accuracy).** `_can_merge_cached` and `_safe_partition` docstrings describe `_MIN_SHARED_TOKENS` as "(≥1)". The constant is actually **2**. The soundness argument survives (≥2 shared with every member ⟹ ≥1 ⟹ token-candidate, so the filter's ≥1-overlap test is a valid *relaxation* / superset), but the parenthetical states the wrong value and could mislead a future reader into believing the constant is 1. Suggest rewording to "≥ `_MIN_SHARED_TOKENS` (=2), and the ≥1-shared-token filter is a sound relaxation of that." Pure comment fix — does not affect the cherry-pick.
- **N2 (test gap).** The randomized generator always yields ≥1 significant token and always sets `name_ar_normalized`, so the zero-token / missing-name path (empty `candidates` → new sub-cluster) is not directly exercised by the equivalence sweep. It is trivially correct by inspection (see finding 3), but a single record with an absent/empty name added to `test_safe_partition_singleton_and_empty` would pin it. Low risk for checkpoint recovery — real canonical records carry names.

Neither note blocks the cherry-pick onto 9cd97be. The optimization is output-identical to the naive greedy partition.

— Kavitha Sundaramurthy, Senior Data Engineer

parametrization:
## Review — PR #307 (token-indexed safe-partition + per-record caching, da#306)

Independent review focused on the recovery path: **resume/fingerprint safety** and the **performance/equivalence** claims, since the plan is to cherry-pick this onto run 4's live cluster checkpoint at `9cd97be`. I traced the actual diff hunks and the base predicates, not just the PR description.

**Verdict: APPROVED**

### Findings

**1. Fingerprint / resume safety (#295 invariant) — gating concern, CONFIRMED SAFE.**
The diff is confined to exactly four hunks in `fuzzy_cluster.py`: the telemetry constants (line ~174), the new `_can_merge_cached` (line ~400), the `_safe_partition` rewrite + `_log_group_size_distribution` (line ~480), and the `uf.groups()`-caching in `_drain_completed` (line ~965). Grepping every added/removed line for `fingerprint|hash_strings|max_block|_MAX_|blocking|digest|threshold, 6` returns **nothing** — the blocking token derivation, the da#270/#303 key/token caps, record ordering, and the checkpoint fingerprint function are all untouched. `_safe_partition` runs *after* union-find and only re-shapes the partition; the resume fingerprint is computed over blocks/records upstream of clustering, so run 4's checkpoint fingerprint is unchanged.

**2. Cherry-pick onto `9cd97be` — CONFIRMED clean for code, only generated file conflicts.**
I ran the merge read-only myself: `git merge-tree --write-tree --merge-base=58a0f6a 9cd97be 5b161bf` exits 1 with the **sole** conflicted path being `ontology/structural/llms.txt` (generated). `src/resolve/fuzzy_cluster.py` and `code-graph.json` both report `Auto-merging` with no conflict, and the test file is a pure add. This matches the PR body exactly — resolve the generated conflict with `git checkout 9cd97be -- ontology/structural/` (the recovery runs `resolve --from-step cluster`, no commit, so the structural gate never fires). Non-gating.

**3. Token semantics — CONFIRMED byte-identical, no raw-vs-filtered mismatch.**
Both the guard and the new index derive tokens through the same path. The guard's `_shared_significant_token_count(a,b)` is `len(_significant_tokens(_match_keys(a)) & _significant_tokens(_match_keys(b)))`; the cache is `keys_cache[i]=_match_keys(records[i])`, `tok_cache[i]=_significant_tokens(keys_cache[i])`, and both the candidate index and `_can_merge_cached`'s guard read `tok_cache`. `_significant_tokens` applies no blocking cap (the `_MAX_BLOCKING_TOKENS_PER_RECORD=64` cap lives only in the blocking phase, a different code path), and `_match_keys` applies the same `_MAX_MATCH_KEYS_PER_RECORD` truncation on both sides. So the intersection integer and the `_name_match` key-lists are identical. **Superset property holds:** acceptance needs `≥ _MIN_SHARED_TOKENS (=2)` shared tokens with *every* member, hence `≥1` with the first member, hence the accepting sub-cluster is always in `candidates` (which is the `share-≥1` union). Nothing is dropped. Non-gating.

**4. Performance claim — CONFIRMED, no regression even in the ubiquitous-token pathology.**
`candidates` is a `set` (deduped), and per considered sub-cluster the work is strictly `⊆` the old naive scan while each `_can_merge_cached` is cheaper than `_can_merge` (no re-derivation of keys/tokens, no repeated `fuzz.token_set_ratio`). In the degenerate one-ubiquitous-token case the candidate build is `O(#subclusters)` and adds a `sorted(candidates)` of cheap integer comparisons per member — the *same order* as the old full scan, whose per-subcluster cost was the expensive rapidfuzz call. So new ≤ old in every case, and dramatically faster in the disjoint case (empty candidates ⇒ skip the scan entirely). Non-gating.

**5. Greedy-equivalence (independent second opinion) — CONFIRMED.**
`sorted(candidates)` iterates ascending sub-cluster index = creation order; first-accepting wins via `break`; empty candidates falls through to the `else` opening a new sub-cluster. The equivalence oracle in `test_safe_partition_equiv.py` keeps a verbatim pre-#306 `_safe_partition_reference` calling the real uncached `_can_merge` and asserts identical output over 60 randomized dense-token groups plus the death-year bridge and singleton/empty paths. The "empty candidates ⇒ open new" branch is exercised by the first member of every group. Non-gating.

**6. Telemetry — CONFIRMED sane.** `cluster_group_sizes` after `uf.groups()` and `safe_partition_progress` every 10k members for groups >5k; `logger = get_logger(__name__)`; field style matches the existing `cluster_progress` / `cluster_*` structlog events. `time`/`defaultdict` already imported. Non-gating.

**7. CI — green.** All 7 checks pass (Lint, Type Check, Test, Integration, sync-drift gate, staleness-check, base-image pin).

### Non-gating nits (optional, do not block merge)
- **`_MIN_SHARED_TOKENS` is `2`, not `1`.** The `_safe_partition` docstring writes "`≥ _MIN_SHARED_TOKENS` (≥1)". The parenthetical is describing the index's own `share-≥1` superset threshold and the logic is correct (2 ≥ 1, so a zero-token sub-cluster can never accept), but a reader could misread "(≥1)" as the constant's value. Consider rewording to "the index uses a looser ≥1-shared-token filter, a superset of the guard's `_MIN_SHARED_TOKENS`(=2) requirement."
- A record whose name yields **zero** significant tokens (all connectors / <2-char) is provably equivalent (empty `tok_cache` ⇒ empty candidates ⇒ opens its own sub-cluster, matching the old scan's all-reject) but isn't explicitly synthesized in the test — the generator always emits ≥1 token. The branch itself is covered; adding one explicit zero-token record would pin the guard.

Nice work — the equivalence argument is sound, the fingerprint surface is genuinely untouched, and the cherry-pick is clean for the recovery. Safe to resume run 4's checkpoint onto this.

— Nikolaos Papadopoulos, Senior Data Engineer

parametrization:
Formal approval trailer for my review at comment 4882695330 — verdict unchanged: APPROVED.

---
RequestOrReplied: Approved
TechDebt: none

parametrization:
Formal approval trailer for my review at comment 4882704134 — verdict unchanged: APPROVED.

---
RequestOrReplied: Approved
TechDebt: none

PR #322, opened 2026-07-06
fix(da#321): derive dedup cross_sect from authoritative sect column

### LINKED ISSUE #321: test: test_run_all_promoted_bio_only_narrator_survives green in CI, red locally (embedder-env cross_sect divergence)
## Summary

`tests/test_resolve/test_run_all.py::test_run_all_promoted_bio_only_narrator_survives` **passes in CI but fails deterministically in the local `.venv`** on the same commit (main `19c2ee9b`).

- CI (`ci.yml`) on `19c2ee9b`: **success** (`Test` job green).
- Local pre-push (`pytest`) on the same commit: **FAILED** — `AssertionError: expected a cross-sect PARALLEL_OF edge` (`cross_sect: [[false]]`).

This blocks `git push` for anyone on the affected machine (pre-push runs the full suite; `--no-verify` is correctly blocked), even for changes that don't touch the resolve pipeline.

## Root cause (likely)

The assertion depends on a **sentence-transformers MiniLM** semantic-similarity outcome: the test fixtures `hs1`/`hs2` must score above the `cross_sect` PARALLEL_OF threshold. The pipeline log shows `parallels_detected cross_sect=1` but the composed `parallel_links` ends with `cross_sect=false` — a threshold edge case that tips differently between the CI model/runtime and the local `.venv` model/runtime. Note the `BertModel LOAD REPORT ... embeddings.position_ids UNEXPECTED` on local load.

Same class as the semantic-embedder parity gap (deploy#523 / `project_semantic_embedder_parity`): a semantic outcome that varies by embedder environment.

## Fix direction

Make the test robust to embedder-environment variance: either pin the exact model+runtime the assertion assumes, assert on a deterministic pre-embedding signal, or relax the fixture so the `cross_sect` pair is unambiguously above threshold in both environments. Do NOT weaken it to always-pass.

## Impact

Non-blocking for #723 (prod is loaded + verified; CI green). But it makes local `pre-push` red on this machine, and the local⇄CI parity guarantee (main#684) is violated for this test. Surfaced during P8W23 wrapup when a memory-only commit could not be pushed.

parametrization:
## Summary

Fixes **da#321** — `test_run_all_promoted_bio_only_narrator_survives` passed in CI but failed deterministically in the local `.venv` (`AssertionError: expected a cross-sect PARALLEL_OF edge`, `cross_sect: [[false]]`).

## Root cause (a real correctness bug, not just a flaky test)

The two `PARALLEL_OF` detectors derived `cross_sect` from **different sources of truth**, and `run_all._compose_parallel_links` lets the *semantic* row win on a shared canonical `(hadith_id_a, hadith_id_b)` pair:

| Detector | `cross_sect` derivation |
|----------|-------------------------|
| `parallels.py` (deterministic lexical) | `sect_i != sect_j` — the authoritative **`sect` column** |
| `dedup.py` (semantic embeddings) | `_is_cross_sect(corpus_a, corpus_b)` — inferred from the source-**corpus name** via hard-coded sunni/shia allowlists |

For two hadiths with the **same `source_corpus` but different `sect`** (test fixtures `hs1` sunni / `hs2` shia, both defaulting `source_corpus="sunnah"`), `dedup` returns `cross_sect=False` and overwrites the correct deterministic `True`.

- **Locally** the MiniLM model loads → `dedup` emits the clobbering row → assertion fails.
- **In CI** the model is absent → `dedup` degrades to empty → the deterministic `True` survives → green.

Same class as the semantic-embedder parity gap (deploy#523).

## Fix

Align `dedup` on the same authoritative signal `parallels.py` already treats as canonical:

- `_load_hadith_texts` loads the `sect` column (a **non-nullable** `HADITH_SCHEMA` field) instead of `source_corpus`.
- `_is_cross_sect(sect_a, sect_b)` → `bool(sect_a and sect_b and sect_a != sect_b)`.
- `id_to_corpus` → `id_to_sect` threaded through `_search_and_collect_resumable` / `run_dedup`.
- Dedup checkpoint schema version **1 → 2** so a checkpoint holding old corpus-derived `cross_sect` values is discarded on resume, not restored stale.

This also fixes a **latent production bug**: any corpus absent from the two allowlists, or a mixed-sect corpus, was mislabeled `cross_sect=False`.

Per the issue's guidance ("assert on a deterministic signal / root fix; do NOT weaken to always-pass") — this is the root fix, not a fixture relaxation.

## Verification

- The previously-failing `test_run_all_promoted_bio_only_narrator_survives` now **passes with the model loaded** (0 skipped — ML deps present locally).
- Full local `pre-push` suite (mypy-strict + pytest) **green** — the local⇄CI parity guarantee (main#684) is restored for this test.
- `ruff format`/`ruff check`/`mypy` clean; structural ontology index regenerated for the shifted `dedup.py` symbols.

Closes #321


parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

**Review: Root-cause fix verified — dedup now derives cross_sect from the authoritative sect column, matching parallels.py exactly**

I verified this against the diff at HEAD (50daabde46346db8573af6049c9748380ef332d9) and read the full files, not just the changed hunks.

- `_is_cross_sect` in `src/resolve/dedup.py` now takes `(sect_a, sect_b)` and returns `bool(sect_a and sect_b and sect_a != sect_b)`. I read `src/resolve/parallels.py` line 322 and it uses the identical guard `bool(sect_i and sect_j and sect_i != sect_j)` against the same `sect` column — both detectors now agree by construction because they read the same field with the same formula, not just coincidentally similar logic.
- `sect` is confirmed non-nullable in `HADITH_SCHEMA` (`src/parse/schemas.py` line 57, `pa.field("sect", pa.string(), nullable=False)`), so the null-guard in `_is_cross_sect` is defensive, not load-bearing for real staging data.
- Read `src/resolve/__init__.py::_compose_parallel_links` in full — confirms the PR's root-cause narrative precisely: deterministic (parallels) rows are inserted first, semantic (dedup) rows second, and the semantic row wins on a shared canonical `(hadith_id_a, hadith_id_b)` key. Before this fix, dedup's corpus-allowlist-derived `cross_sect` could clobber a correct deterministic `True` with a wrong `False` whenever two hadiths shared a `source_corpus` but had different `sect` — exactly the `hs1`/`hs2` fixture in `test_run_all_promoted_bio_only_narrator_survives` (both default `source_corpus="sunnah"` via the `write_hadiths` helper, but `sect="sunni"`/`"shia"` explicitly). That test is not mocked for dedup, so it exercises the real embedding path locally and the composed-artifact path in CI — both now correctly agree.
- Checkpoint schema_version bump 1→2 is wired correctly: `run_dedup`'s resume gate computes `layout_ok = ckpt.get("schema_version") == _DEDUP_CHECKPOINT_SCHEMA_VERSION`, and on mismatch calls `clear_checkpoint` before cold-starting — so a v1 checkpoint holding corpus-derived `cross_sect` values cannot be silently restored. Both hand-built checkpoint fixtures in `tests/test_resolve/test_dedup_faiss_resume.py` (`test_stale_fingerprint_cold_starts` and `test_resume_false_ignores_checkpoint`) were updated from `1` to `2`; `test_stop_after.py`'s hand-built dict has no `schema_version` literal so nothing was missed there.
- Test changes are a genuine strengthening, not a weakening to always-pass: `test_sunni_sunni_not_cross`/`test_unknown_corpus` were replaced with `test_same_sect_not_cross`/`test_missing_sect_not_cross` (covering `None` and empty-string sect on either side), which is a superset of the old coverage under the new signature.
- Ran a repo-wide code search for the removed symbols (`_SUNNI_SOURCES`, `id_to_corpus`) across the whole repository at this ref — zero hits. No stale references left in files outside the PR's changed set (checked `test_dedup_parallel.py`, `test_dedup_resume.py`, `src/cli.py` specifically since they touch dedup/run_all but aren't in the diff).

No remaining path found where the two detectors could disagree on `cross_sect` — they now read the identical authoritative column with the identical formula. This closes the same class of parity bug as the semantic-embedder gap (deploy#523), but here the fix is the correct one: align the derivation, not paper over the composition order.

Must-fix: None
Tech-debt: None

parametrization:
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

**Review: LGTM — `sect` is genuinely the single authoritative source, and the fix makes the two PARALLEL_OF detectors agree.**

I read the diff at HEAD (50daabd) plus full context of `parallels.py`, `src/parse/schemas.py`, `src/resolve/__init__.py::_compose_parallel_links`, and every `src/parse/*.py` corpus parser, and confirmed the design end to end rather than taking the PR description on faith.

Confirmed independently:
- `parallels.py` already derives `cross_sect` as `bool(sect_i and sect_j and sect_i != sect_j)` from the hadith `sect` column — `dedup.py`'s new `_is_cross_sect` uses the byte-identical formula, so the two detectors now genuinely agree instead of merely both compiling.
- `src/parse/schemas.py::HADITH_SCHEMA` declares `sect` `nullable=False`, and every corpus parser that emits `hadiths_*.parquet` (bihar, sunnah_scraped, halimbahae, thaqalayn, fawaz, tusi, lk_corpus, open_hadith, sunnah_api, sanadset, mis) sets a concrete `"sunni"`/`"shia"` value — none leaves it blank. `itqan.py` and `muhaddithat.py` are narrator-bio-only producers with no `hadiths_*.parquet` output, so they're outside this surface entirely.
- The old `_SUNNI_SOURCES`/`_SHIA_SOURCES` allowlists were strictly worse, not just narrower: `fawaz.py::_detect_sect` assigns sect **per edition** (Shia indicators like `kafi`/`nahj` route to `"shia"`) while every fawaz hadith shares one `source_corpus == "fawaz"`. Under the old code a sunni-edition/shia-edition fawaz pair had `a_sunni=b_sunni=True` (since `"fawaz"` was in `_SUNNI_SOURCES`) and would be scored **not cross-sect** — a real mislabel, not a hypothetical. `bihar`, `halimbahae`, `tusi`, `mis` were never in either allowlist at all, so any pair touching those corpora was unconditionally `cross_sect=False` regardless of truth. The new column-based check fixes all of this correctly; I don't see any legitimate classification capability the allowlists provided that `sect` doesn't already cover strictly better.
- `_compose_parallel_links` (`src/resolve/__init__.py`) unions the two detectors' `parallel_links.parquet` outputs keyed on `(hadith_id_a, hadith_id_b)`, letting the semantic row win on collision. Since both detectors now compute `cross_sect` identically, the composed value can no longer flip depending on which detector wins the race — the exact defect da#321 reported.
- The checkpoint `schema_version` bump 1→2 is correct and necessary, not cosmetic: `_search_and_collect_resumable`'s checkpoint literally persists the computed `cross_sects` list (not just raw scores), so a v1 checkpoint holds values keyed on the old corpus-allowlist formula. The version gate (`layout_ok = ckpt.get("schema_version") == _DEDUP_CHECKPOINT_SCHEMA_VERSION`) will correctly discard and cold-start on any pre-existing v1 checkpoint rather than resurrecting stale values. This matches the repo's established checkpoint-invalidation convention.
- Test updates are consistent with the new signature everywhere (`test_dedup.py`, `test_dedup_faiss_resume.py`, `test_stop_after.py`, `test_negative.py`), including the two new `TestIsCrossSect` cases for missing/empty `sect` — `_is_cross_sect(None, "sunni")` and `_is_cross_sect("sunni", "")` both correctly resolve to non-cross via the `bool(sect_a and sect_b and ...)` short-circuit. No leftover references to `id_to_corpus`/`_SUNNI_SOURCES`/`_SHIA_SOURCES` anywhere in the tree.
- CI is green across Lint/Type Check/Test/Integration/Base-image/sync-drift; ontology structural diff is a clean line-number-only regeneration matching the code shift.

Production impact: this is a strictly-more-correct change. I found no real cross-sect pair the corpus-allowlist caught that the `sect` column would now miss — if anything the allowlist actively mislabeled fawaz's mixed-sect editions and silently dropped four whole corpora from cross-sect consideration. No null/empty-`sect` gap exists in the current parser set that would regress recall.

Must-fix: None
Tech-debt: None

PR #323, opened 2026-07-06
fix(graph-load): split multi-statement validation Cypher + register inventory checks as informational (da#319)

### LINKED ISSUE #319: fix(graph-load): validation harness emits multi-statement Cypher → false-FAIL rc=1 on clean load
## Summary

The graph-load post-load validation harness emits **multi-statement Cypher** for two checks, which the deployed Neo4j build rejects, so both fail with a syntax error rather than running:

```
neo4j.exceptions.CypherSyntaxError: {neo4j_code: Neo.ClientError.Statement.SyntaxError}
  {message: Expected exactly one statement per query but got: 5 (line 9, column 0 (offset: 513))}
```

Affected checks (seen in the 2026-07-05 prod full reload):

- `graph_integrity_deferred_inventory` → `[FAIL] query execution failed`
- `sanadset_orphan_inventory` → `[FAIL] query execution failed`

These are **false FAILs**: the underlying data is fine (the same reload passed every #723 record-level acceptance criterion and is at exact stg parity). But because the harness marks the overall load `validation_passed=False`, the loader exits `rc=1`, which makes `scripts/load_staging.sh` report `LOAD FAILED` even though the load converged fully (1,585,981 nodes / 8,629,160 edges).

## Fix

Split the offending validation queries into one statement per `session.run` / `tx.run` call (or wrap in a single `CALL {} ... CALL {}` subquery form that this Neo4j version accepts). Same multi-statement quirk I hit in manual verification queries — this Neo4j build enforces "exactly one statement per query".

## Impact / severity

Non-blocking (data is correct; both are inventory/orphan *reporting* checks, not gates), but it makes a clean load look failed and masks whether these two integrity checks would have surfaced anything real. Pairs with the existing `chain_integrity` timeout downgrade (da#259).

Ref: noorinalabs-main#723 prod-reload verification.

parametrization:
## Summary

Post-load validation (`run_validation()` in `src/graph/validate.py`) read each `queries/validation/*.cypher` file as ONE query string and executed it via a single `execute_read()` -> `tx.run()`. Two files are 5-statement scripts (semicolon-separated blocks): `graph_integrity_deferred_inventory.cypher` (ADR-004) and `sanadset_orphan_inventory.cypher` (ADR-003). Neo4j rejects a multi-statement string (`CypherSyntaxError: Expected exactly one statement per query but got: 5`), which was caught by the broad `except Exception` -> `passed=False` -> `is_fatal=True` -> `validation_passed=False` -> `sys.exit(1)` on a **clean, fully-converged load**.

**Part 1 — split-and-run.** `_split_statements()` splits each file's text into top-level statements on `;` and `run_validation` executes every statement as its own bounded query (same per-statement timeout budget as da#259), aggregating the per-statement rows back into a single `ValidationResult` per FILE (the `query_name` scheme is unchanged).

`_mask_line_comments()` blanks out `//`-to-end-of-line comment content before scanning for the statement-separating `;`. This turned out to be essential: both files' prose comments themselves contain semicolons (e.g. `// ...decision must cover; it is contractual, not a defect.`), so a naive `text.split(";")` cut a single comment sentence into two "statements" and a real Neo4j driver rejected the resulting fragment as invalid Cypher. This exact edge case was caught by the new **real-Neo4j integration test**, not the canned-row unit stubs — which is exactly how the original bug shipped undetected.

This is generic: any future multi-statement file dropped into `queries/validation/` is handled the same way, not just the two named here. A single-statement file (the other 3: `orphan_narrators`, `chain_integrity`, `collection_coverage`) round-trips through `_split_statements` unchanged, so their behavior is byte-for-byte identical to before.

**Part 2 — informational classifiers.** `graph_integrity_deferred_inventory` and `sanadset_orphan_inventory` are owner-accepted inventory/sizing reports (ADR-003, ADR-004) whose whole purpose is to SIZE a tracked, nonzero count — not a pass/fail signal. `_classify_informational_inventory` is registered for both in `_CLASSIFIER_REGISTRY`: an unconditional PASS that reports the per-block row counts in `details`, mirroring the `hadith_number_coverage` precedent in `src/parse/validate.py`.

## Test additions

- `tests/test_graph/test_validate.py`:
  - `TestSplitStatements` — `_split_statements` unit coverage, including the comment-semicolon regression case that mirrors the real file content.
  - `TestInformationalInventoryClassification` — both new classifiers always PASS regardless of row counts.
  - `TestMultiStatementExecution` — proves `run_validation` executes a multi-statement file as N separate `execute_read` calls (via a call-counting stub), preserves single-statement behavior, handles an unregistered future multi-statement file generically, and still hard-fails on a real per-statement error.
- `tests/integration/test_graph_loading.py` (`pytest.mark.integration`, skips cleanly without Docker): `TestPostLoadValidationAgainstRealFiles` runs `run_validation` against the **actual checked-in `.cypher` files** over a real `testcontainers` Neo4j — this is what caught the comment-semicolon edge case a synthetic fixture would have missed.

## No data re-run

Pure loader/validation code change. No data pipeline was run.

## Verification

- `uv run ruff format --check src/ tests/` — clean
- `uv run ruff check src/ tests/` — clean
- `uv run mypy src/` — clean (strict)
- `ENVIRONMENT=test uv run pytest tests/ -m "not integration"` — 1615 passed, 7 skipped
- `ENVIRONMENT=test uv run pytest tests/integration/test_graph_loading.py -m integration -k PostLoadValidationAgainstRealFiles` — 2 passed, against a real dockerized Neo4j and the actual repo files (confirmed the pre-fix `CypherSyntaxError` reproduces, then confirmed the fix resolves it)
- Regenerated `ontology/structural/` per this repo's own staleness gate

Closes #319

https://claude.ai/code/session_011hfJHDPdFJciPKPudSn7Q7


parametrization:
TechDebt: none

Reviewed at head 265a9492298d (`src/graph/validate.py`, both test files) via `gh api contents?ref=`. The split/mask/comment-semicolon handling is solid — `_mask_line_comments` + `_split_statements` correctly isolate top-level `;` from prose semicolons inside `//` comments, the header-comment-stays-attached and comment-only-segment-dropped edge cases are covered, and the real-Neo4j integration test against the actual checked-in `.cypher` files is exactly the right guard for a driver-level syntax error a canned-row stub can't reproduce. `_classify_informational_inventory` also lines up well with the `hadith_number_coverage` precedent in `src/parse/validate.py` — unconditional PASS while still reporting the metric (per-block counts + total) instead of silently swallowing it, and the two ADR-tied names are registered with exactly the file-stem keys `run_validation` uses as `query_name`.

One defect blocks merge, on the "generic default handling" claim itself (the PR states "any future multi-statement file dropped into `queries/validation/` is handled the same way, not just the two named here"):

`run_validation` builds `rows_for_classifier` for any multi-statement file (`len(statements) != 1`, `src/graph/validate.py` ~432-436) as one block-summary dict per statement: `{"block": idx, "row_count": len(rows)}`. That list is handed to whatever classifier is registered for `query_name` — and for a THIRD multi-statement file that isn't explicitly registered (i.e. anything other than the two ADR-003/004 names), that's `_classify_default`:

```
def _classify_default(query_name, rows, deviation_threshold):
    count = len(rows)
    passed = count == 0
    ...
```

`rows` here is the block-summary list, not the underlying query rows, so `count` is the **number of statement blocks**, not the number of result rows. Since this branch only runs when `len(statements) >= 2`, `count` is never 0 — `_classify_default` will report FAIL/`is_fatal=True` for every future unregistered multi-statement file **regardless of what the statements actually returned**, including a file where every block legitimately returns zero rows. That's the same shape of false failure this PR is fixing (a clean load tripping `sys.exit(1)` on an unrelated validation artifact), just relocated to the next person who drops in a multi-statement file without remembering to register a custom classifier for it.

The test that should catch this doesn't: `TestMultiStatementExecution::test_unregistered_multi_statement_file_executes_without_error` (lines ~527-538) drives a 2-statement file where both blocks return 0 rows via `_SequencedClient(rows_per_call=[[], []])`, but only asserts `len(client.queries_seen) == 2` and `query_name` — it never checks `.passed`/`.status`. Adding that assertion would have red-flagged this immediately (it would assert `passed is True` and fail against current behavior).

Suggested fix: give the default (or a new shared helper) real block-summary awareness — sum `row_count` across blocks and apply the zero-rows-is-pass rule to that total, rather than to `len(rows)` of the block list. Something like detecting the `{"block", "row_count"}` shape in `_classify_default` (or routing multi-statement unregistered files through a dedicated `_classify_default_multi_statement` that sums and compares to zero) would restore parity with the single-statement default behavior.

Everything else here is sound: per-statement timeout reuses the da#259 budget correctly (worth noting only that a 5-statement file's worst-case aggregate wall-clock is now up to 5x a single query's budget, not a blocking concern given these are lightweight aggregate MATCH/RETURN/count queries), the hard-failure-mid-file behavior (no silent partial pass) is correctly tested, and the concatenated per-block `details` string still surfaces full per-block visibility to the CLI's one-line-per-check print (`src/cli.py` `_cmd_validate`, `src/graph/__init__.py` `load_all`) since neither consumer truncates `.details`.

Please fix the default-classifier block-summary handling (with a test asserting `.passed is True` on the currently-uncovered all-zero unregistered-multi-statement case) and I'll take another pass.

parametrization:
TechDebt: none

Independent pass at head `265a9492298d` — read `src/graph/validate.py` plus both test files via `gh api contents?ref=`, and pulled the actual `queries/validation/*.cypher` files to check the split logic against real content rather than take the docstrings on faith.

On my assigned lens:

1. **Comment masking / string-literal risk.** `_mask_line_comments` walks the text char-by-char and only ever overwrites with a space, so length and every newline offset are preserved — indices found against the masked copy slice cleanly into the original. I grepped all five checked-in `.cypher` files for quote characters: every string literal present (`'unknown'`, `''`, `'sanadset'`, etc.) is a bare word or empty, none contains a semicolon. So for the two real multi-statement files the "no semicolon inside a string literal" claim in the docstring holds today. It's still a real gap for a hypothetical future file with a `;` embedded in a string — the split would silently mis-cut mid-statement — but that's a pre-existing generic limitation of the approach, not something this PR introduces or worsens, and it's disclosed in the docstring rather than glossed over.

2. **Single-statement graceful degradation.** Confirmed both by code (no `;` outside comments means the loop appends the whole trimmed text once) and by `TestSplitStatements::test_single_statement_no_semicolon_is_unchanged` plus `TestMultiStatementExecution::test_single_statement_file_still_executes_exactly_once`. Byte-for-byte identical to pre-da#319 for the 3 unaffected files.

3. **Informational-classifier scoping.** `_CLASSIFIER_REGISTRY` keys the always-PASS classifier to exactly the two literal `query_name` strings (`graph_integrity_deferred_inventory`, `sanadset_orphan_inventory`), so it can't silently swallow a future file's real regression — a new file gets a fresh `query_name` and falls through to `_classify_default` unless someone explicitly registers it. Scoping itself is fine.

4. **Per-statement error propagation.** `TestMultiStatementExecution::test_error_in_second_statement_is_hard_failure_for_whole_file` is a solid, real test — genuine `RuntimeError` on the 2nd of 2 calls, confirms `calls == 2` (no silent skip to a 3rd) and `is_fatal is True`. No partial-pass masking.

5. **Test adequacy.** `tests/integration/test_graph_loading.py::TestPostLoadValidationAgainstRealFiles` does load a real sample graph into a containerized Neo4j and run `run_validation` against the actual `queries/validation/*.cypher` files on disk (not a synthetic fixture) — exactly the guard needed for a driver-level `CypherSyntaxError` that a canned-row unit stub can't reproduce. That gap is closed.

Where this lands on point 3, though, is where I hit the same wall Alejandra already flagged, and I want to confirm it independently rather than just co-sign: it's real. `run_validation` (`src/graph/validate.py` ~430-436) builds `rows_for_classifier` for ANY file with `len(statements) != 1` as `[{"block": idx, "row_count": len(rows)}, ...]` — one dict per statement block, regardless of whether a classifier is registered for that `query_name`. If a third multi-statement file lands without a dedicated classifier, that block-summary list falls through to `_classify_default`, which does `count = len(rows)` — here `rows` is the list of block dicts, so `count` is literally the number of statement blocks (always >= 2 in this branch), never the number of returned rows. `passed = count == 0` can therefore never be true, so `_classify_default` will FAIL every future unregistered multi-statement file unconditionally, independent of what the underlying queries actually returned — including one where every block legitimately returns nothing. That reproduces the exact failure mode da#319 exists to fix, just one file-name away from being caught by the registry.

I traced the same test gap Alejandra called out: `test_unregistered_multi_statement_file_executes_without_error` (`tests/test_graph/test_validate.py:527`) drives a 2-statement, both-zero-row file through `_SequencedClient(rows_per_call=[[], []])` but only asserts call count and `query_name` — never `.passed` or `.status`. Add that assertion and it fails today, which is the tell.

Given a second independent read lands on the identical root cause, I'd treat this as confirmed rather than a one-reviewer false positive — please land the fix (summing `row_count` across blocks for the zero-rows-is-pass check, rather than counting blocks) plus a `.passed is True` assertion on that currently-uncovered path, and I'll re-review.

parametrization:
RequestOrReplied: Approved
TechDebt: none

Re-verified at `2283993` against my original blocking finding.

`_is_block_summary_rows` distinguishes the per-block summary dicts from real query rows via an exact `{"block","row_count"}` key-set match. Checked this against every registered classifier's actual output columns (`narrator_id`/`name`, `narrator_id`/`cycle_length`, `collection_id`/`expected`/`actual`/`deviation_pct`) — none alias to those two names, so the collision risk is low, and the new defensive case (a row carrying `block`/`row_count` plus an extra key) confirms the match stays exact rather than subset-based.

The single-statement path is untouched: `run_validation` still hands the raw row list straight to the classifier when there's exactly one statement, and the default classifier's non-summary branch keeps `count = len(rows)` exactly as before.

The summed-total semantics check out: all-zero blocks pass, and a mixed case with one block returning a nonzero count correctly fails on the sum rather than the block count (verified with a 2-block case landing on the actual row total, not `2`).

The new regression tests assert real outcomes (`.passed`/`.status`/`.row_count`/`.is_fatal`), not just call counts, and the all-zero-blocks case would have failed under the pre-fix `len(rows)` logic — so it actually covers the bug I flagged rather than just exercising the new code path.

No new issues introduced. Clearing my block.

parametrization:
RequestOrReplied: Approved
TechDebt: none

Re-verified at `2283993` against my original finding. The fallback classifier's blind spot is closed correctly.

`_is_block_summary_rows()` gates on an exact `{"block", "row_count"}` 2-key match, so it only ever fires on the summary dicts `run_validation` itself constructs for a multi-statement file — a real query row that coincidentally carries a `row_count` column alongside other columns is excluded (covered by `test_real_rows_named_block_and_rowcount_with_extra_keys_not_summary`), and the single-statement path is untouched (raw rows passed straight through, never wrapped).

`_classify_default` now sums `_row_count_of(row) for row in rows` when the summary shape is detected, mirroring the existing `_classify_informational_inventory` total — same aggregation idiom, no new pattern introduced. Since summed counts are non-negative, the total is zero iff every block's `row_count` is zero, so an unregistered multi-statement file with any genuine problem row in any block still sums to a nonzero total and correctly FAILs (`test_unregistered_multi_statement_nonzero_rows_fails`: block totals 3+0 → FAIL, `row_count == 3`, not the block count 2). The reported `row_count` on the result is the summed total in both the pass and fail cases, not `len(rows)`.

The registered ADR-003/ADR-004 classifiers (`graph_integrity_deferred_inventory`, `sanadset_orphan_inventory`) are unmodified — still routed to `_classify_informational_inventory`, still always-PASS, tests for both still green.

All 7 CI checks SUCCESS at this head. This fully resolves the blocking defect I raised.

PR #324, opened 2026-07-06
fix(parse): strip residual <NAR>/<SANAD>/<MATN> markup from sanadset matn (da#318)

### LINKED ISSUE #318: Hadith matn contains raw <NAR>/<SANAD>/<MATN> structural markup (parser leak)
## Summary

At least one hadith's displayed matn contains **raw structural markup** (`<NAR>`, `</NAR>`, `<NAR/>`, `<SANAD>`, `</SANAD>`, `<MATN>`, `</MATN>`) that leaked into the stored/served text instead of being stripped during parsing.

## Repro (staging)

1. Narrators → search `الزهري` → open **al-Zuhri** (`nar:5504c419-04c8-5f6d-b80a-91ee971246c8`)
2. Scroll to **Hadith Narrated (20 total)** → last row, hadith `hdt:sanadset:0:0:107270`
3. Displayed text contains literal tags, e.g. `… <NAR/> عاءشة <NAR> … </SANAD> </MATN> <NAR/> وهشام <NAR> …`

## Likely cause

The `<NAR>/<SANAD>/<MATN>` tags are structural annotations that should be consumed and stripped during the parse phase (matn/sanad segmentation). Their presence in the served matn indicates the parser retained the annotation markup in the `text`/`matn` field for at least this sanadset source. May affect a broader set of `sanadset:*` hadiths — needs a corpus-wide count.

## Investigation / acceptance criteria

- [ ] Count hadiths whose stored matn contains `<NAR|SANAD|MATN` markup (corpus-wide, not just this one)
- [ ] Fix the parser to strip structural tags from the matn text (or the loader to sanitize), idempotent re-run
- [ ] Unit test over a fixture matn containing the tags → clean output
- [ ] Frontend defense-in-depth: escape/strip any residual tags before render (belongs to isnad-graph if we want a display-side guard — cross-link)
- [ ] Re-verify `hdt:sanadset:0:0:107270` renders clean on staging

## Notes

Found during the #723 staging reload UI review. Not a reload regression — stored-data/parse issue surfaced by inspecting hadith display.

parametrization:
## Summary

Fixes **da#318** — hadith `matn_ar` leaks raw `<NAR>`/`<SANAD>`/`<MATN>` structural markup (parser leak).

### Root cause (verified at HEAD)
`_MATN_RE = re.compile(r"<MATN>(.*?)</MATN>", re.DOTALL)` + `_extract_tag()` do a single `pattern.search()`, pairing the **first** `<MATN>` with the **first** `</MATN>`. Malformed upstream `sanadset.csv` rows (e.g. `sanadset:0:0:107270`) carry **two** `<SANAD>…<MATN>` segments but only **one** closing `</MATN>`, so the capture spans first-open→lone-close and swallows a second literal `<MATN>` open, an orphan `</SANAD>`, and `<NAR>…</NAR>` tags into `matn_ar` (written verbatim). Measured blast radius: **87,739 / 650,986 = 13.5%** of hadiths carried `<NAR|SANAD|MATN` markup in `matn_ar`.

## Changes (one PR, two parts)

1. **Parser tag-strip** (`src/parse/sanadset.py`) — new `_STRUCTURAL_TAG_RE = re.compile(r"</?(?:NAR|SANAD|MATN)\s*/?>")` (covers self-closing `<NAR/>`) + `_strip_structural_tags()`, applied to the extracted matn **before store**. Idempotent; a **well-formed matn is returned unchanged** (early `search` guard — no whitespace normalization on clean text); markup rows have tags replaced with a single space then whitespace-collapsed so Arabic tokens are not silently concatenated. Mirrors the narrator-side `name_quality._MARKUP_RE` idiom from da#247. **`isnad_raw_ar` is NOT touched** — it legitimately carries `<NAR>` tags (74.5% of rows) that `_extract_narrator_mentions` consumes.

2. **Post-hoc scrub** (`scripts/scrub_matn_markup.py`) — sibling of `scrub_matn_canonical.py`; **imports and replays the parser's own `_strip_structural_tags`** over the persisted `matn_ar` of `data/staging/hadiths_sanadset.parquet`, writing to a **separate `--output-dir`** (read-only over input, refuses `output==input`). Intended for a **later owner-gated reload — NOT run here** (code + tests only, exactly as da#311).

## Tests
- Parser: malformed doubled-SANAD row (à la `sanadset:0:0:107270`) → stored `matn_ar` has **no** residual `<NAR>/<SANAD>/<MATN>` markup and retains both Arabic segments; **regression guard** that `isnad_raw_ar` still carries its `<NAR>` tags; well-formed row untouched; idempotency; self-closing tag; `None`/clean/all-markup passthrough.
- Scrub script: strip-matn + preserve-isnad, schema/row-count preserved, idempotency, no-input-mutation, `output==input` rejected.
- The parse-through markup test **FAILS against pre-fix code** and PASSES after (verified).

## Green-before-push
`ruff format --check` + `ruff check` clean, `mypy src/` (strict) clean, full non-integration suite **1631 passed / 7 skipped**. Structural ontology index regenerated + committed (pre-commit + pre-push hooks green).

## Out of scope (follow-up)
The secondary effect — the ~2 narrator mentions/malformed row that the doubled-SANAD segment contributes to the chain — is **not** recovered here; that is semantic re-segmentation, not tag-strip, and is out of the da#318 AC. Should be tracked as a follow-up if the chain-completeness matters for the affected rows.

## Reviewers
Suggested: Kavitha Sundaramurthy + Oyunbileg Batbayar (orchestrator coordinates the actual verdicts).

Closes #318


parametrization:
RequestOrReplied: Approved
TechDebt: none

Reviewed at head `eb82d3f4c2767f9b611cb5bfd43499248b7a05f1` via the GitHub contents API (no local checkout), focused on parser correctness for `src/parse/sanadset.py`, `scripts/scrub_matn_markup.py`, and the two test files.

**Regex correctness/completeness.** `_STRUCTURAL_TAG_RE = re.compile(r"</?(?:NAR|SANAD|MATN)\s*/?>")` covers open, close, and self-closing (`<NAR/>`, `<NAR />`) forms of all three known tags, matching the existing case-sensitive uppercase convention already assumed by `_SANAD_RE`/`_MATN_RE`/`_NAR_RE` elsewhere in this module — no new assumption introduced. I don't see a plausible over-strip case: the tag vocabulary (Latin `NAR`/`SANAD`/`MATN` inside angle brackets) has no legitimate collision with Arabic hadith prose, so scoping the strip to these three names is actually safer than a generic `<[^>]*>` sweep would be — a broader match risks eating unrelated bracketed content that might appear in a matn. Worth a very minor note: the docstring's "mirrors the narrator-side `name_quality._MARKUP_RE` idiom" is true of the general shape (replace-with-space, collapse-whitespace) but not the literal regex (`_MARKUP_RE` is the unscoped `<[^>]*>` plus a raw-bracket fallback) — not a correctness issue, just a slightly imprecise cross-reference.

**Idempotency and placement.** `_strip_structural_tags` guards on `search()` before doing any work, so a clean matn is returned byte-identical (confirmed by `test_strip_passes_through_clean_and_none`), and the cleaned output can never re-match (confirmed by `test_strip_is_idempotent`). It's wired in at exactly the right point — `matn_text = _strip_structural_tags(_extract_tag(_MATN_RE, full_text))` — after extraction, before the row dict is built and appended for store.

**`isnad_raw_ar` isolation.** Verified in the source: `sanad_text = _extract_tag(_SANAD_RE, full_text)` is assigned straight to `isnad_raw_ar` with no strip call anywhere in that path, and `_extract_narrator_mentions` / `_NAR_RE.findall` both still operate on the tag-bearing string. `test_malformed_row_isnad_raw_ar_still_carries_nar_tags` is a real regression guard here, not just an assertion of the status quo.

**Root cause vs. scope.** The fix removes the literal markup leak, which is the da#318 AC; it doesn't re-derive tag boundaries. One nuance beyond what's already flagged as follow-up (the ~2 lost narrator mentions/row): for a doubled-SANAD malformed row, the two swallowed matn fragments get space-joined into one string. In the test fixture (the actual "actions are by intentions" hadith) that happens to reconstruct a coherent sentence, and the root-cause section cites a real corpus row id, so this looks grounded in actual data rather than a hypothetical. But I'd flag for awareness (not blocking, and arguably out of the stated AC same as the mention-loss item) that if a doubled-segment row's two matn fragments are NOT a single continuous statement, the joined string could read as a garbled hybrid. This seems like the same class of "re-segmentation, not tag-strip" concern the PR already calls out of scope for narrator mentions — reasonable to fold into that same follow-up rather than block this PR on it.

**Scrub script.** `scripts/scrub_matn_markup.py` imports `_strip_structural_tags` directly from `src.parse.sanadset` rather than reimplementing the regex — single source of truth, so a scrubbed row is provably identical to what a fresh parse would produce. Read-only-over-input and `output==input` rejection are both exercised by tests. Not run against real data here, consistent with the stated owner-gated-reload intent.

CI is green across the board (Lint/Type Check/Test/Integration/staleness/sync-drift). No blocking issues from the parser-correctness lens.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

QA pass on `eb82d3f`, code read via `gh api contents?ref=`.

**Verified empirically, not just by inspection.** Cloned the repo fresh to scratchpad, fetched `pull/324/head`, ran the full `test_sanadset_parser.py` + `test_scrub_matn_markup.py` — 29 passed. Then swapped `src/parse/sanadset.py` back to its pre-PR parent (`ec0b2cb`, no `_strip_structural_tags`/`_STRUCTURAL_TAG_RE`) and re-ran just `TestMatnMarkupStrip`: `test_malformed_row_matn_has_no_residual_markup` genuinely fails against pre-fix code (`<SANAD>` survives in the captured matn, assertion trips on the literal match) — the claim in the PR body is real, not a tautology. The other three tests in that class fail on pre-fix code too, but only because `_strip_structural_tags` doesn't exist yet — expected. Restored the post-fix file afterward, diff clean.

Traced the regex composition by hand against the doubled-SANAD fixture to confirm the root-cause narrative: `_SANAD_RE`'s non-greedy first-open→first-close correctly isolates only the first segment for `isnad_raw_ar`; `_MATN_RE`'s equivalent non-greedy search is the one that spans across the second `<SANAD>…<MATN>` because there's only one `</MATN>` in the string — matches the PR's stated mechanism exactly.

Edge cases in `_strip_structural_tags` / `_STRUCTURAL_TAG_RE`: well-formed matn untouched including no whitespace/newline normalization (asserted directly), self-closing tag, idempotency, all-markup collapsing to None, None passthrough — all covered and all pass. The regex's `\s*/?>` tail means a stray substring like `<NARROW>` (leading with a tag name but not actually the tag) does not false-match, so there's no over-stripping risk on look-alike text; not explicitly tested but sound by construction. One minor coverage gap: no test exercises `_strip_structural_tags`/the parser against a mismatched case tag (`<nar>`) or a genuinely non-structural stray `<`/`>` in matn text — low risk given the corpus is documented as uppercase-tag-only and consistent with the existing `_NAR_RE`/`_SANAD_RE`/`_MATN_RE`, not worth a followup.

`isnad_raw_ar`-untouched regression guard is real and useful as a guard (it would also pass pre-fix since sanad extraction is a separate code path untouched by this change — that's expected for a guard against future regressions in this PR's diff, not a proof of the fix itself).

Scrub script (`scripts/scrub_matn_markup.py`): read-only over `--staging-dir` (verified by test + by reading the code — it never opens the input path for write), refuses `output-dir == staging-dir` via resolved-path comparison, preserves schema and row count (`set_column` swaps only `matn_ar`), leaves `isnad_raw_ar` and all other columns untouched, and is idempotent (re-scrubbing scrubbed output changes nothing further, asserted by dict-equality of two independent runs) — all backed by passing tests I re-ran locally. One data-safety observation, not blocking: the script only ever rewrites `hadiths_sanadset.parquet` into `--output-dir` — the sibling `narrator_mentions_sanadset.parquet` / `collections_sanadset.parquet` / `narrators_bio_kaggle.parquet` are never copied there. That's consistent with the PR's stated scope (matn tag-strip only, no re-segmentation), but whoever runs the eventual owner-gated reload needs to know `--output-dir` is a partial staging tree (hadiths only), not a drop-in full replacement for `data/staging` — worth a one-line callout in the reload runbook when that follow-up happens, not something this PR needs to change.

Statuses: all 7 CI checks green (`statusCheckRollup`) as of review time. No blocking gaps found.

PR #327, opened 2026-07-07
enrich(metrics): sampled betweenness centrality for tractability (da#326)

### LINKED ISSUE #326: Narrator centrality/community enrich not applied to current graph → no 'choke point' analysis (betweenness/pagerank/community/degree absent)
## Symptom (user priority: "choke points")

Narrator "choke point" / bottleneck analysis (high-betweenness transmitters, articulation points) cannot be surfaced — the API reads centrality fields that are **null for every narrator**.

## Root cause — enrich phase not applied to the current graph

Live check (stg + prod, read-only): all 150,187 `Narrator` nodes are missing `betweenness_centrality`, `pagerank`, `community_id`, `in_degree`, `out_degree` — the property **keys do not exist in the graph** (Neo4j "property key does not exist" warning), i.e. `src/enrich/metrics.py` (GDS betweenness / PageRank / Louvain + degree) was never run against this reload.

`src/api/routes/graph.py` reads these fields (`_NARRATOR_FIELDS`) and renders null importance for every node -> the graph looks like an undifferentiated hairball with no way to see who the pivotal narrators are.

The underlying edge data **is present** (2,682,069 `TRANSMITTED_TO` edges), so centrality is computable — it just hasn't been computed.

## Fix / rollout

Run the enrich metrics phase (`make enrich` / `src/enrich/metrics.py`) against the loaded graph so betweenness/pagerank/community/degree are written, on stg then prod. Confirm `metrics.py` GDS projection succeeds at prod scale (watch for the da#250-class intractable-query / timeout risk). Sequence AFTER the chain-id fix reload so enrich runs on a correct graph.

## Acceptance

- [ ] `betweenness_centrality`, `pagerank`, `community_id`, `in_degree`, `out_degree` populated on Narrator nodes
- [ ] `GET /graph/narrator/{id}/network` returns non-null centrality; top-betweenness query returns a ranked list of plausible choke-point narrators
- [ ] A "choke points" view (top betweenness / articulation) is demonstrable in the UI

parametrization:
## Summary

Makes the enrich **betweenness centrality** step tractable at prod scale so the metrics enrich can actually be applied to the current graph (root cause of #326: every `Narrator` has null `betweenness_centrality`/`pagerank`/`community_id`/`in_degree`/`out_degree` because `src/enrich/metrics.py` was never run).

**Why it was never run:** exact Brandes betweenness is O(V·E). At 150,187 Narrator nodes × 2,682,069 `TRANSMITTED_TO` edges that is ≈ 4×10¹¹ operations — computationally intractable (the da#250-class timeout/OOM risk the issue flags). The unsampled `gds.betweenness.write` call would run for many hours or exhaust memory.

**Fix:** run GDS's **sampled (approximate) Brandes** over a configurable number of pivot nodes. Approximate betweenness with a few thousand pivots is the standard, well-accepted method for choke-point **ranking** on graphs this size — the ranking is what the "choke points" view needs, not exact scores.

### Sampling default: 2000 pivots (rationale)
- `Settings.betweenness_sampling_size = 2000` (~1.3% of the 150k graph), deterministic via `betweenness_sampling_seed = 42`.
- 2000 pivots is ≈ **75× cheaper** than exact (2000·2.68M vs 150k·2.68M) and sits comfortably in the literature's accepted range (Brandes–Pich) for stable top-k ranking on 10⁵-node graphs.
- Config- and CLI-overridable (`--betweenness-sampling-size`); `0` = exact betweenness (small graphs / tests only).

### Design
- `run_metrics(..., sampling_size, sampling_seed)` — positive size ⇒ sampled Brandes with `samplingSize` + `samplingSeed`; `None`/`0` ⇒ exact. The **function** default is exact (backward-compatible, keeps small-graph/test paths exact); the tractable **2000** default lives at the config/CLI layer where the real large-graph run happens.
- Threaded through `run_all` → CLI `enrich`.

### Verified, not changed
- pageRank / louvain / degree are near-linear (or cheap) per iteration and already tractable at this scale — left unchanged.
- The GDS projection `gds.graph.project($name, 'Narrator', 'TRANSMITTED_TO')` and the write-back property names (`betweenness_centrality`, `pagerank`, `community_id`, `in_degree`, `out_degree`) already match the API's `_NARRATOR_FIELDS` **exactly** — no rename needed.

### Tests
Sampled path emits `samplingSize`/`samplingSeed`; exact path on unset and on explicit `0`; result-object flags on a sampled run; direct `_write_betweenness` helper coverage for both branches. Full enrich suite green (42 passed), mypy-strict + ruff clean via pre-commit/pre-push.

### Rollout note
This PR does **not** run enrich against any environment. The stg→prod enrich run is performed by the orchestrator after the da#325 chain-id reload lands (so enrich runs on a correct graph).

## Reviewers
- **Ivana Horvat**
- **Oyunbileg Batbayar**

Closes #326


parametrization:
Requestor: Ivana Horvat
RequestOrReplied: Approved
TechDebt: #127

## Review — sampled betweenness for tractability (da#326)

Reviewed the artifact at head `216a06b`, full end-to-end chain, ML/correctness lens.

### Statistical defensibility (2000 pivots, seed 42)
Sampled Brandes (Brandes–Pich) with ~2000 pivots over a ~150k-node graph is a sound default for **stable top-k choke-point ranking**. High-betweenness nodes lie on many shortest paths, so they are hit by many pivots and are precisely the sub-population whose estimates converge fastest — rank order of the top narrators stabilizes well below 2000 pivots at this scale. 2000 (~1.3% of nodes) is a defensible, slightly conservative default, and it is overridable via config and CLI, so if a later stability check wants more pivots that is a one-line change rather than a re-review. Not a blocker.

### Seed reproducibility — verified, not assumed
Threaded end to end. The GDS call embeds `samplingSeed` in the config map and binds it as a parameter, `run_metrics` and `run_all` carry `sampling_seed=42`, and the CLI path binds `settings.betweenness_sampling_seed`. Successive runs are deterministic.

### Exact path preserved
The guard `sampling_size and sampling_size > 0` routes None and explicit 0 to the exact branch, which emits only the write-property config with a name-only param map. `run_metrics` and `run_all` both default to None, so the small-graph and test behavior is genuinely unchanged.

### API property contract
The write-back property stays `betweenness_centrality` in both branches (no rename), and pagerank / community_id / in_degree / out_degree are untouched — the API read path over `_NARRATOR_FIELDS` is safe.

### GDS call shape
`gds.betweenness.write(graphName, {config})` with `samplingSize` / `samplingSeed` is the correct shape and is supported by the 2.13.8 target. `samplingSize` greater than node count degrades gracefully to effectively-exact inside GDS.

### Tests
Both branches are exercised — sampled emits size and seed (orchestration and helper level), exact is covered on both unset and explicit 0, and the result flags are asserted on a successful sampled run. Good coverage.

### One optional nit (not blocking)
A negative sampling size falls through the `> 0` guard to the **exact** path. That is the intractable branch this PR exists to avoid, so a fat-fingered `--betweenness-sampling-size -2000` would silently route prod into the da#250-class timeout rather than failing loudly. Exact is a safe-but-slow fallback so this is not a correctness defect on the intended-use path, but consider clamping-to-exact-with-a-warn-log or rejecting negatives at the CLI in a followup. Your call.

TechDebt refs the pre-existing deprecated `gds.graph.project` TODO already tracked by #127 — not introduced by this PR, carried forward.

Approving.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none (one non-blocking follow-up recommended below — reviewer's discretion to file)

## QA / data-quality review — da#326 sampled betweenness

Reviewed the artifact at head `216a06bf462cc112edff85ed85125374cd82a22b` (full `src/enrich/metrics.py`, `src/enrich/__init__.py`, `src/cli.py`, `src/config.py`, `src/models/enrich.py`, and `tests/test_enrich/test_metrics.py` read at the head SHA, not the PR-body framing). Verdict: Approved. The change is correct, the unit tests exercise the real emitted GDS call rather than a vacuous mock, and the critical production seam is safe.

**1. Test coverage is real (strong).** The new tests assert the actual Cypher emitted and the actual param dict, not just that a method was invoked. The sampled test checks both `samplingSize` and `samplingSeed` appear in the query AND that the param values are 2000 / 42. The unset path asserts the sampling keys are absent from the query and that the param dict is exactly the bare graph-name form. The explicit-0 path asserts the exact branch is taken. The `_write_betweenness` helper is tested directly for both branches (positive size, and None). The `_betweenness_call` helper asserts exactly one betweenness call, which guards against a double-write masking a wrong branch. This is the right shape.

**2. Determinism.** The fixed seed reaching the GDS call is asserted at the param level (42 in the run path, 7 in the helper path). True run-to-run reproducibility is a GDS-library property that a mock cannot exercise; asserting the seed is threaded to the call is the correct unit-level proof, so this is adequate.

**3. Regression safety — confirmed no rename.** The write properties the API reads are all intact at head: betweenness_centrality, pagerank, community_id, out_degree, in_degree. The sampled write still targets writeProperty betweenness_centrality. No property-name drift.

**4. Production seam is safe (the thing I most wanted to confirm).** The full-pipeline command routes stage 5 through `_cmd_enrich()` with no args, so the sampling arg is None and resolves to the configured default of 2000 → sampled path. No production entrypoint calls `run_all` / `run_metrics` in a way that would silently run intractable exact Brandes at 150k-node scale. The `run_all` default of None (exact) only affects small-graph/test callers, which is the intended safe default.

**5. Observability.** An operator can see a sampled run ran sampled via the structured log event carrying mode plus size plus seed. Note the result model itself does not carry the mode — visibility is log-side only.

**6. CI.** All seven checks green (Lint, Type Check, Test, Integration Tests, staleness-check, base-image, sync-drift). Structural ontology index was regenerated and committed to match the new symbols.

### Non-blocking follow-up (recommended, not required for this merge)

There is one genuine coverage seam: the override-precedence logic in `_cmd_enrich` (CLI arg wins, else configured default) and the parameter threading through `run_all` are correct by inspection but not covered by a test. If that wiring is later refactored, nothing would catch a silent propagation regression — the classic failure mode for a config/CLI knob. I verified it correct at head, so this is a hardening item, not a defect. Suggest a small end-to-end test (configured default → CLI override → value reaching the GDS call) as a follow-up issue at the author's/reviewer's discretion. Optionally, surfacing the sampling mode on the result model would let an operator read it without grepping logs. Neither blocks this merge.

PR #328, opened 2026-07-07
fix(graph): canonicalize TRANSMITTED_TO.hadith_id to match Hadith.id (da#325)

### LINKED ISSUE #325: Isnad chain empty for every hadith: TRANSMITTED_TO.hadith_id bypasses hadith_node_id() → id-namespace mismatch with Hadith.id
## Symptom (user-reported, prod + stg)

Selecting any search result shows an **empty isnad chain** — for every hadith, in both environments.

## Root cause — hadith-id namespace mismatch between node and edge loaders

`GET /graph/hadith/{id}/chain` reconstructs the chain from `TRANSMITTED_TO` edges filtered by `hadith_id`:

```
MATCH (src:Narrator)-[t:TRANSMITTED_TO {hadith_id: $id}]->(tgt:Narrator) ...
```

But the two loaders build the hadith id differently:

| | value | via |
|---|---|---|
| `Hadith.id` (node) | `hdt:sanadset:0:0:2326` | `hadith_node_id()` — adds `hdt:`, collapses the main#139 double-prefix |
| `TRANSMITTED_TO.hadith_id` (edge) | `sanadset:sanadset:0:0:2326` | **raw** `resolved[i].get("hadith_id","")` — no canonicalization |

Live evidence (both boxes, read-only): intersection of the two id sets is **0** (0 of 523,273 distinct edge ids match any of 776,239 Hadith ids). So the join returns nothing for 100% of hadiths.

Transform that reconciles them: `hdt:` + collapse-double(edge_id) lands exactly on a real node — e.g. edge `sanadset:sanadset:0:0:2326` -> node `hdt:sanadset:0:0:2326` (verified present).

## Fix

`src/graph/load_edges.py` `_build_chain_pairs()` stores `resolved[i].get("hadith_id","")` verbatim. Route it through `hadith_node_id()` (the same canonicalizer the node loader already uses — its own docstring: "THE one prefixing rule ... every ingest path must route through this"). `hadith_node_id()` both adds `hdt:` and collapses the double-prefix via `bare_source_id`, so one change fixes both facets. Audit the sibling edge writers in the same file (`load_edges.py:167,175`) for the same bypass.

## Scope / rollout

Code fix is small; requires re-loading `TRANSMITTED_TO` edges (or a full graph reload) so edges carry canonical ids, then verify chain non-empty on stg, then prod. Distinct from da#282 (multi-isnad flattening) — that is chain topology; this is the id join returning empty.

## Acceptance

- [ ] Edge `hadith_id` == `Hadith.id` for the same hadith (id-set intersection > 0, ~= hadith count)
- [ ] `GET /graph/hadith/{id}/chain` returns nodes+edges for a known multi-narrator hadith
- [ ] Regression guard in `src/graph/validate.py`: assert TRANSMITTED_TO.hadith_id references an existing Hadith node

parametrization:
## Summary

Fixes the isnad chain coming back **empty for every hadith** (stg + prod). Root cause is a hadith-id namespace mismatch between the node and edge loaders: `GET /graph/hadith/{id}/chain` joins `TRANSMITTED_TO.hadith_id` against `Hadith.id`, but the chain loader stored the **raw** staging id (`sanadset:sanadset:0:0:2326` — double-prefixed, no `hdt:`) on each edge while Hadith nodes are canonical (`hdt:sanadset:0:0:2326`). The two id sets never intersect, so the join returns nothing.

Closes #325

## Changes

1. **Loader fix** — `src/graph/load_edges.py` `_build_chain_pairs()` now routes the edge `hadith_id` through `hadith_node_id()` (the same canonicalizer the node loader and every sibling edge writer use — it adds `hdt:` **and** collapses the main#139 double-prefix). Empty/missing ids are preserved as `""`, never a bare `"hdt:"`.
2. **Regression guard** — `queries/validation/transmitted_to_hadith_ref.cypher` + `_classify_transmitted_to_hadith_ref` in `src/graph/validate.py` assert every distinct `TRANSMITTED_TO.hadith_id` references an existing `Hadith` node (0 dangling = pass). Auto-wired via the `run_validation` glob and the classifier registry. This is the check that would have caught the bug.
3. **In-place migration** — new `migrate-transmitted-hadith-ids` CLI command (logic in `src/graph/migrate.py`) remediates the ~2.68M edges already in a live graph (a re-load from parquet isn't possible against a live box). It reads `DISTINCT t.hadith_id`, maps raw→canonical in Python, and applies a batched `UNWIND $rows … MATCH ()-[t:TRANSMITTED_TO {hadith_id: r.raw}]->() SET t.hadith_id = r.canon`. Idempotent (already-canonical ids are no-ops via `hadith_node_id`, empty/null skipped), batched, prints a summary.

   Invocation: `uv run isnad-ingest migrate-transmitted-hadith-ids [--batch-size 1000]`

4. **Tests** — loader emits canonical id and preserves `""`; validate guard passes on 0 / fails on a dangling id; migration maps raw→canonical incl. the `sanadset:sanadset:…` → `hdt:sanadset:…` collapse and is a no-op on re-run. Updated the two existing chain-pair equality assertions and the integration validation-file-set assertion.

## Rollout note

This PR is code only. The stg→prod graph remediation (running the migration against each box) is sequenced by the orchestrator — **not** run here.

## Verification

- `ENVIRONMENT=test uv run pytest` — 1693 passed, 7 skipped (incl. the real-Neo4j integration validation suite; the new guard runs green against a real box).
- Full local pre-commit + pre-push hook set (ruff format/lint, gitleaks, actionlint, cspell, dockerfile-base-pin, fixture-realism, structural-ontology-staleness, mypy-strict, pytest) — all Passed. Structural index regenerated and committed.

## Reviewers

Requesting review from **Nikolaos Papadopoulos** and **Alejandra Reyes-Fuentes**.


parametrization:
RequestOrReplied: Approved
TechDebt: none

Reviewed at head `01c536a08a7b21223198e60e985814e2f26841b3` (data-engineering + correctness lens). This is a correct, well-scoped, idempotent fix. Approving.

## Write-site (load_edges.py)
`_build_chain_pairs` is the single funnel into the `TRANSMITTED_TO` write batch — `all_pairs` -> `valid_batch` -> `execute_write_batch(_TRANSMITTED_TO_QUERY)`. The emitted edge id now routes through `hadith_node_id(raw)` with a `raw or ""` guard, so a missing id stays `""` and never becomes a bare `hdt:`. The other `hid` read inside `_load_transmitted_to` (the `row.get("hadith_id") or row.get("source_hadith_id")` at the top of the loop) is only a grouping key for `by_hadith`, never written onto the edge — so it is not a second raw write-site. Grepped every hadith-id edge writer in the module: the four siblings (NARRATED, APPEARS_IN, PARALLEL_OF, GRADED_BY) already canonicalize through `hadith_node_id`. No raw write-site missed; the module is now consistent.

## Canonicalizer
Confirmed `hadith_node_id` -> `bare_source_id` -> strip-leading-`hdt:` then `_collapse_double_corpus` is idempotent and double-corpus-collapsing, so re-deriving a canonical id is a no-op. Single source of truth, reused by both the write-site and the migration.

## Migration (migrate.py)
- Single-source transform: `compute_rewrites` calls `hadith_node_id`, not a re-implemented rule.
- Idempotent: rows where `canon == raw` are dropped, so an already-canonical graph yields `[]` and issues no write. `test_idempotent_second_run_writes_nothing` locks this in.
- Transaction size is bounded and safe. The batch loops over DISTINCT raw ids (default 1000 per write), and each id's `MATCH ()-[t {hadith_id: row.raw}]->() SET ...` rewrites only the edges that share that one hadith id, i.e. (chain length - 1) edges. Isnad chains are bounded to ~tens of narrators, so the worst-case write tx is ~batch x max-chain = low tens of thousands of edges, not the hundreds-of-thousands single-tx blow-up. At 2.68M edges / ~523k distinct ids (~5 edges/id avg) this is ~523 small write txns. No OOM risk.
- Edge counters correctly come from `RETURN count(t)` rather than the driver's created-counters (which stay 0 on a property SET).

## Guard (validate.py + cypher)
The cypher takes DISTINCT `TRANSMITTED_TO.hadith_id`, OPTIONAL MATCHes a `Hadith {id}`, and returns the ids with no node; classifier passes only at 0 rows. Registered in `_CLASSIFIER_REGISTRY` and added to the integration file-run set. This asserts every non-empty edge id references a real Hadith, so the empty-chain regression cannot silently return. The `<> ""` filter intentionally exempts the legitimate missing-id sentinel, consistent with the write-site guard.

## Tests
Cover write-site canonicalization, already-canonical passthrough, missing-id-preserved-as-empty, migration raw->canon mapping in the write batch, skip-empty/null, and the validation pass/fail classification. Good coverage of the stated surface.

## Advisory (optional followups, not blocking)
1. Migration read cost, not correctness: `MATCH ()-[t:TRANSMITTED_TO {hadith_id: row.raw}]->()` has no endpoint label. Absent a relationship-property index on `TRANSMITTED_TO(hadith_id)`, each distinct id may scan the relationship store, so the one-time run could be slow (correct, just slow). Worth confirming/creating that index before running on the live box, or accepting the runtime. Operational note only.
2. Pre-existing asymmetry (not introduced here): `_build_chain_pairs` reads only `hadith_id`, while `by_hadith` grouping and the NARRATED loader fall back to `hadith_id or source_hadith_id`. Any resolved mention carrying only `source_hadith_id` would get an empty edge id (and the guard's `<> ""` filter would not flag it). Prod edges are populated, so this does not manifest today and is orthogonal to da#325 — noting for awareness only.

parametrization:
RequestOrReplied: Approved
TechDebt: none

Reviewed at head 01c536a08a7b21223198e60e985814e2f26841b3 against the artifact, with an identity/id-correctness and completeness lens.

## Canonicalization correctness — confirmed

The load-bearing question was whether, after migration, a `TRANSMITTED_TO.hadith_id` equals the `Hadith.id` it must join against. It does, because both sides now derive through the single `hadith_node_id` canonicalizer:

- Hadith **node** id at `src/graph/load_nodes.py::_load_hadiths` (line 363) is `hid = hadith_node_id(sid)`.
- The edge write-site (`_build_chain_pairs`) and the migration (`compute_rewrites`) both call the same `hadith_node_id`.

Since `hadith_node_id` = `hdt:` + `bare_source_id`, and `bare_source_id` strips a leading `hdt:` then collapses via `_collapse_double_corpus`, the raw `sanadset:sanadset:0:0:2326` and the node built from the same staging `source_id` converge on `hdt:sanadset:0:0:2326`. Verified in `src/parse/identity.py`.

The collapse is corpus-general, not sanadset-specific: `_collapse_double_corpus` collapses any doubled leading segment where `segments[0] in SOURCE_CORPORA`. So the concern about a partial fix that handles `sanadset:sanadset:` but misses another double-prefixed collection does not apply — any known-corpus double is collapsed the same way. The transform is exercised against the realistic `sanadset:sanadset:...` shape and the single-prefix `sanadset:...` shape in the migration and loader tests, asserting the concrete `hdt:sanadset:0:0:2326` result rather than a mere `hdt:` prefix check.

## Idempotency / no-op — confirmed

`compute_rewrites` skips empty/None (`if not raw or not isinstance(raw, str): continue`) and skips already-canonical values (`if canon != raw`), so a re-run over a migrated graph returns `[]` and issues no write. The DISTINCT source query also filters `IS NOT NULL AND <> ""`, double-guarding the empty case. On the loader side, `hadith_id = hadith_node_id(raw_hadith_id) if raw_hadith_id else ""` preserves a missing id as `""` and never a bare `hdt:`. The idempotent and empty-skip behaviors are directly asserted in the tests.

## Completeness of the write-site fix — confirmed

`TRANSMITTED_TO` gets its edge `hadith_id` only from `_build_chain_pairs`, which is fixed. In `_load_transmitted_to` the raw `hid` at line 255 is used purely as a `by_hadith` grouping key and provenance filter — it is not written to any edge. The sibling edge writers already canonicalize: NARRATED (`hadith_node_id(hid)`), APPEARS_IN (`hadith_node_id(sid)`), grading and chain-node loaders likewise. No remaining raw-write path re-introduces the defect on the next load.

## Double-prefix (main#139) — confirmed clean

This fix routes through the same `_collapse_double_corpus` that resolves main#139; it does not reintroduce or mishandle that class, and the guard cypher plus `_classify_transmitted_to_hadith_ref` give a standing regression check (0 dangling edge ids = pass).

## Minor, non-blocking observations

- The migration's `edges_updated` summation (`result[0].get("updated")`) is not exercised by the unit suite — `MockNeo4jClient.execute_write` returns `[]`, so only the distinct-id counts are asserted. The production path (`RETURN count(t) AS updated`) is trivially correct, so this is a coverage note rather than a defect; no followup needed.
- `transmitted_to_hadith_ref.cypher` uses `LIMIT 100`, so if more than 100 distinct edge ids ever dangle the reported magnitude caps at 100. The pass/fail semantics are unaffected (any count > 0 fails), which is what the guard needs.

Approving. The fix is correct at the identity layer, complete at the write-site, idempotent, and the migration produces the same canonical id the loader and existing Hadith nodes carry.

PR #334, opened 2026-07-08
fix(compose): apply canonical-hadith gate to chain edges — drop fawaz six-books orphans (da#333)

### LINKED ISSUE #333: Composition dedup gate not applied to chain edges → fawaz six-books orphaned chains (dup of lk)
## Bug

The owner-confirmed corpus composition (da#191, `src/parse/composition.py`) makes `lk` the six-books canonical spine and loads `fawaz` for its UNIQUE collections only (Nawawi, Dehlawi, Qudsi) — fawaz's six-books hadiths are deliberately deduped to `lk`. The **node** loader (`src.graph.load_nodes`) correctly applies `is_canonical_hadith` / `HADITH_COMPOSITION` and excludes fawaz's six-books nodes.

But the **chain-edge** path never applies that gate. So fawaz's NER-derived six-books chain mentions load anyway, producing ~196,250 orphaned `TRANSMITTED_TO` edges keyed `fawaz:<book>:<n>` (bukhari/muslim/nasai/abudawud/ibnmajah/tirmidhi/malik) whose Hadith nodes don't exist. These duplicate `lk`'s canonical chains (verified: lk six-books hadiths already carry their own chains at 99.8% coverage), so a fresh `run_all` re-creates the orphans.

## Fix

Apply the composition gate to the chain-edge / narrator-mention loading path so a chain mention whose hadith would NOT be loaded as a canonical node (per `is_canonical_hadith` / `HADITH_COMPOSITION`) is excluded from `TRANSMITTED_TO` (and NARRATED, if applicable) edge production — mirroring the node dedup.

**Critical carve-out:** `mis` is intentionally chains-only for Sahih Muslim ("`network_edges_mis` still loads as narrator-transmission edges" — composition.py), and its chains currently resolve (there are NO `mis` dangling edges — they key to canonical/lk-resolving ids). The gate MUST drop fawaz's six-books orphans WITHOUT dropping mis's intended chains. Verify the exact keying of mis chain edges before choosing the gate predicate (e.g. gate on "the mention's hadith_id resolves to a canonical node per composition" rather than a blunt per-corpus exclude).

## Acceptance criteria

- [ ] A fresh chain-edge build excludes fawaz six-books chain mentions (their hadiths aren't canonical nodes).
- [ ] `mis` multi-isnad chains for Muslim are RETAINED (regression test).
- [ ] `lk` six-books chains and all other canonical chains are unaffected.
- [ ] Unit tests cover: fawaz-six-books mention dropped, mis mention kept, lk/canonical mention kept.
- [ ] CI green; local hooks mirror CI.

## Notes

Pairs with a graph-ops migration (deploy) to delete the 196,250 already-loaded orphaned edges. Root-cause detail: `.claude/memory/project_fawaz_lk_namespace_orphan_chains.md` (org repo). Owner approved the drop-duplicates direction 2026-07-07.

parametrization:
Closes #333

## Problem

The owner-confirmed composition dedup (da#191, `src/parse/composition.py`) makes `lk` the six-books canonical spine; `fawaz` loads only its unique collections (Nawawi/Dehlawi/Qudsi). The **node** loader applies `is_canonical_hadith` / `HADITH_COMPOSITION` and excludes fawaz's six-books nodes — but the **chain-edge** path never did. So fawaz's NER-derived six-books chain mentions loaded as ~196,250 orphaned `TRANSMITTED_TO` edges keyed `fawaz:<book>:<n>` with no Hadith node, duplicating lk's canonical chains.

## Fix

- Add `composition.is_canonical_hadith_id(hadith_id)` — the id-parsing complement to `is_canonical_hadith`. It recovers `(corpus, collection)` from the `source_id` grammar (`<corpus>:<collection>[:<part>...]`, handling `hdt:` prefix + main#139 double-prefix collapse) and applies the same gate.
- `_load_transmitted_to` and `_load_narrated` now skip a mention whose hadith would **not** be a canonical node — mirroring the node dedup. A fresh `run_all` no longer re-creates the orphans. Drops are counted (`dropped_noncanonical`) in the loader logs.

## mis carve-out (verified)

I investigated how mis chains are keyed before choosing the predicate. Finding:

- mis's transmission edges are produced as **`network_edges_mis.parquet`** (`mis.py` → `NETWORK_EDGE_SCHEMA`, keyed `mis:sahih_muslim:<book>:<n>` from its own `id_map`), **not** as narrator mentions.
- NER (`resolve/ner.py`) produces `narrator_mentions_resolved.parquet` only for `sanadset, lk, thaqalayn, open_hadith, fawaz, sunnah` — **mis is absent from every source list**. So the mention path this gate touches contains **zero** mis rows.

**Subtlety worth noting** (the issue anticipated mis might carry `lk:muslim` ids — it does not): mis is keyed `mis:sahih_muslim`, which is genuinely non-canonical (`is_canonical_hadith("mis", …)` is `False`, since mis loads no Hadith nodes). The carve-out therefore holds **structurally**, not by the predicate keeping mis: mis edges never traverse `_load_transmitted_to`/`_load_narrated`, so the gate cannot drop them. The network-edge loader routes them by declared relation exactly as before (da#133). A regression test pins this.

## Tests

- `test_composition.py::TestIsCanonicalHadithId` — fawaz six-books dropped, fawaz-unique kept, lk/sanadset kept, mis flagged non-canonical, `hdt:`/double-prefix/bare-corpus handling.
- `test_load_edges.py::TestCompositionGateOnChainEdges` — fawaz six-books → 0 `TRANSMITTED_TO` + 0 `NARRATED`; fawaz-unique & lk kept; **mis network-edge carve-out** (co-present mis file untouched while fawaz mention is dropped).
- Full suite green: `mypy src/` clean; `pytest -m "not integration"` 1666 passed / 7 skipped.

## Follow-up

Pairs with a graph-ops migration (deploy) to delete the 196,250 already-loaded orphaned edges.

## Reviewers

Requesting review from **Nikolaos Papadopoulos** and **Alejandra Reyes-Fuentes**.


parametrization:
## Review — da#333: composition gate on chain edges (head `8ee0a90`)

Read at HEAD via the GitHub API (`composition.py`, `identity.py`, `load_edges.py`, `enums.py`, `base.py`, `fawaz.py`, and the added tests). Reviewed against the five predicate-correctness lenses.

**1 — Id-grammar parse (correct).** `is_canonical_hadith_id` routes through `bare_source_id`, which strips a leading `hdt:` graph prefix and then collapses a doubled leading corpus via `_collapse_double_corpus`, so both the raw staging id and the `hdt:`-prefixed node id canonicalize to the same body before the split. Taking `corpus = seg[0]` / `collection = seg[1]` recovers exactly what `generate_source_id(corpus, collection, *parts)` encoded as the leading two segments — I confirmed the builder joins `[corpus, collection, *parts]`, so the recovered pair is byte-identical to the `source_corpus`/`collection_name` columns the node loader reads. Verified on `fawaz:bukhari:1` → (fawaz, bukhari), `lk:bukhari:1` → (lk, bukhari), `sanadset:0:0:2326` → (sanadset, 0), and the main#139 doubled form `fawaz:fawaz:bukhari:1` → (fawaz, bukhari).

**2 — fawaz six-books reliably dropped (correct).** The fawaz parser derives its collection slug from the edition key (`eng-bukhari` → `bukhari`, via `_collection_name_from_key`) and builds ids through `generate_source_id`, so every six-books hadith is keyed `fawaz:<book>:<n>`. `fawaz` is present in `HADITH_COMPOSITION` with the allowlist `{nawawi, dehlawi, qudsi}`, so any six-book slug (bukhari/muslim/nasai/abudawud/ibnmajah/tirmidhi/malik) is not in the set and resolves to False. The added test loops the full seven-book set and asserts the drop.

**3 — No over-drop of canonical chains (correct).** `lk` is absent from the composition map, so `is_canonical_hadith` short-circuits to True for all lk collections including muslim — I confirmed `lk:muslim:1` is kept. fawaz-unique (nawawi/dehlawi/qudsi) is in the allowlist and kept; `sanadset` maps to `None` (admit-all, da#219) and is kept for any collection segment including the numeric-book form. halimbahae six-books resolving False is correct, not an over-drop — those books are also deduped to the lk spine, so the edge path rightly mirrors the node loader there.

**4 — Default-keep for bare/unknown ids (correct and conservative).** A collection-less id (`h1`, `hdt:h1`) or an unlisted corpus defers to `is_canonical_hadith`, which keeps anything not named in the map — the right direction for an irreversible load. A real fawaz six-books mention can never slip through this path: fawaz source_ids are always minted with a collection segment via `generate_source_id`, so a six-books mention always carries `seg[1]` and hits the explicit-drop branch, never the default-keep branch. The keep branch only fires for genuinely id-less/toy values.

**5 — Loader integration (clean, no collateral).** The gate sits ahead of the provenance/pair-build logic in both `_load_transmitted_to` and `_load_narrated`, with `dropped_noncanonical` counted and surfaced in the structured log. The da#228 curated muhaddithat orphan-links are unperturbed: their `hadith_id` is a real `sunnah` id, `sunnah` is not in the composition map, so the mention passes the canonical gate and continues to the existing provenance branch that keeps it NARRATED-only. Missing-endpoint accounting improves rather than regresses — mentions that were previously counted as `missing_endpoints` against a never-loaded node are now attributed to `dropped_noncanonical` at the source, which is the accurate classification. `mis` never reaches this narrator-mention path (its edges come from `network_edges_mis.parquet`), so its empty-frozenset entry is a harmless no-op here.

Test coverage is comprehensive for the predicate: fawaz-drop (all seven books), fawaz-unique-keep, lk-keep (incl. muslim), mis-drop, sanadset-keep, `hdt:` handling, double-prefix collapse, and the bare/unknown default-keep are each asserted. The change is a faithful edge-path mirror of the node dedup and closes the ~196k orphaned `fawaz:<book>:<n>` `TRANSMITTED_TO`/`NARRATED` sibling defect. No changes requested.

RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review — da#333 composition gate on chain edges

Reviewed at head `8ee0a90fdf95c16d3b05a13aedaf3800d7c16331`, reading every file via the GitHub API at that SHA (no working-tree checkout). Focus per the brief: independently verifying the `mis` carve-out, integration soundness, test coverage, complementarity with deploy#539, and the regenerated structural index.

### The mis carve-out — independently verified at source (not just PR body)

I confirmed all three legs of the carve-out from the real code, not the PR description:

- **(a) NER never runs over `mis`.** `src/resolve/ner.py` drives extraction over exactly four explicit corpus sets — `_PHASE1_MENTION_SOURCES` (`sanadset`, `lk`), `_ARABIC_SOURCES` (`thaqalayn`, `open_hadith`, `fawaz`), `_ENGLISH_SOURCES` (`sunnah`), and `_SKIP_SOURCES` (`muhaddithat`). The `run` driver iterates only those sets (Steps 1-3); there is no all-corpora glob. `mis` appears in none of them, so no row with `source_corpus="mis"` is ever written to `narrator_mentions_resolved.parquet`.
- **(b) `mis` emits its edges elsewhere.** `src/parse/mis.py` writes `network_edges_mis.parquet` (line 496) via its own id_map, keyed `mis:sahih_muslim:<book>:<n>` through `generate_source_id("mis","sahih_muslim",...)`, with each row declaring `relation = EDGE_RELATION_TRANSMITTED_TO` (line 332) — never a narrator mention.
- **(c) The gated path never reads that file.** The gate lives in `_load_transmitted_to` / `_load_narrated`, which read only `_resolved_mentions_files(...)` (the curated resolved-mentions + staging `narrator_mentions_*`). `network_edges_mis.parquet` is read only by `_load_studied_under`, which routes by declared relation and skips the `TRANSMITTED_TO` rows (da#133). The mention-path gate cannot touch it.

Conclusion: the gate drops the co-present fawaz six-books chain and leaves the mis network-edge file untouched. The carve-out is real and structural, exactly as claimed.

### The dependency behind point 2, stated explicitly

`is_canonical_hadith_id("mis:sahih_muslim:...")` returns False, because `mis` maps to `frozenset()` in `HADITH_COMPOSITION`. That value is harmless ONLY because leg (a) holds — mis never reaches the gate. If a future change ever added `mis` to any NER source set, the gate would silently drop every intended Sahih Muslim chain. The safety here rests entirely on the NER exclusion, not on any property of the gate itself. It is well-documented in the code comments; I flag it so the coupling is explicit for whoever next touches the NER source lists.

### Same-authority check

The edge gate uses the identical map as the node loader: `_load_hadiths` (`src/graph/load_nodes.py:344`) gates on `is_canonical_hadith(source_corpus, collection_name)`, and `is_canonical_hadith_id` is a thin id-parsing wrapper over that same function. Node dedup and edge dedup therefore agree by construction — the correct single-source-of-truth pattern. Parsing edge cases check out: `hdt:` prefix stripping, main#139 double-prefix collapse, and the bare/unlisted-corpus conservative-keep default all behave as the tests assert.

### Test coverage

`TestCompositionGateOnChainEdges` is solid. Drop-assertions are real, not vacuous — the fawaz cases assert that no `TRANSMITTED_TO ... MERGE` / no `NARRATED` query is ever issued to the mock client, not merely that a counter is zero. Positive cases (fawaz unique collections kept, lk kept for both edge types) prove the gate is selective rather than a blanket drop. `test_mis_network_edges_unaffected_by_mention_gate` does exercise genuine co-presence: a fawaz six-books mention file AND a `network_edges_mis.parquet` in the same run, asserting the fawaz chain is dropped while the mis file is routed off STUDIED_UNDER without error.

One minor, non-blocking suggestion: the mixed-selectivity case (a single resolved-mentions file holding both a canonical kept mention and a non-canonical dropped mention) is covered across separate test runs rather than within one file read. The per-row logic makes that equivalent, so this is a nice-to-have, not a gap I would hold the PR on.

### One pre-existing observation, out of scope for this PR

While tracing leg (c), I noticed that `mis`'s `network_edges_mis.parquet` rows (declared `TRANSMITTED_TO`) are not consumed by any loader in `load_all_edges` — `_load_studied_under` is the only reader of `network_edges_*` and it loads STUDIED_UNDER rows only. So mis transmission edges do not currently land in the graph via any path. This is pre-existing behaviour, unchanged and uncaused by this PR, and may well be intentional; I raise it only so the team can confirm separately whether mis transmission edges are meant to be loaded. It does not affect this PR's correctness or the carve-out.

### deploy#539 complementarity

The two are complementary with no overlap and no gap: this PR stops NEW mention-derived orphans at load time (chiefly the ~196k `fawaz:<book>:<n>` TRANSMITTED_TO edges, whose narrator-narrator endpoints both resolve, so the endpoint check never caught them), while deploy#539's prune-orphans removes the orphans ALREADY loaded. A complete fix needs both — this one for future re-loads, that one for the current graph.

### Structural index

`ontology/structural/{code-graph.json,llms.txt}` changes are consistent with a clean regeneration: the new `is_canonical_hadith_id` node, the new import edges (`load_edges → composition`, `composition → identity`), the new test methods, the mechanical line-number shifts, and the header counts (nodes 2850→2867, edges 4373→4394) all match the source deltas. No hand-edit signature.

Approving.

RequestOrReplied: Approved
TechDebt: none

PR #336, opened 2026-07-08
fix(pipeline): make enrich audit-write best-effort (da#335)

### LINKED ISSUE #335: enrich: read-only-fs audit-write crashes a successful centrality run (graph-ops)
## Summary

The `enrich` CLI subcommand aborts with `OSError: [Errno 30] Read-only file system: 'data'` when run inside the `graph-ops` container (deploy `graph-ops.yml`, operation=enrich). The crash is in `write_audit_entry` (`src/cli.py:429` → `src/pipeline/audit.py:40`), which runs **after** `enrich_all` has already completed and written centrality to Neo4j.

## Impact

- A **successful** GDS centrality computation (sampled betweenness + pagerank + louvain + degree, da#326) is reported as a **failed run** (rc=1) purely because a filesystem audit-record write failed.
- Because the process exits non-zero, the workflow's `enrich` **verify step never runs** — so the top-k choke-point echo and the `betweenness_centrality IS NOT NULL` gate are skipped even though the data is (likely) already in the graph.
- Blocks the da#326 choke-points RUN on stg (and therefore prod promotion).

## Root cause

`write_audit_entry(data_dir, entry)` does `audit_dir.mkdir(parents=True, exist_ok=True)` unconditionally. In the `graph-ops` container `data_dir` (= `staging_dir.parent`, default `data/`) is on a **read-only** filesystem, so the mkdir raises `OSError`. The audit JSON written there is throwaway in the container anyway (it vanishes when the container exits) — it should never gate an expensive, already-completed graph operation.

## Observed

deploy graph-ops run 28913510321 (env=stg, operation=enrich, dry_run=false):
```
OSError: [Errno 30] Read-only file system: 'data'
  File "/app/src/pipeline/audit.py", line 40, in write_audit_entry
    audit_dir.mkdir(parents=True, exist_ok=True)
ERROR: [stg] graph op 'enrich' (enrich) failed (rc=1).
```

## Proposed fix

Make the filesystem audit-write **best-effort**: catch `OSError` in `write_audit_entry` (and/or at the `_cmd_enrich`/`_cmd_load` call sites), emit a warning, and continue. On the build host (writable `data/`) behaviour is unchanged; in a read-only container the completed operation succeeds and the workflow's verify step runs.

## Acceptance criteria

- [ ] `enrich` (and other audited subcommands) no longer crash when `data/` is read-only; a warning is logged instead
- [ ] Build-host behaviour unchanged (audit still written when the dir is writable)
- [ ] Unit test covering the read-only / OSError path
- [ ] stg graph-ops enrich run completes green and the verify step echoes top-k choke-points

parametrization:
## Summary

- `write_audit_entry` crashed with `OSError: [Errno 30] Read-only file system: 'data'` when run in the graph-ops container, whose filesystem is read-only by org convention (`ontology/conventions.md` § Container security). The crash surfaced from `_cmd_enrich` **after** `enrich_all` had already completed and written centrality to Neo4j — a throwaway audit-file write was failing an expensive, already-completed graph operation.
- Made the filesystem audit-write best-effort: `write_audit_entry` now catches `OSError` around the `mkdir` + file-write, logs a `warning` (`audit_write_failed`), and returns `None` instead of raising. Happy-path behaviour (dir writable → audit file written, `Path` returned) is unchanged.
- Return type changed `Path` → `Path | None`. Both call sites (`_cmd_enrich`, load cmd in `src/cli.py`) already discard the return value, so this is a no-op for callers; verified under `mypy --strict`.
- Regenerated the structural ontology index (`ontology/structural/{code-graph.json,llms.txt}`) per the repo's local `structural-ontology-staleness` pre-commit hook.

## Test plan

- [x] Added `test_write_read_only_dir_is_best_effort` — monkeypatches `Path.mkdir` to raise `OSError(30, "Read-only file system", ...)`, asserts `write_audit_entry` returns `None` (does not raise) and emits an `audit_write_failed` warning via `structlog.testing.capture_logs()`.
- [x] Added `test_write_failure_on_file_write_is_best_effort` — same contract when `Path.write_text` fails (permission denied) after a writable dir.
- [x] Confirmed existing happy-path tests (`test_write_creates_file`, `test_write_creates_audit_dir`, round-trip) still pass unmodified.
- [x] `make lint` — passed
- [x] `make typecheck` (mypy --strict) — passed
- [x] `make test` (full suite) — 1715 passed, 7 skipped (pre-existing, Docker-gated)
- [x] pre-commit + pre-push hooks (ruff-format, ruff-lint, gitleaks, mypy-strict, pytest) — all green on push

Closes #335


parametrization:
RequestOrReplied: Request
TechDebt: none

Please review — fix for da#335 (enrich CLI crash on read-only audit-write). Best-effort OSError catch in write_audit_entry, tests for the read-only-dir and permission-denied-file paths.

parametrization:
RequestOrReplied: Request
TechDebt: none

Please review — fix for da#335 (enrich CLI crash on read-only audit-write). Best-effort OSError catch in write_audit_entry, tests for the read-only-dir and permission-denied-file paths.

parametrization:
Reviewed at head `96d00b62cce3048decb8c85e71cee098fe4bc7e2` via the contents API (no local checkout).

`write_audit_entry` in `src/pipeline/audit.py` now wraps `audit_dir.mkdir(...)` + `path.write_text(...)` in `try/except OSError` — a targeted catch, not a bare except, so non-filesystem exceptions still propagate. On failure it logs `logger.warning("audit_write_failed", stage=..., path=...)` and returns `None`; the happy path is unchanged (still returns the written `Path`). Both call sites in `src/cli.py` (load cmd line 236, `_cmd_enrich` line 429) invoke `write_audit_entry(data_dir, audit)` as a bare statement — return value fully discarded, so the `Path -> Path | None` signature change is a no-op for callers and can't introduce an Optional-dereference. CI's mypy job only runs `mypy src/` (tests/ excluded), which matches the PR's "verified under mypy --strict" claim scoped to the call sites, not the test file.

The two new tests monkeypatch `Path.mkdir`/`Path.write_text` to raise `OSError` and assert (a) `write_audit_entry` returns `None` without raising and (b) an `audit_write_failed` warning is captured via `structlog.testing.capture_logs()`. Neither test uses `pytest.raises`, so both would fail with an unhandled `OSError` against the pre-fix code — they genuinely exercise the new behavior. Existing happy-path/round-trip tests are untouched. No secret/credential handling touched, no scope creep beyond the audit.py fix + structural-index regen noted in the PR body.

CI is green across Lint, Type Check, Test, Integration Tests, Base-image pin + fixture realism, structural-ontology staleness-check, and the Pre-commit ⇄ CI sync-drift gate.

RequestOrReplied: Approved
TechDebt: none

parametrization:
Reviewed at head sha `96d00b62` via the contents API (no local checkout).

**The fix.** `write_audit_entry` now wraps only the `mkdir` + `write_text` pair in `try/except OSError`, logs `audit_write_failed` with `stage` + `path`, and returns `None`. The diff confirms the old code had no try/except at all — `mkdir` and `write_text` were unguarded — so the two new monkeypatch tests (`test_write_read_only_dir_is_best_effort`, `test_write_failure_on_file_write_is_best_effort`) would genuinely raise/fail against the pre-fix code; they pin real behavior, not a tautology.

**Is best-effort right here, and is it scope-appropriate for both callers?** Both `_cmd_enrich` (cli.py:429) and the load command (cli.py:236) call `write_audit_entry(data_dir, audit)` and discard the return value entirely — verified by reading both call sites, so the `Path` → `Path | None` signature change is a true no-op for both. More importantly, the audit write happens *after* the real work is durable in both cases: load already ran `save_manifest` + graph writes, enrich already wrote centrality to Neo4j (per da#335, confirmed via GDS betweenness/pagerank/louvain/degree completing before the crash). So this isn't a one-off carve-out for enrich — it's a correct, symmetric fix for a shared function whose contract (forensic JSON, gone when the container exits) is genuinely non-critical for both stages.

**Does swallowing `OSError` mask real bugs?** I don't think so here specifically. `ontology/conventions.md` § Container security confirms read-only filesystems are the deliberate org-wide convention for application containers, not an anomaly — so a read-only `data/` mount in `graph-ops` is expected, and hard-failing an already-completed centrality run on every such invocation is the actual bug. The `logger.warning` with stage+path keeps the failure observable (not silent) for anyone watching structured logs, which is the right tradeoff for throwaway bookkeeping data vs. a correctness-bearing write. If audit trail durability ever becomes a hard requirement, that'd need a different mechanism (e.g. writing to a guaranteed-writable location), but that's out of scope for this bug fix.

Exception scope is `OSError` only (not bare `except`), diff is minimal (4 files: source, tests, and the two auto-regenerated structural-ontology artifacts required by the repo's own staleness hook — no scope creep), no secret/credential handling. All 7 CI checks green (Lint, Type Check, Test, Integration Tests, Base-image pin, structural-ontology staleness, pre-commit⇄CI sync-drift).

RequestOrReplied: Approved
TechDebt: none

Issue #337, opened 2026-07-08
Narrator over-merge inflates choke-point centrality — disambiguate generic-name nodes (blocks prod enrich)


parametrization:
## Summary — blocks prod promotion of narrator centrality (choke-points)

The 2026-07-08 stg `enrich` run computed betweenness centrality across all
150,187 narrators (mechanism verified sound — real mega-hubs **al-Zuhrī,
Abū Hurayra, Shuʿba, Sufyān, Sālim b. ʿAbd Allāh** all surface near the top).
**But the very top of the choke-point ranking is contaminated by over-merged
generic-name nodes**, which outrank the real hubs:

| rank | name_ar | betweenness | note |
|-----:|---------|------------:|------|
| 1 | أبو عبد الله (Abū ʿAbd Allāh) | 10,589,305 | generic kunya — ~2× Abū Hurayra |
| 2 | عبد الله (ʿAbd Allāh) | 4,908,456 | generic given-name |
| 5 | أبو جعفر (Abū Jaʿfar) | 2,629,615 | generic kunya |
| 8 | عليم (ʿAlīm) | 1,858,660 | fragment |
| 9 | ابن عمران (Ibn ʿImrān) | 1,769,812 | partial name |

## Root cause

Canonical Narrator nodes are keyed `nar:<uuid5(normalized-name)>`
(`src/graph/load_edges.py` / `load_nodes.py`). Narrators who share a common
kunya or single-token name (e.g. many distinct "أبو عبد الله") therefore collapse
into **one** node, which aggregates the transmission paths of many different
people → artificially inflated betweenness. This is a narrator **entity-resolution
/ disambiguation** problem upstream of the centrality compute — the centrality is
correct *for the graph it was given*; the graph over-merges.

## Why it matters now

Owner decision (2026-07-08): **hold the prod enrich promotion until this is
fixed**, so the first choke-points a prod user sees are all real, disambiguated
narrators — not meaningless generic-name aggregates. stg enrich stays as-is for
internal validation. This issue is the gate.

## Direction (to scope)

Disambiguate same-name narrators using signals beyond the normalized name —
e.g. generation/ṭabaqa, teacher–student context, corpus co-occurrence — so
distinct people sharing a kunya get distinct `nar:` ids. Relates to the ongoing
narrator name-quality work (da#247 NER name-quality, da#258 mention-count tail).

## Acceptance (high level)

[...]

PR #338, opened 2026-07-08
da#337 PR-1: narrator id-rekey foundation helpers

parametrization:
## Summary

Part of #337 — **PR-1: the pure foundation helpers only, no pipeline wiring.**

da#337 root cause: canonical narrator ids are `nar:uuid5(CANONICAL_NAMESPACE, name_ar_normalized)`, a pure function of the name. Distinct people who share a normalized name (bare kunyas like `أبو عبد الله`, single-token mononyms, captured fragments) collapse onto ONE `nar:` node and inflate its betweenness centrality. The full split lands in PR-2 (a new resolve stage); this PR ships only the two pure helpers it will build on, plus tests.

### 1. `src/parse/identity.py` — `make_discriminated_canonical_id(name_normalized, discriminator="")`
Sibling of `make_canonical_id` for *discriminated* ids.
- **Empty discriminator → byte-identical to `make_canonical_id(name_normalized)`** — the backward-compat contract, so wiring this in re-keys nothing for the un-split majority.
- Non-empty → `nar:uuid5(CANONICAL_NAMESPACE, name + U+001F + discriminator)`. The U+001F unit separator gives the name↔discriminator boundary an unambiguous split.

### 2. `src/resolve/generic_name.py` (new) — `is_generic_name(name_ar_normalized, mention_count)` + `GENERIC_MIN_MENTIONS = 50`
Pure eligibility screen flagging over-collapsing generic names for the PR-2 split. Returns `True` iff ALL of: **thin name** (≤2 significant tokens, so a full ism+nasab like `محمد بن اسماعيل البخاري` is never flagged), **generic shape** (bare kunya / bare nasab / single-token mononym / ≤2-token fragment), and **aggregation gate** (`mention_count >= GENERIC_MIN_MENTIONS`). Reuses `fuzzy_cluster._CONNECTOR_TOKENS` / `_significant_tokens` and `arabic.normalize_arabic` so tokenization matches the resolve pass. Pure, deterministic, no IO.

### Tests
- `make_discriminated_canonical_id`: empty-disc byte-identical to `make_canonical_id`; determinism; distinct discriminators → distinct ids; separator boundary-disambiguation; no collision with any real normalized name.
- `is_generic_name`: full ism+nasab → False; each generic shape above the floor → True; below the floor → False; empty/whitespace → False (no crash).

## Reviewer note — separator-collision test deviation (please confirm)
The brief specified a literal assertion `make_discriminated_canonical_id("a\x1fb","") != make_discriminated_canonical_id("a","b")`. This is **mathematically unsatisfiable** under the required byte-identical backward-compat contract: the empty-disc path is `make_canonical_id("a\x1fb") = uuid5(NS,"a\x1fb")`, and `("a","b")` is `uuid5(NS,"a"+"\x1f"+"b") = uuid5(NS,"a\x1fb")` — the same input. It also cannot occur in production: **`normalize_arabic` maps U+001F to a space** (Python `\s` matches it), so no real normalized name contains the separator. I kept the brief's exact minting formula and instead asserted the separator's *true* guarantees (boundary disambiguation + no collision with any real normalized name), documenting the synthetic edge case as a precondition violation. Flagging for a design confirmation.

## Scope / non-goals
No changes to `RESOLVE_STEP_ORDER`, `disambiguate.py`, `bio_promote.py`, the loaders, the graph, or `make_canonical_id` itself. Structural ontology index regenerated (new module/symbols).

## Green before push
`ruff check` + `ruff format --check`, `mypy` (strict), full `pytest` (1737 passed, 7 skipped), and the structural-ontology staleness check all pass locally; pre-commit + pre-push hooks passed on commit/push.


parametrization:
## Architectural review — da#337 PR-1 (pure narrator id/detection helpers)

RequestOrReplied: Approved
TechDebt: none

Reviewed at HEAD `314beb3` via the GitHub contents API — `src/parse/identity.py`, `src/resolve/generic_name.py`, both test modules, plus the two dependencies the design leans on (`src/utils/arabic.normalize_arabic`, `fuzzy_cluster._CONNECTOR_TOKENS`/`_significant_tokens`). Approving. The canonical-id contract is upheld and the scope is clean. Details below, including my call on the separator-vs-namespace question you flagged.

### Backward-compat contract — holds
The empty/falsy-discriminator path short-circuits to `make_canonical_id(name_normalized)` verbatim, so no existing narrator re-keys when this is wired. Byte-identity is locked by a test asserting equality against `make_canonical_id` for both the default and explicit-`""` forms. `make_canonical_id` and `CANONICAL_NAMESPACE` are untouched. Good.

### Determinism — holds
uuid5 inputs are pure functions of the arguments; no ordinals, clock, or randomness. Re-run stability and same-name-same-discriminator identity are both covered.

### The separator-collision question — my architectural call
You were right to drop the brief's literal assertion — it is unsatisfiable under the byte-identity contract (both sides reduce to `uuid5(NS, "a\x1fb")`), and asserting the separator's *true* guarantees instead is the correct move.

I independently verified the invariant the whole approach rests on. Disjointness between discriminated and plain ids holds **iff** a normalized name can never contain U+001F, and it genuinely cannot — but the mechanism is worth stating precisely because your code comment attributes it to the wrong step:

- `strip_format_marks` does **not** touch U+001F. Its regex range is `​-‏‪-‮⁦-⁩﻿۝` — all above the C0 block. It removes zero C0 control characters.
- The character is eliminated by `clean_whitespace`, whose `\s+` collapse space-maps it: in CPython `\s` matches U+001C–U+001F (same table as `str.isspace`). I confirmed `re.sub(r"\s+", " ", "a\x1fb") == "a b"` on the stdlib alone.

So the guarantee is real, but it is a property of `clean_whitespace`'s `\s` semantics — a CPython-specific quirk in a *different* module — not of the format-mark stripper the comment names. Two hardening items follow from that (neither blocks this PR; both are natural PR-2 scope):

1. **Lock the invariant with a regression test.** The disjointness proof currently lives only in prose and in `test_discriminated_never_collides_with_a_real_normalized_name` (which exercises real inputs but does not pin the normalizer). One line — asserting `"\x1f" not in normalize_arabic("a\x1fb")` (or that it maps to a space) — makes the load-bearing behavior explicit and stops a future `clean_whitespace`/`strip_format_marks` refactor from silently reopening the collision. For a keystone identity module this is worth having.
2. **Fix the mechanism attribution** in the `_DISCRIMINATOR_SEP` comment and the `make_discriminated_canonical_id` docstring: it is whitespace-collapse that removes U+001F, not the control/format-mark strip.

On **separator + shared `CANONICAL_NAMESPACE` vs. a dedicated discriminated namespace**: I confirm the shared-namespace-with-separator call is correct and safe for PR-1. But my recommendation for PR-2 — where discriminators actually get minted — is to weigh a dedicated `DISCRIMINATED_NAMESPACE` for the non-empty path (empty stays on `CANONICAL_NAMESPACE` via the existing short-circuit, so the byte-identity contract is untouched and there is zero migration cost, since no discriminated ids exist yet). The reason is architectural robustness: a dedicated namespace makes plain⇄discriminated disjointness a property of **id construction** (cryptographic namespace separation) rather than an incidental property of the input **normalizer**'s `\s` behavior in another module. It removes the cross-module coupling entirely. The separator is still useful inside the discriminated space for the name↔discriminator boundary. It is a genuine tradeoff, not a defect in this PR — flagging it now so the decision is made before PR-2 builds on the current shape.

### `is_generic_name` — sound, with one recall note for PR-2
Correct reuse of `_significant_tokens`/`_CONNECTOR_TOKENS`; the thin-name gate (≤2 significant tokens) protects a full ism+nasab, and the `GENERIC_MIN_MENTIONS` aggregation gate is present and inclusive at the floor. The module-load asserts that the normalized kunya/nasab markers subset `_CONNECTOR_TOKENS` are a nice fail-loud guard (minor: `assert` is stripped under `python -O`, so a test would be sturdier than the bare asserts — again PR-2/optional).

Two observations, neither blocking:
- `single_mononym` and `bare_nasab` are logically subsumed by `short_fragment` (`len <= 2`) — every 1- or 2-token name already matches the fragment disjunct. The explicit enumeration is fine for AC-traceability (and you documented it), just noting the redundancy.
- Recall gap for PR-2 to keep in mind: a 2-significant-token ism+nasab like `محمد بن علي` (3 raw tokens, ism-led, carries a nasab) is **not** flagged, even though it is a classically ambiguous over-collapsing name. That is the conservative, precision-first choice for a foundation screen and I agree with it here — but PR-2's split stage should decide whether that recall floor is where it wants to sit.

### Scope discipline — clean
No changes to `RESOLVE_STEP_ORDER`, `disambiguate.py`, `bio_promote.py`, the loaders, or `make_canonical_id`. Structural ontology index regenerated. Foundation-only, as advertised.

Approving.

parametrization:
## Review — entity-resolution / Arabic-text lens (Ivana Horvat)

Reviewed at HEAD `314beb3`. The id-helper mechanics are clean and the tests are thorough; the concern is narrow but substantive and sits squarely in the detector's linguistic soundness — the crux for PR-2 over-split safety.

### What is solid (no change needed)
- **`make_discriminated_canonical_id`** — determinism, the empty-discriminator byte-identity contract, and the U+001F boundary guarantee all hold. Your reviewer note is correct: the brief's literal `!=` assertion is unsatisfiable under the byte-identity contract (both sides reduce to `uuid5(NS,"a\x1fb")`), and it cannot arise in production because `normalize_arabic` collapses U+001F to a space. Keeping the mint formula and asserting the separator's *true* guarantees (boundary disambiguation + no collision with a real normalized name) is the right call. I agree with your deviation; Jean-Claude owns the namespace question.
- **Normalization consistency (the silent-never-fires hazard the brief flagged): correctly handled.** `_KUNYA_MARKERS`/`_NASAB_MARKERS` are built through `normalize_arabic` (أبو→ابو, أبي→ابي, أم→ام, ابن/بن) and the input is pre-normalized, so raw-token membership matches. The import-time `assert _K/_N <= _CONNECTOR_TOKENS` fail-loud guard is sound and a nice touch.
- Purity/determinism: no IO, no randomness, uuid5 stable. Degenerate input (`""`/blank) returns False without raising. Good.

### Change requested — `short_fragment` breadth contradicts the stated contract
`short_fragment = len(raw_tokens) <= 2` subsumes `single_mononym` (1≤2) and `bare_nasab` (len==2≤2), so the generic-shape clause reduces to `bare_kunya OR (raw_tokens ≤ 2)`. Two consequences:

1. **The docstring overclaims.** It states the screen is "deliberately conservative … so it never nominates a name that safely denotes one person." But a 2-token **ism + nisba** — e.g. `سفيان الثوري` (Sufyan al-Thawri), `مالك الاصبحي` — has 2 significant tokens, no kunya/nasab lead, and (for a real hub) far more than 50 mentions, so it **is** flagged. Those are specific, well-known individuals and legitimate high-betweenness hubs — the exact opposite of the same-name collapse this targets. The mention floor gives no protection here; real hubs have thousands of mentions. So the function does precisely what its contract says it never does.

2. **`single_mononym` and `bare_nasab` are dead clauses** — they never change the result given `short_fragment`. As written, the four-shape enumeration reads as four independent guards but is really `bare_kunya OR ≤2-raw-tokens`, which is misleading to a PR-2 author relying on this as the split-eligibility contract.

This is the sharpest false-positive class for PR-2: an over-broad screen that includes legit hubs is only safe if PR-2 **never splits on the screen alone** and always gates the actual split on independent multi-referent evidence (conflicting death-years/nisbas/generations among the mentions). That contract is load-bearing and currently the docstring implies the opposite.

Minimal, non-redesign ask (author's choice of direction):
- **Reconcile contract with behavior.** Either (a) narrow the shape logic so a 2-token name is flagged only when it is genuinely a bare kunya/nasab/mononym or an ism-only pair — i.e. don't auto-flag ism+**nisba**, where the nisba is a disambiguating token — **or** (b) if the broad recall-first screen is intentional, soften the docstring (drop "never nominates a name that safely denotes one person"; state it is recall-oriented and that PR-2 MUST gate the split on independent multi-referent evidence, not this screen alone), and either drop or genuinely load-bear the redundant `single_mononym`/`bare_nasab` clauses.
- **Pin the decision with a test.** Add an ism+nisba case (`سفيان الثوري` at ≥ floor) asserting whatever you decide, so the nisba tradeoff is a conscious, regression-guarded choice rather than a side effect of `short_fragment`. Today's tests pin the compound-ism `عبد الله` (fine) but leave the ism+nisba class uncovered.

### FP-risk assessment for PR-2 (brief asked)
Recall on the genuine over-collapse shapes (bare kunya reaching past 2 tokens, bare nasab, mononym) is good. Precision on 2-token names is the exposure: the `raw_tokens ≤ 2` catch-all flags ism+nisba/ism-pair forms that often denote one specific person, including hubs. Safe **only** if PR-2 treats this as recall and gates the split on independent evidence — which is exactly the contract the current docstring undercuts. Once the contract is corrected/pinned per above, I'm satisfied.

Requestor: Ivana Horvat
TechDebt: none

parametrization:
## Re-review at `584dad1` — entity-resolution / Arabic-text lens (Ivana Horvat)

The fixup reconciles the contract to my satisfaction. Verified at the new head:

- **Redundant clauses gone.** Generic-shape is now `short_fragment or kunya_led` — two genuinely load-bearing forms, no dead `single_mononym`/`bare_nasab`. `raw_tokens[0]` is safe (the blank-input guard already returned, so ≥1 token). Behavior preserved on the boundary cases: `أبو بكر بن محمد` still screens OUT (4 raw tokens ⇒ not short_fragment; `بن` present ⇒ not kunya_led), and `أبو عبد الله` still screens IN via `kunya_led`.
- **Docstring honest.** The "never nominates a name that safely denotes one person" overclaim is removed. It now states plainly this is a recall-first candidacy screen that over-includes 2-token names (ism+nisba included, by design), NEVER decides a split, and that the split is gated downstream in PR-2 on independent multi-referent evidence (≥2 well-separated death-year bands), abstaining on a single-band person like سفيان الثوري. The mention floor is correctly re-labeled a size cut-off, not a precision guard.
- **Nisba-vs-theophoric rationale is sound.** Not attempting the distinction here is the right call: a naive "2nd token is ال-prefixed ⇒ nisba ⇒ protect" rule would wrongly exclude عبد الله (الله is ال-prefixed too), which we DO want screened in. That judgement genuinely needs downstream evidence, not a name-shape heuristic.
- **Decision is pinned.** `TestRecallFirstCandidacy` regression-locks the deliberate over-inclusion: سفيان الثوري → True (intentional screen-in), عبد الله → True, محمد بن اسماعيل البخاري → False.

The one caveat I'll restate for the PR-2 author (now correctly load-bearing in the docstring, not a defect here): the split MUST be gated on the death-year-band evidence, never on this screen alone — the recall/precision split only holds if that gate lands as documented. Helpers remain pure/deterministic/no-IO. CI 7/7 green. Approving.

Requestor: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

PR #339, opened 2026-07-08
da#337 PR-2: narrator_split resolve stage (generic-name over-merge splitter)

parametrization:
## Part of #337

PR-2 of da#337 (same-name over-merge split). Adds the **`narrator_split`** resolve stage — the *split* mirror of `fuzzy_cluster`'s cross-source *merge* — inserted between `cluster` and `reconcile`. PR-1 (`make_discriminated_canonical_id`, `is_generic_name`) is already merged (#338); PR-3 is the staging re-load.

### Algorithm
For each canonical record screened in by the PR-1 recall-first `is_generic_name` (and **not** owned by `mononym_split`'s registry):
1. Gather all mentions of the id `C` (name `N`).
2. **Isnad-adjacency evidence** — for each mention, map chain neighbours at `position_in_chain` ±1 to their **attested** `death_year_ah` (precision ≠ `tabaqa_estimate`). Teacher (pos−1) dies a transmission gap earlier, student (pos+1) later, so `C`'s per-mention death estimate = `neighbour_death ± MID_GAP` (the midpoint of `mononym_split._MIN_GAP`/`_MAX_GAP`), averaged over attested neighbours.
3. **Cluster** the estimates 1-D, cutting a new death band on any gap > `SPLIT_BAND_GAP`.
4. **Peel** each non-largest qualifying band onto `make_discriminated_canonical_id(N, disc)`; the largest band + undatable/below-support remainder keep the bare id.

### Death-band over-split gate (Ivana's PR#338 constraint)
**The split gates on death-band evidence, NEVER on the `is_generic_name` screen alone.** It ABSTAINS (leaves `C` as one node) unless ALL hold:
- ≥2 bands each with ≥ `SPLIT_MIN_SUPPORT` mentions,
- ≤ `SPLIT_MAX_CLUSTERS` qualifying bands (more ⇒ generic bucket → abstain, logged),
- adjacent midpoints separated by > `SPLIT_MIN_SEPARATION`.

So a name that clusters to a single well-supported band — **Sufyān al-Thawrī (d.161)**, **al-Zuhrī (d.124)** — abstains and is never fragmented, even though both are deliberately admitted by the screen. Locked by `test_thawri_screens_in_but_death_band_gate_abstains` and the protected-hub regression test.

### Peel-not-partition
Mentions with no attested neighbour, and any mention in a below-support band, **stay on the primary** node. Only confidently-dated distinct clusters peel off.

### Outputs
- `narrators_canonical.parquet` — peeled rows appended (band death window; primary `mention_count` reduced).
- `narrator_mentions_resolved.parquet` — peeled mentions remapped. Mirrors `fuzzy_cluster._remap_mention_canonical_ids` (streaming/idempotent/bounded-memory) but **keyed on `mention_id`**, because a split sends *different* mentions of one id to *different* targets — a global id→id remap cannot express that.
- `narrator_splits.parquet` — one-row-per-peel **audit report** (bare name, new id, discriminator, band window, mention_count, exemplar anchors): the owner's review artifact before any prod re-load.

### Discriminator (deterministic — pure function of the data)
Death-band bucket label from `_GENERATION_DEATH_WINDOW_AH` (e.g. `d:150-250`) or a 25-year bucket `d:<floor(mid/25)*25>` when a window is ambiguous; same-label collisions tie-broken by the band's lexicographically-smallest anchor **name** (not id — cascade-free). No running ordinals — reproducible across re-runs.

### Documented v1 limitation (accepted, NOT fixed here)
The chain-edge loaders key on the mention `canonical_narrator_id` and follow the remap correctly. But `load_edges._studied_under_endpoint` and `muhaddithat_links` mint ids from the **name only**, so a STUDIED_UNDER / muhaddithat edge naming a split narrator attaches to the **primary** (bare-name) node. Accepted for v1 (edge never dropped, lands on the dominant referent); documented in the stage docstring per the task.

### Estimated-window provenance
Peeled rows carry `tabaqa_estimate` precision — the enum's generic "derived, not attested" bucket, **pending a first-class `isnad_estimate` value (PR-3 tuning)**. This is load-bearing: it marks the death as an estimate, keeps `tabaqa_dates` from re-touching it, and excludes peeled nodes from the attested-neighbour pool → re-runs are a strict **no-op** (idempotent).

### Thresholds (tunable — to be re-tuned against staging data in PR-3)
| Constant | Value | Meaning |
|---|---|---|
| `MID_GAP` | 47 | transmission-gap offset (midpoint of mononym_split 15–80) |
| `SPLIT_BAND_GAP` | 80 | death-year gap that cuts a new band (~one generation) |
| `SPLIT_MIN_SUPPORT` | 10 | mentions required for a band to qualify as a referent |
| `SPLIT_MIN_SEPARATION` | 50 | min qualifying-midpoint separation |
| `SPLIT_MAX_CLUSTERS` | 6 | more qualifying bands ⇒ generic bucket → abstain |

### Wiring
Added `narrator_split` to `RESOLVE_STEP_ORDER` (between `cluster` and `reconcile`) + `run_all`. `--from-step` handling is automatic (CLI choices derive from `RESOLVE_STEP_ORDER`). Non-resumable (single idempotent pass, like `reconcile`/`tabaqa_dates`). Regenerated the per-repo structural ontology index.

### Tests (13, all green)
al-Zuhrī / Sufyān-al-Thawrī single-band abstain, two-band split + peel, below-support not peeled, ≥MAX_CLUSTERS noise abstain, registered-mononym deferral, label-collision tie-break, undatable remainder stays, remap correctness, re-run idempotence, protected-hub regression (dominant node retains ≥90% of mentions). `make check` (ruff + mypy strict + pytest) green.

### Reviewers should focus on
- The death-band gate never firing on the screen alone (the Sufyān/al-Zuhrī abstain).
- Idempotence via the `tabaqa_estimate`-precision exclusion of peeled nodes from the attested pool (the monotonicity argument in the docstring).
- The `mention_id`-keyed remap vs the merge pass's id→id remap.
- Thresholds — flagged as staging-tunable in PR-3.

https://claude.ai/code/session_01JeGwpLzYXPnv793d9kKzr7


parametrization:
Requestor: Ivana Horvat
RequestOrReplied: Approved
TechDebt: da#337 PR-3 must harden the protected-hub regression fixture with a realistic multi-generation neighbour spread (the current one is a single-anchor delta band) plus a >=90%-retention-with-peel case, and promote peeled-node death precision to a first-class isnad_estimate value.

Approving on entity-resolution grounds. The split decision is correctly gated on independent death-band evidence and never on the recall-first screen alone, the evidence is non-circular, and the peel-not-partition structure bounds the blast radius. Detail below, with two things I want addressed in the staging pass.

What genuinely holds up:

- The abstain-on-single-band guard is real, not decorative. `test_thawri_screens_in_but_death_band_gate_abstains` proves Sufyan al-Thawri is admitted by `is_generic_name` yet clusters to one band (d.161) and is left whole; al-Zuhri (d.124) does the same. Gate 1 (>=2 bands each >= SPLIT_MIN_SUPPORT) is the load-bearing line and it fires correctly. This is the protection I insisted on and it is present.
- Evidence is not circular. `_attested_death` excludes `TABAQA_ESTIMATE` precision, so only attested neighbour death-years feed the estimate, and peeled rows (which this stage stamps `tabaqa_estimate`) are excluded from the attested pool. That makes a re-run a strict subset of the first run's evidence, so splitting only ever gets harder. Idempotence is genuinely tested.
- The +/-MID_GAP correction is the right mechanism: it maps a teacher-neighbour and a student-neighbour of the same person back onto that person's own death year, which is exactly what keeps a wide-neighbour hub in one band. Peel-not-partition (undatable + below-support stay on the primary, largest band keeps the bare id) means the dominant referent structurally cannot be shattered below its largest band. `test_stage_undatable_remainder_stays_on_primary` confirms undatable mentions are never peeled.

The one substantive reservation (why the TechDebt line, not a block):

The protected-hub regression test is close to tautological. `test_stage_protected_hub_keeps_dominant_node` builds al-Zuhri from 100 mentions all pointing at a single anchor at one death year, so every per-mention estimate is the identical value 124 -- a delta function that trivially forms one band. That exercises nothing the plain single-band test does not already cover, it only touches al-Zuhri (not Abu Hurayra / Shu'ba / Sufyan / Salim b. Abd Allah as the brief scoped), and the advertised ">=90% mention retention" property is never actually exercised because abstain yields 100%. A real hub has neighbours spanning several generations; whether the +/-MID_GAP collapse holds under a bimodal early-companion-plus-late-student neighbour distribution is precisely the empirical question, and this fixture does not ask it. I am not blocking because no data reaches prod on this PR: the `narrator_splits.parquet` audit artifact plus the record-level staging check plus my review of PR-3 are the real validation gate, and the peel-not-partition floor caps any bad split. But hardening this fixture into a genuine multi-generation hub (and one case where a hub peels a small band and still keeps >=90%) is a non-negotiable PR-3 deliverable, not optional polish -- it is the empirical proof of the exact constraint this whole stage exists to satisfy.

Secondary notes for PR-3 (not blocking):

- The generation windows in `_GENERATION_DEATH_WINDOW_AH` overlap heavily, so most band midpoints match two windows and `_band_label` falls through to the 25-year bucket much more often than the docstring's `d:150-250` examples suggest. That is deterministic and fine, but it also makes the same-label tie-break unreachable in practice: two distinct bands are >SPLIT_BAND_GAP (80) apart by construction, while any shared label requires them within 25 years (same bucket) or inside a single non-overlapping window region (max width ~78 < 80). `test_same_band_label_collision_tiebreak_by_anchor` does not actually trigger the tie-break -- its two peel bands land in different 25-year buckets (d:150 and d:225) and the assertion passes trivially. The anchor-name tie-break is harmless defensive code, but the test claiming to cover it does not, and the PR body over-states that coverage. Either drop the branch or add a fixture that genuinely collides.
- On the `tabaqa_estimate` precision overload: from a provenance standpoint it does mislabel an isnad-derived window as a tabaqa-derived one, but it is load-bearing for idempotence and correctly marked-as-estimate for display, so I am comfortable deferring the dedicated `isnad_estimate` value to PR-3 as the author proposes. Flagging it so it does not get lost.
- Memory footprint of `_build_chain_index` (the full `hadith_id -> [(pos, cid)]` map is the entire mentions table resident) is worth a glance against real corpus size given prior OOM history, but that is Jean-Claude's architecture lane, not an ER concern.

The band thresholds (BAND_GAP 80 / MIN_SUPPORT 10 / MIN_SEPARATION 50 / MAX_CLUSTERS 6) are defensible starting defaults and correctly flagged staging-tunable. Good, careful work on the evidence gate.

parametrization:
## Architecture review — da#337 PR-2 (`narrator_split`)

Reviewed at HEAD `6598d33` (source read via API, no checkout). Verdict: **approve**. This is a clean, well-reasoned split pass — the *merge* mirror argument holds, the id contract is airtight, and the over-split guard is exactly where it needs to be. Findings against the five focus areas:

**1. Death-band gate is the sole trigger — CONFIRMED.** `plan_split` never splits on the `is_generic_name` screen; the screen only selects candidates. The three guards are AND-composed (each independently returns `abstain`): ≥2 bands each ≥ `SPLIT_MIN_SUPPORT` (via the `qualifying` filter), ≤ `SPLIT_MAX_CLUSTERS`, and adjacent midpoints > `SPLIT_MIN_SEPARATION`. A single well-supported band (Sufyān al-Thawrī d.161, al-Zuhrī d.124) hits `len(qualifying) < 2` and abstains. Locked by tests. This is the load-bearing property and it is solid.

**2. `mention_id`-keyed remap — CONFIRMED correct.** `_remap_split_mentions` streams row-group by row-group into a sibling `.split.tmp` then `replace`, keyed on `mention_id`, remapping only when `target is not None and target != cid`. Every row is written (no drop); mention_ids are globally unique and each belongs to one band of one candidate, so `remap.update()` across plans cannot collide (no double-map). Peeled mentions get the discriminated id; undatable/below-support/retained-band mentions keep the bare id. Bounded memory, idempotent (guarded on empty remap). The author's "a global id→id map can't express peel-not-partition" reasoning is correct.

**3. Ordering + `tabaqa_estimate` overload — the interesting part.** Two things I checked hard:

- *Evidence availability at step 3.55.* `narrator_split` runs before `reconcile`/`tabaqa_dates`, but its anchor evidence only needs a populated `death_year_ah` with precision ≠ `tabaqa_estimate`. `disambiguate` and `bio_promote` both populate `death_year_ah` on canonical rows before this stage, and `tabaqa_estimate` is minted only by `tabaqa_dates` (step 3.65, later), so on a first run no anchor can be excluded. The stage is not a no-op-by-ordering. Good.
- *Downstream safety of the overload.* `reconcile` skips peeled ids (they are discriminated ids, never bio-keyed → not in `obs_by_cid`, precision preserved, year preserved). `tabaqa_dates._death_is_undated` sees the peeled row's populated death year → skips it regardless of precision. No corruption of reconcile, tabaqa_dates, or the graph date fields.

My call on the overload + idempotence: **sound and safe to ship, with one precise caveat.** The direct stage re-run is a genuine strict no-op (proven by test + the structural fact that the primary has exactly one qualifying band left after peeling, and every peeled node is a single band excluded via `tabaqa_estimate`). The one place the "strict no-op on *any* re-run" claim is very slightly too strong: the `run_all(from_step="narrator_split")` resume path interposes `reconcile`, whose `_reconcile_open` can newly-populate an open-bound anchor's death *point* that `bio_promote` left null — so the attested pool is not *strictly* monotone across that specific resume path. It is not data-corrupting (still deterministic and evidence-gated) and it is moot for the prod re-load path (a full deterministic rebuild). I do not consider it merge-blocking. The dedicated `isnad_estimate` enum deferral is acceptable now (both values mean "derived, marked-as-estimate", satisfying the #673 display contract), but it is real debt — the peeled nodes currently *display* as ṭabaqa estimates though they are isnad-derived. I filed #340 to track the enum plus the docstring/monotonicity tightening for PR-3.

**4. Id determinism & backward-compat — CONFIRMED.** `make_discriminated_canonical_id(name, disc)` is a pure uuid5 of `name + U+001F + disc`; an empty discriminator returns the byte-identical bare `make_canonical_id`, so primary and non-split nodes are unchanged. Discriminator is a pure function of band label + smallest anchor **name** (not id → cascade-free), no running ordinals. Reproducible.

**5. Blast radius — CONFIRMED documented, not broken.** Chain edges (`_build_chain_pairs` at `load_edges.py:161-162`, `_load_narrated:373`) key on `canonical_narrator_id` and follow the remap. `_studied_under_endpoint:732` and the muhaddithat links mint name-only ids via `make_canonical_id`, so they attach to the primary node — accurately captured in the stage docstring as an accepted v1 limitation (edge never dropped, lands on the dominant referent). Scope is clean: no graph/load/deploy touched, `make_canonical_id` unchanged, only the new stage + wiring + regenerated structural index.

Nice work, Kavitha — the peel-not-partition discipline and the `tabaqa_estimate`-exclusion idempotence trick are both well judged.

---

RequestOrReplied: Approved
TechDebt: #340

PR #341, opened 2026-07-08
da#337 PR-3a: narrator_split hardening (review follow-ups) + Closes #340

### LINKED ISSUE #340: da#337 follow-up: first-class isnad_estimate DatePrecision (retire tabaqa_estimate overload on split peels)
## Context

`src/resolve/narrator_split.py` (da#337 PR-2, #339) peels an over-merged generic-name node into distinct canonical nodes and stamps each peeled row's death window with `DatePrecision.TABAQA_ESTIMATE`. That value is **overloaded**: the peeled window is an *isnad-adjacency* estimate (derived from attested chain-neighbour death years ± a transmission gap), not a ṭabaqa-class estimate.

The overload is currently load-bearing and correct for the pipeline math:
- `_attested_death` excludes `tabaqa_estimate` rows from the attested-neighbour pool → re-runs don't cascade.
- `tabaqa_dates._death_is_undated` treats a peeled row (death year present) as already-dated → never re-touches it.
- `date_reconcile` skips peeled ids (not bio-keyed) → precision preserved.

## The debt

There is no first-class `isnad_estimate` value in `DatePrecision`, so a narrator split by **isnad** evidence **displays** to users as a *ṭabaqa* estimate — a provenance conflation. Both are "derived, clearly-marked-as-estimate" windows (satisfies the owner display contract on noorinalabs-main#673), so this is acceptable for v1 but should not persist.

## Ask (PR-3 tuning scope)

1. Add a dedicated `DatePrecision.ISNAD_ESTIMATE` enum value.
2. Have `narrator_split._peeled_record` stamp it instead of `TABAQA_ESTIMATE`.
3. Re-point the two exclusion checks (`narrator_split._attested_death`, `tabaqa_dates._death_is_undated`) to treat `ISNAD_ESTIMATE` the same way (exclude-from-pool / already-dated) so idempotence and cascade-freedom are preserved.
4. Add `ISNAD_ESTIMATE` to `date_reconcile._CLOSED_PRECISIONS` if a peeled id could ever appear as a bio observation (it cannot today, but keep the set complete).
5. Surface the new precision in the date-display layer distinct from ṭabaqa.

## Secondary note (idempotence caveat, non-blocking)

The stage's direct re-run is a proven strict no-op. But the `run_all(from_step="narrator_split")` **resume** path interposes `reconcile`, whose `_reconcile_open` can newly-populate an open-bound anchor's death *point* that `bio_promote` left null — so the attested-neighbour pool is not *strictly* monotone across that specific resume path (a candidate could, in a rare edge, split further on a from-step resume). Not data-corrupting (still deterministic + evidence-gated) and moot for the prod re-load path (full deterministic rebuild). Worth either tightening the docstring's "strict no-op on any re-run" wording to "direct stage re-run / full rebuild" or adding a cheap guard while doing the enum work above.

Filed as TechDebt from the #339 architecture review.

parametrization:
Part of #337 — PR-3a hardening (Ivana + Jean-Claude review follow-ups required before the staging run).
Closes #340 — first-class `isnad_estimate` DatePrecision.

### 1. Protected-hub regression fixtures made real (Ivana's non-negotiable)
The old `test_stage_protected_hub_keeps_dominant_node` built al-Zuhrī from 100 mentions all pointing at ONE anchor at ONE death year — a delta-function that trivially forms one band and never exercises the ±MID_GAP correction or the ≥90%-retention property (an abstain gives 100% for free). Replaced with two empirical fixtures:
- **`test_stage_protected_hubs_multigen_spread_abstain`** — two hubs (al-Zuhrī d.124, Shuʿba d.160), each with **teacher-generation AND student-generation** attested neighbours spanning ~110 AH of raw death years. The ±MID_GAP correction is genuinely exercised and must fold that spread back into ONE band per hub → **abstain** (hub stays whole).
- **`test_stage_hub_peels_small_spurious_band_keeps_ge_90pct`** — the case that actually proves the guard: a dominant band (92%, 184 mentions) + a small well-separated **qualifying** band (8%, 16 ≥ `SPLIT_MIN_SUPPORT`). Asserts the primary retains **≥90%** AND the small band correctly peels to its own node.

### 2. Same-band-label tie-break — confirmed reachable, not dead (option b)
Ivana's "two distinct bands >`SPLIT_BAND_GAP` (80y) apart can never share a bucket label" holds for the 25-year buckets but **misses the wide `LATER` generation window (280-400)**: its uniquely-matching sub-range 301-400 is 99 AH wide, so two bands e.g. d.310/d.395 (85y apart) **both** label `d:280-400`. The old `test_same_band_label_collision_tiebreak_by_anchor` never triggered the branch (its 160/245 bands land in *distinct* 25-year buckets → passed trivially). Rewrote it to **provably fire** the tie-break (asserts `discriminator == label|anchor`), and added a reachability note to the module docstring. The branch is genuinely load-bearing, so it stays.

### 3. Closes #340 — retire the `TABAQA_ESTIMATE` overload
Peeled rows stamped `TABAQA_ESTIMATE` but the window is an *isnad-adjacency* estimate, not a ṭabaqa one. Added `DatePrecision.ISNAD_ESTIMATE` (`(str, Enum)` / `StrEnum`), wired into the exhaustiveness test and `date_reconcile._CLOSED_PRECISIONS`. Peeled rows now carry `ISNAD_ESTIMATE`. **Idempotence preserved**: the evidence pool now excludes *any* non-attested precision via the new `_ESTIMATED_PRECISIONS = {tabaqa_estimate, isnad_estimate}`, so a re-run's evidence stays a strict subset → no new splits → no-op (asserted in `test_stage_idempotent`).

### 4. Chain-index memory audit — OOM flag addressed (cf. #723)
`_build_chain_index` kept **every** hadith's full chain resident — i.e. the entire mentions table, exactly the full-table-resident class that OOM'd resolve in #723. Refactored to **two bounded streaming passes**: pass 1 collects candidate mentions + the set of hadiths containing them; pass 2 fills `chains` for **only** those candidate-touching hadiths (a hadith with no candidate mention can never supply a neighbour lookup). 

**Memory characteristic:** resident footprint is now `O(mentions in candidate-touching hadiths) + O(candidate mentions)` — a small fraction of the multi-million-row corpus (generic-name candidates are a curated minority) — instead of `O(all mentions)`. Cost: 2× sequential read, one-row-group working set (time-for-memory, mirroring `fuzzy_cluster`'s streaming IO shell). The staging run should not OOM on this structure.

## Verification
- `make check` green — **1707 passed**, 7 skipped.
- Structural ontology index regenerated (`python3 scripts/structural_ontology.py emit`), staleness gate current.
- Pre-commit + pre-push (ruff/gitleaks/mypy-strict/pytest/structural) all pass.

## Reviewer focus
- **Item 2**: please sanity-check the reachability claim (LATER window 280-400, unique sub-range 301-400 width 99 > 80). If you still consider a real collision impossible, the alternative is removing the branch — but the new test demonstrates it firing.
- **Item 4**: confirm the two-pass restriction to candidate-touching hadiths is a correct bound (pass 2 retains every position in those hadiths, so non-candidate ±1 neighbours are still present for the lookup).


parametrization:
## Review — da#337 PR-3a: narrator_split hardening + Closes #340

Reviewed at HEAD `bc39273` from a data-flow / schema-evolution stance (enums.py, narrator_split.py, date_reconcile.py, identity.py, schemas.py, tabaqa_dates.py, and the two test modules). Structural / schema soundness holds on all four focus areas.

### 1. Schema evolution — `DatePrecision.ISNAD_ESTIMATE`
Added consistently everywhere the enum is exhaustively matched:
- Enum member + `__all__`-adjacent docstring distinguishing isnad-adjacency provenance from the ṭabaqa layer (enums.py).
- `_CLOSED_PRECISIONS` (date_reconcile.py) — sits with `TABAQA_ESTIMATE` as a central-estimate-asserting precision; the set stays exhaustive over that class.
- `_ESTIMATED_PRECISIONS` (narrator_split.py) — the evidence gate excludes *any* non-attested precision, so a peeled row can never re-seed a further split.
- Exhaustiveness test in `test_all_models.py` — both the `== "isnad_estimate"` value assertion AND the exhaustive member-set literal are updated.

**No migration needed on widening.** `death_date_precision` persists as a plain nullable `pa.string()` (schemas.py) — the column carries the StrEnum *value*, so widening is purely additive; no existing serialized row changes, and no row carries `isnad_estimate` until this stage writes it. The pydantic widening only *admits* a new accepted value, so no round-trip breaks.

The `TABAQA_ESTIMATE` / `ISNAD_ESTIMATE` split is semantically clean, not a rename: `TABAQA_ESTIMATE` is retained and still emitted by `tabaqa_dates`; `ISNAD_ESTIMATE` is minted only by the peel. The two provenances are genuinely distinct (generation-layer window vs isnad-neighbour adjacency), and the code keeps them apart end-to-end.

### 2. Canonical-id discriminator threading — backward-compat holds
`make_discriminated_canonical_id(name, disc="")` short-circuits `if not discriminator: return make_canonical_id(name)` → **byte-identical** to the pre-existing id. Load-bearing property confirmed on both sides:
- Every peeled band receives a **non-empty** `d:<lo>-<hi>` / `d:<bucket>` label (`_band_label` always returns a non-empty string), so no peel path can produce an empty discriminator that would collide with the primary.
- The **retained** (largest) band keeps `primary_id = make_canonical_id(N)`.

So an un-split narrator — and the dominant referent of a split one — keeps its exact prior `nar:` id; the graph reloads without orphaning unchanged narrators. `test_two_separated_supported_bands_split` pins both halves (`new_id == make_discriminated(...)` and `!= make_canonical_id(...)`). The `\x1f` unit-separator join also correctly prevents name/disc conflation.

### 3. Two-pass chain-index refactor — output-equivalent
Data-flow-correct. Pass 1 records candidate mentions + the set of candidate-touching hadith ids; pass 2 fills `chains` for *only* those hadiths, retaining every position (so non-candidate ±1 neighbours are still present). Since `_collect_datable` is only ever called on a candidate's own mentions — whose hadith ids are all candidate-touching by construction — no neighbour lookup is ever missed relative to the old full-table build. The `chains` value is a strict subset (dead non-candidate hadiths dropped), and neighbour matching is `npos == pos±1` filtering, so intra-list ordering is irrelevant. Downstream `chains` structure and completeness are unchanged; only the resident footprint shrinks. Pass naming/comments are clear.

### 4. Naming + documentation
At your usual bar — the module docstring's reachability note (LATER 280-400 window, 99-AH sub-range → two 80y-apart bands share `d:280-400`) is accurate and `test_same_band_label_collision_tiebreak_by_anchor` provably fires it. Idempotence and estimated-window-provenance sections are precise.

### Non-blocking notes
- **Same-label tie-break residual collision (module-owned):** the tie-break appends only `anchors[0]` (band's lexicographically-smallest anchor name). Two same-label bands that *also* share their smallest anchor would mint an identical discriminated id → two canonical rows with the same `canonical_id` (under-split / duplicate row, not a crash — Neo4j MERGE folds them). Extraordinarily narrow (needs the wide LATER-window label AND a shared smallest anchor across two ≥80-AH-apart bands) and self-limiting, and the `narrator_splits.parquet` audit surfaces it before any prod reload — but a second tie-break key (e.g. band midpoint) would close it fully.
- Cross-repo: confirm the isnad-graph API/frontend DatePrecision display maps `isnad_estimate` (same follow-up `tabaqa_estimate` implied); no exhaustive consumer breaks in *this* repo.

Schema evolution and id backward-compat both hold. Approving.

RequestOrReplied: Approved
TechDebt: Tie-break uses only smallest-anchor; two same-label bands sharing that anchor could mint a duplicate discriminated canonical_id — add a secondary tie-break key (band midpoint) to fully close it.

parametrization:
## Review — da#337 PR-3a narrator_split hardening (Ivana Horvat)

I re-verified all four of my flagged items against the code at head `bc39273`, tracing the numeric logic myself rather than trusting the PR narrative. All four hold.

### Item 2 — same-label tie-break reachability (my "dead branch" call): my premise was WRONG, the branch is reachable

My original claim was "two distinct bands >80y apart can never share a bucket label." That is true for the 25-year fallback buckets, but FALSE for the generation-window labels. Traced numerically:

- `_GENERATION_DEATH_WINDOW_AH`: `ATBA_TABA_TABIIN=(220,300)`, `LATER=(280,400)`. `_band_label` returns `d:<lo>-<hi>` only when the midpoint uniquely matches one window. LATER's uniquely-matching sub-range is therefore 301–400 (ATBA ends at 300) — 99 AH wide, wider than `SPLIT_BAND_GAP`(80). So two bands at d.310 and d.395 (85 AH apart, both cut into distinct bands by `_cut_bands`) BOTH label `d:280-400`.
- The subtlety the tie-break actually needs: it fires only among **peel** bands (`label_counts[label] > 1`), and the largest band is retained under the bare id. So a genuine collision requires ≥3 qualifying bands — two colliding peelers + a distinct retained one. The rewritten test gets this right: `_dm(120,30,"ret") + _dm(310,15,"aaa") + _dm(395,15,"bbb")` — d.120 (30) is retained, and the two LATER-window peelers collide on `d:280-400`, forcing `discriminator = d:280-400|aaa` / `d:280-400|bbb`. It asserts both the collision (`"|" in discriminator`) and the distinct resulting ids. This provably fires the branch; the old 20/160/245 fixture landed in distinct 25-year buckets and passed without touching it.

Reachability claim: HELD under my own trace.

### Item 1 — protected-hub fixtures made empirical

Both new tests genuinely drive the ±MID_GAP arithmetic (MID_GAP = (15+80)//2 = 47) over a real multi-generation neighbour spread, not a delta-function:
- `test_stage_protected_hubs_multigen_spread_abstain`: al-Zuhrī teachers d.70/85 (+47 → 117/132) and students d.165/180 (−47 → 118/133); Shuʿba analogous. A ~110 AH raw neighbour-death spread folds to a ~16 AH single band via the directional correction → one qualifying band → abstain. Both sign branches of `_collect_datable` are exercised (teacher at pos−1 → +MID_GAP; student at pos+1 → −MID_GAP), which the old 100-identical-mention fixture never did.
- `test_stage_hub_peels_small_spurious_band_keeps_ge_90pct`: 184/16 (92%/8%) with the 8% band ≥ SPLIT_MIN_SUPPORT — asserts the primary retains exactly 92% AND the small band peels to its own node. This is a real test of the ≥90%-retention behaviour (an abstain keeps 100% and would not exercise it).

These are real, not trivial. One residual, filed as follow-up debt below (not blocking): the multigen-abstain assertion is somewhat insensitive to a hypothetical same-sign inversion of MID_GAP, because even the mis-signed estimates land with a max consecutive gap of exactly 80 (`_cut_bands` cuts on `> 80`, so 80 does not cut) and still collapse to one band. The production signs are correct and both directions are driven, so this is a test-sensitivity nit, not a code defect.

### Item 4 — chain-index OOM refactor: the bound is correct

Traced the two-pass `_build_chain_index`. Pass 1 collects candidate mentions and `candidate_hadith_ids` (every hadith containing ≥1 candidate mention). Pass 2 fills `chains` for exactly those hadiths, retaining EVERY position in them. A candidate's per-mention estimate in `_collect_datable` derives solely from ±1 neighbours within the SAME hadith; any hadith holding a candidate is candidate-touching, so all its positions — including non-candidate ±1 neighbours — are retained. No ±1 neighbour is dropped. It is also behaviour-preserving versus the old full-table build: the old `chains` held every hadith, but `_collect_datable` only ever looked up candidate-touching hadith_ids, so the non-candidate entries were dead weight. Chain-boundary positions (position 0 has no teacher, last has no student) naturally yield fewer neighbours, which is correct, not an off-by-one; the `npos == position ± 1` matching is unchanged from the prior version. Bound: HELD.

### Item 3 — ISNAD_ESTIMATE precision (Closes #340)

`DatePrecision.ISNAD_ESTIMATE = "isnad_estimate"` added; peeled rows carry it via `_peeled_record`. `_ESTIMATED_PRECISIONS = {tabaqa_estimate, isnad_estimate}` and `_attested_death` excludes both from the attested-neighbour evidence pool, so a re-run's evidence is a strict subset → each peeled node is a single band → abstain → no-op. `test_stage_idempotent` confirms the peeled precision is `isnad_estimate` and the second run leaves both parquet tables byte-identical. Exhaustiveness (`test_values` + `test_member_set_is_exhaustive`) and `date_reconcile._CLOSED_PRECISIONS` are both updated; adding ISNAD_ESTIMATE to the closed set is consistent — a peeled row carries point + earliest + latest, i.e. a central estimate with bounds. No other DatePrecision match/exhaustiveness site is broken by the new member.

All 7 CI checks report SUCCESS (non-empty statusCheckRollup).

Approving. The item-2 reachability claim and the item-4 memory bound both held under independent verification.

Requestor: Ivana Horvat
RequestOrReplied: Approved
TechDebt: Multigen-abstain hub test is insensitive to a same-sign MID_GAP inversion (mis-signed spread still folds to one band at a max gap of exactly 80); widen the fixture so a sign flip provably splits.

PR #349, opened 2026-07-09
fix(loader): make save_manifest best-effort on OSError (read-only /app/data)

### LINKED ISSUE #348: bug(loader): save_manifest writes into read-only /app/data — 'load' can never run in graph-load container
## Problem

`isnad-ingest load` cannot run inside the `graph-load` container. The container mounts its Parquet input dir read-only (`/app/data:ro`, `read_only: true` rootfs — org convention), but `_cmd_load` writes into that same directory:

```
OSError: [Errno 30] Read-only file system: 'data/.manifest.json'
```

From `noorinalabs-deploy` run `29030931321` (stg). The loader dies before `load_all`, so the graph is untouched — but no load can ever complete.

## Root cause

`src/cli.py::_cmd_load` performs three writes into `data_dir` (= `settings.data_staging_dir`'s parent = `/app/data`):

| Line | Call | Status |
|------|------|--------|
| 178 | `save_manifest(current_manifest, data_dir / MANIFEST_FILENAME)` | **crashes** — before `load_all` |
| 222 | `save_manifest(current_manifest, data_dir / LAST_LOADED_MANIFEST_FILENAME)` | would crash — after `load_all` |
| ~236 | `write_audit_entry(data_dir, audit)` | already best-effort (da#335) |

Line 178 runs **unconditionally**, before the `incremental` branch, so every invocation dies there regardless of flags. `graph-ops.yml` shares this image but only exercises `enrich`, which never touches the manifest — which is why the read-only assumption went unnoticed.

## Fix

Make `save_manifest` best-effort on `OSError`, exactly mirroring the `write_audit_entry` precedent established by da#335 in `src/pipeline/audit.py`:

```python
    try:
        audit_dir.mkdir(parents=True, exist_ok=True)
        path.write_text(json.dumps(asdict(entry), indent=2) + "\n")
    except OSError:
        logger.warning("audit_write_failed", stage=entry.stage, path=str(path))
        return None
```

Apply the same shape in `src/pipeline/manifest.py::save_manifest`: catch `OSError`, emit a structured warning, return `None`; widen the return type accordingly.

### Why swallowing is correct here

The manifest is a **change-detection optimization**, not a correctness requirement:

- `MANIFEST_FILENAME` is informational.
- `LAST_LOADED_MANIFEST_FILENAME` only feeds `--incremental`'s skip-unchanged path. If it cannot be written, the next `--incremental` run does *more* work, never *wrong* work.

Additionally, `deploy-data-load.yml` does `rm -rf "${LOAD_DATA_DIR}"` before every run, so the manifest already has **zero** cross-run persistence in the containerized path. Nothing is lost.

`save_manifest` has no callers outside `_cmd_load` and its own tests (verified), so this does not weaken error reporting anywhere else.

## Scope

- `src/pipeline/manifest.py` — `save_manifest` best-effort on `OSError` + structured warning.
- Tests in `tests/test_pipeline/test_manifest.py` — add a read-only-directory case asserting `save_manifest` returns `None`, warns, and does **not** raise. Existing write-path tests must still pass unchanged.
- Do **not** change the compose mount. The mount stays `:ro` by owner decision (2026-07-09); the loader is what must stop writing into its inputs.

## Acceptance

- `uv run isnad-ingest load` proceeds past manifest generation when `data_dir` is read-only.
- Full local CI parity green before push (ruff format + lint, mypy, pytest, and every other check CI runs).

## Related

- deploy#553 (the blocker as seen from the deploy side; keeps `:ro`, corrects a false comment)
- da#335 (the `write_audit_entry` precedent this mirrors)

parametrization:
Closes #348

## What changed

`src/pipeline/manifest.py::save_manifest` is now best-effort on `OSError`: it catches the error, emits a structured `manifest_write_failed` warning, and returns `None`. The return type widens from `None` to `Path | None` (the written path on success).

This mirrors the `write_audit_entry` precedent established by da#335 in `src/pipeline/audit.py` — same `try` shape around `mkdir` + `write_text`, same structured-warning-then-`None` contract, same docstring framing.

The compose mount is **unchanged**. It stays `/app/data:ro` by owner decision; the loader is what must stop writing into its inputs.

## Why swallowing `OSError` is correct here

The manifest is a **change-detection optimization**, not a correctness requirement:

- `MANIFEST_FILENAME` (`.manifest.json`) is purely informational.
- `LAST_LOADED_MANIFEST_FILENAME` only feeds the `--incremental` skip-unchanged path. If it cannot be written, the next `--incremental` run does *more* work, never *wrong* work.
- `deploy-data-load.yml` runs `rm -rf "${LOAD_DATA_DIR}"` before every run, so in the containerized path the manifest already has **zero** cross-run persistence. Nothing is lost by not writing it.

**No error reporting is weakened elsewhere.** `save_manifest` has exactly two non-test callers, both in `_cmd_load` (`src/cli.py:178` and `src/cli.py:222`), plus its own tests in `tests/test_pipeline/test_manifest.py`. Verified with a repo-wide grep; there are no other call sites.

Before this change, `src/cli.py:178` ran unconditionally *before* the `incremental` branch and before `load_all`, so every `isnad-ingest load` inside `graph-load` died with `OSError: [Errno 30] Read-only file system: 'data/.manifest.json'` — no containerized load had ever succeeded. `graph-ops.yml` shares the image but only exercises `enrich`, which never touches the manifest, which is why the read-only assumption went unnoticed.

## Tests

Added to `tests/test_pipeline/test_manifest.py`:

- `test_returns_written_path` — the success path returns the `Path` (covers the widened return type).
- `test_read_only_dir_is_best_effort` — `write_text` raises `OSError(30)`; asserts `save_manifest` returns `None`, logs `manifest_write_failed` at `warning`, writes no file, and does not raise.
- `test_read_only_parent_mkdir_is_best_effort` — same contract when the failure surfaces on the parent `mkdir` instead.

Existing write-path tests pass unchanged.

**Monkeypatch, not `chmod` — and why.** The brief allowed either. I chose `monkeypatch.setattr(Path, "write_text", ...)` raising `OSError` because **a test process running as root does not observe `EACCES` on a `chmod`-ed directory** — root bypasses the permission bits, the write succeeds, and the test passes green *without ever exercising the `except OSError` branch*. That is a silently-vacuous test, and container CI running as root is exactly the environment this fix targets. It also leaves no unwritable directory behind on failure. This is the same approach the existing `write_audit_entry` best-effort tests take (`tests/test_pipeline/test_audit.py:54` and `:76`), so the two now match.

I separately verified the **real** read-only case by hand as an unprivileged user (uid 1000), against an actual `chmod a-w` directory:

```
warning  manifest_write_failed  path=/tmp/…/data/.manifest.json  service=isnad-graph-ingestion
result: None
file exists: False
rc=0
```

So the genuine `OSError` — not just the injected one — is caught, warned, and swallowed.

## CI / lint commands run, and their results

Run from the worktree, before push. Full local⇄CI parity, not a subset.

| Command | Result |
|---|---|
| `uv run ruff format --check .` | `239 files already formatted` |
| `uvx ruff@0.15.11 format --check .` (pinned, matches hooks-lint CI) | `239 files already formatted` |
| `uv run ruff check .` | `All checks passed!` |
| `uv run mypy src/` | `Success: no issues found in 97 source files` |
| `ENVIRONMENT=test uv run pytest tests/ -x --tb=short -q -m "not integration"` | `1777 passed, 7 skipped, 47 deselected` |
| `python3 scripts/structural_ontology.py check` | initially stale → regenerated via `emit`, now `OK: structural ontology index is current with source` |

`src/pipeline/manifest.py` changed, so the structural-ontology staleness gate fired; `ontology/structural/{code-graph.json,llms.txt}` are regenerated and committed in this PR.

Full pre-commit and pre-push hook sets ran and passed on the real commit/push (ruff-format, ruff-lint, gitleaks, actionlint, cspell, dockerfile-base-pin, fixture-realism, structural-ontology-staleness; then mypy-strict + pytest at push). No `--no-verify`.

## Related

- deploy#553 — the blocker as seen from the deploy side (keeps `:ro`, corrects a false comment)
- da#335 — the `write_audit_entry` precedent this mirrors

https://claude.ai/code/session_01BmCDgMue9jRtYqqUxvbDN1


parametrization:
RequestOrReplied: Approved
TechDebt: none

Reviewed the artifact at `bf0972329722930470729bb0dbb8b0543ef90e9b` (not the PR body), plus the two call sites, the enrich path, the compose mount, and both deploy workflows.

## The fix is correct, and — more to the point — it is *complete*

The thing I most wanted to disprove was that this just relocates the crash. `_cmd_load` writes into `data_dir` in three places: the manifest at `src/cli.py:178`, the last-loaded manifest at `src/cli.py:222`, and `write_audit_entry(data_dir, audit)` at `src/cli.py:236`. If that third one still raised, `load` would die under `/app/data:ro` anyway, 14 lines past the fix.

It doesn't. `src/pipeline/audit.py:38-59` already wraps `mkdir` + `write_text` in the same `except OSError` -> structured warning -> `None` shape (da#335). So the full `load` path under a read-only mount is now: `generate_manifest` (reads only) -> warn -> `load_all` (reads only) -> warn -> warn -> exit 0. I grepped `src/graph/` for `write_text` / `open(...,"w")` / `.mkdir(` / `to_parquet` / `json.dump(` and found no writers, so `load_all` contributes nothing. The stage genuinely survives. Mirroring the `write_audit_entry` precedent rather than inventing a new contract was the right instinct.

## On the specific things worth challenging

**Does a caller depend on the swallowed error?** No. Both non-test call sites discard the return value, so widening `None` -> `Path | None` is a pure superset and cannot break a caller or the type contract. Nothing re-reads either manifest within the same process.

**Incremental semantics under a silent write failure.** `load_manifest` returns `{}` for a missing file, so an absent previous manifest makes every file count as *added*, `has_changes` is `True`, and the next incremental run does a **full** load. More work, never wrong work — and loaders are idempotent by repo convention. The failure direction is safe, which is the property that actually licenses swallowing the error.

**Exception breadth.** `OSError` is right. `PermissionError`, `IsADirectoryError`, and `ENOSPC`/`EROFS` are all subclasses, so nothing in the intended class escapes. Notably `json.dumps` sits inside the `try`, but a non-serializable manifest raises `TypeError`, not `OSError` — so a genuine serialization bug still surfaces loudly instead of being masked as a filesystem hiccup. That boundary lands in the right place.

**Monkeypatch vs `chmod`.** Your reasoning is correct and I want to underline it, because the opposite choice is the kind of thing that passes review and rots. Container CI runs as root, root bypasses the mode bits, the `chmod`-ed write would succeed, and the test would go green *without ever entering the `except` branch* — a vacuously-passing test guarding the exact environment the fix targets. Patching `write_text` (and `mkdir` separately) forces the branch. Both tests assert no exception escapes, the return is `None`, and the `manifest_write_failed` event is emitted; the `write_text` one additionally pins `log_level == "warning"`.

## One correction — to the PR body, not the code

The body says `.manifest.json` "is purely informational." That is not true. `src/cli.py:381-382`, inside `_cmd_enrich`'s `if incremental:` branch, reads it as the **fallback previous-manifest** when `.last_loaded_manifest.json` is absent:

```python
previous_manifest = load_manifest(data_dir / LAST_LOADED_MANIFEST_FILENAME)
if not previous_manifest:
    previous_manifest = load_manifest(data_dir / MANIFEST_FILENAME)
```

I traced both directions before deciding whether this bites:

- **Both writes fail** (the actual read-only-mount case): previous is `{}`, everything reads as added, enrich runs in full. Safe.
- **`.manifest.json` written but the `:222` write fails** — which needs a *mid-run* filesystem transition, e.g. `ENOSPC` during `load_all`, since a static `:ro` mount fails at `:178` first and never creates the file: a later `enrich --incremental` would read `.manifest.json`, find it identical to current, and silently skip. Pre-PR that raised loudly at `:222`.

Operational exposure is **nil**, which is why this isn't a change request. `deploy-data-load.yml` passes `load` or `load --nodes-only` and never `--incremental`, and it `rm -rf`s `LOAD_DATA_DIR` before every run (`:309`), so no manifest has cross-run persistence there. `graph-ops.yml` invokes a bare `enrich` (`:189`). The fallback branch is unreachable in both deployed paths. Worth fixing the sentence in the body so the next person doesn't inherit the claim as license to lean on it; the code stands as-is.

Nit, non-blocking: `test_read_only_parent_mkdir_is_best_effort` checks the event name but not `log_level`, unlike its sibling. And `assert not path.exists()` is close to vacuous once `write_text` is stubbed. Neither is worth another round.

Approving. Good, disciplined fix — narrow blast radius, correct failure direction, tests that actually exercise the branch they claim to.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: #350

## QA review — test quality and data-integrity risk

Reviewed the artifact at `bf09723`, not the PR body. Nikolaos has the correctness/call-graph lens; mine is whether the new tests actually test anything and whether a silently-failing manifest write can mask a bad load.

### The tests are not vacuous, and the monkeypatch is the *more* faithful choice

I went looking for the `feedback_test_mock_masks_prod_failure` failure mode — an injection mock that bypasses the very condition it claims to exercise. It does not apply here, and I want to be precise about why rather than wave the memory around.

The injected `OSError(30, "Read-only file system", ...)` and a real `EROFS` traverse **identical** code, because nothing branches on the error's identity. `save_manifest` catches bare `OSError`, and `grep -rn "errno\|EROFS" src/` returns **zero hits** across the whole package — no code inspects `.errno`, `.strerror`, or `.filename`. The mock therefore carries everything the real error carries *that this code path can observe*. The condition under test is exactly `except OSError`, and the test exercises exactly that.

A `chmod`-based test would have been **worse**, on two independent counts:

1. Container CI runs as root, and root bypasses permission bits. The write would succeed, the `except` branch would never execute, and the test would pass green while asserting nothing. That is precisely the vacuous test I was sent to look for — and it is the shape the alternative would have taken.
2. `chmod a-w` yields `EACCES` (errno 13), not `EROFS` (errno 30). It would model a *different* error than the one that killed the stg load.

Hand-verifying the genuine read-only case as uid 1000 and reporting the actual log line is the right way to close the residual gap between the injected and real error. Noted and appreciated.

### The data-integrity question: no, a silent manifest failure cannot mask a bad load

This was the one that mattered, so I traced `_cmd_load` at head rather than taking the PR body's word.

The manifest is consulted to decide *what to load* in exactly one place — `load_manifest(data_dir / LAST_LOADED_MANIFEST_FILENAME)` at `src/cli.py:184`, whose result feeds `compare_manifests` and then `skip_files` at `:216`. Every failure of that read degrades toward **more** loading, never less:

- File absent, because the write was swallowed. `load_manifest` returns `{}` at `manifest.py:118`.
- Empty previous manifest. `compare_manifests` puts every current key in `added`, none in `unchanged`.
- `skipped_files` is therefore empty, and `load_all` skips nothing.

So the fail-open direction is a full reload. The write at `:178` (`.manifest.json`) is never read back by anything, and the write at `:222` only ever *records*. The best-effort contract is sound, and the widened return type has no production consumer — both call sites discard it, which is fine and makes the `Path | None` purely a testability affordance.

I also verified the `rm -rf` claim rather than assuming it: `noorinalabs-deploy/.github/workflows/deploy-data-load.yml:309` does `rm -rf "${LOAD_DATA_DIR}"` before every run, and that workflow's `load_args` input never passes `--incremental`. Cross-run manifest persistence in the containerized path is genuinely zero. Both claims in the PR body hold.

### What I did find — filed as #350, not a blocker here

Tracing `:222` surfaced a **pre-existing** hazard that this PR neither introduces nor worsens. `save_manifest(current_manifest, LAST_LOADED_MANIFEST_FILENAME)` is unconditional — no `nodes_only` guard — and `current_manifest` is the snapshot taken at `:177`, *before* `load_all` runs. It records what was **present on disk**, not what was **ingested**. After `load --nodes-only`, the file asserts the edge-bearing Parquet was loaded; a later `load --incremental` then finds those files `unchanged` and passes them in `skip_files`, so edges silently never load while the run reports success. `strict=False` at `:213` means nothing raises.

It is latent only because of the two incidental mitigations above. It is live today for a local operator running nodes-only then incremental. I filed #350 with the trace and a suggested durable fix (have `LoadSummary` report ingested files and persist only those keys).

Worth being explicit about the direction, since it is the thing that decides blocker-vs-followup: this PR **reduces** exposure to #350. A manifest write that fails and is swallowed cannot poison a subsequent incremental run. Blocking on it would be scope creep, and per the org's own guidance I am not going to hold a correct, narrowly-scoped fix hostage to an orthogonal pre-existing defect.

### Test-hygiene nits, none blocking

- `test_read_only_dir_is_best_effort` asserts `log_level == "warning"`; `test_read_only_parent_mkdir_is_best_effort` asserts only the event name and drops the level check. Harmless asymmetry, but the level is the part an operator's alerting keys on — worth mirroring.
- The mkdir test does not assert `not path.exists()`, where its sibling does. Trivially true (the parent is never created), so the assertion is cheap to add and makes the pair symmetric.
- `monkeypatch` auto-reverts the class attribute and nothing is written, so both tests leave the filesystem clean. Confirmed — no `chmod`-ed directory left behind for a later test to trip over, which is the other reason the monkeypatch approach wins.
- `test_returns_written_path` covers the widened return type on the success path. That was the regression I was asked to check for, and it is present.

One operational note for whoever watches the stg load: on a read-only mount this now emits `manifest_write_failed` **twice** per `load` invocation, from `:178` and `:222`. Two warnings per run is expected, not a symptom of a retry loop.

### Verdict

The fix is minimal, matches the `write_audit_entry` precedent, the tests exercise the real branch, and the fail-open direction is toward reloading rather than skipping. Approving.

PR #357, opened 2026-07-09
fix(parse/sanadset): name-key the books join, populate hadith_number, content-derive source_id (da#353)

### LINKED ISSUE #353: bug(parse/sanadset): column-name mismatch nulls hadith_number+book_number for 76% of corpus, degrades source_id to a row offset, and fakes a double-prefix
## Corrected diagnosis (2026-07-09) — the original framing in this issue was wrong

This issue was filed as *"the batch path constructs the `sanadset` hadith id without routing through `hadith_node_id()`"*. **That is not the bug.** The batch path routes through `generate_source_id()` and `hadith_node_id()` correctly. Verified at `main` (`src/parse/sanadset.py:453`, `src/graph/load_nodes.py:363`).

The real defect is a **column-name mismatch in the Sanadset parser** that silently degrades the id, the collection, and two node properties for **76% of the entire corpus** (650,986 of 853,218 hadiths). The 16M warnings are a *symptom*, and the collapse guard is firing on an id that is not actually double-prefixed.

## Ground truth

`data/raw/sanadset/sanadset.csv` header:

```
Hadith,Book,Num_hadith,Matn,Sanad,Sanad_Length
```

`data/raw/sanadset/books.csv` header — a bare list of 956 Arabic book names:

```
Book
```

## Three chained failures

**1. `_parse_books` bails; da#219 has never activated.**
`_parse_books` (`src/parse/sanadset.py:348`) resolves its id column from `("book_id", "id", "bookid")`. `books.csv` has only `Book`. So `id_col is None` → it logs `books_csv_no_id_column` once → returns `{}` → `has_books = False`. The da#219 book→collection mapping is dead code in production. The warning has been firing into the void.

**2. The hadith CSV column mapping misses both real columns.**

| Parser looks for | Actual column | Result |
|---|---|---|
| `hadith_id` / `id` / `Hadith_ID` | `Num_hadith` | `hadith_number = NULL` |
| `book_id` / `book` / `Book_ID` | `Book` (case) | `book_number = NULL` |

Measured on `data/staging/hadiths_sanadset.parquet`: `hadith_number` null in **650,986/650,986** rows; `book_number` null in **650,986/650,986**. `Book` is additionally an Arabic *name*, not an int — so even a case fix leaves `safe_int()` returning `None`. Every other corpus is clean:

```
hadiths_sanadset.parquet    rows=650986  hnum_null=650986  bnum_null=650986  distinct_coll=1
hadiths_lk.parquet          rows= 34088  hnum_null=   129  bnum_null=     0  distinct_coll=6
hadiths_thaqalayn.parquet   rows= 33190  hnum_null=     0  bnum_null=     0  distinct_coll=33
...
```

**3. The id degrades to a row offset, and only *looks* double-prefixed.**
With `book_num = None`, `_resolve_collection_name` falls back to the per-CSV-stem `default_name`. The file is literally `sanadset.csv`, so `collection_name == "sanadset" == _SOURCE_CORPUS`. Hence:

```
generate_source_id("sanadset", "sanadset", "0", "0", row_offset)
  -> "sanadset:sanadset:0:0:645817"
```

That is a **legitimate `corpus:collection` id whose collection happens to equal the corpus name** — not the main#139 `corpus:corpus:` double-prefix. `_collapse_double_corpus` cannot tell the two apart, eats the collection segment, and warns. 16,000,000+ times.

## What is and is not broken

**Not corrupted.** I checked for the failure this guard exists to prevent:

```
distinct source_ids: 650986   distinct node ids: 650986
COLLISIONS introduced by collapse: 0
```

The collapse is uniform, so no two hadiths merge. The graph is not currently wrong *in the way ig#63 was wrong*.

**But three real consequences:**

- **The id is positional.** `sanadset:sanadset:0:0:{row_offset}` is keyed on nothing but the row's index in a 1.4 GB CSV. Re-acquire the file with one row inserted, removed, or reordered and **every subsequent hadith is re-keyed** — the graph reloads them as new nodes and every existing `TRANSMITTED_TO` / `PARALLEL_OF` / `APPEARS_IN` edge dangles. This voids the determinism contract da#82 was written to establish, for 76% of the corpus.
- **All 650,986 sanadset hadiths land in ONE `Collection` node** instead of 956, because `collection_name` is constant. The APPEARS_IN breadth da#219 was meant to preserve does not exist.
- **`hadith_number` and `book_number` are NULL for 76% of the corpus** — these are citation fields; a hadith that cannot be cited to book+number is not usable for scholarly reference.

## The correct key exists in the data

`(Book, Num_hadith)` is a **perfect unique key** over the raw CSV:

```
rows:                       650986
distinct books:                956   (== books.csv row count)
Num_hadith nulls:                0
distinct (Book, Num_hadith): 650986
duplicate keys:                  0
```

956 distinct `Book` values, exactly matching `books.csv`'s 956 names. The book→collection join key is the **name**, not an id.

## Proposed fix

1. **`_parse_books`** — accept a name-only `books.csv`. Key the mapping on a **stable, content-derived** book key (e.g. a short digest of the normalized Arabic name), never on `books.csv` row order — a positional book index reintroduces exactly the instability this issue is about, one level up. Carry the Arabic name onto the `Collection` node as `name_ar`.
2. **Hadith column mapping** — read `Num_hadith` as `hadith_number`, and `Book` as the collection **name**, joining to the `_parse_books` mapping to obtain the book key. Match column names case-insensitively (as `_parse_books` already does) rather than by enumerating spellings.
3. **`source_id`** — emit `sanadset:{book_key}:{num_hadith}`. Content-derived, stable under re-acquisition, and no longer collides with the corpus name. Drop `row_offset` from the id entirely.
4. **Guard the class of bug, don't repair it.** Once (3) lands, no producer emits `corpus:corpus:`. Then make an unexpected doubled prefix in the *input* a hard error rather than a silent collapse (see the follow-up issue on `_collapse_double_corpus`'s false-positive: it cannot distinguish a real double prefix from a collection legitimately named after its corpus).
5. **Fail loudly on a dead mapping.** `books_csv_no_id_column` returning `{}` silently disabled a feature for months. A parser that cannot resolve its join key for a corpus of this size should raise, not warn.

## Blast radius / sequencing

This **re-keys 650,986 hadith node ids**, so it must land *before* the pending wipe-and-rebuild, not after. It also means the pipeline re-run starts at **`parse`**, not `resolve` — one stage earlier than currently planned, since `resolve` consumes `source_id` from staging.

## Acceptance

- `hadiths_sanadset.parquet`: `hadith_number` non-null for all 650,986 rows; `distinct collection_name == 956`.
- No `collapsed_double_corpus_prefix` warning during a full load (count == 0, not "reduced").
- `source_id` for a fixed `(Book, Num_hadith)` is byte-identical across two parses of CSVs that differ by an inserted row.
- `tests/test_parse/test_sanadset.py` covers the name-keyed books join and the row-insertion stability property.

## Lineage

`main#139` is the P4W2 end-to-end test issue in which the *streaming* double-prefix was found; `ig#63`/`ig#72` fixed that path; `da#82` settled the collision-safe id scheme. None of them touch this: the Sanadset parser never had the double-prefix bug — it has a **column-mapping bug that produces an id resembling one**.

parametrization:
## Summary

The Sanadset parser looked for `hadith_id`/`book_id` columns that **do not exist** in the real CSV (header `Hadith,Book,Num_hadith,Matn,Sanad,Sanad_Length`). Consequence, on **650,986 / 853,218 hadiths (76% of the corpus)**: `hadith_number`/`book_number` NULL; `collection_name` fell back to the CSV stem `"sanadset"` (== the corpus), collapsing every row into ONE Collection and yielding positional ids `sanadset:sanadset:0:0:{row_offset}` that (a) re-key on any re-acquisition and (b) falsely trip `_collapse_double_corpus` 16M+ times per load.

This is a column-mapping bug, not the main#139 double-prefix — verified against the issue's corrected diagnosis before implementing.

## What changed

- **`_parse_books`** — supports the real **name-only** `books.csv` (single `Book` column, 956 Arabic names). Keyed on a stable content digest of the `normalize_arabic`-normalized name (`_book_key`, BLAKE2b) — **never** books.csv row order. A **present-but-unmappable** `books.csv` now **raises** instead of silently returning `{}` (the `books_csv_no_id_column` foot-gun that shipped dead for months); a genuinely absent file remains a legitimate pre-B1 fallback.
- **`_process_chunk`** — resolves source columns **case-insensitively** (not by enumerating spellings). `Num_hadith` → `hadith_number`; `Book` (a NAME) → `_book_key` → `collection_name`. `book_number` is left null (Sanadset has no book ordinal — `Book` is a name).
- **`source_id`** = `generate_source_id("sanadset", book_key, num_hadith)` — content-derived and stable under re-acquisition. **`row_offset` dropped entirely** (and the `global_row_offset` bookkeeping with it).
- `generate_source_id` fail-fast (da#89) preserved — `_parse_narrator_bios` still joins its non-corpus `bio_id` manually.

## Acceptance evidence

Measured on a **representative 40,000-row slice** of the real `data/raw/sanadset/sanadset.csv` + the **full real `books.csv`** (not the full 650,986-row parse — stated plainly). The full `books.csv` parsed to exactly **956** name-keyed entries.

| Check | Result |
|---|---|
| `hadith_number` nulls | **0 / 40,000** |
| distinct `collection_name` in slice | **59** (the 59 books present in the slice; was **1** before) — per-book breadth restored |
| distinct `source_id` | **40,000 == rows** → 0 collisions |
| `source_id` starting `sanadset:sanadset:` | **0** |
| `is_double_prefixed` count | **0** |
| `_double_corpus_collapse_count` after parse | **0** (was 16M+) |
| Collections emitted with `name_ar` | **59 / 59** (Arabic names flow onto Collection nodes) |

Sample id: `sanadset:b52612b6fc65c4c8:1` (book-name digest + `Num_hadith`, no offset). The full-corpus `hadith_number`-non-null (650,986) and `distinct collection_name == 956` follow by construction: every sliced row got a non-null number, and `books.csv` name-keys to all 956 distinct names.

New/updated unit tests (`tests/test_parse/test_sanadset_parser.py`, real-schema fixtures): name-keyed join breadth, `hadith_number` populated, no-double-corpus, **row-insertion stability** (a fixed `(Book, Num_hadith)` yields a byte-identical `source_id` across two parses differing by an inserted row), and present-but-unmappable-`books.csv`-raises. Full local gate green: ruff 0.15.11 format+lint, mypy strict, and 888 parse+graph tests.

## APPEARS_IN + null-MERGE finding (requested)

Read `src/graph/load_edges.py::_load_appears_in` and `_APPEARS_IN_QUERY` at HEAD, cross-checked against `feedback_appears_in_merge_null`. **The null-MERGE-abort bug does not apply at HEAD** — the query is already the da#77 null-safe form: `MERGE (h)-[r:APPEARS_IN]->(c)` then `SET r.hadith_number_in_book = coalesce(row.hadith_number, r.hadith_number_in_book)` (positional props SET *after* the MERGE with coalesce-preserve, never inside the MERGE pattern). So **populating `hadith_number` does not change edge-creation behavior**: the edge now simply carries a real `hadith_number_in_book` instead of a null. `book_number` stays null → `coalesce(null, r.book_number)` leaves the key absent (per `feedback_neo4j_set_null_removes_property`), which is correct — Sanadset has no book ordinal. The join key still agrees on both endpoints (`collection_name` and the Collection's `collection_id` both derive from `_book_key`).

## Contradictions with the diagnosis

None. Every measured fact matched the corrected issue writeup (name-only `books.csv`, `(Book, Num_hadith)` unique, collapse-introduces-0-collisions, 956 == books.csv rows).

## Sequencing / out of scope

Re-keys 650,986 hadith ids → the pipeline re-run starts at **`parse`**, before the pending wipe-and-rebuild. Changing `_collapse_double_corpus` to assert rather than repair is out of scope (da#355), sequenced after this; `src/parse/identity.py` is untouched.

Closes #353

---

## Review round 2 — books.csv-absent identity collapse (blocking, fixed)

Both reviewers independently found a blocking defect in the first cut, and they were right.

The `has_books == False` fallback emitted `sanadset:sanadset:{Num_hadith}`, which `hadith_node_id` collapses to `sanadset:{Num_hadith}`. Measured on the real 1.4 GB `sanadset.csv`: **650,986 rows but only 37,239 distinct `Num_hadith`**, so the loader would have MERGEd **613,747 hadiths (94.3% of the corpus)** on top of one another — signalled only by a `collapsed_double_corpus_prefix` warning this PR's summary had described as eliminated noise. The pre-PR code emitted an unstable-but-*unique* row-offset id, so the first cut was strictly more destructive than what it replaced.

**Why this could not be deferred to da#355:** `src/acquire/sanadset.py:174-179` skips the entire Mendeley download whenever `dest.glob("*.csv")` is non-empty and never checks that `books.csv` landed, and `_download_mendeley` has no rollback — and **`books.csv` itself matches that glob, so a lone `books.csv` left behind by a failed run suppresses the whole download.** A transient failure fetching the 48 KB `books.csv` after the 1.4 GB `sanadset.csv` is written therefore poisons every later run, permanently. `has_books == False` was the actual production state until this PR.

### Fix

`books.csv` supplies Collection **metadata**; the join key comes from the hadith row's own `Book` column. The first cut gated the *key* on `books.csv`'s presence.

1. `_resolve_collection_name` — drop the `has_books` conjunct; the row's own `Book` digest is the key whenever it exists.
2. `_collection_row` — re-key the label branch on `meta is None`, so a digest-keyed collection never takes the raw digest as its `name_en`.
3. `_process_chunk` — raise when the `Book` column is absent entirely, matching the fail-loud standard this PR already applies to an unmappable `books.csv`.

Net effect: a `books.csv`-absent run yields the same 650,986 stable, distinct, non-doubled ids as the happy path, with Collections labelled `Sanadset book {key}` instead of an Arabic name. **Degraded metadata, intact identity.**

### Verification on the real corpus

Raw `sanadset.csv`: 650,986 rows · 37,239 distinct `Num_hadith` · 956 distinct `Book` · `(Book, Num_hadith)` unique at 650,986 · **0** blank `Book` cells.

Full parse with `books.csv` withheld:

```
hadith rows parsed        : 650986
distinct source_id        : 650986
distinct hadith_node_id   : 650986
double-prefixed source_id : 0
collections emitted       : 956
sample collection name_en : 'Sanadset book 00f99e275dde120f'
IDENTITY PRESERVED: True
```

### Tests

`TestBooksCsvAbsentIdentity` parses a fixture named literally `sanadset.csv` with no `books.csv`, asserting distinct `source_id` count == row count and no doubled corpus segment. Two design points, both load-bearing:

* The fixture **must** be named `sanadset.csv` — only then does the CSV stem equal the corpus name and reproduce the doubled prefix. The prior `test_collections_emitted_without_books_csv` used `hadiths.csv` and was structurally incapable of failing.
* Its two books **reuse** `Num_hadith` 1 and 2. With globally unique ordinals the ids stay distinct even under the collapsing code and the test proves nothing.

All five new/changed tests fail against the pre-fix parser (`4 rows -> 2 distinct ids`; `sanadset:sanadset:1`; `DID NOT RAISE ValueError`).

The fail-loud guard then exposed **five** further fixtures — in `conftest.py`, `test_real_data_flow.py`, `test_sanadset_lightup.py`, and `test_pathb_integration_verify.py` — built on a synthetic `hadith_id,book_id,hadith` header that exists in no edition of the corpus. All migrated to the real `Hadith,Book,Num_hadith`. Two of them (`test_sanadset_lightup`, `test_pathb_integration_verify`) were asserting *through* the CSV-stem fallback rather than the per-book name join they claimed to exercise; `test_sanadset_lightup` now asserts 2 Collections (البخاري, مسلم) instead of 1.

`book_number` stays null: `books.csv` is an unordered name list, so its row index is exactly the re-acquisition-unstable positional value this PR removes. Accepted tech debt on both verdicts.

### Also

`.claude/memory/feedback_appears_in_merge_null.md` is annotated stale-at-HEAD (retained, not deleted). Its body already recorded the da#77 fix, but its frontmatter `description` and `MEMORY.md` pointer still asserted the bug in present tense — and those are the only lines a reader sees before deciding whether to open the file. `src/graph/load_edges.py:440-448` has been the null-safe coalesce-after-MERGE form since da#77, so emitting a null `book_number` is safe.


parametrization:
Verdict: Changes Requested
TechDebt: Sanadset has no book ordinal, so `APPEARS_IN.book_number` stays null on all 650,986 edges — book identity is reachable only via the Collection endpoint's `name_ar`; accept for now, revisit if a UI needs book ordering.

Reviewed the diff and the code at `0766147`, not the PR body. Ran the checks against the real `data/raw/sanadset/{sanadset.csv,books.csv}` rather than a slice.

## Everything the PR claims about the happy path is true, and true at full scale

The author's evidence was a 40,000-row slice. I re-measured on all 650,986 rows:

| Check | Full-corpus result |
|---|---|
| rows | 650,986 |
| null/empty `Book` (via `safe_str`) | 0 |
| null `Num_hadith` (via `safe_int`) | 0 |
| distinct `(book_key, num_part)` | 650,986 — zero collisions |
| distinct hadith-side `Book` names | 956 |
| hadith rows whose `Book` is absent from `books.csv` | 0 |
| `books.csv` names → distinct `_book_key` | 956 → 956, zero digest collisions |
| `normalize_arabic` idempotent on all 956 names | yes |

So the uniqueness claim generalizes from the slice. `_book_key` is a fixed 16-hex-char BLAKE2b digest (`digest_size=8`) that can never contain the `:` delimiter, and no two of the 956 book names normalize to the same string. Items 2 and 3 of your brief clear.

I also independently confirmed item 5. At `0766147`, `src/graph/load_edges.py:440-448` MERGEs the bare pattern and then does `SET r.hadith_number_in_book = coalesce(row.hadith_number, r.hadith_number_in_book)`. No property lives inside the MERGE pattern, so populating `hadith_number` for 650,986 rows changes only the value the edge carries, never whether the edge is created. `book_number` staying null coalesces to the existing value and leaves the key absent, which is the behavior the null-removes-property memory describes. The `feedback_appears_in_merge_null` memory is stale at HEAD and should be retired or annotated.

Item 7 clears: `src/parse/identity.py` is untouched. The PR's four files are `src/parse/sanadset.py`, `tests/test_parse/test_sanadset_parser.py`, and the two regenerated `ontology/structural/` artifacts.

Item 6 clears. `test_source_id_stable_under_row_insertion` (`tests/test_parse/test_sanadset_parser.py:565`) is not vacuous — `csv_b` inserts a row ahead of the target, which under the old row-offset id would move the target from offset 0 to offset 1 and produce a different string, so the test fails when the fix is reverted. `test_present_but_unmappable_books_raises` (line 580) writes a `serial`-only header, `pick()` returns `None`, and the raise fires for real. Nit only: line 583 re-imports `pytest` inside the function body, though the module already imports it (it is used at line 184).

## Blocking: the fallback branch is now strictly more destructive than the code it replaces

This is your item 1, and having measured it I think it is worse than "latent," so I am holding the merge on it rather than deferring to da#355.

Before this PR, when `has_books` was false the id was `sanadset:sanadset:0:0:{row_offset}` — unstable under re-acquisition, but **unique per row**: 650,986 rows produced 650,986 nodes. After this PR the same branch produces `sanadset:sanadset:{Num_hadith}`. `Num_hadith` is an in-book ordinal, and the corpus has only **37,239 distinct values of it**. `hadith_node_id` collapses the doubled corpus segment, so those 650,986 hadiths resolve to 37,239 node ids and the loader MERGEs **613,747 hadiths (94.3% of the corpus) on top of each other**, silently. The only signal is the `collapsed_double_corpus_prefix` warning — the very warning this PR's summary presents as the noise it eliminated, so an operator seeing it would now reasonably read it as a stale symptom.

That branch is reachable, and not by an exotic route. `src/acquire/sanadset.py:174-179` skips the entire Mendeley download when `dest.glob("*.csv")` is non-empty, and its only post-condition is `if not hadith_csvs: raise` — it never checks that `books.csv` landed. `_download_mendeley` iterates `_MENDELEY_FILES` and downloads `sanadset.csv` and `books.csv` in sequence with no rollback, so a transient failure on the 48 KB `books.csv` after the 1.4 GB `sanadset.csv` has been written leaves a raw dir that every subsequent `make acquire` treats as complete. Pre-PR that produced unstable ids. Post-PR it produces a 94% silent node collapse. `has_books == False` was the *actual production state* until this PR — this is the branch we were living in.

No test covers it, and the existing one structurally cannot. `test_collections_emitted_without_books_csv` (`tests/test_parse/test_sanadset_parser.py:228-246`) names its fixture `hadiths.csv`, so `default_name` is `"hadiths"` and `collection_id` is `sanadset:hadiths` — a stem that can never equal the corpus. The production file is `sanadset.csv`.

### The fix is smaller than the guard you might expect

`book_key` is computed at `src/parse/sanadset.py:539` from the hadith row's **own** `Book` column. It does not depend on `books.csv` at all. `books.csv` supplies only `name_ar` / `name_en` / `compiler_name` for the Collection node. So `has_books` is gating the wrong thing: it gates the *join key* when it should gate only *metadata enrichment*.

In `_resolve_collection_name` (`src/parse/sanadset.py:439-450`), drop the `has_books` conjunct:

```python
if book_key is not None:
    return book_key
return default_name
```

`_collection_row` already handles the missing-metadata case, but its label branch at `src/parse/sanadset.py:470-472` keys off `has_books` too and must follow the same predicate, or a digest collection would take the digest itself as its `name_en`. Key that branch on `meta is None` instead.

That single change makes the books.csv-absent run produce the same 650,986 stable, distinct, non-double-prefixed ids as the happy path, with Collections that carry a `Sanadset book {key}` label instead of an Arabic name. Degraded metadata, intact identity — which is the right failure shape.

Then the only surviving `default_name` path is "the hadith CSV has no `Book` column at all," which is a broken corpus by the same standard this PR already applies to `books.csv`. Make `_process_chunk` raise when `book_col is None`, matching the fail-loud contract the PR itself establishes. Please add a test that parses a fixture literally named `sanadset.csv` with no `books.csv`, asserting distinct `source_id` count equals row count and that no id starts with the doubled corpus segment.

Minor, non-blocking: the `_content_digest` tail is only reached when `Num_hadith` is null, where two byte-identical texts in one book would share an id. Those are genuine duplicates, so that is fine.

## On item 4, book_number

I would leave it null and not synthesize an ordinal. `books.csv` is an unordered list of names; its row index is exactly the re-acquisition-unstable positional value this PR exists to remove. Writing it as a "merely informative" property still bakes an arbitrary upstream ordering into the graph, and `Collection.name_ar` already answers "which book." Recorded above as the debt line rather than a change request.

parametrization:
Requestor: Oyunbileg Batbayar
Verdict: Changes Requested
TechDebt: `collection_id` is an opaque `sanadset:<16-hex>` digest with `book_number` null — a graph node cannot render a numeric book citation, only the Arabic `name_ar`; file a follow-up to carry a stable human-readable book slug alongside the digest.

Reviewed for data quality, evidence, and test integrity at `07661477f07d4cb73ff96d13f00e4fdd0f4fa146`. Co-reviewer Kavitha Sundaramurthy has the parse-logic lens; I did not duplicate it.

The parse change is right and the tests are real. I am requesting changes on one thing only: the `books.csv`-absent fallback still mints the exact id shape this PR exists to eliminate, it is reachable through a mundane operational path, and the one test that covers that path is constructed so it cannot see the defect.

## What I verified, and what held

I ran the checks against the **full 650,986-row raw corpus** in `data/raw/sanadset/`, not a slice.

**Collision safety — clean, decisively.** All 956 `books.csv` names are distinct after `normalize_arabic`, and their BLAKE2b-64 digests are distinct. On the hadith side, all 650,986 rows carry a non-empty `Book`, those resolve to exactly 956 distinct book keys, and every one of them is present in `books.csv` (zero unmatched). Zero rows have a null or unparseable `Num_hadith`. The `(book_key, Num_hadith)` pair is unique across the corpus — **0 duplicate pairs, 0 rows involved**. The uniqueness claim in the PR body is true, and it is now measured rather than asserted.

Note this is an *empirical* property of today's edition, not a structural one. `_process_chunk` has no guard: a future edition shipping one duplicate pair silently mints two hadiths with the same `source_id`, and one of them MERGE-collapses at load with no warning. That is worth a guard someday, but it is not what I am blocking on.

**Test integrity — `test_source_id_stable_under_row_insertion` is not vacuous.** I mutation-tested it. Reverting only the id tail at `src/parse/sanadset.py:550` to a positional `rows.index(row)` and leaving the rest of the PR intact produces `sanadset:2dd4bdb6b7fa0e74:0` versus `sanadset:2dd4bdb6b7fa0e74:1` for the same `(Book, Num_hadith)` across the two fixtures. The test kills that mutant. It proves stability under row insertion, which is the property `da#353` is about, and it pins the exact id form at `tests/test_parse/test_sanadset_parser.py:578` so a shape change cannot pass silently. This is the strongest test in the PR.

**`test_present_but_unmappable_books_raises` (`tests/test_parse/test_sanadset_parser.py:580`) is specific, not a bare "something raised."** It asserts `ValueError` and matches on `book-name column`, so a different `ValueError` from elsewhere in the parse would not satisfy it. It correctly distinguishes present-but-unmappable (raises, `src/parse/sanadset.py:401-406`) from absent (falls back, covered by `test_collections_emitted_without_books_csv`). One gap: the *second* raise at `src/parse/sanadset.py:429-434` — a present file with a name column but no usable rows — has no test.

**Ontology was regenerated, not hand-edited.** I validated every `src/parse/sanadset.py` node in the PR's `ontology/structural/code-graph.json` against the source at head: all 13 function line numbers land exactly on their `def`, and both new symbols (`_book_key`, `_content_digest`) are indexed. That is generator output. CI rollup is 7/7 green.

## Why I am blocking

`_resolve_collection_name` (`src/parse/sanadset.py:439-450`) branches on `has_books`, **not on whether the individual row has a book name**. So the fallback is not gated by "rows where `Book` is missing" — there are none. It is gated by whether `books.csv` was read at all. When it was not, *every* row takes the fallback and gets the CSV stem, and production's hadith file is named `sanadset.csv`.

I ran the PR's parser at head against a raw dir holding `sanadset.csv` and no `books.csv`:

```
FALLBACK source_id       = sanadset:sanadset:7
is_double_prefixed(sid)  = True
collection_id            = sanadset:sanadset
```

That is the 16M-warning id shape, restored in full, across all 650,986 rows, with `name_ar` null on the single collection node (`src/parse/sanadset.py:469`). The parser does not raise. It does not warn. `_collapse_double_corpus` absorbs it downstream, once per row.

This is reachable without anybody doing anything unusual. `download_sanadset` at `src/acquire/sanadset.py:175-179` short-circuits the Mendeley download when `dest.glob("*.csv")` finds *any* CSV — so a raw dir that has `sanadset.csv` and lost `books.csv` (interrupted download, partial restore, selective cleanup) will never re-fetch it. The acquisition validator at `src/acquire/sanadset.py:203-206` then checks only that *some* CSV exists; it never checks for `books.csv`. Acquisition reports success, parse reports success, and the corpus reloads with positional-equivalent collapsed ids.

And the reason nobody would notice: `test_collections_emitted_without_books_csv` (`tests/test_parse/test_sanadset_parser.py:228`) writes its fixture to **`hadiths.csv`**, so the stem is `hadiths` and the test asserts a clean `sanadset:hadiths` at line 243. Rename that one fixture to `sanadset.csv` — production's actual filename — and the same assertion yields `sanadset:sanadset`. The fallback test passes because of the filename it chose, not because the fallback is safe. This is the pattern the org has already been bitten by twice: a green repro that used a different invocation form than production.

The headline "zero `collapsed_double_corpus_prefix` warnings" is therefore true of the 40,000-row slice *and* of the full corpus as it exists today, and I confirmed that independently. But it is a property of the input, not of the code. The code still contains the defect; the current `books.csv` is what hides it.

## What I need

1. Make the fallback structurally incapable of minting a double-prefixed id. Either fail loud in `parse_sanadset` when `books.csv` is absent (the Mendeley corpus always ships it, and `_parse_books` already fails loud on a *present-but-unmappable* file — an absent file is not obviously the more benign case), or, if the fallback must survive, guard the id: reject a `row_collection` equal to `_SOURCE_CORPUS`, or assert `not is_double_prefixed(source_id)` before append. The helper is already imported in the test module.
2. Rename the fixture in `test_collections_emitted_without_books_csv` from `hadiths.csv` to `sanadset.csv` so the fallback test exercises production's filename, and let it assert whatever the chosen fix guarantees.
3. Add the `books.csv` presence check to `download_sanadset`'s validation block so a partial raw dir fails at acquisition rather than silently degrading at parse.

Items 1 and 2 are the block. Item 3 is small and in an adjacent file; if you would rather split it into its own issue against `src/acquire/sanadset.py`, that is fine by me — say so and I will re-review on 1 and 2 alone.

Everything else here is good work. The name-keyed join is the correct call over a positional book index, the fail-loud contract on an unmappable `books.csv` is exactly right, and populating `hadith_number` restores the citation ordinal that `da#353` nulled. The stability test is the kind of test I wish more PRs had.

parametrization:
## Re-review requested — Kavitha Sundaramurthy and Oyunbileg Batbayar

The blocking defect you both found is fixed. CI is green at `35eed73`: all 12 checks SUCCESS, zero non-terminal, zero failing.

### What changed since your review

`books.csv` supplies Collection *metadata*; the join key comes from the hadith row's own `Book` column. The first cut gated the key on `books.csv`'s presence.

1. `_resolve_collection_name` — dropped the `has_books` conjunct.
2. `_collection_row` — label branch re-keyed on `meta is None`, so a digest-keyed collection never takes the raw digest as its `name_en`.
3. `_process_chunk` — raises when the `Book` column is absent entirely.

### Evidence

Measured on the real 1.4 GB `sanadset.csv`: 650,986 rows, **37,239** distinct `Num_hadith`, 956 distinct `Book`, `(Book, Num_hadith)` unique at 650,986, and **zero** blank `Book` cells. Your 613,747-row loss figure is exact.

Full parse with `books.csv` withheld now yields 650,986 rows, **650,986 distinct `source_id`**, 650,986 distinct `hadith_node_id`, **0** double-prefixed ids, and 956 Collections labelled `Sanadset book {digest}`. Degraded metadata, intact identity.

`TestBooksCsvAbsentIdentity` fails against the pre-fix parser — verified by reverting `src/parse/sanadset.py` and running it (`4 rows -> 2 distinct ids`; `sanadset:sanadset:1`). Two properties of that fixture are load-bearing and should not be "simplified": it is named literally `sanadset.csv` (only then does the CSV stem equal the corpus name), and its two books **reuse** `Num_hadith` 1 and 2 (with globally unique ordinals the ids stay distinct even under the collapsing code).

### Two things worth your attention on re-read

The guard flushed out **five** fixtures across the unit and integration suites built on a synthetic `hadith_id,book_id,hadith` header that exists in no edition of the corpus. All are migrated to the real `Hadith,Book,Num_hadith`.

Two of those fixtures were asserting *through* the CSV-stem fallback rather than the per-book name join they claimed to exercise. `test_sanadset_lightup` asserted exactly one Collection — true only because its fixture had no `Book` column and every row collapsed onto the stem of `bukhari.csv`. It now asserts two (البخاري, مسلم). That same test's `not h["id"].startswith("hdt:sanadset:sanadset:")` assertion was **vacuous** under the old fixture, since a `bukhari` stem can never produce a doubled corpus prefix; it only became a real check once the fixture used the production filename.

`book_number` remains null, as agreed on both verdicts.

parametrization:
Verdict: Approved
TechDebt: A blank `Book` cell (0/650,986 today) still routes to `default_name`, so a future edition with blanks would emit `sanadset:sanadset:{Num_hadith}` and collide those rows — the residue of the branch this PR closed; the tail should be `_content_digest(full_text)` when `book_key is None`.

Re-reviewed at `35eed73`, against the code and the incremental range `0766147..35eed73`, not the PR body. The blocking defect is closed, and closed at the root rather than papered over.

## The hole is closed, not moved

`_resolve_collection_name` (`src/parse/sanadset.py:439`) now takes `(book_key, default_name)` and returns the key whenever the row has one. `books.csv` no longer participates in identity at any point: `has_books` survives only as the `from_books_csv` log field at `src/parse/sanadset.py:876`. `_collection_row`'s label branch keys on `meta is None` (`:471-481`), not on a presence flag in disguise, so a digest-keyed collection with no metadata row gets `Sanadset book {key}` rather than the bare digest as its `name_en`. That was the specific way this fix could have gone wrong, and it did not.

I did not take that on inspection. I loaded the parser at `35eed73` as a module and ran it against a fixture named `sanadset.csv` with no `books.csv` and `Num_hadith` reused across two books, then re-ran it with only `_resolve_collection_name` restored to its old `has_books`-gated behavior and nothing else changed:

```
AS-WRITTEN (35eed73): rows=4 distinct_ids=4 double_prefixed=0
    ['sanadset:2dd4bdb6b7fa0e74:1', 'sanadset:2dd4bdb6b7fa0e74:2',
     'sanadset:8b1b051b73a27366:1', 'sanadset:8b1b051b73a27366:2']
REVERTED (has_books gate): rows=4 distinct_ids=2 double_prefixed=4
    ['sanadset:sanadset:1', 'sanadset:sanadset:1',
     'sanadset:sanadset:2', 'sanadset:sanadset:2']
```

That is the 94.3% collapse reproducing on demand and then not reproducing. It also settles the non-vacuity question for `TestBooksCsvAbsentIdentity` empirically: reverting the fix breaks both `test_every_hadith_keeps_a_distinct_source_id` (4 → 2) and `test_no_source_id_is_double_prefixed` (0 → 4). The class fails when the fix is absent.

Scaling that to the corpus needs no new measurement. The id path no longer reads `books.csv` at all, so the full-corpus numbers I took on my first pass — 650,986 distinct `(book_key, Num_hadith)` pairs, 0 blank `Book`, 0 null `Num_hadith`, 956 names to 956 distinct digests — now describe **both** branches identically. `books.csv`-absent yields 650,986 distinct, non-double-prefixed ids.

The surviving `default_name` paths are exactly two, and I exercised both. No `Book` column at all raises (`src/parse/sanadset.py:546-556`): `Sanadset hadith CSV has no 'Book' column …`. A blank `Book` **cell** still falls through to the stem, and I confirmed two blank-cell rows sharing an ordinal produce `['sanadset:sanadset:1', 'sanadset:sanadset:1']`. Today that is 0 rows out of 650,986 and it is documented in the docstring, so I am not blocking on it — but the blast radius is no longer 613,747 hadiths, it is however many blank cells a future edition ships, and `_content_digest` already exists as the right tail for that case. Carried as the debt line above.

## What the migrated fixtures now prove

Alejandra's report is right in substance and slightly misattributed, which is worth getting on the record since "passed for the wrong reason" is the failure mode we keep finding.

`test_pathb_integration_verify.py` never had a per-book Collection **assertion**. It asserts `assert emitted` and that every hadith's rebuilt endpoint is a member of the emitted set (`:170-175`). Both hold trivially when all rows collapse onto one stem collection — one Collection is emitted, and the single endpoint every hadith rebuilds is that one. The per-book breadth lived only in the file's *comment*. So its old form was a membership tautology: it could not distinguish a healthy 2-collection join from a total collapse. With the real header it now discriminates, because the two books yield two distinct digests and two endpoints that both must resolve.

The genuine wrong-reason pass was in `test_sanadset_lightup.py`, which asserted `len(collections) == 1` (now `== 2`). Under the old synthetic `hadith_id,book_id,hadith` header there was no `Book` column, every row fell back to the CSV stem, and one Collection was the *correct* count for a corpus that had lost its breadth. The test was pinning the bug. It now proves that two distinct `Book` names produce two Collection nodes through a real name-digest join, end to end into Neo4j with `APPEARS_IN` resolving on both.

One caveat on that file, non-blocking. Its fixture is still written to `bukhari.csv` (`:66`), so its neighbouring assertion `not h["id"].startswith("hdt:sanadset:sanadset:")` remains structurally incapable of catching a collapse — the stem can never equal the corpus. It is not wrong, it is just not load-bearing there. `TestBooksCsvAbsentIdentity` is where that property is actually tested, and it uses the production filename.

`test_real_data_flow.py` had the same missing-`Book` header and, since it only asserted that parsing produced outputs, never claimed breadth in the first place. What it now adds is that the production header parses at all through the whole flow.

`TestBooksCsvAbsentIdentity`'s ordinal reuse is real and its docstring says why: البخاري and مسلم both carry 1 and 2, and the docstring states outright that globally unique ordinals would keep ids distinct under the collapsing code and prove nothing. Confirmed against the fixture at `:_CSV`, and confirmed by the revert run above.

## Not covered by CI, worth knowing

Nothing the fix broke. Two things it changes that no check will tell you about:

The missing-`Book`-column raise now applies to **every** `*.csv` in the raw dir except `books.csv`, since `parse_sanadset` globs them all. A stray or partially-written CSV in `data/raw/sanadset/` that used to parse into junk rows now hard-fails the whole parse. That is the intended direction and matches the standard `_parse_books` already sets, but it is a new way for a dirty raw dir to stop the pipeline rather than corrupt it quietly.

Separately, `num_col` still accepts `hadith_id` as a spelling for the citation ordinal (`src/parse/sanadset.py:541`). Harmless against today's header, but a future edition shipping a surrogate `hadith_id` alongside `Num_hadith` would have that surrogate silently written to `hadith_number` and onto `APPEARS_IN.hadith_number_in_book`, which by `project_hadith_number_collection_vs_inbook` must be the in-book ordinal. Cheap to drop that alias while you are here; not a merge condition.

Updating `feedback_appears_in_merge_null` rather than leaving it to mislead the next reader was the right call, and the lesson you added to it — that a fixed-bug memory must say so in its `description`, because that is what recall surfaces — is the more valuable half of this PR.

parametrization:
Requestor: Oyunbileg Batbayar
Verdict: Approved
TechDebt: `_assert_b1_orphan_free` in `tests/integration/test_pathb_integration_verify.py:161-175` asserts only `needed ⊆ emitted`, which is a tautology of the parser's own construction — it cannot detect a collapse of per-book breadth; add a `len(collections) == 2` breadth assertion so the Path-B test guards what its docstring claims.

Re-reviewed at `35eed7397453ab836ff23bb3b4c43091622b51ec`. My earlier Changes Requested against `0766147` is cleared. Data quality, evidence, and test integrity lens; parse logic is Kavitha's.

The fix is better than what I asked for. I proposed failing loud when `books.csv` is absent. `edad7d1` instead removed the `has_books` gate from the join key entirely (`src/parse/sanadset.py:439-461`), so the collection key now comes from the hadith row's own `Book` column and a missing `books.csv` degrades **metadata** while identity stays intact. That preserves per-book breadth in the degraded state, which my proposal would have thrown away. Correct call.

## The most important thing I found

**The severity of the original defect was worse than I reported, and the new docstring's numbers are exactly right.** I verified them independently against the full 650,986-row corpus:

- `Num_hadith` has **37,239 distinct values** across 650,986 rows.
- `hadith_node_id("sanadset:sanadset:7")` and `hadith_node_id("sanadset:7")` both return `hdt:sanadset:7` — `_collapse_double_corpus` folds the doubled segment away.

So the `books.csv`-absent path did not merely emit 16M warnings, as I characterized it last round. It would have MERGEd **613,747 hadiths (94.3% of the corpus)** into 37,239 namesakes at load. A 48 KB metadata file failing to download would have silently destroyed the corpus. The `src/parse/sanadset.py:445-455` docstring states this, and every figure in it checks out.

## Answers to what you asked

**1. What each migrated test now proves — and one correction to the brief.** The "per-book Collection assertion passing through the CSV-stem fallback" is in **`test_sanadset_lightup.py`, not the Path-B test**. It asserted `len(collections) == 1`; the fake `hadith_id,book_id,hadith` header carried no `Book` column, so every row fell back to the CSV stem (`bukhari.csv` → `"bukhari"`) and collapsed into one collection. The assertion was right about the number and wrong about the reason. It now asserts `len(collections) == _EXPECTED_COLLECTIONS == 2` through the real name-digest join, and Alejandra's inline comment documents precisely this. That is a genuine new property: per-book breadth, end-to-end, in a live graph.

`test_real_data_flow.py::test_parse_sample_sanadset_data` gains less. Its assertions are unchanged (outputs present, `num_rows == 3`); only the fixture header moved to the real one. What it newly proves is that the parser accepts the production header at all — non-trivial now that a missing `Book` column raises. It does not pass for a wrong reason; it simply under-asserts (nothing about `source_id`, `collection_name`, or `hadith_number`). Not blocking.

`test_pathb_integration_verify.py` is the one with a residual weakness, and I audited it the way you asked. Its B1 check is `_assert_b1_orphan_free` (`:161-175`), which asserts `emitted` is non-empty and `needed ⊆ emitted`. Both sides are derived from the *same* `collection_name` values the parser wrote, so the subset relation holds by construction no matter what the join key is. I proved this: running the `0766147` parser against the old fixture yields a single `sanadset:hadiths` collection, and the coverage assertion **still passes**. It would pass today with one collection too. It does guard one real thing — that the collections parquet exists and its ids agree with what `load_edges` rebuilds, which is the original ADR-003 orphan defect — but it is blind to breadth. Migrating the fixture was necessary and correct; the assertion behind it is still weaker than the module docstring implies. Filed as the tech-debt line above rather than blocking, because breadth *is* asserted at unit level in `test_name_keyed_join_one_collection_per_distinct_book` (5 books → 5 collections) and `TestBooksCsvAbsentIdentity.test_collection_key_is_the_book_not_the_csv_stem`. The suite as a whole catches a regression here; the Path-B test just isn't the one that would.

**2. `TestBooksCsvAbsentIdentity` is non-vacuous. I confirmed it from the fixture, then mutation-tested it.** The fixture (`tests/test_parse/test_sanadset_parser.py`) puts `Num_hadith` 1 and 2 in البخاري and 1 and 2 again in مسلم, in a file named literally `sanadset.csv` with no `books.csv`. Both properties are load-bearing and both are needed. I re-applied the `edad7d1` defect (re-gating the join key on `books.csv` presence) to the head parser and ran that fixture:

```
source_ids: ['sanadset:sanadset:1', 'sanadset:sanadset:2',
             'sanadset:sanadset:1', 'sanadset:sanadset:2']
rows: 4   distinct: 2
```

`test_every_hadith_keeps_a_distinct_source_id` fails (2 ≠ 4) and `test_no_source_id_is_double_prefixed` fails. With globally-unique ordinals the first test would pass under the mutant, and with any filename other than `sanadset.csv` the second would — so the class docstring is correct that both are required. It records exactly that, naming the 37,239-distinct-ordinals fact and the "any fixture named otherwise is structurally incapable of reproducing the collapse" rationale, which is what stops a future cleanup from simplifying it away. This is the strongest test in the PR.

**3. `test_no_book_column_raises` (`:592-605`) asserts the specific exception and message** — `pytest.raises(ValueError, match="no 'Book' column")`. A different `ValueError` raised elsewhere in the parse would not satisfy it. It is paired with `test_present_but_unmappable_books_raises`, so absent-column (raises) and present-but-unmappable-`books.csv` (raises) are distinguished from absent-`books.csv` (degrades cleanly).

**4. `tests/conftest.py` blast radius is nil — but for an unwelcome reason.** Exactly one fixture changed, `sample_sanadset_csv` (single hunk at `:281`), migrated to the real header and to `sanadset.csv`. No other fixture was touched. However, **no test anywhere requests `sample_sanadset_csv`** — I grepped the whole suite at head and on main; the only near-match is a local `_create_sample_sanadset_csv` helper in `test_real_data_flow.py:61`, a different function. So nothing unrelated silently changed what it exercises. The fixture is dead code that is now correct dead code. Worth deleting or wiring up in a follow-up; not this PR's problem.

**5. `ontology/structural/` was regenerated, not hand-edited.** All 13 `src/parse/sanadset.py` nodes and all 45 `tests/test_parse/test_sanadset_parser.py` nodes in `code-graph.json` land exactly on their `def`/`class` line at head, zero mismatches. `TestBooksCsvAbsentIdentity` and the re-signatured `_resolve_collection_name` are both indexed. That is generator output.

**6. `src/parse/identity.py` and `src/resolve/**` are untouched.** Confirmed against the PR file list; da#355 and da#356 stay clean.

**On the check-set.** Your 12-vs-7 catch is the right instinct and matches `feedback_statuscheckrollup_ci_clean` — an EMPTY or short rollup is not evidence of green. I re-pulled it myself: 12 terminal SUCCESS, including the five that only appeared once the memory and markdown edits landed.

## One residual, non-blocking

`_resolve_collection_name` still returns `default_name` when a row's `Book` cell is blank (`src/parse/sanadset.py:459-461`). For `sanadset.csv` that is the corpus name, so such a row would still mint `sanadset:sanadset:<N>`. The blast radius is now one row rather than 650,986, the real corpus has zero such rows (I re-verified: 0/650,986 blank `Book`), a missing `Book` *column* raises, and the docstring calls the case out explicitly. Not worth holding the PR. If a future edition ships a blank cell, the row-level guard is a two-line addition.

Good work — you fixed the root cause rather than the symptom I pointed at, and the fail-loud guard earning its keep by immediately exposing two more fake fixtures is exactly what a guard is for.

Issue #361, opened 2026-07-09
resolve: missing required input silently yields an empty success across 11 sites (same class as #309)


parametrization:
## Summary

Several `src/resolve/` stages return a success-looking empty result when a **required input artifact is missing**. A missing input means an upstream stage did not run (or ran against the wrong directory) — an upstream defect. It is not the same thing as an input that ran and legitimately produced nothing, but the two are currently indistinguishable to every caller and to the process exit status.

This is the same defect class as #309 (dedup's missing-dependency skip), one layer up: instead of a missing *dependency* it is a missing *input*.

## Sites (complete sweep of `src/resolve/`, 2026-07-09, at 2cf5968)

Missing required input → empty/None return:

| File:line | Condition | Returns |
|---|---|---|
| `src/resolve/dedup.py:139` | no `hadiths_*.parquet` in staging | `[], [], []` → empty `parallel_links.parquet` |
| `src/resolve/ner.py:71` | `narrator_mentions_{corpus}.parquet` absent | `[]` |
| `src/resolve/ner.py:124` | no `hadiths_{corpus}*.parquet` | `[]` |
| `src/resolve/disambiguate.py:487` | `narrator_mentions_resolved.parquet` absent | `[]` |
| `src/resolve/disambiguate.py:1048` | no candidates loaded | `[]` |
| `src/resolve/disambiguate.py:1056` | `_count_mentions() == 0` | `[]` |
| `src/resolve/bio_promote.py:143` | no `narrators_bio_*.parquet` | `None` |
| `src/resolve/narrator_split.py:606` | `narrators_canonical.parquet` absent | `None` |
| `src/resolve/narrator_split.py:609` | `narrator_mentions_resolved.parquet` absent | `None` |
| `src/resolve/date_reconcile.py:483` | `narrators_canonical.parquet` absent | `None` |
| `src/resolve/tabaqa_dates.py:200` | `narrators_canonical.parquet` absent | `None` |

Each logs at `warning` and returns. None raises; none causes a non-zero exit.

`date_reconcile.py` and `tabaqa_dates.py` document the empty return as an intentional no-op ("is a no-op until disambiguate/bio_promote have produced the table"), so those two are plausibly correct by design and should be confirmed rather than changed blindly.

## Not in this class

- `src/resolve/dedup.py` "no English matn on any row" — files exist and were read; a genuinely empty input for an English-matn semantic detector. Split out as its own branch in the #309 fix.
[...]

Issue #360, opened 2026-07-09
resolve: run_all swallows every stage exception and still reports resolve_pipeline_complete (exit 0)


parametrization:
## Summary

`src/resolve/__init__.py: run_all` wraps **every** stage in `except Exception: logger.error("resolve_step_failed", ...)` and then continues to the next stage. It always reaches `logger.info("resolve_pipeline_complete", ...)` and returns `results`. `src/cli.py: _cmd_resolve` prints `"Resolution complete."` and exits **0**.

A stage that crashes is therefore indistinguishable, at the process level, from a stage that ran and found nothing.

## Evidence

Observed directly at `origin/main` (2cf5968) by patching `src.resolve.dedup.run` to raise `RuntimeError("BOOM-dep-missing")` and calling `run_all(..., from_step="dedup")`:

- `resolve_step_failed step=dedup` (error log)
- `resolve_step_failed step=parallels` (error log — a *second*, unrelated crash)
- `resolve_pipeline_complete ... parallel_links=0` (info log)
- `run_all` returned normally

Two stages crashed and the pipeline reported completion with zero parallel links.

This is the reason da#309's fix raises a `BaseException` subclass (`MissingDependencyError`) rather than an `Exception` — an `Exception` would have been swallowed here and the guard would have been inert. `StopAfterReached` (`src/resolve/_checkpoint.py:207`) already subclasses `BaseException` for the same documented reason. Two workarounds for the same orchestrator behaviour is a signal the orchestrator is wrong, not the exceptions.

## Sibling: the staging pre-flight

`src/resolve/__init__.py:365-372`:

```python
if not staging_dir.exists() or not _has_staging_parquets(staging_dir):
    logger.warning("resolve_preflight_failed", ...)
    logger.warning("resolution_skipped", reason="no staging Parquet files found")
    return results          # empty dict; caller exits 0
```

An unrunnable resolve exits 0 with an empty `results`.

## Why this is not folded into da#309

Changing `run_all` to fail on any stage error is a behaviour change across all nine stages and every caller (CLI, integration tests, the VM pipeline runner). It needs a decision on: which stages are genuinely tolerable-to-fail (`bio_promote`? `muhaddithat_links`?), whether a partial `results` should still be returned, and what exit status the CLI should use. That is a design decision, not a bug fix.

## Proposed direction (for discussion)

- `run_all` collects per-stage failures and re-raises (or returns a failure-carrying result) at the end, rather than discarding them.
[...]

PR #362, opened 2026-07-09
fix(resolve): dedup fails loud when an enabled stage is missing a declared dep

### LINKED ISSUE #309: resolve: dedup silently skipped — sentence-transformers declared but missing from venv (fail-loud on missing declared dep)
## Problem

Run 4b's dedup stage was **silently skipped**:

```
[error] dedup_missing_deps  msg='sentence-transformers or numpy not installed -- skipping dedup'
```

numpy IS installed (2.4.4). `sentence-transformers` is **declared in `pyproject.toml`** (`"sentence-transformers>=2.5"`, a main dependency, present in `uv.lock`) but is **not installed in the resolve host's `.venv`** — the venv drifted from the lockfile. NER didn't surface it because NER uses CAMeLBERT via `transformers`; dedup is the only importer of `sentence_transformers`.

## Impact

Hadith semantic dedup (near-duplicate matn across collections, da#245 memmap encode) does not run → duplicate hadith nodes reach the graph. The stage degrades gracefully (pipeline continues, parallels still runs), so the miss is invisible unless you grep the log — a **silent-skip** anti-pattern.

## Fix

1. Operational (immediate, before the next resolve run): `uv sync` on the resolve host so the venv matches `uv.lock`.
2. Guardrail: dedup should **fail loud** (or the pipeline should refuse to report success) when a *declared* dependency for an enabled stage is missing, instead of `error`-logging and continuing. A missing declared dep is an environment defect, not a valid skip. Distinguish from a genuinely optional stage.
3. CI/env: ensure the resolve/runbook environment provisioning runs `uv sync` (no partial installs) so ML deps are present wherever resolve runs.

## Acceptance

- `uv sync` on the resolve host installs sentence-transformers; `resolve` dedup stage runs.
- dedup raises (or pipeline exits non-zero / marks incomplete) on a missing declared dep, rather than silent-skip.
- Note in the runbook: verify `sentence-transformers` importable before a reload-grade resolve run.

Refs: da#245 (dedup memmap harden), #723 reload.

parametrization:
Closes #309

## The defect

`src/resolve/dedup.py` caught `ImportError` on its declared ML dependencies and returned `_write_empty_output(staging_dir)` — a schema-valid, **zero-row** `parallel_links.parquet`. `run_all` then logged `resolve_pipeline_complete` and the CLI printed `Resolution complete.` and exited **0**. Every downstream consumer accepts the empty table as a genuine "no parallels found".

Impact is **zero `PARALLEL` edges** (production carries ~4.49M) presented as a negative result — not duplicate hadith nodes. `dedup` writes only `parallel_links.parquet` (lines 779, 812); node-level hadith dedup is `canonical_matn_identity` in `load_nodes` (da#220), a different stage. The org ontology (`ontology/repos/ingestion.yaml`) independently records `dedup: output: parallel_links.parquet`.

## The two legs, verified

**Operational leg — the correction needs a correction.** `sentence-transformers` 5.3.0 imports in the *main checkout's* `.venv`, but `ml` is a **non-default `[dependency-groups]` entry** and there is no `[tool.uv] default-groups`. A bare `uv sync --frozen` — what CI runs, and what a fresh worktree gets — installs **neither `sentence-transformers` nor `faiss`**:

```
$ cd .claude/worktrees/0309-dedup-fail-loud && uv run python -c "import sentence_transformers"
ModuleNotFoundError: No module named 'sentence_transformers'
```

So the silent skip is not merely latent on an unprovisioned box — it is the **current behaviour of every fresh environment**, including CI. If the authorized full re-run is provisioned with a bare `uv sync`, dedup emits zero parallel links and reports success.

**Guardrail leg.** Red against pristine `origin/main` (2cf5968), embedder present in the venv and removed only via `patch.dict(sys.modules, {"sentence_transformers": None})`:

```
FAILED tests/test_resolve/test_da309_red_probe.py::test_run_dedup_raises_when_embedder_missing
  Failed: DID NOT RAISE <class 'BaseException'>

FAILED tests/test_resolve/test_da309_red_probe.py::test_no_empty_output_is_written_when_dep_missing
  AssertionError: SILENT SKIP: run_dedup wrote parallel_links.parquet with 0 rows
```

with the captured log showing the mechanism:

```
[error] dedup_missing_deps  msg='sentence-transformers or numpy not installed -- skipping dedup'
[info ] dedup_empty_output  path=.../parallel_links.parquet
```

## A raise alone would have been inert

`run_all` wraps **every** stage in `except Exception: logger.error("resolve_step_failed", ...)` and continues. Patching `dedup.run` to raise `RuntimeError` at `origin/main`:

```
[error] resolve_step_failed        step=dedup
[error] resolve_step_failed        step=parallels
[info ] resolve_pipeline_complete  parallel_links=0
```

Two stages crashed; `run_all` returned normally and the CLI would exit 0. So `MissingDependencyError` subclasses **`BaseException`**, following the precedent `StopAfterReached` (`_checkpoint.py:207`) already documents for exactly this reason. `test_missing_dependency_error_is_not_an_exception_subclass` pins that contract, and `test_missing_dependency_error_propagates_through_run_all` proves the error survives the orchestrator.

The orchestrator's blanket swallow is itself the deeper defect — filed as #360, since fixing it changes behaviour across all nine stages.

## What changed

- **`src/resolve/_deps.py`** (new) — `MissingDependencyError` + `require_dependencies` / `missing_dependencies`. The message names the stage, the missing deps, and the remediation (`uv sync --group ml`).
- **`dedup`** checks its declared deps at **stage entry**, before reading input: an enabled stage with an absent dep is a defect regardless of whether its input happens to be empty. Both `except ImportError` sites (embedder at 657, faiss at 698) are gone.
- **`DEDUP_REQUIRE_ML`** (default `true`) makes the distinction explicit *in code*: `true` → raise; `false` → an auditable opt-in to the deterministic no-model fallback (`parallels.py`), logged as `dedup_skipped_degraded`, never as a negative result. The test suite sets it `false` once, in `tests/conftest.py`, because CI genuinely cannot embed — and `test_dedup_requires_ml_by_default` pins the production default so that opt-out can never quietly become the shipped behaviour.
- **CLI** maps the error to `EXIT_MISSING_DEPENDENCY` (4), distinct from `EXIT_STOPPED_AT_LIMIT` (3).
- **`_load_hadith_texts`'s two collapsed empty cases are now separate branches**: no input files at all (upstream defect, `dedup_no_hadith_files`) vs files present with no English matn (a true negative, `dedup_no_texts`).

## On `logger.warning("dedup_no_texts")` — reasoned, not lumped in

The brief called this a legitimately-empty input. That is **half right**. `_load_hadith_texts` returned `[], [], []` through **two** distinct paths, and `run_dedup` could not tell them apart:

- **Case A** — no `hadiths_*.parquet` files at all (`dedup.py:139`). Parse never ran, or `staging_dir` is wrong. An upstream defect, structurally identical at the call site to a true negative.
- **Case B** — files present, every row's `matn_en` null/empty. Genuinely empty input for an English-matn detector.

Only Case B is a true negative. They are now separate branches with separate log events. Case A is left non-fatal (unit tests call `run_dedup` on empty `tmp_path` dirs), and making it fatal is #361 — a behaviour change, not a bug fix.

## Sibling sweep — complete, `src/resolve/` only

Same defect, **fixed here**:

- `dedup.py:657-665` — `except ImportError` → `_write_empty_output` (embedder/numpy)
- `dedup.py:698-705` — `except ImportError` → `_write_empty_output` (faiss). **Not named in the brief**; the same defect, a second site.

Same class, **design decision → #360**:

- `__init__.py:384,410,441,474,501,525,550,580,601,624` — ten `except Exception` → `logger.error("resolve_step_failed")` → continue
- `__init__.py:365-372` — staging pre-flight returns an empty `results` and exits 0

Same class, **design decision → #361** (missing required input → empty success): `dedup.py:139`, `ner.py:71,124`, `disambiguate.py:487,1048,1056`, `bio_promote.py:143`, `narrator_split.py:606,609`, `date_reconcile.py:483`, `tabaqa_dates.py:200`.

**Ruled out, with reasons** (not silently dropped):

- `parallels.py:353` — `log = logger.warning if indexed and not records else logger.info` already discriminates "had candidates, found none" from "had none". This is the pattern the others should adopt.
- `dedup.py:220` — `except Exception: pass` around `torch.set_num_threads`. A best-effort thread-limit tweak with no output effect.
- `__init__.py:160,177,252,261,275` — five `except Exception` in `_collect_*_metrics` / `_read_parallel_links`. These degrade *reporting*, not pipeline output. Worth revisiting, but a metrics read failing is not a stage producing a false negative.
- `fuzzy_cluster.py:644` — `if len(flat_keys) < 2: return []`. An arithmetic guard on a pure function.
- `disambiguate.py` / `fuzzy_cluster.py` are Ivana's (da#356); listed, not touched.

Nothing was capped or truncated: the sweep is every `except ImportError`, every `logger.error`, every `_write_empty_*`, every `except Exception`, and every `return []` / `return None` in `src/resolve/`.

## Test-side defects found while writing the guard

The repo shipped an inert guard against this very class of bug:

- **`test_output_pairs_are_canonically_ordered`** wrote `hadith_test.parquet` (singular) while the stage globs `hadiths_*.parquet` (plural). It matched **zero** files, produced an empty table, and its ordering assertion sat behind `if table.num_rows > 0:` — so it **never executed**. Verified: `Path.glob("**/hadiths_*.parquet")` does not match `hadith_test.parquet`. Fixed the filename, marked `@ml`, and removed the conditional so the assertion actually runs.
- **`test_cross_sect_flagging`** was conditionally inert the same way; now asserts `num_rows >= 1` first.
- **`TestGracefulFallback.test_empty_output_on_no_texts`** encoded the defect as expected behaviour ("No hadith files -> empty output regardless of ML availability") and, despite its name, exercised Case A. Replaced by `TestEmptyInput` covering Cases A and B separately.

## Verification

Both dependency environments, full tree, no `-m` filter:

- `uv run pytest tests/ -q` (fresh venv, no `ml` — **what CI runs**): `1841 passed, 10 skipped`
- same tree driven by the `ml`-provisioned venv (so the `@ml` tests I changed actually execute): `1847 passed, 4 skipped`

`uvx ruff format --check` clean, `ruff check` clean, `mypy src/` clean (98 files). Commit-stage and `--hook-stage pre-push` pre-commit hooks all pass, including `structural-ontology-staleness` (index regenerated) and `gitleaks`/`actionlint`/`cspell`. `pre_commit_ci_sync.py` reports `OK: pre-commit config mirrors all CI-enforced checks`.

## Reviewer notes

The one judgement call worth scrutiny is `tests/conftest.py` setting `DEDUP_REQUIRE_ML=false` suite-wide. CI cannot import the embedder, so *some* opt-out is unavoidable; the mitigation is that the production default is pinned by its own test and the raise is driven explicitly with `require_ml=True`, bypassing the fixture. If you would rather see `DEDUP_REQUIRE_ML: "false"` declared in `ci.yml`'s test job instead, say so — I chose `conftest.py` so the local gate and CI cannot drift.


parametrization:
Requestor: Tarek Mansour
Verdict: Approved
TechDebt: Both default-pinning tests reach `Settings()`, which reads `env_file=".env"` — `test_dedup_requires_ml_by_default` directly, `test_default_settings_make_run_dedup_raise` via `get_settings()` — so a developer `.env` copied from the new `.env.example` (which sets `DEDUP_REQUIRE_ML=true`) masks a flipped production default locally, blinding both; use `Settings(_env_file=None)` and neutralize `.env` before `get_settings()` in the second. Also `_read_parallel_links`'s `except Exception -> None` silently drops the semantic half at composition time — same defect class as da#309, route to da#360.

Reviewed at `f8f24eb5f5e54eab7eb6d2129c4d33a932678ce4`; **re-verified at `cf5e5de0b5bacee3a59cb006762c9a73e481201d`** after two pushes landed four seconds and four minutes after this comment was first posted. The verdict carries. This comment has been edited in place — the debt line above now reflects the later head; see the amendment at the bottom for what changed and why. Lens: provisioning and operational blast radius.

## 1. Provisioning hazard — confirmed independently

`pyproject.toml` declares `ml = [sentence-transformers, faiss-cpu, transformers, torch, camel-tools]` under `[dependency-groups]`, and there is **no `[tool.uv]` table at all** — so no `default-groups`. All four `ci.yml` jobs run bare `uv sync --frozen`, so CI and every fresh worktree lack the ML deps. That is exactly the environment in which the pre-PR code wrote a zero-row `parallel_links.parquet` and reported success.

`Dockerfile.load` builds with `uv sync --frozen --no-dev --no-install-project` — no `--group ml` — so **that image also lacks the ML deps**. It does not, however, reach `dedup`: its `CMD` is `load --nodes-only`, `scripts/load_staging.sh` overrides only to `load`, and `_cmd_load` never imports `src.resolve`. The `src.resolve` import in `_cmd_resolve` is function-local, so the image doesn't even import the guard. Net: the loader image is not a blast-radius surface, but if anyone ever runs `python -m src.cli resolve` (or `pipeline`) inside it, this PR turns a silent zero into exit 4. That's the right direction.

## 2. Does the fix fire where it matters

Traced all three states at `dedup.py:687-707`:

- **(a) deps present** — guard is a no-op; `missing_dependencies` returns `[]`; stage proceeds. The 7.5-hour `resolve` re-run from the main checkout's `.venv` (which does carry the ML deps) is unaffected. The guard does eagerly `importlib.import_module("torch")` at stage entry, costing a few seconds and ~0.5 GB — but the stage imports torch moments later anyway, so it's not new cost.
- **(b) deps absent + `require_ml=True`** — raises before any input is read and before any checkpoint or output file is touched. `run_all`'s `except Exception` at line 583 does not catch it; `_cmd_resolve` catches it and exits 4. `_cmd_pipeline` calls `_cmd_resolve` directly, so `SystemExit(4)` propagates and **stage 4 `load` never runs**. That is the operationally important property: a dep-less box can no longer load a graph missing 4.49M edges.
- **(c) deps absent + `require_ml=False`** — writes the empty parquet, logs `dedup_skipped_degraded` with a message that explicitly says "This is NOT a 'no parallels found' result."

Case (c) is still a zero, but it is not *silent* and it is not *lossy at the artifact*: `_compose_parallel_links` unions the two detectors, so `parallels.py`'s deterministic lexical links survive an empty semantic table. It is reachable in production only by an operator explicitly setting `DEDUP_REQUIRE_ML=false`. That is the one legitimate skip, correctly separated from the environment defect. Defensible.

## 3. `BaseException` — audited

The cited precedent is real, not asserted: `_checkpoint.py:207` has `class StopAfterReached(BaseException)` with the same documented rationale, and `_cmd_resolve` already catches it alongside the new error. So this is consistent with a pattern the repo already ships, not a novel escape hatch.

Cleanup-bypass check: there is **no `finally:` anywhere** in `src/resolve/__init__.py`, `src/resolve/_checkpoint.py`, or `src/resolve/dedup.py`. The raise happens at stage entry — before `_hadith_files`, before the encode memmap, before `CheckpointController` is constructed, before any `ProcessPoolExecutor` is opened. So no half-written checkpoint, no leaked pool, no partial parquet. Confirmed by `test_no_empty_output_is_written_when_dep_missing`, which asserts the output file does not exist at all.

One latent hazard, not this PR's problem: `src/graph/validate.py:74` catches `BaseException` to marshal it across a thread boundary. Nothing in the resolve path crosses it today, but any future `BaseException` guard placed inside a graph-validation worker would be captured there.

## 4. Is the conftest mitigation inert — the question I ran an experiment on

`tests/conftest.py` sets `DEDUP_REQUIRE_ML=false` suite-wide via an autouse fixture. `test_dedup_requires_ml_by_default` requests the same function-scoped `monkeypatch` instance and calls `delenv("DEDUP_REQUIRE_ML", raising=False)`, so it does undo the autouse `setenv`. Fixture ordering is fine.

I did not take that on inspection. I copied the worktree tree to a scratch dir, ran the file, then mutated `src/config.py:71` from `dedup_require_ml: bool = True` to `False` and re-ran:

- baseline: `10 passed`
- mutated default: `1 failed, 9 passed` — and the one failure is `test_dedup_requires_ml_by_default`, with `AssertionError: assert False is True`

**So yes — the guard goes red on a default flip, and it is the only test that does.** The nine behavioural tests keep passing because they pass `require_ml=True` explicitly, which is precisely the property that makes this a real mutation detector rather than a restatement of the fixture.

The one flaw, and it is why the TechDebt line above exists: `Settings` has `model_config = SettingsConfigDict(env_file=".env", ...)`. `delenv` removes the process env var, so pydantic falls through to `.env`. With the default mutated to `False` **and** a `.env` containing `DEDUP_REQUIRE_ML=true`, the same test goes back to `1 passed` — I reproduced that too. Neither the main checkout nor the worktree currently has a `.env`, and CI's `uv sync --frozen` never creates one, so **the guard is live where it must be**. But `.env.example` now instructs every developer to write exactly the line that blinds it. `Settings(_env_file=None)` closes it in one character-ish. Not blocking; it fails safe (the masking direction requires someone to have *already* flipped the default).

## 5. Sweep classifications — spot-checked four of the ruled-out sites

- `parallels.py:353` — `log = logger.warning if indexed and not records else logger.info`. Correctly ruled out; it is the only site in the family that already discriminates "had candidates, found none" from "had no candidates." Agreed it's the pattern the others should adopt.
- `dedup.py:220` (head 234) — `except Exception: pass` wrapping `torch.set_num_threads`. Correctly ruled out: `_set_thread_limit` has already written the four `*_NUM_THREADS` env vars before the try, so the failure mode is a suboptimal thread count, never an output difference.
- `fuzzy_cluster.py:644` — `if len(flat_keys) < 2: return []` inside the per-block `cdist` worker. Correctly ruled out: a pure function, and a block with one key genuinely has no pairs.
- `__init__.py:160,177,252,261,275` — four of these five (`_collect_dedup_metrics`, `_collect_narrator_metrics` ×2, `_collect_ner_metrics`) are reporting-only, correctly ruled out. **The fifth is not.** `_read_parallel_links` (head line 180) returns `None` on an unreadable parquet, and `run_all` feeds that `None` straight into `_compose_parallel_links`, where a `None` semantic table is skipped and the composed artifact silently loses the entire semantic half. It logs a warning, the deterministic links survive, and it needs `pq.read_table` to fail on a file `dedup` wrote seconds earlier — so it is narrow and low-probability. But the classification "degrades *reporting*, not pipeline output" is wrong for that one line, and the failure shape is the same one this PR exists to kill. Not a blocker; captured in the debt line for da#360.

Everything else in the sweep I sampled routes sensibly. Two additions I'd have made are already there: `dedup.py:698-705` (the faiss `except ImportError`, a second instance the brief didn't name) was found and folded in, and Case A / Case B at `dedup.py:139` are now structurally distinct rather than sharing one silent return.

## Summary

The fix fires at the right place (stage entry, pre-side-effect), survives the orchestrator by a mechanism the repo already uses, halts the pipeline before `load`, and its anti-inertness guard is genuinely load-bearing — I flipped the default and watched it go red. The conftest opt-out does not reach it in CI. Approving.

---

## Amendment — re-verified at `cf5e5de0`

The original text above was written against `f8f24eb5`. Two commits landed after it (`268e00e9`, `cf5e5de0`), touching the very test file I mutation-tested. An approval does not transfer across a head move that modifies the reviewed artifact, so I re-ran the delta. Everything above stands; three things are now more precisely true.

**A new detector, and a genuinely independent one.** `cf5e5de0` adds `test_default_settings_make_run_dedup_raise`. I applied the same standard rather than reading it: restored the default to `True` (11 passed), mutated `src/config.py:71` to `False`, re-ran. Two tests fail, **for different reasons** — `assert False is True` for the value test, `DID NOT RAISE <class 'src.resolve._deps.MissingDependencyError'>` for the new one. The new test calls `run_dedup(staging, threshold=0.70)` with no `require_ml` argument, which is the shape `run_all` uses, so it drives the `require_ml is None -> get_settings().dedup_require_ml` branch at `dedup.py:687-688` and asserts the output file was never created. It catches a defect the value test cannot see: a default left at `True` while the `None`-resolution branch is broken, so the flag never reaches the guard. Two detectors, not two assertions.

**The `.env` masking now covers both of them.** The new test never constructs `Settings()` itself, but `run_dedup` with no `require_ml` calls `get_settings()`, which does. Same door, one level of indirection. Reproduced: with the default mutated to `False` **and** a `.env` carrying `DEDUP_REQUIRE_ML=true`, the file returns to 11 passed — both guards blind at once. `_env_file` appears nowhere in the file at `cf5e5de0`. This is why the debt line was widened; it is still not blocking, and still fails safe, since the masking only bites someone who has already flipped the default and CI never materializes a `.env`.

One note on the new comment block's claim that the `delenv` is "verified load-bearing." It is — but that control demonstrates the *conftest* env var reaches `Settings()`. It says nothing about `.env`, which is precisely the path that survives `delenv`. Accurate and incomplete, not wrong; the two findings describe different doors into the same room.

**The `.pre-commit-config.yaml` change widens the watched paths, and closes a real parity hole.** The `structural-ontology` hook's `files:` filter went from `^(src/.*\.py|scripts/structural_ontology\.py|ontology/structural/.*)$` to `^((src|tests|scripts|queries|\.claude)/.*\.(py|cypher)|ontology/structural/.*)$`. I regenerated the counts from the committed index rather than accepting them: `llms.txt` at `cf5e5de0` carries 248 sections — 127 `tests/`, 98 `src/`, 10 `scripts/`, 7 `queries/`, 6 `.claude/`; 241 `.py` and 7 `.cypher`. Measured against those paths, the **old** filter misses **149 of 248**; the **new** filter misses **0**. So the local gate could not fire on a `tests/`-only commit — exactly the shape of the commit that exposed it — while CI's whole-tree check runs regardless. Local green, CI red, which is the divergence noorinalabs-main#684 exists to forbid. The change moves in the direction that rule demands, and I would have asked for it had it not shipped.

Net: the two later commits strengthen the fix and close a parity hole. Approving at `cf5e5de0`.

---
Requestor: Tarek Mansour
RequestOrReplied: Approved
TechDebt: Both default-pinning tests reach `Settings()`, which reads `env_file=".env"` — `test_dedup_requires_ml_by_default` directly, `test_default_settings_make_run_dedup_raise` via `get_settings()` — so a developer `.env` copied from the new `.env.example` (which sets `DEDUP_REQUIRE_ML=true`) masks a flipped production default locally, blinding both; use `Settings(_env_file=None)` and neutralize `.env` before `get_settings()` in the second. Also `_read_parallel_links`'s `except Exception -> None` silently drops the semantic half at composition time — same defect class as da#309, route to da#360.

parametrization:
Requestor: Oyunbileg Batbayar
Verdict: Approved
TechDebt: `require_dependencies()` is exported in `_deps.__all__` but has no caller and no test (dedup raises `MissingDependencyError` directly) — either route dedup through it or drop it; `missing_dependencies` catches only `ImportError`, so a dependency that is installed-but-broken (native lib raising `OSError`/`RuntimeError` at import) escapes the guard and is then swallowed by `run_all`'s `except Exception`, which is the very silent-success this PR closes; and the composed `parallel_links.parquet` carries no detector provenance, so the `DEDUP_REQUIRE_ML=false` artifact is byte-shaped identically to a true negative (a `detector` column or a sidecar manifest recording semantic/deterministic row counts would close it).

Reviewed at `cf5e5de0b5bacee3a59cb006762c9a73e481201d` via `gh api contents?ref=…`; no checkout. Check-runs at that sha: 12/12 success, rollup non-empty. Second approval under the charter, after Tarek's at the same sha. I did not re-run his sentinel mutations or re-audit the `BaseException` choice.

## 1. Nothing on the resolve call path can swallow it, and `pipeline` aborts before `load`

Confirmed, and the answer is stronger than "no `finally:` masks it" — there is no `finally:` on the dedup path **at all**. `src/resolve/dedup.py` contains zero `finally:` blocks. The only two in the package are `disambiguate.py:567` and `fuzzy_cluster.py:1164`, both a bare `writer.close()` with no `return` and no `raise` inside the block, so neither can suppress a propagating exception even in principle — and neither sits on the dedup call path. `dedup.py`'s single handler is the `except Exception` at :234, which is `torch.set_num_threads` best-effort inside the thread-pinning helper; it catches `Exception`, not `BaseException`, and in any case runs after the guard.

All fifteen handlers in `src/resolve/__init__.py` are `except Exception`, including the dedup step's own at :604. `MissingDependencyError` derives from `BaseException` and is not an `Exception` subclass (pinned by `test_missing_dependency_error_is_not_an_exception_subclass`), so none of them can catch it. There is no bare `except:` and no `except BaseException` anywhere in `src/resolve/**`, and no `contextlib.suppress`. The `with ProcessPoolExecutor(...)` at `dedup.py:355` is downstream of the entry guard, and its `__exit__` calls `shutdown(wait=True)` without suppressing, so it is not an escape either.

`_cmd_pipeline` genuinely aborts. It calls `_cmd_resolve()` at Stage 3 with no `try` around it; `_cmd_resolve` catches `MissingDependencyError` and calls `sys.exit(EXIT_MISSING_DEPENDENCY)`, raising `SystemExit`. `_cmd_pipeline` has no handler, `main()` has no handler, and cli.py's only other handlers (`:31`, `:65`, `:74`) are `except Exception` inside `_cmd_info`/`_cmd_acquire`/`_cmd_parse`. `SystemExit` is itself a `BaseException`, so nothing intercepts it. `_cmd_load()` on the next line is never reached, and the process exits 4. `test_cli_maps_missing_dependency_to_distinct_exit_code` pins the code and that it is non-zero.

There is one thing that could have made the whole guard inert, so I checked it. `tests/conftest.py` now sets `DEDUP_REQUIRE_ML=false` autouse for the entire suite. If the fail-loud tests inherited it, they would be testing the degraded path while claiming to test the guard. They do not — `test_dedup_requires_ml_by_default` and `test_default_settings_make_run_dedup_raise` both `monkeypatch.delenv("DEDUP_REQUIRE_ML", raising=False)` first and assert against `Settings()` directly, and the class-scoped raise tests pass `require_ml=True` explicitly. The fixture docstring flags this hazard itself and the comment at line 62 records that the `delenv` was verified load-bearing. That is the right instinct and it is correctly executed.

## 2. No — the composed artifact is not distinguishable from a true negative

This is the real finding. With `DEDUP_REQUIRE_ML=false` and the ML group absent, `run_dedup` takes the degraded branch and returns `_write_empty_output(staging_dir)`, so the file exists with zero rows. `run_all` then reads it back at `__init__.py:600`, giving a **0-row table, not `None`**. `detect_parallels` overwrites the shared artifact, `_compose_parallel_links` unions the two, and the union of a 0-row semantic table with an N-row deterministic table is exactly N deterministic rows.

Now take the other run — ML group installed, embedder loads, and the corpus genuinely has no semantic parallels. Semantic table is 0 rows; deterministic is the same N rows; the union is the same N rows. **The two `parallel_links.parquet` files are identical** — same schema, same rows, same counts. `load_edges` consuming that artifact cannot tell an environment defect from a true negative.

What separates them today is a `dedup_skipped_degraded` warning and `parallel_links_composed` logging `semantic=0`. That is a distinction *by log level* — which is precisely what `_deps.py`'s own module docstring says this change replaces ("The two conditions are now distinct *in code* rather than by log level"). The escape hatch preserves the original silent-zero shape.

I am not blocking on it, for two reasons that I think are the right ones rather than convenient ones. `dedup_require_ml` defaults to `True`, so reaching this state requires an operator to deliberately set the flag, which is the "explicit, auditable choice" the design names as the one legitimate skip. And the deterministic detector still runs, so the degraded artifact is not empty — the failure is an understated PARALLEL edge count, not a zero graph. So the framing holds: this PR makes one silent zero loud, and leaves a quieter second one behind an opt-in. Worth noting that because the autouse fixture sets the flag false, every other test in the repo composes its artifacts through this exact indistinguishable shape.

## 3. Yes — a `None` silently drops the entire semantic half, and it is reachable by more than a corrupt file

`_read_parallel_links` returns `None` when the file is absent **or** when `pq.read_table` raises (broad `except Exception` → `parallel_links_read_failed` warning → `None`). In `_compose_parallel_links` the `for table in (deterministic, semantic)` loop does `if table is None: continue`. So the semantic half is dropped, a composed table is still written from the deterministic rows, a path is still returned, and `run_all` reports completion. Nothing downstream sees a difference.

The part worth carrying into da#360: the corrupt-file case is not the likely trigger, because `dedup.run` writes `parallel_links.parquet` on every non-raising path, so `path.exists()` is almost always true. The reachable trigger is the crash. `semantic_links` is initialized to `None` at `__init__.py:593` and assigned only inside the `try`. If `dedup.run()` raises any ordinary `Exception` — FAISS OOM, a torch CUDA fault, a parquet write failure — the `except Exception` at :604 logs `resolve_step_failed`, `semantic_links` stays `None`, composition proceeds on the deterministic half alone, and the run reports `resolve_pipeline_complete`. Every dedup failure mode *except* the missing-dependency one still ends in a successful-looking run with the semantic half gone.

That is not a regression here, and I confirmed it: `_read_parallel_links` and `_compose_parallel_links` are untouched at this sha, and the entire `src/resolve/__init__.py` diff is the `_deps` import plus two `__all__` entries. Tarek was right to route it out rather than block. My one ask is on the scope of the ticket, not on this branch — da#360 should be written against the dedup-crash path, not against "unreadable file," or it will be closed by a fix that never covers the trigger that actually fires. Same reasoning applies to the `ImportError`-only detection in the TechDebt line: a broken-but-installed native dep takes that identical crash path.

The rest is clean. The guard sits at stage entry before any input is read, `test_no_empty_output_is_written_when_dep_missing` proves no artifact is written on the raise, the error message names stage, missing modules, and `uv sync --group ml`, and exit code 4 is distinct from `EXIT_STOPPED_AT_LIMIT`. Good work.

---
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: `require_dependencies()` is exported in `_deps.__all__` but has no caller and no test (dedup raises `MissingDependencyError` directly) — either route dedup through it or drop it; `missing_dependencies` catches only `ImportError`, so a dependency that is installed-but-broken (native lib raising `OSError`/`RuntimeError` at import) escapes the guard and is then swallowed by `run_all`'s `except Exception`, which is the very silent-success this PR closes; and the composed `parallel_links.parquet` carries no detector provenance, so the `DEDUP_REQUIRE_ML=false` artifact is byte-shaped identically to a true negative (a `detector` column or a sidecar manifest recording semantic/deterministic row counts would close it).

parametrization:
**HOLD — do not merge.** This PR carries two approvals and was next in the merge queue. It is held on an exit-code collision, not on anything wrong with the code in it.

`src/resolve/_deps.py:39` in this PR declares `EXIT_MISSING_DEPENDENCY = 4`.
`src/graph/__init__.py:31` in open PR#372 (da#354) declares `EXIT_VALIDATION_FINDINGS = 4`.

Neither file exists on `origin/main`, so **each PR's suite is green against a tree in which the other's constant is absent.** Whichever merges second, `4` means two things — and `_cmd_pipeline` runs `resolve` and `load` in a single process, so after both land `isnad-ingest pipeline` exiting `4` means either *"resolve is missing its ML dependency group"* or *"load succeeded, validation has findings."* An environment defect and a clean load with a data finding, indistinguishable.

Found by Oyunbileg Batbayar while reviewing #372 — a collision no CI on either PR can observe.

Tracked as **#384**: one module owns the pipeline exit-code space, imported by `src.resolve`, `src.graph`, `src.cli`, so a second claim on a value is an **import-time collision** rather than a production discovery. Renumbering `4` → `5` on one side buys one merge and re-arms the trap for the third module.

**#384 merges first. Then this PR rebases** — a one-line change: import the constant instead of declaring it.

Nothing here reflects on the work. da#309 is a launch-gate item and its approvals stand; the hold is mine and the sequencing is mine. I will lift it when #384 lands.

— Nadia Khoury

parametrization:
## Merged `main` after PR #363, at `b6acb49`

`main` @ `70b66ad` (da#356 + da#376) rewrote `src/parse/identity.py` and the structural index while this branch sat approved, turning it `CONFLICTING`. Resolved by **merge, not rebase**, so the existing approvals stand.

`EXIT_MISSING_DEPENDENCY = 4` is untouched — `src/resolve/_deps.py` has no diff across either new commit. `src/parse/identity.py` is byte-identical to `origin/main`; this branch never touched it. da#384 will make the constant an import; that rebase stays import-only.

Only `ontology/structural/llms.txt` actually conflicted. `code-graph.json` **auto-merged textually**, which is not a valid resolution for a generated graph, so both files were **regenerated** with `ontology_gen` rather than hand-resolved, and I confirmed the result is a genuine union — this branch's `MissingDependencyError(BaseException)` and `run_dedup(…, require_ml)` alongside main's `test_canonical_identity_invariant.py`. Full suite re-run on the **merged** tree: 1,878 passed, 10 skipped. Rollup is 12 checks, non-empty, all green.

### Two changes made after approval, both disclosed

**1. The two default-asserting tests were machine-dependent.** `config.py:86` sets `env_file=".env"` and `.env.example:26` documents `DEDUP_REQUIRE_ML`. On any machine carrying that opt-out, `Settings()` reads the developer's `.env` instead of the shipped default. A/B'd against a real `.env` containing `DEDUP_REQUIRE_ML=false`:

```
without the pin:  test_dedup_requires_ml_by_default          AssertionError: assert False is True
                  test_default_settings_make_run_dedup_raise Failed: DID NOT RAISE MissingDependencyError
with the pin:     2 passed
```

Pinned with `Settings(_env_file=None)` and `monkeypatch.chdir(tmp_path)`. To be exact about the severity: this produced a false **red** on such a machine, not a silent green, so it was never masking a defect — but the tests that assert the shipped default should not be reading a `.env`.

**2. A comment, corrected — and the correction that prompted it was itself off by five.** The guard's `checked at stage *entry*, before any input is read` was read as a claim about `resolve`. It is **dedup's** entry, now stated explicitly.

The framing handed to me was that "dedup is Step 4 of `run_all`", with eight stages listed as already-run. Those two cannot both hold. `RESOLVE_STEP_ORDER` is `ner, disambiguate, bio_promote, cluster, narrator_split, reconcile, tabaqa_dates, muhaddithat_links, dedup, parallels` — **dedup is the ninth of ten**, and the count of eight prior stages is the part that is right. The comment now says ninth, names them, and records the operational consequence: exit `4` means *a stage aborted mid-pipeline with prior artifacts on disk*, not *the pipeline refused to start*. An operator seeing exit 4 must not assume nothing was written. `MissingDependencyError` subclasses `BaseException` precisely so it escapes those stages' handlers.

Neither change touches `dedup`'s behaviour. No renumbering, no value change.

parametrization:
## Fresh verdict at head `8d567eea37178257390f4c12df0f7b14aa76fb57`

My earlier approval was against `b6acb49e` and `src/` has changed since. It is void, and it was right to treat it as void rather than as permission — a verdict is a statement about a sha, and the hook counting comments instead of shas is the fail-open that main#935 exists to close.

`MERGEABLE` / `CLEAN`, 12 runs, 0 not-success, resolved with `-R noorinalabs/noorinalabs-data-acquisition`. Read at the sha; throwaway detached worktree; no checkout in the shared clone.

**Approved.**

### The contribution delta is four lines out and one line in

I compared contribution against contribution — `merge-base(b6acb49e, main)..b6acb49e` versus `1abb7e25..8d567eea` — because an `old-head..new-head` diff sweeps in everything `main` brought and proves nothing. That diff names four `src/` files; two of them (`src/exit_codes.py`, `src/resolve/_checkpoint.py`) are byte-identical to `main` and are da#387's, not hers.

Per file, comparing only the added and removed lines of each contribution:

```
src/cli.py                 contribution IDENTICAL
src/resolve/dedup.py       contribution IDENTICAL
src/config.py              contribution IDENTICAL
src/resolve/__init__.py    contribution IDENTICAL
src/resolve/_deps.py       contribution CHANGED
```

And the whole of that change:

```
removed:  #: Process exit status the CLI returns when an enabled stage is missing a declared
          #: dependency. Distinct from ``EXIT_STOPPED_AT_LIMIT`` (3) so an operator — or a
          #: wrapping script — can tell an environment defect from a bounded partial run.
          EXIT_MISSING_DEPENDENCY = 4

added:    from src.exit_codes import EXIT_MISSING_DEPENDENCY
```

A declaration-site move. **The value is unchanged — `4` before, `4` after — and it is now the registry's object rather than a coincidentally equal integer:**

```
_deps value           : 4
registry value        : 4
_deps IS registry obj : True
```

That last line is the one that matters. Before the rebase, `src/cli.py`'s `sys.exit(EXIT_MISSING_DEPENDENCY)` emitted an `int` that happened to equal `ExitCode.MISSING_DEPENDENCY`. Now it emits the member. `4 is <ExitCode.MISSING_DEPENDENCY: 4>` is exactly the assertion that reddened on the rebase, and it is exactly what da#384 was for.

### The offender set is a reading, not a silent zero

`tests/test_exit_codes.py` is `62 passed` at this head. That is only worth something if the guard can still see an offender in *this* file, so I planted two into `src/resolve/_deps.py`, each asserted present before running:

```
re-declare EXIT_MISSING_DEPENDENCY = 4    ->  2 failed, 60 passed
declare a new EXIT_PLANTED = 99           ->  2 failed, 60 passed
```

Both reds cite the sole-declarer scan and the resolves-to-the-registry test. The empty offender set at the unplanted head is therefore a measurement.

### What I originally approved is intact

`MissingDependencyError` still subclasses `BaseException` (`_deps.py:39`) — the property the whole da#309 fix rests on, since `run_all` wraps every stage in `except Exception`. `src/cli.py:157` still carries the `except MissingDependencyError` handler and its exit. `tests/test_resolve/test_dedup_fail_loud.py` is `11 passed`. The `dedup.py` guard, `config.py`'s `dedup_require_ml`, and the `resolve/__init__.py` re-export are contribution-identical to the tree I read the first time.

### On `git merge-tree` calling `src/cli.py` conflict-free

Kavitha is right that a clean auto-merge is not evidence of a correct merge, and `src/cli.py` is the demonstration. Worth adding what caught it here, because it was not a human: **da#387's self-expiring pin doubles as a merge-integrity check on that exact file.** `test_the_live_tree_has_exactly_one_unnamed_exit_and_da372_owns_it` asserts the live tree contains precisely one unnamed exit. Had the merge reverted da#387's routing of `_check_neo4j`, `_cmd_validate` and `_cmd_enrich` — a change textually confined to lines `git merge-tree` would happily take from either side — the count would be three or four and that test would red. It passes. A guard written to expire on one PR's rebase turns out to detect a semantic revert on another's.

### One near-miss of my own

My first per-file comparison ran the loop, reported `_deps.py` as CHANGED, and then a follow-up `diff` of the two contributions printed **nothing** — which reads as "identical," contradicting the loop. Two empty inputs also print nothing. I checked the line counts (`116` and `113`, both non-zero) before believing either result. Had I not, I would have reported the contributions identical and approved on a comparison that never happened.

A `diff` with no output and a `diff` that never ran are the same silence, exactly as an unapplied mutation and a surviving mutant are the same green.

---
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: The three items from my first verdict are unchanged and now tracked — `require_dependencies()` remains exported with no caller (dedup raises `MissingDependencyError` directly), `missing_dependencies` still catches only `ImportError` so an installed-but-broken native dep escapes the guard and is swallowed by `run_all` (da#360), and the composed `parallel_links.parquet` still carries no detector provenance (da#378).

parametrization:
## Re-verdict at `8d567eea` — supersedes my earlier approval

My prior approval on this PR was posted against `cf5e5de0b5bacee3a59cb006762c9a73e481201d`. The head is now `8d567eea37178257390f4c12df0f7b14aa76fb57`. I re-reviewed from scratch. No branch was moved: files were read via `contents?ref=<sha>` and the tree exported with `git archive`.

### 1. `_deps.py` declares nothing

At this head the name appears three times and none is an assignment — an import at `_deps.py:29`, the `__all__` entry at `:32`, and a docstring reference at `:50`. The `EXIT_MISSING_DEPENDENCY = 4` literal is gone. The re-export chain terminates at the registry: `cli.py:123` imports from `src.resolve`, `resolve/__init__.py:30` imports from `._deps`, `_deps.py:29` imports from `src.exit_codes`, and `exit_codes.py:251` binds the alias to `ExitCode.MISSING_DEPENDENCY`.

### 2. The value is unchanged — including the part a value-equality test cannot see

`ExitCode.MISSING_DEPENDENCY` is `4` and its member docstring cites da#309 (`exit_codes.py:198-199`). The real risk was not the integer. `cli.py:157` calls `sys.exit(EXIT_MISSING_DEPENDENCY)`, and substituting an enum member for an `int` literal is only behaviour-preserving if the enum is an `IntEnum` — under a plain `enum.Enum`, `sys.exit` prints the repr and exits **1**. So I executed it rather than reasoning about it:

```
ExitCode bases: ['IntEnum', 'int', 'ReprEnum']
alias is registry object: True
ACTUAL PROCESS EXIT STATUS = 4
```

`@enum.unique` is applied at `exit_codes.py:145`. The CLI's observable contract is intact.

### 3. Offender scan across all of `src/`, with a positive control

I scanned all 99 `.py` files under `src/` at this head for any `Assign`/`AnnAssign`/`AugAssign` binding a `Name` beginning `EXIT_`, excluding the registry. Before trusting an empty result I planted a fake binding into `src/resolve/_deps.py` and re-ran the same predicate:

```
CONTROL: planted EXIT_PLANTED=99 into src/resolve/_deps.py
CONTROL: detected 1 planted binding(s): [('src/resolve/_deps.py', 116, 'EXIT_PLANTED')]
CONTROL: PASS
CONTROL: total hits with plant present = 1

SCAN: files scanned = 99
SCAN: offenders = 0
```

The only offender the predicate ever found was the one I introduced. Zero real offenders, from an instrument shown able to see one.

### 4. Contribution diff — and my baseline was not the one I was handed

I was briefed to diff `merge-base..b6acb49e` against `1abb7e25..8d567eea`. Done, and against that pair the delta is exactly 7 lines, all inside the `_deps.py` hunk: two added (the import, a blank), five removed (a blank, a three-line comment, the assignment). `cli.py`, `config.py`, `resolve/__init__.py` and `dedup.py` all have byte-identical *contribution hunks* across that pair.

But `b6acb49e` is **not** the sha I approved. I approved `cf5e5de0`, which is an ancestor of `b6acb49e`, and 418 lines of `src/` moved between them. Almost all of that is main flowing in (`identity.py`, `disambiguate.py`, `fuzzy_cluster.py`, `arabic.py` — da#356/PR#363), now sitting below the merge-base and outside her contribution. One piece is not: **her `dedup.py` contribution grew from 99 to 108 insertions after I approved it.** The briefed diff window could not have shown me that, so I diffed her contribution against my real baseline.

That growth is comment-only. Across `cf5e5de0 → 8d567eea`, per file:

```
  cli.py       contribution IDENTICAL since my approval
  config.py    contribution IDENTICAL since my approval
  __init__.py  contribution IDENTICAL since my approval
  dedup.py     CHANGED; new '+' lines=10, of which non-comment code=0
  _deps.py     CHANGED; new '+' lines=1,  of which non-comment code=1
      CODE: +from src.exit_codes import EXIT_MISSING_DEPENDENCY
```

Since my approval, the single new line of executable code anywhere in this PR is that import. The ten `dedup.py` lines are a comment explaining that exit 4 fires at *dedup's* entry rather than *resolve's*, so an operator seeing it must not assume nothing was written. That is a correct and useful clarification, and I checked the fact it asserts: `dedup` really is the ninth of the ten `RESOLVE_STEP_ORDER` steps.

### 5. CI

The rollup is non-empty — 12 checks, all `success` — and I confirmed each is bound to `head=8d567eea` via `commits/<sha>/check-runs` rather than trusting the PR-level rollup.

### On the registry's self-justifying docstring

I did not weigh it. The sentence arguing that the alias form "lets an already-approved PR rebase without invalidating its approvals" is prose written by an interested author, and this review is its counterexample: the rebase was value-preserving, yet re-review still surfaced a post-approval change to `dedup.py` that the rebase had nothing to do with. My approval rests on the diff and on the executed exit status above.

That `git merge-tree` called `cli.py` conflict-free while `cli.py` was the file that broke is the same lesson from the other side: a clean auto-merge is a statement about text, not about meaning. The load-bearing checks here are the offender scan and the executed exit status.

### Assessment

Minimal, correct, confined. It resolves the da#384 double-booking by making the registry the sole declarer, while leaving `_deps.py`'s public surface and the CLI's exit status unchanged. I approve at this head.

Requestor: Tarek Mansour
RequestOrReplied: Approved
TechDebt: Exit 4 is documented under two colliding step-numbering schemes — `exit_codes.py:42` says "dedup is Step 4" and its member docstring says "Steps 1-3.7 already ran", while the new `dedup.py` comment says dedup is the ninth of ten `RESOLVE_STEP_ORDER` steps (which is what the code says). All three are defensible in their own scheme, but an operator who hits exit 4 and greps for what is on disk gets contradictory guidance. Reconcile on `RESOLVE_STEP_ORDER` as the single source of truth. (Related, and already self-documented at `exit_codes.py:112-120`: the uniqueness guard is `EXIT_`-prefix-scoped and cannot see a second `IntEnum` outside the registry whose members carry neither that prefix nor a registry member name — my AST scan inherits the same blind spot.)

PR #359, opened 2026-07-09
fix(identity,parse): assert the id contract instead of silently repairing it (da#355, da#358)

parametrization:
## Summary

Both issues are the same theme: **stop silently repairing bad ids; start asserting.** Sequenced after da#353 (PR#357), which removed the last producer of the doubled form.

Closes #355
Closes #358

## da#355 — `_collapse_double_corpus` asserts instead of repairing

`corpus:collection:...` is the *defined grammar* of a `source_id`. The collapse could not distinguish a genuinely double-prefixed id from a collection legitimately named after its corpus, and when it guessed wrong it performed a **lossy rewrite of a valid identifier** without failing. It fired 16M+ times in a single load.

- `bare_source_id` raises `DoubledCorpusPrefixError` (a `ValueError` subclass) rather than collapsing.
- The `hdt:` strip stays idempotent — genuine, unambiguous prefix-stripping.
- The occurrence-throttled warning (`_double_corpus_collapse_count % 100_000`) is removed with the guard.
- `hadith_node_id("lk:lk:1")` is no longer silently rewritten to `hdt:lk:1`.

### Is a `migrate.py` migration needed?

**No new migration is needed — but the existing one had to be made self-contained.**

`src/graph/migrate.py::compute_rewrites` (da#325) called `hadith_node_id` and **depended on the collapse**: its entire purpose is rewriting legacy `TRANSMITTED_TO.hadith_id` values like `sanadset:sanadset:0:0:2326`. Leaving the assertion in the constructor would have made the migration abort on exactly the ids it exists to fix.

So the repair moved to `migrate.py` as an explicit, one-shot `canonicalize_legacy_hadith_id`. It is sound there where it was not in the constructor: only values a *historical* loader already persisted reach it, and for those the doubled corpus is **known** — not guessed — to be the main#139 defect (that loader doubled the corpus onto ids whose collection segment was `0`, never a corpus-named collection).

No **new** migration is required for stg/prod: the graph is scheduled to be wiped and rebuilt from a fresh da#353 parse, so the legacy ids disappear at the source. `migrate.py` remains correct for any legacy graph and is a no-op on a rebuilt one.

## da#358

**(a) `num_col` no longer accepts `hadith_id`.** A row with no recognized ordinal now falls back to `_content_digest` — content-derived and stable, never a surrogate.

[...]

Issue #356, opened 2026-07-09
resolve: disambiguate keys canonical identity on the matched bio candidate's name, not the mention's — 21,644 chimeric nodes, 92% via crossref (Stage 5), not fuzzy


parametrization:
> **DIAGNOSIS CORRECTED 2026-07-09.** The original body of this issue (below, preserved) named `fuzzy_cluster` + `_choose_representative` as the site. **That was wrong**, and it was my error, not the investigator's. Ivana Horvat disproved it before writing any code. The symptom, the severity, and the acceptance criteria all stand. The mechanism and the fix location do not.

## Actual root cause: `disambiguate` Stage-2 fuzzy bio-match

The collapse happens when a mention is first assigned a canonical id — **upstream of any cluster merge**. There was never a separate correct-spelling node to merge away.

`merge_log.parquet` is decisive:

| node (`name_ar`) | mention_count | `merge_stage` of its mentions | modal real mention text |
|---|---:|---|---|
| `عائذة` | 35,436 | **fuzzy 35,434**, exact 0 | `عاءشه` (ʿĀʾisha) — 35,416 |
| `ابن عبس` | 50,962 | **fuzzy 49,886**, crossref 1,005, exact 4 | `ابن عباس` (Ibn ʿAbbās) — 49,826 |
| `الأعلم` | 19,296 | **fuzzy 19,253**, crossref 31, exact 6 | `الاعمش` (al-Aʿmash) — 19,234 |

The **itqan rijāl database carries OCR/variant-corrupt spellings** (`عاءذه`, `ابن عبس`, `الاعلم`) and does **not** carry the correct forms (0 exact bio candidates). So a mention of the real narrator finds no exact bio, falls through to `_fuzzy_match_blocked`, and matches the *corrupt* bio at `fuzz.ratio >= 80` with `Levenshtein <= 2` — one Arabic letter (`fuzz.ratio("عاءشه","عاءذه") == 80.0`, distance 1). `disambiguate.run` (`src/resolve/disambiguate.py:1203,1223`) then keys the mention on the **candidate's** normalized name and stamps the candidate's `name_ar`, demoting the real spelling to an alias (`:1250`).

That also explains the `source_corpus: fawaz` / `source_ids: ['itqan:*']` disagreement: the corpus comes from the mention, the identity from the bio.

### Two hypotheses from the original body, both refuted

[...]

PR #363, opened 2026-07-09
fix(resolve): canonical identity is a function of the mention, never the bio (da#356)

parametrization:
Part of #356. Not `Closes` — this needs post-reload verification on the re-run before we close it.

## What was wrong

`disambiguate` derived a mention's canonical **identity and display name** from the matched biographical candidate rather than from the mention itself. A mention `عائشة` fuzzy-matching the OCR-corrupt itqan bio `عائذة` produced a node keyed `uuid5(عاءذه)`, displayed `عاءذه`; the mention's own correct spelling was demoted to `aliases`.

Measured independently on the current curated artifacts: **14,316 chimeric canonical nodes holding 593,456 mentions.** (Not the 14,223 / 593,314 in the issue — that figure compares against the *stored* `name_normalized`; the disambiguator re-normalizes before matching, so the re-normalized surface is identity-relevant.) Full measurement report: https://github.com/noorinalabs/noorinalabs-data-acquisition/issues/356#issuecomment-4929258717

## What changed

**1. Identity invariant.** Canonical id and display name are a pure function of the mention's normalized form. A bio match may attach `external_id`, `birth_year_ah`, `death_year_ah`, `generation`, `gender`, `trustworthiness`, `source_ids`. It may not rename, and it may not re-key. `_upsert_canonical` enforces this *at the sink* — `name_ar`/`name_en` are never back-filled from the candidate — so no future call site can regress it. The bio's spelling is kept as a searchable alias instead. The two branches of the streaming loop (bio-resolved vs self-canonicalized) are unified, which is what makes the invariant structural rather than incidental.

Stated precisely, with its one carve-out:

> `norm_name ∈ { mention_norm, refine_mononym_name(mention_norm, chain_evidence).norm_name }` — and **never** `candidate.name_ar_normalized`.

The carve-out is load-bearing. `refine_mononym_name` (da#248) *deliberately* re-keys away from the mention surface, driven by **chain evidence** rather than by a bio name. A naive "a node's name must appear among its own mentions" assertion false-positives on every mononym split by design — and an assertion that fires on correct behaviour is an assertion the next person deletes. The chimera invariant test encodes the carve-out explicitly (3 nodes in the current corpus).

**2. Corroboration gate** (`_bio_corroborated`). A confident-by-score match must also clear: chain-neighbour temporal agreement, or — absent temporal evidence — near-identity of names (`score >= 0.90`, `Levenshtein <= 1`). Contradicting temporal evidence vetoes outright.

This still matters once identity is safe, because an attached `death_year_ah` enters `death_year_index`, which feeds `_temporal_filter` for chain neighbours **and** `refine_mononym_name`'s da#248 evidence — bad metadata propagates *back into* identity.

## Two premises in the issue were wrong

**`crossref` (Stage 5), not `fuzzy` (Stage 2), is the dominant mechanism.** Chimeric `merge_log` nodes by dominant stage:

| stage | rows | chimeric nodes | chimeric mentions |
|---|---|---|---|
| `exact` | 1,266,245 | 3 | 1,365 |
| `fuzzy` | 742,607 | 1,741 | 137,422 |
| **`crossref`** | 937,539 | **19,900 (92%)** | **474,126 (77%)** |

`fuzzy` requires `ratio >= 80 AND lev <= 2` then runs `_temporal_filter` + `_geographic_filter`. `crossref` required only `ratio >= 60`, had **no Levenshtein bound**, and ran through **neither filter**. With a 2-char blocking prefix every `ابو …` kunya lands in one block — one node had accreted **607 distinct names**, merging Imam al-Bāqir with `ابو نصر احمد بن سهل الفقيه`. Scoping the gate to Stage 2 would have fixed ~8% of the nodes: a gate that doesn't cover the path the defect actually takes is not a gate. crossref now runs through both filters and the gate.

**Corroboration criterion (a) — "the mention's own form is itself bio-registered" — is vacuous.** Survival: **0 of 742,607** fuzzy, **0 of 937,539** crossref. Provable a priori: Stage 1 returns early on any exact bio hit, so nothing reaching Stage 2/5 can be bio-registered. Adopting it is byte-equivalent to deleting both stages.

**Criterion (b) alone was also rejected, and this is the important part of the reasoning:** only **26.2%** of the 140,174 bio candidates carry a `death_year_ah` at all. Requiring positive temporal agreement therefore rejects ~70% of `fuzzy` **on grounds of missing data, not on evidence of wrongness.** That distinction is the whole argument for the composite. Criterion (c) alone is insufficient in the other direction: at `score >= 0.85` it still admits 26.8% of `crossref` — the 92%-chimera stage — with no temporal check whatsoever.

`_CONFIDENCE_THRESHOLD` is **unchanged at 0.70**. It was never the defect. With the branches unified it now governs only whether bio metadata attaches — what it should always have meant.

## Does this fragment the graph? — derived, not asserted

An earlier version of this PR asserted a 89.4% / 10.6% split from a **lexical proxy** (`ratio>=90 & lev<=2` against each group's modal form). Two reviewers falsified the load-bearing half of it. It is replaced here by the **production predicate itself** — `fuzzy_cluster._can_merge` at the production threshold (90.0) and `_MIN_SHARED_TOKENS = 2`. Since `_can_merge` requires ≥2 shared significant tokens, which is exactly the condition for sharing a `combinations(tokens, 2)` blocking key, `_can_merge(a,b) == True` implies the pair is scored together. No proxy.

Of the **124,834** new nodes de-keying creates (post-fold), measured on `merge_log.parquet`:

| fate | nodes | share | mentions |
|---|---|---|---|
| **A.** `fuzzy_cluster` **will re-merge** (`_can_merge = True`) | 20,692 | 16.6% | 73,811 |
| **B.** scored, and the precision guards **refused** | 96,917 | 77.6% | 289,366 |
| **C.** **never scored** — fewer than 2 significant tokens, no blocking key | 7,225 | 5.8% | 94,657 |

**B is a decision; C is a non-decision.** My old "89.4% are different people" conflated the two, and that conflation is exactly what the reviewers caught. Class B is the graph correctly separating distinct men — `عبيد الله بن العباس` vs `عبد الله بن عباس`, `ابو سعيد المقبري` vs `ابو سعيد الخدري`, `جابر بن یزید` vs `جابر بن زيد`. Class C is the da#376 structural blind spot: `fuzzy_cluster` never looked.

Class C is **reduced but not eliminated** by this PR. Across the whole corpus the never-scored population goes from **7,694 forms** to **7,169**; the fold converges the named inflection families (Abū Hurayra's 54,433 mentions collapse from three nodes to one) but a single-significant-token name still generates no blocking key. `_MIN_SHARED_TOKENS = 2` is a deliberate precision guard ("a bare single-token subset never clusters") and loosening it is not in scope here. What this PR does is make C **countable**: `ClusterMetrics.unblockable_records` and a `cluster_records_unblockable` warning, so "never scored" can no longer be mistaken for "scored and refused."

`أنس بن مالك`/`أنس` (da#347) and bare `عبد الله` (da#346) remain correctly self-keyed separate nodes and are unchanged.

## The mononym-split ordering bug — same defect, one call deeper

`refine_mononym_name` was called on `norm_name` *after* that variable had been overwritten with the bio's name, so the da#248 registry lookup keyed on the **bio** spelling. A mention `سفيان` (a registry key) that fuzzy-matched a bio spelled `سفين` (ratio 0.889, lev 1, *not* a registry key) looked up `سفين`, missed, and the split was silently **suppressed**. Fixed, and given its own test — red against `origin/main`:

```
E  AssertionError: the decoy bio's spelling must never become a canonical name
E  assert 'سفين' not in {'ابن عباس', 'سفين'}
```

## Tests — red-first, and the red count corrected

`tests/test_resolve/test_canonical_identity_invariant.py`, **36 tests**. Fixtures use real corpus spellings and production Parquet schemas.

### Correction (Alejandra Reyes-Fuentes, reviewing `27d20b1`) — this section previously overstated its own evidence

An earlier revision of this body claimed `12 failed, 5 passed` against `origin/main`, of which **7 behavioural**, and claimed the da#376 guards use only symbols that already exist on `main`. **Both claims are now wrong, and one was wrong when written.**

- The `12 failed, 5 passed` figure was measured at commit `47fa5a9`, where the module imported no new symbols and therefore *did* collect against `origin/main`. It is **unreproducible at head**: the module now imports `canonical_surface`, so against `origin/main` it raises `ImportError` at collection and yields **`no tests collected, 1 error`** — not a red count at all.
- The claim that the da#376 guards use only pre-existing symbols is **false**. Three of them require `canonical_surface` or `count_unblockable`. That property held only of a *standalone throwaway* file I used to demonstrate the behavioural red, not of the guards as they landed in the module.

Measured at head against `origin/main`, with the single neutral shim `canonical_surface := normalize_arabic` (which *is* `origin/main`'s identity surface — nothing else stubbed, since a missing `_bio_corroborated` / `count_unblockable` *is* the absence of the feature):

```
26 failed, 10 passed
  ├─ 18  behavioural
  └─  8  symbol / signature / schema absence
         (5× AttributeError _bio_corroborated, 1× count_unblockable,
          1× ClusterMetrics unexpected kwarg, 1× KeyError death_year_provenance)
```

The behavioural reds reproduce the **exact production node ids**, which is what shows the fixture models production rather than a fabricated shape:

```
E  AssertionError: canonical node must be keyed on the MENTION's form 'عاءشه';
   got nodes [('nar:ed1f1633-aa9a-54f0-b411-a65e7c584280', 'عاءذه')]

E  AssertionError: chimeric canonical nodes (name absent from own mentions):
   [('nar:ed1f1633-…', 'عاءذه', ['عاءشه']),
    ('nar:56fdc572-…', 'ابن عبس', ['ابن عباس']),
    ('nar:4382ae39-…', 'ابو جعفر محمد بن علي الكوفي', ['ابو جعفر محمد بن علي الباقر'])]

E  AssertionError: crossref must not re-key the mention onto the bio
E  AssertionError: the decoy bio's spelling must never become a canonical name
E  assert 'سفين' not in {'ابن عباس', 'سفين'}
E  AssertionError: three Abu Hurayras: {'ابي هريره': 'nar:016010e4-…', …}
```

### The fold is mutation-covered — it was not

Alejandra mutated each of `_fold_token`'s three rules independently at `27d20b1`. Killing the kunya fold → 2 red. Killing the glued `عبدال` split → 1 red. **Killing the accusative-stem strip → 30/30 still green.** The rule could be deleted and nothing noticed.

It hid because every fold test was negative-direction or idempotence-only, and **both are satisfied by the rule's absence** — `canonical_surface("عليا")` is a fixed point whether it folds or not. Nothing asserted the rule merges anything. That is the same "gate that cannot fail" this PR exists to remove, inside this PR. The rule covers `انس` (da#347's exact under-merge case), `علي` (a top-20 chimera), `مالك`, `نافع`, `معمر`.

`test_accusative_stems_are_folded` now asserts both directions over the six highest-mention stems. Mutation-verified at head:

| mutant | reds |
|---|---|
| delete accusative strip | **6** (was 0) |
| delete kunya fold | 2 |
| delete `عبدال` split | 1 |

and under the accusative mutant the negative guards stay **green** — the proof they never could have caught it.

Guards green **before and after** (real regression guards, not tautologies): the da#248 mononym split, and the fold hazards `زكريا` / `عمرا` / `عبدان` / `عبدوس`. My first draft of the da#248 guard was wrong and abstained — year 131 is temporally plausible for *both* Sufyāns (gaps 30 and 67) — so it now derives a discriminating year and asserts that precondition.

Post-fix: 36/36 green. Full tree → **1869 passed, 7 skipped**.

## End-to-end verification against the real bio corpus

Ran the fixed stage against the **actual production** `narrators_bio_*.parquet` (140,174 candidates):

```
mention                       node name                     ok    ext_id
عائشة                         عاءشه                         PASS  None
ابن عباس                      ابن عباس                      PASS  8572
الأعمش                        الاعمش                        PASS  None
أنس بن مالك                   انس بن مالك                   PASS  54058
أنس                           انس                           PASS  16576
عبد الله                      عبد الله                      PASS  59596
أبو جعفر محمد بن علي الباقر   ابو جعفر محمد بن علي الباقر   PASS  None

canonical nodes: 7   chimeric: 0
```

Ibn ʿAbbās keeps his own spelling **and** picks up the itqan `external_id` — the strict near-identity path attaching metadata, as designed (`ابن عبس` *is* his OCR-corrupt bio). al-Aʿmash keeps his spelling with metadata correctly withheld (score 0.833 < 0.90).

## Adjacent finding, not fixed here

Identity is keyed on `normalize_arabic(mention_text)` rather than on the stored `name_normalized`. That column is written by `clean_narrator_name(normalize_arabic(x))` (name-quality, da#247 family) but re-derived downstream by `normalize_arabic(x)` alone — **two different functions**, so the stored value and the re-derived value disagree on **47,586 of 129,234 canonical rows (36.8%)**, and `make_canonical_id` could mint two ids for one name. Keying on the re-derived form makes `disambiguate` internally consistent; the provenance mismatch itself is parse-side and is tracked separately.

Weighted by `mention_count` the drift is **53,933 / 3,249,721 = 1.7%** (20,134 of the drift rows are `mention_count == 1` singletons) — it is the OCR/matn tail seen from the other side, **not** a graph-scale defect, and it does not move any ratio in this PR. `normalize_arabic` itself is provably idempotent (0 counterexamples over all 129,234 names); an earlier framing of mine that blamed it was wrong.

## ⚠ Merge-order constraint — read before resolving conflicts with #359

`git merge-tree` reports three conflicts with **#359**: `src/parse/identity.py`, `ontology/structural/llms.txt`, `ontology/structural/code-graph.json`.

**`src/parse/identity.py` is dangerous.** This PR *adds* `from src.utils.arabic import canonical_surface` immediately beside the `logger` import that **#359 deletes**. Resolving that hunk in favour of #359 wholesale drops the import and leaves `make_canonical_id` raising `NameError` at runtime — and CI may stay **green on #359's own suite** while doing it. **Whoever rebases second must keep the `canonical_surface` import.**

The two `ontology/structural/*` conflicts are by construction: both PRs commit a regenerated structural index, which per `CLAUDE.md` is generated, never hand-edited. **Regenerate after the second merge (`python3 scripts/structural_ontology.py emit`); never hand-resolve them.**

Separately: this PR rewrites **every `nar:` id in the corpus**, because the fold sits inside `make_canonical_id` and every minting site routes through it. It therefore **cannot ship as an incremental load onto the existing graph** — only as part of the wipe-and-reload from `parse`. No in-repo consumer holds a `nar:<uuid>` literal (grep: 0), so the constraint is external: prod/stg and `noorinalabs-isnad-graph`'s narrator model resolve by that id.

## Scope caveat

These artifacts predate the da#353 parse fix (~76% of corpus), so absolute counts will move on the re-run. The **ratios and the mechanism** are what this PR asserts. `src/parse/**` (Nikolaos, da#355/#358) and `src/graph/**` are untouched.

TechDebt: #371, #376

- **#376 residual** — the structural blind spot is *counted*, not closed. 7,169 canonical forms still produce no blocking key and are never offered to `_can_merge`. `_MIN_SHARED_TOKENS = 2` is a precision guard; loosening it needs its own precision/recall harness.
- **#371** — `name_ar_normalized` is still written by `clean_narrator_name(normalize_arabic(x))` and re-derived by `normalize_arabic(x)`. `disambiguate` is now internally consistent, but the column is not. Parse-side; `src/parse/**` untouched apart from the two-line `make_canonical_id` fold.
- The `+14,146 new orphans` figure cited in review is **not reproduced here** — I could not derive it from the artifacts under any definition I tried (never-scored forms = 7,694; new nodes = 127,434; `mention_count==1` forms = 94,073). Asking for its derivation rather than repeating a number I cannot stand behind.


parametrization:
Requestee: Ivana Horvat
Verdict: Changes Requested
TechDebt: 4,547 permanently-duplicated canonical nodes (35,558 mentions) that `fuzzy_cluster` structurally cannot re-merge, plus up to 14,146 new zero-degree bio_promote orphans feeding da#352 — both currently undeclared.

Reviewed at `47fa5a9d9eac68da930791b8475fb6284da0aedd`, architectural / downstream-consequence lens (Oyunbileg has the fixture-realism lens). Everything below is measured against the same curated artifacts the PR measured on: `data/curated/{merge_log,narrators_canonical,narrator_mentions_resolved}.parquet`. I reproduce your 34,915 → 162,349 and your 13,565 exactly, so we are reading the same numbers.

**The identity invariant is right and I want it to land.** Keying canonical identity on the mention and enforcing it at the sink in `_upsert_canonical` is the correct architecture, the unification of the two streaming branches is what makes it structural rather than incidental, and finding that `crossref` rather than `fuzzy` is the dominant chimera engine is a real result. My objections are to three load-bearing claims in the body that the launch plan is about to consume, and to one downstream consequence the body does not mention. None of them require you to redesign the fix.

## 1. The corroboration gate is self-defeating, and it disarms the guard the fragmentation argument depends on

Replaying `_bio_corroborated`'s no-temporal-evidence branch (`score >= 0.90 AND Levenshtein <= 1`) over `merge_log`, approximating each match's bio name by the pre-fix chimeric node's `name_ar_normalized` — an upper bound on rejection, since it cannot credit the positive temporal branch:

| stage | rows | rejected |
|---|---|---|
| `exact` | 1,205,424 | 0 (0.0%) |
| `fuzzy` | 709,622 | 542,895 (76.5%) |
| `crossref` | 824,442 | 822,411 (**99.8%**) |

Of 98,874 post-fix forms that carry a bio-attached death year today, **86,363 (87.3%) lose it entirely.**

Three consequences, in ascending order of seriousness:

- You reject criterion (a) because "adopting it is byte-equivalent to deleting both stages." By that same standard the composite gate deletes `crossref`'s metadata contribution — 99.8% rejection, ~2,031 surviving attachments out of 824,442. Post-fix `crossref` no longer contributes identity (identity is mention-derived), so metadata was the only thing it still did.
- It starves `death_year_index`, which is exactly the feedback path your docstring names: `_temporal_filter` is a documented no-op on empty evidence, so it passes more candidates through, and `refine_mononym_name`'s da#248 evidence thins, so *fewer* mononyms get split. Because the index is populated incrementally in stream order, this is monotone — withholding a year can only reduce the adjacent evidence available to later mentions, never increase it. Sufyān stays over-merged.
- The one that decides the PR: `fuzzy_cluster._death_years_conflict` returns `False` when either record's `death_year_ah` is `None`, so a missing year makes the merge *permitted*. Stripping the year from 87.3% of year-bearing forms disarms the precision guard on the clustering pass — the pass this PR nominates to safely re-merge the shattered nodes. The gate makes the re-merge that does happen less safe.

The gate governs two different things through one switch: *what enters the identity feedback loop* (`death_year_index`, da#248 evidence) and *what is stored on the node and shipped to the UI, `tabaqa_dates`, `date_reconcile`, the historical overlay, and `_death_years_conflict`*. Only the first needs the strictness. **Ask:** gate `death_year_index` and the da#248 evidence as you do now, but persist `death_year_ah` on the record with a provenance/confidence marker rather than dropping it. That keeps the feedback loop clean at ~zero product cost.

## 2. `fuzzy_cluster` does not re-merge the 10.6%. It cannot reach a third of them, and it can never reach the corpus's most-cited narrators

`cluster_records` builds its candidate blocks as `itertools.combinations(sorted(sig_tokens), 2)` — a record joins a block only via an **exactly shared token pair** — and `_can_merge` additionally requires `_MIN_SHARED_TOKENS = 2`. `_CONNECTOR_TOKENS` contains `ابن ابو ابي بن ال ام بنت عن`. So **a name with fewer than two significant tokens generates zero blocking keys, joins no block, and is never scored.** Your own code comment states the consequence: "Within a composite block every pair already shares the two key tokens, so the ≥2-shared-token guard holds by construction."

Measured on the post-fix form set:

- **7,694 forms (4.7%), carrying 1,065,833 mentions (36% of all `merge_log` mentions), are structurally unclusterable.** They include `ابن عباس` (49,826 mentions), `ابي هريره` (46,563), `عاءشه` (35,418), `سفيان` (30,182), `ابن عمر` (30,065), `شعبه` (29,251). These are the highest-traffic narrators in the corpus, and `fuzzy_cluster` has never once scored them.
- Of your 13,565 "spelling variant, `fuzzy_cluster` re-merges these": **4,547 (33.5%) cannot be re-merged**, on the name path, ever. They carry 35,558 mentions; 840 have more than five mentions, 48 have more than a hundred. The heaviest are grammatical-case variants of one person — `ابو الزبير` (2,026 mentions) stranded from `ابي الزبير`; `ابي الاحوص` (1,340) from `ابو الاحوص`; `ابو الزناد` (605) from `ابي الزناد`; `ابا الدرداء` (490) from `ابو الدرداء`. Each is one narrator becoming two permanent nodes. That is fragmentation, not an over-merge being undone.

The alias bridge does not rescue them. `bio_alias` is attached only when `_bio_corroborated` passes, whose non-exact branch requires Levenshtein ≤ 1; 1,705 of the 4,547 are at distance 2 and are denied the alias by the gate in item 1. And where the alias *is* attached, `bio_alias=attached.name_ar` passes the **raw** name, un-normalized, into a list that `fuzzy_cluster._match_keys` and `bio_promote._merge_aliases` both consume as a normalized key space. A raw spelling contributes tokens that match nothing. That is a one-line fix in this PR (`attached.name_ar_normalized`), and it is the only path by which the alias could have supplied the second shared token.

The classifier is also wrong in the other direction, which matters because it means the 89.4% is not the number you think either: `ratio >= 90 AND lev <= 2` against the modal form calls `الحسين`/`الحسن` (869 mentions) a spelling variant. So are `محمد بن بشر`/`محمد بن بشار` (1,118), `علي بن حرب`/`علي بن حجر` (708), and `يحيى بن ابي بكير`/`يحيى بن ابي كثير` (769). Those are four different pairs of different people. The bucket you labelled "will re-merge" contains persons who must not merge, and `fuzzy_cluster` is the only reason they don't.

**Ask:** correct the body, and file the `fuzzy_cluster` connector-token blocking gap as its own issue. It sits *at* the resolve stage, so under the owner's standing rule (a gate downstream of a defect is not a gate) it belongs in the main#928 pre-re-run set, not after it. da#295 covers cap tuning and da#337 covers generic-name over-merge; neither covers this.

## 3. "Zero graph-metric weight" for the singletons is not true

`load_edges` builds `TRANSMITTED_TO` from consecutive resolved mentions of each chain. Joining the singleton forms back to `narrator_mentions_resolved`:

- **100%** of the 94,073 singleton forms appear in at least one chain, so every one of them gets at least one `TRANSMITTED_TO` edge.
- **67,128 (71.4%)** sit strictly interior to a chain of length ≥ 3. Degree ≥ 2, on shortest paths, carrying betweenness.

`enrich/metrics.py` projects `('Narrator', 'TRANSMITTED_TO')` and runs `gds.betweenness.write` with a fixed `samplingSize`. Inflating the node set 4.65× against a fixed pivot count degrades the estimate for every *real* narrator, which is precisely the metric da#337 is open against. The singletons are not weightless; they are noise injected into the choke-point ranking.

To be clear about direction, because this is the question the lead asked and the answer is in your favour: the PR makes matn pollution **better**, not worse. Sentence-like forms (≥ 6 significant tokens) are 16,240 forms / 33,980 mentions, and 9,281 of those mentions are today swept onto short-named real-narrator nodes, inflating their `mention_count`, polluting their aliases, and attributing chain edges to them. Post-fix they self-key. That converts invisible contamination of real narrators into visible isolated junk. It is the right trade. It is just not weightless, and 12.7% of the new nodes — not the majority — are matn-derived.

## 4. Undisclosed: this PR feeds da#352 up to 14,146 new orphans

`bio_promote` unions by `make_canonical_id(bio_norm_name)`. Pre-fix, a mention re-keyed onto the bio's spelling minted exactly that id, so the promoted bio node inherited the mention's edges. Post-fix the mention keeps its own key, and the bio node is promoted alone.

Counting the chimeric nodes with a linked `bio_id`: **14,146** bio names lose all their mentions and come back as zero-mention, zero-degree catalog nodes. da#352 is currently `[FAIL]` at 70,855 orphans and is one of the three failures making `load` exit `rc=1`. This PR grows population 1 of that issue by ~20%. It is a defensible consequence — those bios genuinely have no chain participation, and da#352's own acceptance contemplates reclassifying them — but it must be stated, and da#352's triage has to be done knowing the number moves because of this change.

## 5. `Part of #356` — residual

Correct call, and the residual is real: post-reload verification cannot run until the re-run. da#356's title asserts 14,223 / 593,314; you measure 14,316 / 593,456 and explain the delta. I reproduce 14,147 (my join drops nodes the cluster stage remapped), which corroborates the magnitude. Please leave da#356 open and add items 2 and 4 to it, or to the new blocking-gap issue, so the pre-re-run gate is complete.

## What I need to approve

1. `bio_alias=attached.name_ar_normalized`, not `attached.name_ar`.
2. Decouple the gate: keep `death_year_index` and the da#248 evidence gated; persist `death_year_ah` with provenance rather than dropping 87.3% of it. If you'd rather defer this, say so and I'll take it as a separate issue — but then it is a declared, sized regression, not "identity is safe so the gate is free."
3. Body corrections: the 13,565 do not all re-merge (33.5% cannot); the singletons are not weightless (71.4% are interior chain nodes); the +14,146 orphan consequence for da#352.
4. File the `fuzzy_cluster` sub-two-significant-token blocking gap, and put it in the main#928 pre-re-run set.
5. Replace the tech-debt declaration — items 2 and 4 are introduced residuals, not pre-existing ones.

None of this touches the invariant. Fix the four and I approve on sight.

parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Ivana Horvat
Verdict: Changes Requested
TechDebt: The 0.90/lev-1 corroboration constants are a populated cliff (a ~1% surface sample puts ~12 surfaces in [0.88,0.92) at lev≤1 and ~63 at score≥0.90/lev 2, mixing true variants like `يحيي بن ايوب`~`يحيى بن ايوب` with distinct people like `الاسم`~`القاسم`); they are correctly named+documented, but want a labelled sample before the next tuning, and `_bio_corroborated`'s "vetoed outright when temporal evidence contradicts" promise is not true of the `exact` stage, which short-circuits ahead of the veto across 1,675 bio names carrying conflicting death years.

Reviewed at head `47fa5a9d9eac68da930791b8475fb6284da0aedd` (7/7 SUCCESS, MERGEABLE), read via `gh pr diff` / contents-at-sha; no branch switch. Independent runs were done in throwaway detached worktrees.

## The core fix is correct, and I verified the evidence

The identity re-keying is right and the chimera it removes is real. Re-running the 17 new tests against `origin/main`'s `disambiguate.py` (new test file dropped onto an unmodified main worktree) reproduces **12 failed, 5 passed** exactly as reported, and the chimera assertion prints the production node ids verbatim:

```
('nar:ed1f1633-aa9a-54f0-b411-a65e7c584280', 'عاءذه',  ['عاءشه'])
('nar:56fdc572-0317-5f9b-b36c-7247b62ab799', 'ابن عبس', ['ابن عباس'])
('nar:4382ae39-4b2c-50f4-aeaa-b45a5d9374b8', 'ابو جعفر محمد بن علي الكوفي', ['ابو جعفر محمد بن علي الباقر'])
```

Two honest qualifications. Five of the twelve reds are `TestBioCorroboratedUnit`, which fail on main with `AttributeError: module 'src.resolve.disambiguate' has no attribute '_bio_corroborated'` — a symbol-existence failure, not a behavioural reproduction. The behavioural red count on main is **7**. And the three `test_canonical_id_is_uuid5_of_the_mention_form` parametrizations assert only that `uuid5` is injective over two different strings; they never touch `disambiguate`. They are a useful *fixture precondition* (they prove `normalize_arabic` does not collapse the pairs, which would render the chimera test vacuous), but the docstring "the identity contract, stated directly" oversells them — they cannot go red on any regression of this code.

The da#248 guard is a genuine regression guard, not a green-both-sides accident. I confirmed 131 is plausible for both Sufyāns (so the first draft would indeed have abstained), the derived `candidates[0] = 81` uniquely selects `سفيان الثوري`, and the `(target,)` unpack enforces that uniqueness. Neutering `refine_mononym_name` to always abstain at head turns the test red with `da#248 split must survive: expected 'سفيان الثوري'`. It discriminates.

Corroboration gate and re-keying cover crossref on both paths by construction: `_bio_corroborated` and the `mention_norm` keying both live in `run()` and are stage- and matcher-agnostic, and the temporal+geographic filters were added to the crossref stage in `_disambiguate_mention` *and* `_disambiguate_mention_indexed` (only the latter is live; the former is exercised solely by `tests/test_resolve/test_negative.py`). `_CONFIDENCE_THRESHOLD` is untouched at 0.70, as are `_FUZZY_RATIO_THRESHOLD`=80 and `_LEVENSHTEIN_MAX_DIST`=2. The corpus statistics in the docstring check out exactly against `data/staging/narrators_bio_*.parquet` — 140,174 candidates, 26.2% carrying a `death_year_ah`. `src/parse/**` and `src/graph/**` are untouched; the diff is 4 files.

## What blocks this: the fragmentation claim is false, and it costs the top narrator

The 89.4% / 10.6% split is not substantiated anywhere in the diff, and the half that matters is wrong. `fuzzy_cluster._can_merge` requires `_shared_significant_token_count >= _MIN_SHARED_TOKENS` (=2), where significant tokens are exact-identity, ≥2 chars, and non-connector — and `ابو`, `ابي`, `ابن`, `بن` are all connectors. A spelling variant of a **single-significant-token** name therefore has 0 or 1 shared tokens and can *never* re-merge, whatever its ratio. This is structural, not statistical.

Measured on real data, against the head matchers: taking the top 3,000 mention surfaces by mention count, 2,419 groups that `main` merged onto one bio-keyed node, of which head splits 282, and in **232** of those at least one member is never re-merged by `fuzzy_cluster`. The head of that list is not marginal:

```
54,433 mentions  ابي هريره (46,563) | ابا هريره (5,691) | ابو هريره (2,179)   -> 3 nodes, pairwise can_merge=False
17,972 mentions  انس (16,881)  | انسا (1,091)                                  -> 2 nodes, shared_sig=0
17,164 mentions  علي (14,030)  | عليا (3,134)                                  -> 2 nodes, shared_sig=0
17,162 mentions  مالك (16,542) | مالكا (620)                                   -> 2 nodes, shared_sig=0
12,970 mentions  جابر (12,176) | جابرا (794)                                   -> 2 nodes, shared_sig=0
```

Those are one person each — Abū Hurayra's kunya declined nominative/accusative/genitive, and the accusative tanwīn alif on `انسا`/`عليا`/`مالكا`/`جابرا`. `normalize_arabic` collapses none of them (`N('انس') != N('انسا')`, `N('ابو هريرة') != N('ابي هريرة')`, `N('عبد الله') != N('عبدالله')`). The cleanest single witness: `جابر بن عبد الله` (9,375 mentions) fuzzy-matches bio `itqan:104493` at 0.9677 and `جابر بن عبدالله` (3 mentions) matches the same bio exactly, so `main` correctly merged them into `nar:5191ed39-920d-5236-af51-4d9323e5f757`. At head they become `nar:83777117-…` and `nar:5191ed39-…`, and `_can_merge` returns False even when I load the pair with their real production aliases — `بن` is a connector, so the only shared significant token is `جابر`.

So the bio key, for all the damage it did, was also incidentally absorbing Arabic inflection and spacing. Making `mention_norm` the sole identity source removes that absorption and hands the job to a `normalize_arabic` that is not case- or spacing-invariant. Over-merge becomes under-merge, `fuzzy_cluster` demonstrably does not catch it, and it lands on the #1 narrator by mention count — inside the same 7.5-hour resolve run this PR is meant to unblock, and squarely on the top-N deliverable the wave exists to produce. Per the sweep-the-expensive-stage rule (main#928), a defect at-or-before the costly stage has to be merged before that stage is launched; a fix downstream of the split is not a gate.

Worth saying plainly, because the same experiment shows it: the PR *also* correctly tears apart real chimeras in that population — `ابن عباس`+`ابى بن كعب` and `الزهري`+`البهي` were merged on main and are different people. `معمر`/`معتمر` are correctly kept apart. The mechanism is right. The normalization underneath it is not yet strong enough to carry identity alone.

## What I'm asking for

Do **not** revert the re-keying — restoring the bio key would reinstate the chimera. Before this becomes the identity key, make the mention surface inflection- and spacing-invariant: strip the accusative tanwīn alif, unify the kunya case forms `ابو`/`ابا`/`ابي`, and collapse `عبد الله`↔`عبدالله`, either in `normalize_arabic` or in a mention-side canonicalization applied where `mention_norm` is computed. Pin it with a test asserting `ابو هريره`, `ابا هريره`, `ابي هريره` land on one canonical id, and that `جابر بن عبد الله` and `جابر بن عبدالله` do too. Then re-derive the node count — 162,349 will drop, and the 89.4/10.6 split should be replaced by the reproducible measurement rather than asserted.

Two smaller items to fold in while the branch is open. Tighten the chimera carve-out: the test exempts any node whose name is in the *global* `MONONYM_REGISTRY` person set, which is broader than the stated invariant `norm_name ∈ {mention_norm, refine_mononym_name(mention_norm, evidence).norm_name}`. The registry holds only 5 person names, but three of them exist as real bios — `muhaddithat:20012` is `سفيان الثوري` and both `muhaddithat:11062` and `itqan:104662` are `يحيى بن سعيد الانصاري` — so a re-key onto one of those bio names would pass the assertion today. Restrict `refined` to persons reachable from that node's own mention surfaces. And either implement the exact-stage temporal veto that `_bio_corroborated`'s docstring promises, or amend the docstring to say the `exact` stage is exempt and why: `_exact_match_indexed` returns `exact[0]` unfiltered, and 1,675 normalized bio names carry mutually conflicting death years, so an arbitrary homonym's year is stamped and flows into `death_year_index` — which, by your own argument, is not inert.

Happy to re-review quickly once the normalization lands; the rest of this is ready.

parametrization:
Requestor: Oyunbileg Batbayar
Requestee: Ivana Horvat
Verdict: Approved
TechDebt: `death_year_provenance` is written but never read — `_death_years_conflict` and `_genders_conflict` are provenance-blind, so a year sourced from a non-exact bio match (≤20.8% of surfaces in a 2,500-surface sample) is read as merge-refusing evidence exactly as a corroborated one is, while the ~32.2% of surfaces with no bio match keep `death_year_ah` NULL and leave that guard blind (it returns `False` when either side is `None`); separately the `alias=` argument to `_upsert_canonical` is dead on the non-split path (`alias == norm_name` after the fold, so nothing is appended) — verified identical before the delta, so pre-existing, not introduced here.

Re-reviewed the delta `ed3e1fc..27d20b1` only, at head `27d20b117d0852105c123e14af82299e6333d14c`. Two commits: the da#376 fold and a memory correction. Read via `gh api .../contents?ref=27d20b1` and `git diff` of the two refs; no checkout in the shared clone — the runs below are in throwaway `git worktree add --detach` trees, removed after. Check-runs at head: 12/12 success, none failing. The full `tests/test_resolve tests/test_parse tests/test_utils` suite is 1,335 passed at head.

## 1. Eight minting sites — confirmed, and the count is load-bearing

I enumerated them with an AST walk over `src/` rather than by grep, because the docstrings mention the functions too. Callers of `make_canonical_id` / `make_discriminated_canonical_id` outside `identity.py`, classified into sites that *mint and keep* an id versus sites that only key a lookup:

```
MINT (8)                                           LOOKUP-ONLY (1)
  src/resolve/disambiguate.py:684                    src/resolve/bio_promote.py:101
  src/resolve/bio_promote.py:188                       rec = canonical_map.get(make_canonical_id(norm))
  src/resolve/fuzzy_cluster.py:1096
  src/resolve/narrator_split.py:320
  src/resolve/narrator_split.py:366  (discriminated)
  src/resolve/date_reconcile.py:457
  src/resolve/muhaddithat_links.py:177
  src/graph/load_edges.py:732
```

Nine call sites, eight mints — `bio_promote:101` computes an id purely to index `canonical_map` and is exactly what makes the number eight rather than nine. Your claim survives its own test.

The property that actually matters is that no site can bypass the fold, and none can: both mint functions route the name through `canonical_surface`, including the non-empty-discriminator branch of `make_discriminated_canonical_id`, which folds the name half before joining `_DISCRIMINATOR_SEP`. The only other `uuid5(CANONICAL_NAMESPACE, …)` in the tree is `muhaddithat_links.py:233`, and it mints a `mention_id`, not a narrator id — a namespace reuse across id kinds, harmless because the prefix and the payload differ. So the boundary crossing into `identity.py` is the right call, and folding at the seven call sites instead would indeed have left the eighth minting unfolded ids.

One pre-existing wrinkle, not in this delta and not yours to fix here: `load_edges._studied_under_endpoint` falls back to `narrator_node_id(external_id)` when the name is absent, minting a `nar:<external-id>` outside the fold entirely. Its own docstring records that such ids "match no canonical node, so every edge was silently dropped." That is a ninth way to produce a `nar:` string, it is not a name-keyed mint, and `load_edges.py` is untouched between `ed3e1fc` and `27d20b1`.

## 2. The precision guards go red — under exactly the mutation each one guards

Green-before-and-after was the right thing to be suspicious of. I mutated `src.utils.arabic._fold_token` at head, twice and independently, and ran `TestInflectionFolding` against each.

Mutating only the accusative rule to the naive "strip a trailing alif from any token of length ≥ 4":

```
test_hazards_are_not_folded[زكريا-زكري]  FAILED  Zakariyya is not the accusative of a name 'Zakari'
test_hazards_are_not_folded[عمرا-عمر]    FAILED  'Amr (عمرو) accusative is not 'Umar (عمر) — different men
test_hazards_are_not_folded[عبدان-…]     PASSED
test_hazards_are_not_folded[عبدوس-…]     PASSED
```

Mutating only the `عبد` rule to trigger on a bare `عبد` prefix instead of the article `عبدال`:

```
test_hazards_are_not_folded[عبدان-…]     FAILED  'Abdan is a name, not عبد + a second token
test_hazards_are_not_folded[عبدوس-…]     FAILED  'Abdus is a name
test_hazards_are_not_folded[زكريا-زكري]  PASSED
test_hazards_are_not_folded[عمرا-عمر]    PASSED
```

Each hazard reddens under its own mutation and stays green under the other — the guards discriminate rather than failing as a block, which is stronger evidence than a single all-red mutant would have been. `test_abu_hurayra_three_case_forms_mint_one_id` survives both, correctly, since neither mutant touches the kunya rule. `test_fold_is_idempotent` is green under both mutants; it guards a different property and should not be read as a fold-correctness sentinel.

The lexicon design is the right one. `_ACCUSATIVE_STEMS` being frozen and hand-curated rather than corpus-derived is what makes `canonical_surface` a pure function, and the excluded-with-reasons list in the source is the part a future maintainer will actually need.

## 3. The `None` path — the guard is unchanged, and it still permits

Tested directly on two records that clear the name and shared-token guards (`محمد بن ابراهيم التيمي` / `…السامي`, three shared significant tokens):

```
corroborated year vs NO year   -> can_merge = True    (_death_years_conflict = False)
both years, conflicting        -> can_merge = False   (_death_years_conflict = True)
both years, agreeing           -> can_merge = True
corroborated 161 vs UNcorroborated 200 -> can_merge = False
```

So the answer to the question as posed is **no** — two records where one carries a corroborated year and the other carries none still merge, because `_death_years_conflict` returns `False` the moment either side is `None`. That guard is untouched by this delta. What da#376 changes is the *population*, not the predicate: persisting the year instead of dropping it moves bio-matched records out of the blind spot, so the guard fires where it previously could not. That is a real improvement and it is the right diagnosis of why dropping the year was worse than keeping it.

Two things follow that are worth having on the record. The blind spot is not closed, only shrunk: in a 2,500-surface sample against the production bios, 32.2% of surfaces have no bio match at ≥0.70 at all, so their `death_year_ah` stays NULL and every pairwise merge involving them is decided without the death-year guard. And the guard is provenance-blind — the fourth line above shows an `uncorroborated` 200 vetoing a merge against a `corroborated` 161 with exactly the authority a corroborated year would have. `death_year_provenance` is written by `_upsert_canonical`, declared in `NARRATORS_CANONICAL_SCHEMA`, and asserted in two tests; no production code reads it. Up to 20.8% of sampled surfaces take a year from a non-exact (fuzzy or crossref) match and would carry that tag.

I am not blocking on it, and I want to be explicit about why rather than leave it implicit. Relative to `main` — which is the merge target — `_upsert_canonical` already attached candidate years unconditionally, so `_death_years_conflict` already consumed unverified years. This delta does not regress that; it adds the tag that would let a future guard tell the two apart. The clean follow-on is three lines in `_death_years_conflict` to require both years corroborated, or to ignore an uncorroborated one, which is what would make the new column load-bearing instead of documentary. Your `run()` comment is accurate that an uncorroborated year may not re-enter `death_year_index`; it is silent on the fact that the same year still steers `_can_merge`, and that silence is what I would close.

## 4. The red count is 8, not 7 — and the current file cannot produce either number

You corrected 12 → 7 unprompted and in the right direction. The number is now one short, for a reason that is not your error.

The 17-test file I reviewed at `47fa5a9` produced 12 reds against `origin/main`, five of them `AttributeError` on the absent `_bio_corroborated`, giving 7 behavioural. At `ed3e1fc` the file has **18** tests — `TestMononymSplitPreserved::test_mononym_split_keys_on_the_mention_not_the_matched_bio` was added after my review — and against `origin/main` it produces 13 reds. Classified by exception type rather than by arithmetic:

```
5 × AttributeError   (TestBioCorroboratedUnit — symbol-only)
4 × AssertionError   } 8 behavioural
4 × KeyError         }
```

So the honest number for the branch as it stands is **8 behavioural reds, 5 symbol-only, 5 green controls**. Worth stating plainly: the head file cannot be run against `origin/main` at all — line 43 imports `canonical_surface`, which does not exist there, so the module fails at collection and *no* red count derives from the current file. Any count in the PR body is necessarily a statement about the da#356 subset at an earlier sha, and should say so.

Separately, I checked whether the da#376 tests are genuine regression guards or merely new-symbol tests, using only symbols `origin/main` has. `make_canonical_id` exists there:

```
  [main] ابي هريره / ابا هريره / ابو هريره   -> 3 distinct ids   (fix requires 1)
  [main] جابر بن عبد الله vs جابر بن عبدالله -> DIFFERENT ids     (fix requires SAME)
```

Both go red on `main` behaviourally. The fold tests are real guards. End-to-end at head, the three case forms produce one canonical node carrying `mention_count = 3`.

## On the numbers I did not re-derive

I did not touch the lexical proxy, and I did not re-derive 4,547. Your point about `الحسين`/`الحسن` is correct on its face — Ḥusayn and Ḥasan are different men, `_can_merge` refusing them is the guard working, and a proxy that counts that as a duplicate is measuring its own defect.

The `+14,146 new orphans` figure is **not mine** — the string does not occur anywhere in my review of `47fa5a9`, which made no claim about orphans or da#352. I have nothing to retract and I will not defend a number I did not produce; whoever authored it should re-derive it against the production predicate or withdraw it, on exactly the standard being applied to the rest of this.

I also tested, and refuted, a suspicion of my own before bringing it: folding `alias` looked like it would erase the variant surfaces the da#356 docstring promises stay searchable. It does — `aliases` is `[]` for the three Abū Hurayra forms at head — but it is `[]` at `ed3e1fc` too, so the `alias=` parameter was already dead on this path and the delta neither caused nor worsened it. It is in the debt line, not in the findings.

`_MIN_SHARED_TOKENS = 2` stands and da#376 stays open for the residual; `count_unblockable` and the `cluster_records_unblockable` warning are the right shape — a record that was never scored no longer looks like one that was scored and refused, which is the same silent-zero family as da#309 and worth naming as such in the code. Approving this releases da#356 and da#376 together, and I am satisfied both are sound.

---
Requestor: Oyunbileg Batbayar
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: `death_year_provenance` is written but never read — `_death_years_conflict` and `_genders_conflict` are provenance-blind, so a year sourced from a non-exact bio match (≤20.8% of surfaces in a 2,500-surface sample) is read as merge-refusing evidence exactly as a corroborated one is, while the ~32.2% of surfaces with no bio match keep `death_year_ah` NULL and leave that guard blind (it returns `False` when either side is `None`); separately the `alias=` argument to `_upsert_canonical` is dead on the non-split path (`alias == norm_name` after the fold, so nothing is appended) — verified identical before the delta, so pre-existing, not introduced here.

parametrization:
## Second review — verified at head `27d20b1`

I reviewed this independently of Oyunbileg. I agree with his three substantive conclusions, I get a different number than he does on the mint-site count, and I found one thing neither of us had: **a third fold rule with no test coverage at all.**

### The blocker: `_fold_token`'s accusative rule is unguarded

`canonical_surface` folds through three rules in `_fold_token`. I mutated each one independently, disabling it and running the suite at head:

| mutant | suite result |
|---|---|
| kunya folding disabled | 2 failed |
| glued `عبدال` split disabled | 1 failed |
| **accusative-stem strip disabled** | **30 passed** |

The accusative rule can be deleted outright and this PR stays green. It covers a closed lexicon of 25 stems — among them `انس`, `علي`, `مالك`, `نافع`, `معمر`. `انس` is the exact under-merge in da#347, and `علي` is a top-20 chimera. This is the rule that keeps `انسا` and `انس` on one node.

The reason it survives is that every existing fold test is in the *negative* direction or is idempotence-only, and both are satisfied by the rule's absence:

- `test_hazards_are_not_folded` asserts things that must **not** fold (`عمرا`/`عمر`, `زكريا`/`زكري`). Correct and valuable — it is what stops someone widening the stem set to include `عمر`. It cannot notice a rule that stopped firing.
- `test_fold_is_idempotent` includes `عليا`, but `canonical_surface("عليا")` is a fixed point whether it folds to `علي` or stays `عليا`.

Nothing asserts the rule *merges* anything. I verified a three-line addition closes it — red under the mutant, green at head, and the negative-direction guards stay green:

```python
@pytest.mark.parametrize(("acc", "stem"), [("عليا","علي"), ("انسا","انس"), ("مالكا","مالك")])
def test_accusative_forms_fold_onto_their_stem(self, acc: str, stem: str) -> None:
    assert make_canonical_id(acc) == make_canonical_id(stem)
```

I am asking for this rather than filing it because of what this PR is. It exists because a verification gate was trusted that could not fail. Your own body says it: *"A test that cannot fail is not evidence."* Shipping the identity function that gates a 7.5-hour resolve with a third of its fold silently unguarded is that same shape, one level down. The cost is three lines against a stage that costs 7.5 hours to redo.

### Body corrections (no code change)

The evidence section cannot be reproduced from this head, and one claim in it is false.

**The test file cannot be collected against `origin/main` at all.** Line 43 imports `canonical_surface`, which does not exist there. I ran it: `ImportError`, **0 tests collected** — not `12 failed, 5 passed`. That figure necessarily comes from an earlier sha of the file and should say so.

**"The da#376 guards were written to fail behaviourally on origin/main, using only symbols that already exist there (no new-symbol reds)"** is not true at head. Three of the five da#376 tests need symbols absent from `main`: `test_fold_is_idempotent` needs `canonical_surface`, and both `TestUnblockableRecordsAreCounted` tests need `count_unblockable` / the new `ClusterMetrics` kwarg.

The good news, and I checked this rather than assume it: with the import stubbed so the module loads, the substance holds and then some. I measured **21 failed / 9 passed** on `main`, of which **11 are behavioural** — including all three of the genuinely behavioural da#376 guards (`test_abu_hurayra_three_case_forms_mint_one_id`, `test_glued_abd_allah_matches_spaced`, `test_every_id_producer_agrees`). Your body claims 7. You undercounted, in your own disfavour. The mechanism is real; the presentation of it is not reproducible.

### Where I disagree with Oyunbileg

He reports eight minting sites across nine call sites. My AST walk over `src/` finds **twelve** call sites of the three names, of which two are wrapper-internal delegation (`identity.py:268`, `disambiguate.py:684`), leaving **ten in seven files**. And the mint/lookup split is softer than "one lookup": `bio_promote.py:101` is a `canonical_map` lookup (he is right about that one), but `date_reconcile.py:457` computes a join key into `obs_by_cid`, `load_edges.py:732` resolves an existing endpoint, and `disambiguate.py:1447` mints candidate ids into the ambiguous-narrators report rather than assigning a canonical identity. Roughly six are true mints.

None of that changes the conclusion, and the invariant that actually matters is one I verified directly and which holds: **there are zero `uuid5` narrator-id constructions anywhere in `src/` outside `identity.py`**, and no `f"nar:..."` string building. Every producer routes through the folded mint. `test_every_id_producer_agrees` is misnamed for what it does — it proves the two mint functions agree with each other, not that the producers call them — but the AST evidence supports its docstring.

### Verified, and a finding I withdrew

- **`_death_years_conflict` is genuinely untouched.** The diff confirms it; `fuzzy_cluster.py` changes only the metrics and the unblockable counter. It still returns `False` when either year is `None`, so `corroborated year vs no year` remains a permitted merge. da#376 changes the population, not the predicate. Agreed.
- **`death_year_provenance` is unread in production.** Confirmed — the only reads in `src/` are inside its own writer. Filed as #380 rather than blocking, because against `main` the column did not exist to be read.
- **Withdrawn.** I thought `disambiguate` and `bio_promote` could mint different ids for one person, since `bio_promote:188` pre-applies `clean_narrator_name` and `disambiguate` does not. I fed both paths polluted surfaces and they diverged. Then I traced the input: `ner.py:89` and `:178` drop a mention outright when `clean_narrator_name` returns `None`, and I probed every rejection path — it never returns `""`, only `None`. So no mention row ever carries an uncleaned `name_normalized`, the `or name_raw` fallback at `disambiguate.py:1307` is dead for parquet input, and both paths land on `fold(cleaned)`. Not a defect.

### Merge ordering

This conflicts with #359 on `src/parse/identity.py` — you add the `canonical_surface` import next to the `logger` import that #359 deletes. Whoever rebases second must keep the `canonical_surface` import; taking #359's side of that hunk wholesale leaves `make_canonical_id` raising `NameError`. Both PRs also commit regenerated `ontology/structural/{llms.txt,code-graph.json}`, which conflict by construction — regenerate after the second merge rather than hand-resolving.

Approve as soon as the fold test lands. The de-keying itself is right, and the corroboration gate is the correct shape.

---
Requestee: Ivana Horvat
TechDebt: #380

parametrization:
## Second review, round 2 — verified at head `333c4f9`

Approved. The blocker is closed, and I reproduced the fix rather than reading the table.

`src/` is **byte-identical** between `27d20b1` and `333c4f9` (`git diff --quiet 27d20b1 333c4f9 -- src/`), so everything I verified about the production code last round carries unchanged: the de-keying at `disambiguate.py:1373`, the corroboration gate, `_death_years_conflict` untouched, both mint paths folding, and zero `uuid5` narrator-id constructions anywhere outside `identity.py`. This delta is one test plus the regenerated structural index.

### The mutation table reproduces

Run independently at `333c4f9`:

| mutant | result |
|---|---|
| delete accusative strip | **6 failed**, 30 passed |
| delete kunya fold | 2 failed, 34 passed |
| delete `عبدال` split | 1 failed, 35 passed |

The six reds are exactly the six new parametrizations, nothing else. And under the accusative mutant, `test_hazards_are_not_folded` and `test_fold_is_idempotent` run **5 passed** — green, as you said. That is the proof, not the anecdote: those two guards were structurally incapable of noticing, because a deleted fold rule leaves `canonical_surface("عليا")` a fixed point and violates no negative assertion.

### The three checks

**1. The stems that matter are pinned.** Read from the `parametrize` list, not the prose: `عليا`/`علي` at the first case and `انسا`/`انس` at the second, annotated `a top-20 narrator` and `da#347's exact under-merge case`. Both present. The per-case `mentions` string carrying the ratio is a good touch — it makes the failure output say *why the case was chosen*, which is the thing that rots first.

**2. Both directions asserted, per case.** `canonical_surface(accusative) == stem` and `make_canonical_id(accusative) == make_canonical_id(stem)`, both in the test body, so both run for all six. Strictly the id assertion is implied by the surface one — `make_canonical_id` is `uuid5 ∘ canonical_surface` and the surface is idempotent — but it is the right belt-and-braces: it stays meaningful if `make_canonical_id` ever stops folding, which is precisely the da#376 failure mode that motivated putting the fold inside the mint.

**3. Nineteen of twenty-five stems remain unpinned. I accept the line, and here is exactly what it buys.**

I measured the residual instead of asserting it. Removing a single **unpinned** stem (`محمد`) from `_ACCUSATIVE_STEMS` leaves the suite at **36 passed** — survives. Removing a **pinned** stem (`علي`) reds 1. So coverage is per-stem for the six, and the *branch as a whole* is pinned by any one of them.

Nobody should read "mutation-verified" as "all twenty-five verified." Six are load-bearing; nineteen are asserted only by the code that implements them.

And a correction to the advice I would have given: I assumed the cheap total fix was a two-line loop over the lexicon —

```python
for stem in _ACCUSATIVE_STEMS:
    assert make_canonical_id(stem + "ا") == make_canonical_id(stem)
```

I tested it under the `محمد` mutant. **It is inert.** Iterating `_ACCUSATIVE_STEMS` is a tautology against stem-removal: the removed stem is no longer in the set being iterated, so the check reports zero unfolded stems and passes. It is the same "assertion no fixture can violate" trap, one level subtler — and it would have looked like *stronger* coverage than the six literals while pinning nothing at all.

Which means the literal, hand-written list is the **only** shape that pins a stem. Extending it is a matter of typing nineteen more pairs, not of finding a better assertion. Six highest-mention is a defensible place to stop, and I am not asking for the other nineteen.

### Body

The retraction landed and is more honest than it needed to be. It names the sha the stale figure came from (`47fa5a9`), states that it became unreproducible the moment the import was added, gives `no tests collected, 1 error` as the actual result against `origin/main`, and decomposes the re-measurement into behavioural versus symbol-absence. Correcting a number in your own favour is easy; this one moved against her.

### Accepted as-is

`test_every_id_producer_agrees` stays misnamed this round. Right call not to churn the head under review. I have recorded it on #380 with the two ways to fix it, since #380 is otherwise a `death_year_provenance` issue and this would have been lost in a review thread.

### Verified at head

Full tree `1869 passed, 7 skipped` — reproduced exactly. CI rollup: 12 checks, all `SUCCESS`, non-empty.

### Merge ordering — unchanged, and now sharper

Still conflicts with #359, and `src/parse/identity.py` has moved from a clean auto-merge to a **content conflict**. The hazard is the same and worth repeating for whoever resolves it: this PR adds `from src.utils.arabic import canonical_surface` beside the `logger` import that #359 deletes. Taking #359's side of that hunk wholesale drops the import and leaves `make_canonical_id` raising `NameError` — while both PRs' own suites stay green. Keep the `canonical_surface` import. The two `ontology/structural/` files conflict by construction and should be regenerated after the second merge, never hand-resolved.

Clean. Merge it.

---
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: #380

parametrization:
## Re-verdict at head `333c4f92fc7a4d155616cc30ed5169e6bfdfcb2b`

My `Approved` on this PR was posted at `22:07:40` against `27d20b1`; this head landed at `22:39:07`. A new comment rather than an edit, because an approval is a statement about a sha and an edited `updated_at` is exactly the field a naive freshness check would trust.

**It is clean, and the delta is the guard that was missing.** I re-derived the premise rather than taking it: `git diff --quiet 27d20b1 333c4f9 -- src/` confirms `src/` is byte-identical, the delta is `tests/test_resolve/test_canonical_identity_invariant.py` plus the regenerated structural index, and my review of the production code at `27d20b1` therefore stands untouched. The six new parametrizations pin exactly what they claim. Deleting `_fold_token`'s accusative branch at this head reddens **precisely those six and nothing else** — the other 30 tests pass under that mutation, which reproduces Alejandra's finding exactly and confirms that at `27d20b1` the rule could be removed with a fully green suite. The two assertions per case are the right ones: `canonical_surface(accusative) == stem` pins the fold, and `make_canonical_id(accusative) == make_canonical_id(stem)` pins that the fold reaches identity rather than stopping at the surface. Under the same mutant `test_hazards_are_not_folded` stays green, correctly — the absence of a rule cannot cause a false fold — and that is the structural reason the negative-direction tests could never have caught this. 36 pass at head.

The docstring gets the diagnosis right and states it better than I did: a rule whose deletion no test notices is not under test, and `test_fold_is_idempotent` is satisfied by the rule's absence because `canonical_surface("عليا")` is a fixed point either way. That is the same class as the `uuid5` parametrizations I flagged on `47fa5a9`, turned back on the fold itself. I checked one direction of this and missed the other; the mutants I ran proved the *hazard* guards discriminate, and no test I ran proved the *fold* was load-bearing.

Two observations, neither blocking, both in the debt line.

The guard pins the **rule**, not the **lexicon**. `_ACCUSATIVE_STEMS` holds 25 stems; the new test names 6. I removed `سعد` — one of the 19 unnamed — and all 36 tests still pass while `canonical_surface("سعدا")` silently stops folding to `سعد`. The rule's existence is now under test, which is what the finding was about; the membership of the frozen list is not. A single parametrization asserting `len(_ACCUSATIVE_STEMS) == 25`, or driving the list itself, would close that without adding 19 cases.

And the mention counts in the parametrize table reproduce no artifact in the tree. Measured against `data/curated/narrator_mentions_resolved.parquet` (3,249,721 mentions), the six accusative figures are within 1–4% of whole-surface counts (`عليا` = 3,134 against the cited 3,173), but the six stem figures sit within 2–4% of **token** occurrences and 2–9× above the surface count — `علي` is 14,030 as a normalized surface and 127,030 as a token anywhere in a name, against the cited 122,269. The pair therefore compares a surface count to a token count, and it flatters the rule in the direction of the argument: what this fold actually merges is surface `عليا` (3,134) into surface `علي` (14,030), a 22% addition, not a 2.5% shadow of a 122k node — and the ~113,000 token occurrences inside names like `علي بن ابي طالب` belong to other narrators the rule never touches. Nothing is asserted on these numbers; they are the message argument, shown only on failure. But they are what a future reader will quote when deciding whether the lexicon is worth maintaining, and mixing denominators there is the same failure the rest of this PR just spent two rounds correcting.

Neither costs a re-run. Approving; this is the right fixup and it was found the right way.

---
Requestor: Oyunbileg Batbayar
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: The new guard pins the accusative *rule* but not its *lexicon* — 6 of 25 `_ACCUSATIVE_STEMS` are named, and dropping any of the other 19 (verified with `سعد`) leaves all 36 tests green while that stem silently stops folding; and the six mention counts in the parametrize table mix denominators, citing whole-surface counts for the accusatives against token-occurrence counts for the stems (`علي` = 14,030 as a surface, 127,030 as a token, cited as 122,269), reproducing no artifact in the tree.

PR #367, opened 2026-07-09
fix(graph): chain_integrity reports a true cycle population, not a cap

### LINKED ISSUE #250: validate: chain_integrity cycle query intractable at prod scale (>40min, no timeout) — hangs loader
## Problem

The loader's post-load `chain_integrity` validation check does not complete on a
prod-scale graph. During the #723 prod reload (1,665,760 nodes / 8,707,661 edges)
it ran **>40 minutes on the single `chain_integrity` query without finishing** and
had to be stopped manually. The data load itself was unaffected (all nodes + edges
committed with 0 errors; validated independently via direct cypher), but the
loader's own self-validation is effectively unusable at prod scale — and because
there is no timeout, it hangs the loader container indefinitely (and any wrapper
polling it, e.g. `scripts/load_staging.sh`).

## Root cause

`queries/validation/chain_integrity.cypher`:

```cypher
MATCH path = (n:Narrator)-[:TRANSMITTED_TO*1..20]->(m:Narrator)
WHERE n = m
RETURN n.id AS narrator_id, length(path) AS cycle_length
LIMIT 100
```

A variable-length `[:TRANSMITTED_TO*1..20]` expansion from **every** `Narrator`
over ~2.5M `TRANSMITTED_TO` edges enumerates an enormous path space (up to 20 hops
out of each of 232k narrators). `LIMIT 100` bounds only the *output*, not the
*search* — Neo4j still expands the full neighbourhood before limiting. On the
smaller stage graph it completed (reported 100 cycles); at prod scale it does not
finish in any practical time.

Two compounding gaps:
1. **The query is intractable** as written (naive var-length cycle search).
2. **`src/graph/validate.py` has no per-query timeout** (`run_validation` →
   `client.execute_read(cypher_text)` runs unbounded), so a slow/divergent
   validation query hangs the whole loader instead of failing gracefully.

## Proposed fix

- Replace the naive expansion with a tractable cycle check — e.g. a much smaller
  bound (`*1..6` is plenty for isnad chains), or APOC/GDS cycle detection
  (`apoc.nodes.cycles` / GDS) that terminates deterministically.
- Add a **per-query timeout** in `validate.py` (Neo4j transaction timeout /
  `tx.run` with a deadline) so any validation query that exceeds it is recorded as
  `status=TIMEOUT` and the loader continues to the next check rather than hanging.

## Acceptance

- `chain_integrity` completes in seconds on a prod-scale graph (1.6M+ nodes).
- A pathological validation query cannot hang the loader: it times out, is logged,
  and the validation summary reports it without blocking completion.

## Context

- Surfaced during #723 prod reload, 2026-06-29 (meta noorinalabs/noorinalabs-main#723).
- Related: da#248 (transmission cycles — the data condition this check looks for),
  da#249 (validation harness multi-statement `.cypher` bug). This is a distinct
  third validation-robustness issue (query tractability + missing timeout).

parametrization:
## Problem

`queries/validation/chain_integrity.cypher` ended in a load-bearing, undocumented `LIMIT 100`, and `_classify_chain_integrity` classified on `len(rows)`. The gate could not report more than 100 and, on the full graph, always reported exactly `100 cycle(s) detected` — **a constant**. A gate that reports a constant is worse than no gate: it reads as reassurance. It is also the rc=1 that main#723 misread as a data defect.

The classifier additionally **passed on an empty result** (`_classify("chain_integrity", []) → passed=True`), so a query that silently returned nothing looked like a clean graph.

`queries/analysis/chain_integrity_mononym_ranking.cypher` had the same non-terminating shape with **no LIMIT at all**, so the "uncapped cycle total" its header promised was never obtainable from it.

## Measurement (reproduced against stg, not taken on trust)

Graph: `160,614` Narrator, `2,679,527` TRANSMITTED_TO.

| Quantity | Value |
|---|---|
| self-loops (length 1) | **0** |
| reciprocal pairs (length 2), distinct | **23,139** |
| narrators in a reciprocal pair | **5,803** |
| reciprocal *edge instances* | 295,576 |

Two traps hit and defused while measuring, both recorded here because they nearly produced a wrong number:

1. **`count(*)` is not the pair count.** `MATCH (a)-[:TRANSMITTED_TO]->(b) WHERE a.id < b.id AND EXISTS{...} RETURN count(*)` returns **295,576** — edge *instances* (one TRANSMITTED_TO per hadith), not pairs. `count(DISTINCT [a.id, b.id])` returns 23,139.
2. **A missing row is not a zero.** The uncapped `*1..2` form emitted **no rows** and was killed at 150s: `UNCAPPED_1_2_RC=124`. Read from stdout alone that is indistinguishable from "0 cycles". Every remote query in this PR was run under a terminal `printf 'QUERY_RC=%s\n' $?` sentinel; absence of the sentinel is treated as "did not complete."

Instrument verified live before any zero was believed: the same shell returned `count(n)=160614` and `count(t)=2679527` immediately before the `self_loops=0` reading.

## Change

`chain_integrity.cypher` is now a **single-statement census with no LIMIT**, returning one aggregate row (`self_loops`, `reciprocal_pairs`). ~109s at full stg scale, inside the 300s per-query budget (da#259); an overrun downgrades to a non-fatal WARN rather than hanging.

Two signals, previously conflated, are now separated:

- **`self_loops` is the GATE.** A narrator transmitting to himself is unambiguously corrupt. It is `0` on stg and prod today, so this guard is live and non-vacuous rather than decorative.
- **`reciprocal_pairs` is a METRIC and never gates.** "A taught B and B taught A" is historically impossible, so every pair is manufactured upstream by identity collapse in `resolve` (da#356). It is a **thermometer for over-merge**, not a loader defect — to be **re-measured after da#356 lands**, not fixed here, and certainly not deleted.

Coverage is stated out loud in both the query header and the emitted `details` string: **lengths 1 and 2 exact; length >= 3 NOT measured.** Enumerating length >= 3 means the variable-length expansion, which does not terminate at this scale. The limit was not raised — that would have preserved the lie at a larger number.

The classifier no longer passes on `[]`: the query always returns exactly one row, so no row means the census did not run.

The ranking query is rewritten to a terminating degree-bounded form (~110s) that names the worst over-merge offenders — useful input for da#356:

| canonical name | reciprocal partners | mention_count |
|---|---|---|
| عبد الله | 829 | 19,412 |
| سفيان | 422 | 28,323 |
| محمد | 419 | 5,736 |
| شعبة | 360 | 29,280 |
| الحسن | 304 | 11,768 |

The single سفيان node merges Sufyān al-Thawrī (d. 161 AH, Kufa) with Sufyān ibn ʿUyayna (d. 198 AH, Mecca) — exactly the case `src/resolve/mononym_split.py` documents.

## Tests — shown RED against `origin/main` first

Run at `origin/main` (`2cf5968`) with only the test file changed:

```
FAILED TestChainIntegrityClassification::test_query_carries_no_limit_clause
        - AssertionError: chain_integrity.cypher must not cap its own result
FAILED TestChainIntegrityClassification::test_clean_census_passes_and_reports_exact_counts
        - AssertionError: assert False is True
FAILED TestChainIntegrityClassification::test_reciprocal_pairs_alone_never_gate
        - AssertionError: assert False is True
FAILED TestChainIntegrityClassification::test_missing_row_is_fatal_not_a_silent_pass
        - AssertionError: assert True is False
FAILED TestChainIntegrityClassification::test_details_disclose_the_uncovered_cycle_lengths
        - AssertionError: assert False is True
FAILED TestBoundedValidation::test_fast_graph_validates_normally
        - AssertionError: assert 'FAIL' == 'PASS'
FAILED TestMultiStatementExecution::test_single_statement_file_still_executes_exactly_once
        - AssertionError: assert 'FAIL' == 'PASS'
7 failed, 58 passed
```

The old classifier's own log line during that run is the clearest statement of the bug:

```
validation_complete  details='1 cycle(s) detected'  query=chain_integrity  rows=1  status=FAIL
```

It reads the one-row census as "1 cycle" — it was counting rows, never cycles.

Honest accounting of that list: five are new tests that are genuinely red for the reason under test. The last two are **pre-existing** tests that turn red once the fixture returns the *real* query's row shape — that is the finding, not collateral: the old classifier cannot correctly classify an honest query. Two further new tests (`test_self_loop_is_the_gate_and_is_fatal`, `test_malformed_row_is_fatal`) pass on main coincidentally, because `len(rows)==1` also yields FAIL there; they are kept as regression guards but are **not** claimed as red.

`test_registered_classifier_pass_on_empty` is parametrized over the whole registry and asserted "empty rows ⇒ pass" for **every** classifier — it encoded the vacuous pass. `chain_integrity` is now excluded with the reason inline; the convention does not hold for an aggregate query.

Fixtures use the real query text and real Arabic (`_write_query` now mirrors the shipped census verbatim).

## Verification

- `ENVIRONMENT=test uv run pytest tests/ -q` (whole tree, no `-m` filter): **1837 passed, 7 skipped**
- `uv run ruff format --check src/ tests/` → `FMT_RC=0`; `ruff check` → `LINT_RC=0`; `mypy` clean
- `python3 scripts/structural_ontology.py check --require-generator` → `CHECK_RC=0` (index regenerated)
- Candidate query executed against the real stg graph before it was written into the repo

## Not done here, deliberately

`src/resolve/mononym_split.py:19` references this artifact in a docstring and is now slightly stale, but `src/resolve/**` is owned by da#356/da#309 — not touched.

Closes #250
Part of #248


parametrization:
## Architecture + data-integrity review — chain_integrity honest census (closes da#250)

Reviewed at head `72a9647`, reading the code via `gh pr diff` / `gh api contents`.

**The fix is correct and the tests exercise it.** The old gate paired `MATCH … [:TRANSMITTED_TO*1..20]->(m) WHERE n=m … LIMIT 100` with a classifier that counted returned rows, so on the full graph it reported the constant `100 cycle(s) detected` — a cap read as a measurement — and it also passed on an empty result (a zero nobody measured). Both failure modes are now closed:

- `queries/validation/chain_integrity.cypher` is rewritten as two `CALL {}` aggregation subqueries returning exactly one row (`self_loops`, `reciprocal_pairs`), with no `LIMIT`. Two aggregating subqueries each yield one row even on empty input, so the composed result is always one row — the "grouped aggregation over empty input returns zero rows" trap is avoided by construction.
- `_classify_chain_integrity` now treats `len(rows) != 1` as fatal ("an absent row is not a zero"), validates the two integer columns (malformed row is fatal), gates **only** on `self_loops == 0`, and reports `reciprocal_pairs` as an explicitly-labelled metric. The coverage string states out loud that cycles of length ≥ 3 are NOT measured — a PASS cannot be misread as "acyclic".

**Termination reasoning is sound.** The variable-length enumeration genuinely does not terminate at stg scale (even `*1..2` killed at 150s); measuring lengths 1 and 2 exactly and declaring the gap is the right call over capping-and-presenting-the-cap.

**Gate/metric split is architecturally right.** A self-loop is unambiguously corrupt (load-fixable); reciprocal pairs are manufactured upstream by identity collapse in `resolve` (da#356), so failing the load on them would punish the loader for someone else's merge. Not-gating-on-metric is correct, and the memory records the re-measure-after-da#356 intent.

**No regression to the da#354 rc-separation table.** The runbook union kept main's rc 0/1/4 table intact and only swapped the old `100 cycle(s)` bullet for the honest-count bullet; the `MEMORY.md` union is a clean single-line addition. No conflict markers survive in either file. The mononym-ranking diagnostic (`queries/analysis/…`) was correspondingly rewritten to a terminating degree-bounded form and stays out of `queries/validation/` so it never becomes a gate.

**Test quality.** `test_query_carries_no_limit_clause` masks line comments before scanning, so the prose that *discusses* the removed `LIMIT 100` isn't mistaken for a live clause. `test_missing_row_is_fatal_not_a_silent_pass`, `test_malformed_row_is_fatal`, and `test_details_disclose_the_uncovered_cycle_lengths` each pin a distinct real behaviour. `chain_integrity` is correctly excluded from the "empty → pass" parametrization since an empty result now means the census did not run. `_write_query` mirrors the shipped census. CI 12/12 green.

Sound. Approving.

---

RequestOrReplied: Approved
TechDebt: none

parametrization:
Independent second review (data-correctness + test-rigor lens). Read at head `72a9647` via `gh pr diff` and `gh api contents`.

**The fix reports a real population, not a cap.** The old `chain_integrity.cypher` ended in `LIMIT 100` and `_classify_chain_integrity` counted `len(rows)`, so on the full graph it emitted the constant `100 cycle(s) detected` — a cap masquerading as a measurement, and the rc=1 main#723 misread. The replacement is a single-statement census (two `CALL {}` aggregation subqueries, no LIMIT) returning one row of `self_loops` and `reciprocal_pairs`. An aggregate with no grouping always returns exactly one row, which is why the classifier's `len(rows) != 1` correctly reads "the census did not run" rather than "clean graph" — the vacuous pass-on-`[]` the predecessor had is gone.

**The gate/metric split is right.** `self_loops == 0` is the only thing that gates; `reciprocal_pairs` is reported but never fails the load, correctly attributed to upstream over-merge in resolve (da#356) and left for re-measurement, not deletion (da#248). Coverage — lengths 1 and 2 exact, length >= 3 not measured — is stated in both the query header and the emitted `details`, so a PASS cannot be misread as "acyclic".

**Silent-zero family defused.** This is exactly the failure class the prior sessions kept hitting, and each trap is closed: an absent row is fatal (not a zero), a malformed row is fatal, and the ranking query was rewritten from the non-terminating `*1..20` form to a degree-bounded reciprocal-partner count so its promised "uncapped total" is actually obtainable.

**Tests genuinely bite.** I confirmed `_mask_line_comments`, `is_fatal`, and the 300s WARN-downgrade all exist at head. `test_query_carries_no_limit_clause` masks comments first (the header *quotes* the removed `LIMIT 100`), so it checks live clauses, not prose. The five new tests were shown red on `origin/main`, and the fixture `_write_query` mirrors the shipped census verbatim rather than a toy shape — so the two pre-existing tests that flip red are the finding (the old classifier can't classify an honest query), not collateral. `test_registered_classifier_pass_on_empty` correctly excludes `chain_integrity` with an inline reason. `_split_statements` treats the semicolon-free census as one statement, matching the driver's one-statement rule.

CI is CLEAN at head (Test + Integration Tests + sync-drift gate all green). No defect found; the fix is correct and the tests exercise it.

---

RequestOrReplied: Approved
TechDebt: none

PR #372, opened 2026-07-09
fix(graph): separate load-failure from validation-findings exit codes

### LINKED ISSUE #354: bug(loader): 'load' exits rc=1 for validation-only findings — indistinguishable from a genuine load failure
## Problem

`isnad-ingest load` exits `rc=1` when the load itself fully succeeded and only the post-load **validation** flagged findings. The CLI gives the caller no way to distinguish:

- **the load failed** — data did not land, the graph is untouched or half-written; and
- **the load succeeded and mutated the graph**, and validation then reported known/expected findings.

Both surface identically: a red GitHub Actions run and `ERROR: [stg] batch load 'load' failed (rc=1).`

Observed on stg, run 29035731215 (2026-07-09). The graph was fully and correctly mutated — 1,563,783 nodes, 235,039 new edges, `Narrator created=10427 merged=118807` — and the run is marked **failure** because three validation checks were red:

```
[FAIL] chain_integrity: 100 cycle(s) detected                       (see da#248)
[FAIL] orphan_narrators: 70855 orphan narrator(s) found             (see the orphan-population issue)
[FAIL] transmitted_to_hadith_ref: 3 ... (da#325)
```

## Why it matters

This conflation is actively dangerous, and we nearly walked into it. Project memory records loader `rc=1` as "non-blocking, it's the da#319 validation-harness artifact." Under that prior, the correct reflex on seeing `rc=1` is to shrug. If a future load genuinely fails — bad credentials, partial write, wrong artifact — it produces the same red run and the same `rc=1`, and the standing interpretation says ignore it.

A red signal that is known to be red for benign reasons stops being read. That is the failure mode.

## Proposed fix

Separate the exit codes:

- `0` — load succeeded, validation clean.
- `1` — **load failed** (reserve for genuine load failure).
- distinct non-zero (e.g. `2`) — load succeeded, validation reported findings.

Then `deploy-data-load.yml` can treat the validation code as a warning annotation on a green run, and a red run once again means the load did not happen. Related: `strict=False` is already passed at the `load_all` call site, so the loader has already decided validation findings are non-fatal to the load — the exit code should say the same thing.

parametrization:
## Problem

`load` exited `1` both when the load genuinely failed and when the load **succeeded** but post-load validation reported findings. From the exit status alone an operator cannot tell which happened.

This is not hypothetical. main#723's `rc=1` was read as a data defect when the load had in fact succeeded and the non-zero came from a validation-harness `CypherSyntaxError` (da#319) plus a benign orphan tail. `docs/PROD_RELOAD_RUNBOOK_723.md` went as far as documenting an **"Expected `rc=1` false alarm"** — the definition of a gate nobody can act on.

## Change

The load result and the validation verdict are independent facts, so they get independent exit codes — **sourced from the da#384 registry**, not declared here:

| rc | member | meaning |
|----|--------|---------|
| `0` | — | load succeeded, validation clean (downgraded warnings allowed) |
| `1` | `ExitCode.LOAD_FAILED` | **the load failed** — the graph was not fully written |
| `5` | `ExitCode.VALIDATION_FINDINGS` | **the load succeeded**; validation has findings |

`_cmd_load` wraps the load in an explicit failure path that prints the traceback and exits `LOAD_FAILED`, reaching **neither the manifest write nor the audit entry** — a load that raised did not happen. Previously the exception escaped as a bare traceback, which also happened to be `rc=1`.

Post-load bookkeeping (`save_manifest` / `create_audit_entry` / `write_audit_entry`) is wrapped in its own handler. A bookkeeping failure leaves the rc alone rather than turning a committed load into `rc=1` — Ivana Horvat's blocker.

## Behaviour changes, stated explicitly

Two commands are touched, and only one changes its exit code.

**`isnad-ingest load` — exit code changes.** A findings load exited `1` on `main`; it now exits `5`. Anything that treats non-zero as "the load failed" must be updated. This is the point of the PR.

**`isnad-ingest validate` — exit code UNCHANGED; message text changes.** Relative to `main`, `_cmd_validate` already exits `ExitCode.VALIDATION_FINDINGS` (`5`) — da#384 Amendment G routed it. This PR does not move it. It replaces `WARNING: Some validation checks failed.` with a message that names the failing checks and says a finding is not a load failure.

> Relative to *this branch's pre-rebase head* the `validate` exit moved `4 -> 5`. That is a statement about a tree that no longer exists. Against `main` — the only baseline a merge changes — the code is unchanged. Read `validate` for the message, not for the number.

## The rebase found a rival declaration, not a duplicate

This branch carried its own constants in `src/graph/__init__.py`:

```
EXIT_LOAD_FAILED = 1
EXIT_VALIDATION_FINDINGS = 4
```

Since da#387 landed the registry, `4` is `MISSING_DEPENDENCY` (da#309). This is not a race between two PRs to be sequenced — it is a contradiction no merge order fixes. Both bindings are **deleted**; `src/graph/__init__.py` is now byte-identical to `origin/main`.

`TestRegistryIsTheSoleDeclarer` reded on that block. It was not exempted and the constants were not renamed. The cause was removed.

## Two self-expiring pins expired, and are renamed in the commit that falsifies them

**`test_the_live_tree_has_exactly_one_unnamed_exit_and_da372_owns_it`** -> `test_the_live_tree_has_no_unnamed_exit`. It pinned `offenders == {"_cmd_load": ["1"]}` and promised to red when da#372 routed it. It did.

`assert not offenders` alone is the weakest assertion in the file — an empty set is also what a scan that read nothing returns. Two instrument guards are therefore inline: the scan must have read >20 files, and the detector must still red on a planted `sys.exit(9)`.

**`test_enrich_exits_bare_one_after_the_load_has_committed`** -> `test_enrich_leaks_rc_one_after_the_load_has_committed`. da#384 Amendment I gave enrich `ENRICH_FAILED` (`7`), so the bare `1` it asserted is gone. **The leak it documented survived its own cause**: enrich's audit write sits outside any handler, and neither `main()` nor `_cmd_pipeline` has a top-level one, so an escaping exception exits `1` via CPython's default handler — after the load has committed. Filed as **da#394**. The scoped comment in `_cmd_load` now names that mechanism rather than an exit code someone can renumber.

## Prose corrected against the tree

- `src/cli.py`, `_check_neo4j`'s docstring — said "Exits with code 1". It exits `DB_UNREACHABLE` (`8`).
- `src/exit_codes.py`, the module docstring's "Call sites still emitting a bare integer" section — said the emitters of `1` would be `_check_neo4j` **and** `_cmd_load`'s genuine failure, "true for `load` *and* for `pipeline`". Both halves were wrong.

An AST walk of the **rebased** tree, not a forecast about it:

```
EMITTERS OF 1  (count=1):
  src/cli.py  sys.exit(ExitCode.LOAD_FAILED)   # _cmd_load, in the except around load_all

statically unresolved: src/tools/duck.py:340  SystemExit(main())
  -> main() returns 0 on all three paths; not imported by cli.py or src/pipeline/
```

A **singleton**, not an empty set. `_cmd_load`'s genuine-failure path fires from inside the `except` around `load_all`, before any manifest is written — so `rc=1 implies an unwritten manifest` holds for `load`, and is false for `pipeline` for the da#394 reason.

## Verification — exit codes read from real processes

`ExitCode.VALIDATION_FINDINGS == 5` is a statement about a Python object. Under a plain `enum.Enum` that assertion still passes while `sys.exit(member)` prints the repr and exits `1`. Isolated — same assertion, same `sys.exit`, differing only in the base class:

```
intenum     value-equality assertion: PASSED   ACTUAL PROCESS EXIT STATUS = 5
plainenum   value-equality assertion: PASSED   ACTUAL PROCESS EXIT STATUS = 1
            stderr: E.VALIDATION_FINDINGS
```

So every code below is `$?` from an executed process, at `95c2989`:

```
SCENARIO                     ACTUAL PROCESS EXIT STATUS
load_clean                   0
load_findings                5
load_genuine_failure         1
load_bookkeeping_fails       5     <- Ivana Horvat's blocker
validate_clean               0
validate_findings            5
enrich_steps_failed          7
enrich_audit_write_fails     1     <- da#394, not fixed here
```

## Mutation evidence for the emptied pin

Run through `scripts/mutation_guard.sh`, re-run at `12669f0` (the head under review). Two-sided, because a guard that cannot go red has proved nothing:

| mutant | result |
|--------|--------|
| M1 blind the detector (`_unnamed_exits_by_scope -> {}`) | **CAUGHT** — `the detector has gone blind; saw {}` |
| M2 new bare `sys.exit(9)` | **CAUGHT** |
| P1 `sys.exit(1 if a else 2)` | **CAUGHT** |
| P2 `return 3` in a module-local callee | **CAUGHT** |
| P3 `def _die(EXIT_CODE: int): sys.exit(EXIT_CODE)` | **CAUGHT** |
| P4 self-recursion `def f(): return f()` | **CAUGHT** — via the assertion (`found: {'_p4_exit': ['_p4()']}`), not a `RecursionError` |

Under **M1**, `offenders == {}` and `assert not offenders` **passes**. Only the instrument guard reds. That is the proof that the bare assertion is vacuous and the guard is load-bearing.

Every run gated by `--expect-count 1` (a selector matching nothing returns rc=5 and reads as green) and `--expect-text` (a plant that never applied and a surviving mutant are the same green).

## Verification

- `ENVIRONMENT=test uv run pytest -q` (whole tree) -> **1951 passed, 10 skipped, rc=0**
- `ruff@0.15.11 format --check` -> rc=0; `ruff check` -> rc=0; `mypy src/cli.py src/graph/__init__.py src/exit_codes.py` -> rc=0
- `scripts/structural_ontology.py emit` -> header moved `251/3126/4912` -> `252/3151/4969`; staleness gate passed. The index was **regenerated**, never hand-merged.

> Bare `python3 -m pytest` fails `tests/test_tools/test_duck.py` with `ModuleNotFoundError: duckdb`. That is the wrong interpreter, not a defect — `duckdb` is a project dep, and `uv run pytest` is green.

Runbook updated: the "expected false alarm" section is replaced by the exit-code table.

## Note for reviewers

Both Changes-Requested comments cite **`3f55c74a`**, and I originally told you to diff from it. **That instruction was wrong and I am withdrawing it.**

`3f55c74a` is orphaned, and so is `797e302`, the "twin" I offered as a safe substitute:

```
git merge-base --is-ancestor 3f55c74a 12669f0 -> rc=1   orphan
git merge-base --is-ancestor 797e302  12669f0 -> rc=1   orphan
git merge-base --is-ancestor d695104  12669f0 -> rc=0   live  (main)
git merge-base --is-ancestor 203a9c3  12669f0 -> rc=0   live  (the real twin)
```

The blob-identity check I ran (`cli.py` = `b938e3c7cb`, `graph/__init__.py` = `7a613ba707` at both shas) was **true**, and load-bearing for nothing: both shas are unreachable from this head, so a window rooted at either sweeps in `main`'s landings as though they were mine.

```
3f55c74a..12669f0   32 files
797e302..12669f0    18 files
d695104..12669f0     7 files   <- what this PR does
```

**Diff `d695104..12669f0`.** Seven files:

```
docs/PROD_RELOAD_RUNBOOK_723.md
ontology/structural/code-graph.json
ontology/structural/llms.txt
src/cli.py
src/exit_codes.py
tests/test_exit_codes.py
tests/test_graph/test_load_exit_codes.py
```

I verified a premise and never checked the ref it hung from. A true premise under a false instruction is worse than a false one, because checking it feels like diligence.

A validation-harness error (e.g. a `CypherSyntaxError` in a `.cypher` file) and a genuine data finding both surface as `is_fatal` and therefore both yield `5`. Both mean "the load is in the graph; validation is not clean," which is the distinction #354 asks for. Splitting *those two* further needs a new field on `ValidationResult` and is deliberately not done here.

Part of da#392
Closes #354


parametrization:
## Review of `3f55c74a58e315a60d43a62f2dc1141930eae4d1`

Read at head via `gh api contents?ref=3f55c74a`, never checked out. Every claim below was executed, not inferred: the tree was extracted read-only and the tests run against it.

The change is right and the reasoning behind it is right. `rc=1` did conflate "the write failed" with "the write succeeded and the report has findings", and that conflation is exactly how main#723's rc was misread. One blocker, and it is in the load-bearing claim rather than in the mechanics.

### Blocker — `rc=1` does not mean "the load did not happen"

The comment at `src/cli.py:222-225` states that the failure path *"reaches neither the manifest write nor the audit entry below — a load that raised did not happen."* That is true of exceptions raised **inside** the `try`. It is not true of `rc=1`.

`save_manifest(…LAST_LOADED…)` (`cli.py:234`), `create_audit_entry` (`:236`) and `write_audit_entry` (`:248`) all sit **outside** the `try`, and `main()` has **no** top-level handler — so an exception from any of them propagates to Python's default handler, which exits **`1`**. Identical to `EXIT_LOAD_FAILED`, from a load that *did* happen.

Demonstrated, unconfounded (a naive probe that patches `save_manifest` globally is confounded — it is also called at `:179`, *before* the `try`; this one raises only on the second call):

```
call order : ['load_all', 'write_audit_entry']
outcome    : UNCAUGHT OSError -> python default rc=1
last_loaded manifest written before the raise? True

call order : ['save_manifest:.manifest.json', 'load_all', 'save_manifest:.last_loaded_manifest.json']
outcome    : UNCAUGHT PermissionError -> python default rc=1
```

In the first case the graph is written, the load **is** recorded as loaded, and the operator is told `rc=1` — "it did not happen". That is the same false reading da#354 exists to eliminate, moved one step later and made worse: `rc=1` from a *pre*-commit failure is safe to retry; `rc=1` from a *post*-commit failure is not, and nothing distinguishes them.

This matters beyond ergonomics. #359 is being rebased onto this constants block precisely because `rc=1` *binds an invariant* — "the load raised; it did not happen". A leaky binding makes that reasoning unsound, and #359's malformed-id case is exactly a run-to-completion-then-fail shape.

**Ask, ~5 lines.** Wrap the post-load bookkeeping in its own handler that does not exit `1`. The graph is committed by then; a failed audit write is not a failed load. Either warn and exit `0`, or give it a code of its own — but `EXIT_LOAD_FAILED` must not be reachable from a path where the write succeeded. A test in the shape of the probe above pins it.

### Verified, and clean

**The exit-code partition holds.** Grepped every exit site at this ref rather than trusting the constants block:

| site | code |
|---|---|
| `cli.py:135` | `EXIT_STOPPED_AT_LIMIT` (3) |
| `cli.py:230` | `EXIT_LOAD_FAILED` (1) |
| `cli.py:282`, `:334` | `EXIT_VALIDATION_FINDINGS` (4) |
| `cli.py:313`, `:708` | `0` |
| `cli.py:33`, `:483`, `:766` | bare `sys.exit(1)` |
| `graph/validate.py:288` | inside a docstring, not code |

`EXIT_STOPPED_AT_LIMIT = 3` is genuinely live at `src/resolve/_checkpoint.py:54`, and those three are the only `EXIT_*` constants in `src/`. No collision.

**The non-collision test is not a tautology.** It does not derive its expectations from the collection under test, and it reds under mutation — I checked, because the same shape hid a dead rule in #363:

```
mutant: EXIT_VALIDATION_FINDINGS = 1  ->  2 failed  ("assert 1 != 1", "assert 1 not in (0,1,2,3)")
mutant: EXIT_VALIDATION_FINDINGS = 3  ->  1 failed  ("assert 3 not in (0,1,2,3)")
baseline                              ->  8 passed
```

**`test_clean_load_exits_zero` is a deliberate positive control** for the `not (tmp_path / LAST_LOADED…).exists()` assertion in the failure test — proving the negative is a real observation rather than an inert one. That is the trap that has bitten this repo repeatedly and you pre-empted it. Worth saying explicitly.

### Downstream consumer — the improvement does not reach the deploy path

`noorinalabs-deploy/.github/workflows/deploy-data-load.yml:370`:

```sh
${COMPOSE_LOAD} run --rm --no-deps graph-load ${LOAD_ARGS} || RC=$?
if [ "${RC}" -ne 0 ]; then
  echo "ERROR: batch load '${LOAD_ARGS}' failed (rc=${RC})." >&2
  exit "${RC}"
fi
```

Any non-zero is a failure. So a successful load with validation findings now exits `4`, and the deploy still aborts printing **"batch load failed (rc=4)"** — an operator reading the workflow log learns exactly what da#354 set out to stop them concluding. Not a regression (rc=1 aborted it too) and not this PR's file, but the reload runs through this workflow, so the fix is incomplete at the point where it is actually read. Needs a cross-repo change (`RC -eq 4` → warn, not fail) before the reload. Not filed by me.

### Smaller notes, no action required in this PR

- `_cmd_load` calls `_check_neo4j()` (`:156`), which exits a bare `1` (`:33`). Semantically consistent — the load did not happen — but it should be `EXIT_LOAD_FAILED` so the constants block is the whole truth rather than most of it.
- `validate-staging` exits a bare `1` on *findings* (`:766`) while `validate` exits `4` on findings. Same semantic class, two codes — the conflation da#354 removes, still present one subcommand over. Pre-existing.
- The runbook edit is accurate and the `LOAD SUCCEEDED (… nodes, … edges written)` line before an `rc=4` is the right operator affordance.

Once `EXIT_LOAD_FAILED` is unreachable from a committed load, this is an approve from me.

---
Requestor: Ivana Horvat
TechDebt: none

parametrization:
## Second review at head `3f55c74a58e315a60d43a62f2dc1141930eae4d1`

Read at that sha via the contents API and a throwaway detached worktree; no checkout in the shared clone. 12/12 checks green, `mergeable=clean`. Second reviewer after Ivana Horvat.

The design is right and the four things I was asked to attack all hold up. What blocks it is a code collision with another open PR that neither CI can see.

### The blocker: `4` is already claimed by an open PR

`EXIT_VALIDATION_FINDINGS = 4`. Open PR #362 (da#309, two approvals, awaiting merge, base `main`) adds `src/resolve/_deps.py` with **`EXIT_MISSING_DEPENDENCY = 4`**, and `_cmd_resolve` exits on it. Neither PR's tests can see the other, because at `3f55c74a` the file `src/resolve/_deps.py` does not exist, and on `origin/main` it does not exist either. Whichever merges second, `4` means two different things.

That is not a theoretical objection about separate processes, for two reasons. The first is that **this PR's own constants block treats the exit-code space as global**: "3 is already taken by `src.resolve.EXIT_STOPPED_AT_LIMIT`." `EXIT_STOPPED_AT_LIMIT` is a `resolve` code and `load` steps around it. By exactly that rule, `load` must also step around `EXIT_MISSING_DEPENDENCY`, which is likewise a `resolve` code. The PR's stated principle forbids what the PR does, the moment #362 lands.

The second is that the two are reachable from **one process**. `_cmd_pipeline` runs acquire → parse → resolve → load → enrich in a single invocation. After both PRs land, `isnad-ingest pipeline` exiting `4` means either "resolve is missing its ML dependency group" or "load succeeded and validation has findings" — an environment defect versus a clean load with a data finding, which is precisely the conflation da#354 exists to remove, reintroduced one level up.

This also matters right now for da#355: Nikolaos is taking `EXIT_MALFORMED_IDS = 5` on the strength of "4 is validation findings." Half of that is true. The other half is that 4 is also missing-dependency, so the space he is reasoning about has a live double-booking in it.

The cheapest correct fix is not to renumber twice. It is to make the reserved set **derivable instead of hand-written** — one module owning the pipeline exit-code space, imported by `src.resolve`, `src.graph` and `src.cli`. Then a second claim on `4` is a name collision at import time rather than a discovery in production.

### The guard that exists to prevent this cannot see it

`test_validation_findings_code_is_not_reserved` asserts `EXIT_VALIDATION_FINDINGS not in (0, 1, 2, EXIT_STOPPED_AT_LIMIT)`. That tuple is hand-written. It pins the members it names and is blind to every sibling constant nobody remembered to add — which is how `EXIT_MISSING_DEPENDENCY` walks straight past it. It is the same shape as the accusative-stem literals on da#376: six named, nineteen unpinned. Derive the reserved set by scanning the `EXIT_*` constants the CLI can actually emit, and this test reddens on the next collision instead of the one after it.

### The three things I was asked to verify, which are all fine

**The non-collision test can fail.** I did not take this on trust. Setting `EXIT_VALIDATION_FINDINGS = EXIT_LOAD_FAILED` reddens both tests in `TestExitCodesAreDistinct`; setting it to `3` reddens `test_validation_findings_code_is_not_reserved` alone and leaves `test_codes_do_not_collide` green, which is the correct discrimination. It asserts against literals and against the imported constant, not against a set-length or a loop over the module it tests. It is not the tautology.

**`3` genuinely belongs to `EXIT_STOPPED_AT_LIMIT`.** Re-derived rather than quoted: `src/resolve/_checkpoint.py:54` defines it, and importing `src.resolve.EXIT_STOPPED_AT_LIMIT` at this head evaluates to `3`.

**Every exit path in `src/`, not just the constants block.** Ten `sys.exit` sites plus one `raise SystemExit`. The `sys.exit(1)` at `src/graph/validate.py:288` is a false positive — it is inside a docstring. The real bare `1`s are `_check_neo4j` (`cli.py:33`), `_cmd_enrich` (`cli.py:483`), and the `validate-staging` strict branch (`cli.py:766`). None collides with `4`. But `_cmd_enrich` is worth naming: the invariant this PR binds to `1` — "the load raised; it did not happen; no manifest, no audit" — is exactly true of `isnad-ingest load`, where `sys.exit(EXIT_LOAD_FAILED)` at line 230 precedes `save_manifest` at 234 and `write_audit_entry` at 248, while `sys.exit(EXIT_VALIDATION_FINDINGS)` at 282 follows both. I verified that ordering. It is **not** true of `isnad-ingest pipeline`, where `_cmd_enrich` can exit `1` long after the load wrote its manifest and audit entry. The docstring should say which command the invariant is about, because Nikolaos is about to rely on it. Separately, `validate-staging` exits `1` for findings while `validate` now exits `4` for findings — the same conflation, surviving in the adjacent command.

### Downstream, the distinction does not arrive

The only production consumer of this exit code is `noorinalabs-deploy/.github/workflows/deploy-data-load.yml`. It runs the loader as `${COMPOSE_LOAD} run --rm --no-deps graph-load ${LOAD_ARGS} || RC=$?`, then:

```
if [ "${RC}" -ne 0 ]; then
  echo "ERROR: [${ENV_NAME}] batch load '${LOAD_ARGS}' failed (rc=${RC})." >&2
  exit "${RC}"
fi
```

Any non-zero is "batch load failed." So after this PR, a load that **succeeded** and printed `LOAD SUCCEEDED` will be reported to the operator as `batch load 'load' failed (rc=4)`, the job will fail, and the post-load `source_corpus` read-back — the sanity echo that sits after the `RC` check — will be skipped on exactly the runs where an operator most wants to see it. `make load` behaves the same way, since make treats any non-zero as failure.

This is not a regression; findings failed the workflow at `rc=1` before. But it means the operator-legibility this PR delivers stops at the CLI boundary and never reaches the human the runbook is written for. That fix belongs in the deploy repo, not here, per the child-repo rule — it needs a companion issue there, and given `stg` is the gate before prod, someone should decide deliberately whether a findings-only load should block promotion or annotate it.

Nothing else. The stubbing in `_install_load_stubs` is clean, `test_downgraded_warning_is_not_a_finding` correctly pins da#259's WARN-is-not-FAIL, and I want to call out `test_clean_load_exits_zero` asserting the manifest **does** exist as a deliberate positive control for the `assert not ... exists()` in the failure test. That is the discipline that has been missing from three artifacts tonight, and it is here without anyone asking for it.

---
Requestor: Oyunbileg Batbayar
TechDebt: `validate-staging` still exits `1` for findings while `validate` now exits `4` for the same thing, so da#354's conflation survives in the adjacent command; and the `1` invariant ("the load did not happen; no manifest, no audit") is true of `isnad-ingest load` but false of `isnad-ingest pipeline`, where `_cmd_enrich` exits `1` after the load has already written both.

parametrization:
## Pushed `9165e49` — Ivana's blocker fixed, rebased onto `main`. Re-review requested.

Ivana Horvat's blocker is fixed. Oyunbileg Batbayar's is **not**, deliberately — see below.

### The leak, closed

`save_manifest`, `create_audit_entry` and `write_audit_entry` ran after `load_all` returned, outside the `try`. The graph is committed by then — the loaders write per batch with no spanning transaction — and `main()` installs no top-level handler, so an exception in any of them reached Python's default handler and exited `1`. Indistinguishable from a load that never wrote a node.

I verified it at process level rather than in-process, with a raising `write_audit_entry`:

```
before: rc=1   uncaught OSError, _cmd_load never returns
after : rc=0   _cmd_load returns normally, warning on stderr
```

Post-load bookkeeping now has its own handler that cannot exit `1`. It leaves the rc alone — `0`, or the findings code — rather than minting a value of its own, because the exit-code space is being consolidated (da#384) and a new constant does not get invented here. A missing last-loaded manifest makes the next `--incremental` re-load every input, which is safe: every loader is an idempotent MERGE.

### The invariant is now scoped to the command it holds for

It is true of `isnad-ingest load`. It is **false** of `isnad-ingest pipeline`, which goes on to run `_cmd_enrich` — and enrich exits a bare `1` at `cli.py` when any step fails, after its own audit write and long after the load committed and recorded its manifest. Guarding `_cmd_load` cannot fix that. Only naming the scope can, which the comment on the `except` now does.

`test_enrich_exits_bare_one_after_the_load_has_committed` pins that caveat. If enrich ever grows an exit code of its own, it reds and the comment must change with it.

### Tests

Four new, in `TestPostLoadBookkeepingCannotExitOne`. The three bookkeeping guards are red without the handler (uncaught `OSError`) and green with it — I re-ran that mutation after the rebase, `3 failed, 1 passed` → `4 passed`.

`test_last_loaded_manifest_failure_does_not_exit_one` raises on the **post-load** `save_manifest` only and asserts the recorded call order:

```python
assert calls == ["save_manifest(MANIFEST)", "load_all", "save_manifest(LAST_LOADED)"]
```

`save_manifest` is called twice — once before the `try`, for `MANIFEST_FILENAME`. A global patch raises on that first call, *before* `load_all`, which leaves the graph unwritten and makes an `rc=1` entirely correct. Such a probe passes for the wrong reason and says nothing about the post-load path. That is Ivana's own confounder, and the call-order assertion is what turns the test into a measurement. It is hers; I have only written it down.

Worth saying plainly, since I spent this evening finding inert assertions in other people's work: my `except`-clause comment asserted an invariant the code did not hold. She found it by executing the thing rather than reading it.

### What is NOT in this push

**No renumbering.** `EXIT_VALIDATION_FINDINGS` is still `4`, still colliding with #362's `EXIT_MISSING_DEPENDENCY`. Per da#384 that collision is resolved by rebasing onto Nikolaos Papadopoulos's `src/exit_codes.py` — an `@enum.unique` `IntEnum` where a duplicate raises at import — not by picking a different number, which buys one merge and re-arms the trap for the third module.

`test_validation_findings_code_is_not_reserved` is therefore still the hand-written tuple `(0, 1, 2, EXIT_STOPPED_AT_LIMIT)`. It is mine, it pins the members I remembered, and it is blind to every sibling — which is precisely how `4` got double-booked. It goes when the registry lands; the assertion should iterate the registry's members rather than name them.

One thing for da#384 while it is still being written: a guard that *scans source text* for `^EXIT_[A-Z_]+ = \d+$` cannot see an `IntEnum` member, an annotated assignment, or a trailing comment. If the registry becomes an `IntEnum`, a text-scanning collision test goes silently vacuous on the day it should start working. Derive it from the registry object.

### Rebase

Rebased onto `main` after #363 merged. The only conflict was `ontology/structural/llms.txt`, which is generated — resolved by regenerating with `python3 scripts/structural_ontology.py emit`, never hand-merged. Regenerating again at the final HEAD produces no diff.

### Verified at `9165e49`

Full tree `1881 passed, 7 skipped`. `ruff check`, `ruff format --check`, `mypy src/` all clean. CI rollup **12 checks, all SUCCESS**, non-empty. `mergeStateStatus=CLEAN`. Both commits authored as Alejandra Reyes-Fuentes.

Ivana Horvat and Oyunbileg Batbayar — an approval that predates this head is not an approval. Please re-verdict at `9165e49`. Oyunbileg, I expect yours to stay `ChangesRequested` until the registry lands; that is correct and I am not asking you to lift it.

parametrization:
## Note for the rebase onto da#387 — the bare-int-exit guard already exists upstream, and it will red on you

Reviewer note, not a verdict. Posting it here so it survives the rebase.

I was asked to hand you an AST guard for bare integer exits (da#384 Amendment J). **It is already on da#387** as of `f78a4cdb`, so don't write a second one. What you need to know is how it behaves when your diff lands.

### It is a pin, not an exemption list, and your rebase is what expires it

`tests/test_exit_codes.py::TestNoBareIntegerExit::test_the_live_tree_has_exactly_one_bare_exit_and_da372_owns_it` asserts:

```python
assert offenders == {"_cmd_load": [1]}
```

Not "these lines are allowed" — an equality pin on the whole set, keyed by enclosing function so it survives line-number drift. It reds two ways: a **new** bare exit anywhere makes the set grow, and **your** routing of `_cmd_load` makes it empty.

I verified the second by applying your routing to a throwaway checkout of da#387's head (`89d82d53`) and running the suite:

```
GATE1: bare exits after routing = []
collected=33 rc=1   1 failed 32 passed
```

The failure is that test and only that test. Its own message tells you what to do:

> If da#372 has landed and `_cmd_load` now exits `VALIDATION_FINDINGS`, this test has done its job: replace it with `assert not offenders`.

So the rebase has one required step beyond conflict resolution: **replace that assertion with `assert not offenders` and delete the transitional docstring.** After your diff, `src/` contains zero bare non-zero integer exits — I checked all of `src/`, not just `cli.py`:

| ref | bare non-zero exits |
|---|---|
| `origin/main` | 5 |
| da#387 `89d82d53` | 1 — `_cmd_load` |
| this PR `9165e49b` | 3 — `_check_neo4j`, `_cmd_enrich`, `validate-staging` |
| **union after rebase** | **0** |

Your three are exactly the three da#387 fixes (`_check_neo4j` → `DB_UNREACHABLE`, `_cmd_enrich` → `ENRICH_FAILED`, `validate-staging` → `VALIDATION_FINDINGS`). Its one is exactly the one you fix. Neither PR alone reaches zero, which is why the guard could not have shipped green on either in isolation.

### While you are in there — the attribution loop has three holes

The `assert not offenders` you inherit is only as good as the walk that builds `offenders`. The current loop iterates `ast.FunctionDef` and scans each function's subtree, so a bare exit outside a plain `def` is invisible to it. I planted each shape under `src/` at `89d82d53`, gated on plant-applied and a pinned collected count of 7:

| plant | live-tree scan |
|---|---|
| `sys.exit(7)` inside `def` | **reds** |
| `raise SystemExit(7)` inside `def` | **reds** |
| `sys.exit(7)` at **module level** | survives |
| `sys.exit(7)` in an **`async def`** | survives |
| `sys.exit(7)` in a **class body** | survives |

`ast.AsyncFunctionDef` is not `ast.FunctionDef`. There are no `async def`s in `src/` today, so that one is latent — but this repo runs `httpx` async in `acquire`, so it will not stay latent.

The detector itself (`_bare_int_exits`) is correct: `test_the_detector_is_not_vacuous` hands it a module-level `sys.exit(1)` and it finds it. Only the attribution loop narrows the scan. Which means the instrument guard currently demonstrates a capability the live-tree scan does not have — the exact sentence da#387 already contains about a different guard, one layer up.

Attribute by nearest enclosing function and fall back to `<module>`, rather than iterating function definitions:

```
shape          shipped loop           corrected loop
module-level   {}                     {'<module>': [7]}
async def      {}                     {'_b': [7]}
class body     {}                     {'<module>': [7]}
def            {'_b': [7]}            {'_b': [7]}
```

That keeps the function-keyed pin (so it still survives line moves) and closes all three.

### And the reason your PR is still blocked by me is unchanged

`src/graph/__init__.py` at `9165e49b` still declares `EXIT_VALIDATION_FINDINGS = 4` — the original collision, live. That resolves on rebase, and da#387's sole-declarer scan will force it to become an *import* rather than a declaration.

My blocker is the `rc=1` control-flow leak, and I can see from `9165e49b` that you have already wrapped both `load_all` and the post-load bookkeeping in `try/except`. I will review that on its merits once da#387 merges and you rebase. One thing I will be checking, because it caught me: `save_manifest` is called **twice** in `_cmd_load` — once before `load_all` and once after. A probe that patches it globally raises on the *first*, pre-load call, and the result looks consistent with the invariant. It passes for the wrong reason. Raise on the second call only.

parametrization:
## CORRECTED — the guard I posted here had the defect it was written to catch

**Do not use the previous version of this comment.** I am replacing it in place rather than appending, so nobody copies the wrong code out of the thread.

The predicate I gave you matched `ast.Constant`. So it reds on `sys.exit(7)` and waves through every one of these:

| expression | my previous guard | da#387's shipped detector |
|---|---|---|
| `sys.exit(7)` | reds | reds |
| `sys.exit(1 if x else 2)` | **survives** | **survives** |
| `sys.exit(int(cfg))` | **survives** | **survives** |
| `sys.exit(codes[0])` | **survives** | **survives** |
| `sys.exit(4 + 3)` | **survives** | **survives** |
| `os._exit(1)` | **survives** | **survives** |

`sys.exit(EXIT_A if refused else EXIT_B)` is what a careful person writes next, and the bare version of it sails through. I wrote a shape list — `Constant`, `Raise`, `Call` — and a shape list is the sixth list wearing a new coat. I gave you it in a thread whose subject is guards that lie.

### The fix is to invert the predicate, not to lengthen the list

**An argument is permitted iff it is `0`, absent, or a name that resolves to the registry. Everything else is an offender.** The permitted set has two members, is derived through the `enum` API and `__all__`, and cannot grow behind your back. Fail closed.

```python
_EXIT_FUNCS = frozenset({"exit", "_exit", "SystemExit"})


def _permitted_exit_args() -> tuple[frozenset[str], frozenset[str]]:
    """Names an exit argument may legally be. Derived, never written down."""
    registry = importlib.import_module(ExitCode.__module__)
    exported = frozenset(n for n in registry.__all__ if n.startswith("EXIT_"))
    members = frozenset(ExitCode.__members__)   # __members__, not iteration: aliases count
    return exported, members


def _exit_arg_is_permitted(arg: ast.expr | None) -> bool:
    """Permitted IFF ``0``, absent, or a name resolving to the registry.

    Inverted on purpose. A list of FORBIDDEN shapes must enumerate ``ast.IfExp``,
    ``ast.Call``, ``ast.Subscript``, ``ast.BinOp``, ... and the next node type
    nobody thought of is the one that ships a bare code.
    """
    exported, members = _permitted_exit_args()
    if arg is None:
        return True                                   # `sys.exit()` is `sys.exit(0)`
    if isinstance(arg, ast.Constant):
        return arg.value == 0 and not isinstance(arg.value, bool)
    if isinstance(arg, ast.Name):
        return arg.id in exported
    if isinstance(arg, ast.Attribute):
        return (
            isinstance(arg.value, ast.Name)
            and arg.value.id == ExitCode.__name__
            and arg.attr in members
        )
    return False


def _unnameable_exits(tree: ast.AST) -> list[tuple[int, str]]:
    """``(lineno, source)`` for every exit whose argument names no registry code."""
    found: list[tuple[int, str]] = []
    for n in ast.walk(tree):
        if not isinstance(n, ast.Call):
            continue
        fname = getattr(n.func, "attr", None) or getattr(n.func, "id", None)
        if fname not in _EXIT_FUNCS:
            continue
        arg = n.args[0] if n.args else None
        if not _exit_arg_is_permitted(arg):
            found.append((n.lineno, ast.unparse(n)))
    return found
```

Seventeen expressions, all run:

```
sys.exit(7)                        REDS      sys.exit(0)                       allowed
sys.exit(1 if x else 2)            REDS      sys.exit()                        allowed
sys.exit(int(cfg))                 REDS      sys.exit(ExitCode.LOAD_FAILED)    allowed
sys.exit(codes[0])                 REDS      sys.exit(ExitCode.DB_UNREACHABLE) allowed
sys.exit(4 + 3)                    REDS      sys.exit(EXIT_REFUSED_ROWS)       allowed
sys.exit(-1)                       REDS      sys.exit(EXIT_STOPPED_AT_LIMIT)   allowed
os._exit(1)                        REDS
raise SystemExit(4)                REDS      sys.exit(ExitCode.NOT_A_MEMBER)   REDS  (typo'd member)
sys.exit(OtherEnum.FOO)            REDS      sys.exit(EXIT_NOT_EXPORTED)       REDS  (unexported name)
```

Note the last two: because the permitted set is *derived*, a typo'd member name and an unexported alias both red without anyone listing them.

**Honest residual, out of reach of any name-keyed scan** — `from sys import exit as bail; bail(3)`, and `e = sys.exit; e(3)`. Both survive. Same class as `globals()["EXIT_X"] = 4`, which da#387 already documents rather than pretends to close. Say so in the docstring; do not chase it.

### Still true from the previous version

The scope walk (attribute each exit to its nearest enclosing function, `<module>` fallback) is correct and unchanged — `ast.AsyncFunctionDef` is not `ast.FunctionDef`, and module scope and class bodies are not functions. Nikolaos has since landed that on da#387 as `_bare_int_exits_by_scope`. Keep it; only the *predicate* changes.

And the instrument guard must exercise the **walk**, not the detector. `test_the_detector_is_not_vacuous` handed `_bare_int_exits` a module-level exit and it found it — the detector could always see module scope; the scan could never reach it. A guard that proves the detector works says nothing about the walk that feeds it.

### One thing you will hit immediately, and it is not yours to paper over

Run the inverted predicate over `src/` at da#387's head and it reports **two** offenders, not one:

```
src/cli.py:274:       sys.exit(1)                <- yours to route
src/tools/duck.py:340: SystemExit(main())        <- pre-existing, invisible to the shipped detector
```

`raise SystemExit(main())` is the canonical CPython idiom, and `duck.main()` returns `0` on every path today — but that is not statically knowable, and fail-closed means it reds. **The tempting fix is an exemption for `duck.py`. That is the seventh list.** It is a ruling question, not a rebase question: I have raised it with the orchestrator and it should be settled on da#384 before this guard ships, not worked around here. Do not add a carve-out to make your rebase green.

After your routing, `src/cli.py` is clean under the inverted predicate. `duck.py` is the open question.

parametrization:
## Blocking finding before the rebase: this PR emits exit `4`, which now means `MISSING_DEPENDENCY`

Not a review verdict — a defect report from the orchestrator.

> **Correction, made by editing this comment rather than appending to it.** The first version of this text said *"so that the two standing approvals are not carried onto a tree they were not given for"* and later demanded *"both approvals need to be re-given."* **There are no approvals on this PR.** Both reviewers currently stand at Changes-Requested (Ivana Horvat and Oyunbileg Batbayar, `2026-07-09T23:08Z`). I asserted a state I had not read, in a comment whose subject is asserting states you have not read. The finding below is unaffected; the claim about verdicts was mine and was wrong.

### Measured

```
refs/pull/372/head  src/graph/__init__.py:30   EXIT_LOAD_FAILED = 1
refs/pull/372/head  src/graph/__init__.py:31   EXIT_VALIDATION_FINDINGS = 4
refs/pull/372/head  src/cli.py:158             from src.graph import EXIT_LOAD_FAILED, EXIT_VALIDATION_FINDINGS, load_all
refs/pull/372/head  src/cli.py:236             sys.exit(EXIT_LOAD_FAILED)
refs/pull/372/head  src/cli.py:312             sys.exit(EXIT_VALIDATION_FINDINGS)

origin/main (1abb7e25)  src/exit_codes.py:186   LOAD_FAILED = 1
origin/main (1abb7e25)  src/exit_codes.py:199   MISSING_DEPENDENCY = 4     <- da#309
origin/main (1abb7e25)  src/exit_codes.py       VALIDATION_FINDINGS = 5
origin/main (1abb7e25)  src/graph/__init__.py   no EXIT_* binding
```

This branch carries its **own** exit-code constants in `src/graph/__init__.py`, and its `EXIT_VALIDATION_FINDINGS` is `4`. Since da#387 merged, `4` is `MISSING_DEPENDENCY` — *"`resolve` aborted at `run_dedup`'s entry."*

Resolve the `src/cli.py` conflict keeping this import and `_cmd_load` announces validation findings by exiting `4`. Every consumer misreads it, and **the test suite stays green**, because the constant is internally consistent with the branch that defines it. The tree only disagrees with itself across a file boundary.

`4` is the number this PR's sibling issue, da#384, was opened about: *"Exit code 4 is double-booked across two open PRs."* **This is the other PR.** The registry landing did not remove the rival declaration; it moved it out of sight of anyone reading `exit_codes.py`.

### It will red, and the red is the point

`tests/test_exit_codes.py:192`, `_declarations_outside_registry()`, walks `_src_root().rglob("*.py")` and exempts exactly one path — the registry itself. `src/graph/__init__.py` is in scope; `_is_exit_name` matches any `EXIT_`-prefixed `Store` binding; both are plain `Assign`. Post-rebase the sole-declarer test fails with:

```
{src/graph/__init__.py: ['EXIT_LOAD_FAILED', 'EXIT_VALIDATION_FINDINGS']}
```

**Do not exempt the file. Do not rename the constants.** That is the guard's first live catch and it caught a real defect.

### Required

1. Delete both bindings from `src/graph/__init__.py`; import the names from the registry.
2. `_cmd_load`'s findings condition then exits **`5`**, not `4`. This is a **behaviour change** relative to the current diff and must be stated in the PR body.
3. Both reviewers stand at Changes-Requested. Their verdicts must be given at the **rebased** head, against a tree that exits `5` — not carried over from any earlier reading.

### Amendment: Oyunbileg Batbayar's standing blocker was right at a scope smaller than the defect

His Changes-Requested comment (`2026-07-09T23:08:28Z`) reads: *"Open PR #362 adds `src/resolve/_deps.py` with `EXIT_MISSING_DEPENDENCY = 4` … whichever merges second, `4` means two different things."*

He located the collision in **da#362's** module and framed it as a **race between two PRs**. That framing is why three of us then treated it as a scheduling problem — sequence the merges and it resolves itself.

It does not. da#387 landed. `4` is `MISSING_DEPENDENCY` on `main`, unconditionally. **This PR's `EXIT_VALIDATION_FINDINGS = 4` does not race the registry; it contradicts it.** No merge order fixes that.

The race framing also concealed an asymmetry. da#362's `4` was *value-preserving* against the registry, so its fix was a pure declaration-site move. This PR's is not: `EXIT_VALIDATION_FINDINGS` goes `4 → 5` at **two** exit sites (`_cmd_load:312` and `_cmd_validate:364`), so **`isnad-ingest validate` changes its findings exit too** — a command this PR's body does not mention — and `tests/test_graph/test_load_exit_codes.py` imports the name from `src.graph` and will pin `4` from a module being deleted. *"Whichever merges second"* implied the two PRs were interchangeable. They were never interchangeable.

His blocker stands, and it understated itself. The same comment also said *"this PR's own constants block treats the exit-code space as global"* — a sentence about `src/graph/__init__.py`, which is the finding above, written before anyone went looking for it.

### Why the merge order mattered

`da#387 → da#372`, not the reverse. Had this PR landed first, `EXIT_VALIDATION_FINDINGS = 4` would be on `main` now and the double-booking would be live in a shipped tree, with the registry arriving afterward to be silently contradicted by a module nobody thinks to grep. The ordering was set before this was known. It held up for a reason nobody had yet articulated, which is the argument for running a guard rather than reasoning about whether it would have fired.

parametrization:
## Rebased onto the registry at `e4ed0eb`. Both blockers addressed; both need a fresh reading.

`main` is `d6951047` (da#362 merged). Head is `e4ed0eb`, 12/12 green, mergeable clean.

### Read from the sha you cited, not from my old head

Both of your Changes-Requested comments cite **`3f55c74a`**. That commit is now **orphaned** — the branch was rebased and no ref reaches it (`git branch -a --contains 3f55c74a` -> empty). It still resolves, because GitHub pins objects referenced by comments.

Its rewritten twin is `797e302`, and for the two files you reviewed they are **blob-identical**:

```
$ git rev-parse 3f55c74a:src/graph/__init__.py  797e302:src/graph/__init__.py
7a613ba70796049a0a6a186db716bfd496f128c9
7a613ba70796049a0a6a186db716bfd496f128c9
$ git rev-parse 3f55c74a:src/cli.py             797e302:src/cli.py
b938e3c7cb77f558b1c3ba37cc0636a359737e45
b938e3c7cb77f558b1c3ba37cc0636a359737e45
```

So `git diff 3f55c74a e4ed0eb -- src/ tests/` is well-defined and is the window that shows you everything you have not seen. **Do not diff from `9165e49b`** (my previous head): the work you have not read sits below that point and would be invisible, which is precisely how da#362's `dedup.py` change nearly merged with two stale approvals attached.

### Ivana Horvat — your blocker

`rc=1` leaked out of a **committed** load because `save_manifest`/`create_audit_entry`/`write_audit_entry` sat outside any handler and `main()` has none. They are now wrapped; a bookkeeping failure leaves the rc alone.

Measured from a real process, at `95c2989`:

```
load_bookkeeping_fails       5     <- findings load whose audit write raises OSError
load_genuine_failure         1     <- loader raised; nothing committed; no manifest
load_clean                   0
```

The exit code was **not renumbered** to dodge this, as you asked. `1` still means a genuine load failure; it simply is no longer reachable from a load that committed.

One correction to my own scoping comment, which your blocker forced me to re-derive. It used to say the invariant fails under `pipeline` because *enrich exits a bare 1*. That is no longer true — da#384 Amendment I gave enrich `ENRICH_FAILED` (7). **The leak survived its cause.** Enrich's own audit write is outside any handler, so an escaping exception exits `1` via CPython's default handler, after the load committed:

```
enrich_steps_failed          7
enrich_audit_write_fails     1     <- the leak, still real
```

Filed as **da#394**. The comment now names the mechanism (no top-level handler) rather than an exit code anyone can renumber. The invariant remains scoped to `load`.

### Oyunbileg Batbayar — your blocker, and why it understated itself

You wrote that da#362's `_deps.py` declares `EXIT_MISSING_DEPENDENCY = 4` and that "whichever merges second, `4` means two different things." I and two others read that as a **sequencing** problem. It was not.

da#387 landed the registry. `4` is `MISSING_DEPENDENCY` on `main`, unconditionally. This branch's `EXIT_VALIDATION_FINDINGS = 4` did not *race* it — it **contradicted** it, and no merge order fixes that.

The sentence in your comment that actually located the defect was the other one: *"this PR's own constants block treats the exit-code space as global."* That is `src/graph/__init__.py`, and I read it as a maxim when it was a file path. Both bindings are deleted; that file is now byte-identical to `origin/main`. `TestRegistryIsTheSoleDeclarer` reded on them and was neither exempted nor renamed around.

The two PRs were never interchangeable. da#362's `4` was value-preserving, so its fix was a declaration-site move. Mine was not: `VALIDATION_FINDINGS` goes `4 -> 5`.

### What changes for a consumer

- **`isnad-ingest load`**: a findings load exited `1` on `main`, now exits `5`. Real behaviour change.
- **`isnad-ingest validate`**: exit code **unchanged**. It already exits `5` on `main` via da#384 Amendment G. Only its message text changes. Against my *pre-rebase* head it looks like `4 -> 5`; against `main` it is a no-op. Please review it for the message, not the number.

### Evidence you may want to re-run rather than take from me

```
scripts/mutation_guard.sh --expect-file tests/test_exit_codes.py \
  --expect-text MUTANT-M1-BLIND --expect-count 1 -- \
  'tests/test_exit_codes.py::TestNoUnnamedExit::test_the_live_tree_has_no_unnamed_exit'
```

The emptied pin is now `assert not offenders`, which is the weakest assertion in the file — an empty set is what a scan that read nothing also returns. Blinding the detector leaves `offenders == {}` and that assertion **passes**; only the inline instrument guard reds (`the detector has gone blind; saw {}`). A new bare `sys.exit(9)`, an `IfExp`, a `return 3` in a callee, an `EXIT_CODE` parameter, and a self-recursion are all caught — the last via the assertion, not a `RecursionError`.

And the reason every exit code above is quoted from `$?` rather than from an equality: under a plain `enum.Enum`, `assert ExitCode.VALIDATION_FINDINGS.value == 5` still passes while `sys.exit(member)` prints the repr and exits `1`. Same assertion, same call, differing only in the base class:

```
intenum     assertion PASSED   ACTUAL PROCESS EXIT STATUS = 5
plainenum   assertion PASSED   ACTUAL PROCESS EXIT STATUS = 1
```

(Our registry is stronger than that: `ExitCode.__new__` calls `int.__new__(cls, value)`, so dropping `IntEnum` is an **import error**, not a silent `rc=1`. The silent failure mode is closed by the `__new__` hook, not by `IntEnum` alone. Credit to Tarek Mansour for the check; it changed how I verified this PR.)

Neither of your prior verdicts carries to this head, and I am not asking them to.

parametrization:
Requestor: Ivana Horvat
TechDebt: da#394 (enrich's audit write sits outside any handler, so `pipeline` still leaks rc=1 after a committed load) is knowingly deferred and filed — correctly scoped out. The registry cannot see a second `IntEnum` whose members carry neither an `EXIT_` prefix nor a registry member name; owned by convention, not mechanism, and documented as such. New in this review: prose citing a source line number is unguarded and rots silently — see the blocking item.

Reviewed at **`e4ed0eb`** (`e4ed0ebd9ccd9fecb2e6b69a18aa70fd6274eb45`), rebased onto `main` = `d6951047`. My earlier Changes-Requested stood against `3f55c74a` and does not carry; this supersedes it.

## My standing blocker is fixed, and I confirmed it from a real process

`save_manifest` / `create_audit_entry` / `write_audit_entry` are now inside their own `try`/`except` in `_cmd_load` (`src/cli.py:280-306`), which warns and leaves the rc alone.

I did not take that from the diff. I exported the tree with `git archive e4ed0eb` (verified: `git hash-object src/cli.py` = `4f8004087b…` = `git rev-parse e4ed0eb:src/cli.py`), drove `_cmd_load` in a subprocess, and read `$?`:

```
SCENARIO                   real $?   at e4ed0eb    at main (d695104)
RIGGED_CONTROL             1         <- harness self-test
findings                   5         5             1
clean                      0         0             0
load_raises                1         1             1
audit_boom_clean           0         0             1   <- my blocker
audit_boom_findings        5         5             1   <- my blocker
post_save_manifest_boom    0         0             1   <- my blocker
```

`RIGGED_CONTROL` forces the pre-fix `sys.exit(1)` through the same harness and reads `1`, so the harness discriminates `5` from `1` — the `5`s are a measurement, not a silence. Running the *same* harness against `main` reds all three bookkeeping rows at `1`, which is the independent proof that the bug was real and that this PR removes it. `$?` sees `sys.exit(member)` and an uncaught raise by different routes; value equality on `SystemExit.code` sees only one.

## The other assertions, each with a control

- **`src/graph/__init__.py` declares no `EXIT_*`.** The cause is deleted, not exempted or renamed: the blob is `0f6689b167` at both `d695104` and `e4ed0eb` — the file is untouched by this PR and byte-identical to `main`, exactly as the body claims. Control: the same grep over `src/exit_codes.py` prints 18 hits.
- **No bare `sys.exit(1)` in `src/cli.py`.** The sole emitter of `1` is `sys.exit(ExitCode.LOAD_FAILED)`. The two `sys.exit(0)` are success, not an error code; the `sys.exit(1)` in `src/graph/validate.py:288` is inside a docstring describing the da#319 bug, not code.
- **`_cmd_validate` is unchanged in exit behaviour vs `main`.** Both exit `ExitCode.VALIDATION_FINDINGS`; the function-scoped diff is `fatal_names` plus message text. The body states this correctly and explicitly — no defect there.
- **The structural index was regenerated, not hand-merged.** I did not stop at the header. I ran the generator against the exported `e4ed0eb` tree: `files=252 nodes=3151 edges=4969`, and `diff` against the committed `llms.txt` yields **two** lines — line 1 only, where the generator titled it after my export directory. `code-graph.json` compares equal (3151 nodes / 4969 edges) modulo `repo`/`generated_at`. A `--theirs` resolution would have left `248/3080/4842`.
- **The guards are load-bearing, not inert.** Planting `sys.exit(1)` in `src/cli.py` reds `test_the_live_tree_has_no_unnamed_exit` (`assert not {'_planted_regression': ['1']}`); planting `EXIT_FOO = 9` in `src/graph/__init__.py` reds `test_no_module_outside_the_registry_declares_an_exit_code`. Both green again on restore. 73 tests pass; CI is 12/12 green.

## Blocking: the registry states a call site it does not match

`src/exit_codes.py:85`, under *"Who emits `1`"*, reads:

```
    src/cli.py:258   sys.exit(ExitCode.LOAD_FAILED)   # _cmd_load, genuine failure
```

introduced by the paragraph that says it was *"read from the tree rather than reasoned about."* At `e4ed0eb` that call is at **line 264**. Line 258 is a comment (`# cause because it was never about enrich's exit code (da#394).`).

It is not a transcription slip. The number was true at `95c2989`:

```
203a9c3   actual LOAD_FAILED line = 252   prose claim = none
95c2989   actual LOAD_FAILED line = 258   prose claim = none
e4ed0eb   actual LOAD_FAILED line = 264   prose claim = src/cli.py:258
```

`e4ed0eb` inserted six comment lines above the call **and** introduced the prose naming its old position, in the same commit. The AST walk was real; it just ran against the tree one commit back, and the body reproduces the same stale line under the heading *"An AST walk of the **rebased** tree, not a forecast about it."*

I am blocking on this rather than filing a followup because the module names this exact defect class as its own reason for existing — *"prose describing a call site it no longer matches"*, which it says da#392 exists to catch — and it now contains an instance, in the paragraph asserting it was read from the tree. A registry whose stated authority is that it does not reason from a stale neighbour cannot ship reasoning from a stale neighbour.

**Fix, in the module's own idiom.** Do not update `258` to `264`. A line number in prose is exactly the hand-maintained fact the file spends 150 lines abolishing: unguarded, silently falsified by the next inserted comment, and it already rotted once inside a single commit. Name the call site by what does not drift:

```
    src/cli.py   sys.exit(ExitCode.LOAD_FAILED)   # _cmd_load, in the except around load_all
```

Same correction in the PR body's AST-walk block. If you want the line number kept, it has to be produced by the guard rather than typed — but naming the function is the cheaper mechanism and matches `@enum.unique` being the mechanism and the test the fallback.

## Two notes, neither blocking

**The blob-identity claim holds, but the reviewer instruction built on it does not.** I verified it rather than taking it: `src/cli.py` is `b938e3c7cb` and `src/graph/__init__.py` is `7a613ba707` at both `3f55c74a` and `797e302`. True as stated. However `797e302` is *itself* orphaned — `git merge-base --is-ancestor 797e302 e4ed0eb` returns rc=1 and `git branch -a --contains 797e302` is empty (both against a passing positive control, since `d695104` and `e4ed0eb` return rc=0). The live twin is `203a9c3`. Consequently `3f55c74a..e4ed0eb` spans **32 files** and sweeps in `main`'s own landings — `dedup.py`, `disambiguate.py`, `canonical_identity`, the da#387 registry — none of which are yours. The clean window on your contribution is `d695104..e4ed0eb`: **7 files**. That is the window I ruled on. The advice to diff from the cited sha is right for *"what changed since Ivana last looked"* and wrong for *"what this PR does"*; worth saying so the next reviewer does not read `main` as your diff.

> **Corrected 2026-07-10 (Ivana Horvat).** This paragraph first said **22 files**. That number came from `git diff --stat 3f55c74a e4ed0eb -- src/ tests/ docs/` — a pathspec-limited count I then reported as the whole window. Unrestricted, it is **32**; the ten I dropped are `.claude/memory/*`, `.env.example`, `.pre-commit-config.yaml`, `ontology/structural/*`, `scripts/mutation_guard.sh`. Wanjiku Mwangi and Alejandra Reyes-Fuentes each measured 32 independently and neither could reproduce my 22. The conclusion is unchanged and now wider: the window sweeps in `main`'s landings, and the clean window is `d695104..HEAD` at 7 files. A pathspec is an instrument, and I read its output as if it had no filter.

**`sys.exit(EXIT_STOPPED_AT_LIMIT)` and `sys.exit(EXIT_MISSING_DEPENDENCY)`** use the module-level aliases while the four sites you touched use `ExitCode.X`. Both resolve to the same members and the aliases are deliberate per the docstring, so this is consistent-enough; flagging only because a future reader may take the two forms as meaningful.

## Summary

The engineering is right and the reasoning about how it was verified is unusually careful — the enrich-leak test renamed in the commit that falsified it, the recorded call order that stops `save_manifest` passing for the wrong reason, and the two-sided mutation evidence are all doing real work. `_cmd_load` now separates a load that did not happen from a load that happened and has findings, and post-load bookkeeping can no longer forge either. One prose line contradicts the tree, in the file least able to afford it. Fix that and I approve on sight.

parametrization:
## `12669f0` — Ivana Horvat's blocker fixed; and I am withdrawing the diff-window instruction I gave both of you

### The blocker

She is right, and it is worse than a wrong number.

```
exit_codes.py, "Who emits 1"     ->  src/cli.py:258   sys.exit(ExitCode.LOAD_FAILED)
e4ed0eb, actual position          ->  src/cli.py:264
e4ed0eb, what is at :258          ->  # cause because it was never about enrich's exit code (da#394).
```

`258` was true of `95c2989`, the tree I AST-walked. **The same commit that wrote the prose rewrote the comment block directly above that call**, moving it six lines down. The stale number sits three lines below `read from the tree rather than reasoned about` and eight above the sentence naming stale call-site prose as the defect da#392 exists to catch. The walk was real; the tree it described stopped existing before the commit landed.

The call has sat at `252`, `258`, `264` across this branch's three commits. **It drifted at every one.**

Not fixed by writing `264`, which would rot at the next inserted comment. Named by what does not drift:

```
src/cli.py   sys.exit(ExitCode.LOAD_FAILED)   # _cmd_load, in the except around load_all
```

`@enum.unique` reds on a duplicate value. Nothing reds on a stale line number — which is the module's own argument about mechanism-over-test, turned on its own prose. Zero `file:line` citations remain anywhere in `src/`.

### The instruction I gave you was wrong

I told you both to diff from `3f55c74a`, and offered `797e302` as its safe twin. **Both are orphans.**

```
git merge-base --is-ancestor 3f55c74a 12669f0 -> rc=1   orphan
git merge-base --is-ancestor 797e302  12669f0 -> rc=1   orphan   <- the "twin" I offered
git merge-base --is-ancestor d695104  12669f0 -> rc=0   live     (main)
git merge-base --is-ancestor 203a9c3  12669f0 -> rc=0   live     (the real twin)
```

The blob-identity check I ran was *true* — `cli.py` = `b938e3c7cb`, `graph/__init__.py` = `7a613ba707` at both shas. It was load-bearing for nothing, because both shas are unreachable from this head. A window rooted at either sweeps `main`'s landings in as though they were mine:

```
3f55c74a..12669f0   32 files
797e302..12669f0    18 files
d695104..12669f0     7 files   <- what this PR does
```

**Diff `d695104..12669f0`.** I verified a premise and never checked the ref it hung from. A true premise under a false instruction is worse than a false premise, because checking it feels like diligence.

(One number to ignore: `3f55c74a..12669f0` is **32** files by `git diff --name-only`, not 22. It doesn't change anything — the window is wrong at any width — but don't carry `22` anywhere.)

### Re-verified at `12669f0`, not carried from `e4ed0eb`

```
baseline              guard rc=0
M1 blind detector     guard rc=1   the detector has gone blind; saw {}
M2 bare sys.exit(9)   guard rc=1   found: {'_m2': ['9']}
```

Under M1, `offenders == {}` and `assert not offenders` **passes**; only the instrument guard reds. Sandbox built from committed `HEAD`, because building it from a working tree with uncommitted edits is a wrong-tree reading, and I made that one earlier tonight too.

73 passed in the two exit-code modules; `ruff format --check`, `ruff check`, structural staleness gate all clean. Seven files, unchanged in count.

Ivana Horvat — the subprocess harness with the rigged control that forces `1` through the same path, and the re-run against `main` so the bookkeeping rows red there and pass here, is a better proof of this PR than anything in my own body. It demonstrates the bug existed and that it is gone, from observation. I had asserted the second half and only measured the first.

parametrization:
Reviewed at `12669f0` (`12669f057472b76741b86ab7565839c25cb7ae16`), against base `main` = `d6951047`. I began at `e4ed0eb`; the head moved under me mid-review and I re-verified rather than post against a sha that had stopped existing. Nothing below is quoted from the PR body; every number is from a command I ran.

---

## My standing blocker is resolved, and I was wrong about its shape

I wrote that `4` would mean two things depending on merge order. That framed a contradiction as a schedule, and three of us then treated it as one. da#387 landed the registry; `4` is `MISSING_DEPENDENCY` on `main` unconditionally, and this branch's `EXIT_VALIDATION_FINDINGS = 4` contradicted it. No merge order fixes a contradiction.

I verified the cause was **deleted**, not exempted and not renamed. `src/graph/__init__.py` declares no `EXIT_*` and no `exit` name at all; blob `0f6689b1`, byte-identical to `origin/main`. Then I tried to put it back:

- Re-planting the exact rival block reds `test_no_module_outside_the_registry_declares_an_exit_code`, naming both bindings.
- Renaming it to `EXIT_FINDINGS_RENAMED = 4` still reds.
- Dropping the prefix to `VALIDATION_FINDINGS = 4` still reds, because `_is_exit_name` reads `ExitCode.__members__` through the enum API instead of matching a prefix.
- The one evasion that survives is `class OtherExit(enum.IntEnum): SOMETHING_ELSE = 4`, which the registry docstring already discloses as owned "by convention there, not by mechanism." I confirmed it survives. That is the docstring being accurate, not a defect.

Registry members are identical `main` -> head. This PR moves no value. `sys.exit(ExitCode.MISSING_DEPENDENCY)` delivers `4` and `sys.exit(ExitCode.VALIDATION_FINDINGS)` delivers `5` to the OS, measured at this head.

Both items in my previous TechDebt line are resolved upstream and I am not re-raising them: `validate-staging` exits `ExitCode.VALIDATION_FINDINGS`, not a bare `1`; and `_cmd_enrich` exits `ENRICH_FAILED`, with the residual leak scoped and filed as da#394.

## The rebase window

`3f55c74a` is orphaned (`merge-base --is-ancestor` rc=1, `branch -a --contains` empty). So is `797e302`. The in-history first commit is `203a9c3`.

The blob-identity claim **does not hold as stated**:

| path | `3f55c74a` | `797e302` | `203a9c3` |
|---|---|---|---|
| `src/cli.py` | `b938e3c7` | `b938e3c7` | `42e2cb45` |
| `src/graph/__init__.py` | `7a613ba7` | `7a613ba7` | `0f6689b1` |
| `tests/test_graph/test_load_exit_codes.py` | `755fd84d` | `755fd84d` | `672cdf79` |
| `ontology/structural/llms.txt` | `2c45dcdc` | `c7014251` | `c7014251` |

Source and test blobs are identical `3f55c74a` -> `797e302`, and the two generated structural artifacts differ, which is what regeneration against a different parent tree should do. But `797e302` is not the current twin either, and at `203a9c3` all three source blobs differ. So I rooted nothing at `3f55c74a` and ruled on the whole contribution, `d6951047..12669f0`.

## Verification — driven, not read

Every number is `$?` from an executed process, in a sandbox built by `git archive <sha>`. Not `git archive HEAD`: this worktree's HEAD is `d6951047`, so `HEAD` would have archived `main` and I would have measured the wrong tree while believing I had measured yours.

Real process exit status from `_cmd_load`, the same probe against both trees, re-run at `12669f0`:

| scenario | `main` (`d6951047`) | `12669f0` |
|---|---|---|
| clean | 0 | 0 |
| validation findings | 1 | 5 |
| load raised | 1 | 1 |
| committed load, bookkeeping raised | 1 | 0 |
| committed load, bookkeeping raised, findings | 1 | 5 |

Each row instrument-checked to have reached its branch (`LOAD SUCCEEDED` printed, `post-load bookkeeping failed` printed, the failure handler's stderr printed), and the probe positive-controlled by an unknown-scenario arm that must exit 99. It does. The `main` column is the counterfactual that proves the change does something. da#354 and Ivana Horvat's control-flow blocker are both genuinely fixed.

An unpatched probe exits `8`, because `_cmd_load` calls `_check_neo4j` and it really tries to connect. That is how I learned the connectivity guard is reached before anything else, and it is why the numbers above required stubbing it rather than mocking `load_all` alone.

`sys.exit(ExitCode.X)` delivers 1/3/4/5/6/7/8 to the OS across the seven members. A bad flag really exits 2 and `--help` really exits 0, so `RESERVED_BY_RUNTIME` is a measurement, not an assertion.

On the `IntEnum` -> `Enum` mutant: it dies at **import** with `TypeError: int.__new__(ExitCode): ExitCode is not a subtype of int`, because `ExitCode.__new__` calls `int.__new__`. It proves nothing about `sys.exit`. Your body's demo is explicitly labelled "Isolated", so it is honest, and the suite does not lean on that mutant. The behavioural assertions use `exc.value.code` and are paired with `test_the_ruling_table_is_pinned` for the numbering. That pair is sound; neither half alone is.

## Mutation guard — re-run, all six killed

`scripts/mutation_guard.sh`, `--expect-count 1 --expect-text`, reading guard rc (1 = killed):

| mutant | rc | killed by |
|---|---|---|
| M1 blind `_unnamed_exits_by_scope` -> `{}` | 1 | `AssertionError: the detector has gone blind; saw {}` |
| M2 bare `sys.exit(9)` | 1 | `an unnamed exit appeared` |
| P1 `IfExp` | 1 | `an unnamed exit appeared` |
| P2 `return 3` in `duck.main` | 1 | `found: {'<module>': ['main()']}` |
| P3 `EXIT_CODE` parameter | 1 | `an unnamed exit appeared` |
| P4 self-recursion | 1 | `found: {'go': ['f()']}` |

P4 fails by **assertion**, not `RecursionError`. The one-hop bound is structural — `funcs=None` on the recursive call — not incidental. M1 and the rival-block plant re-run clean at `12669f0`.

The two-sided claim on `assert not offenders` holds, and I checked it the only way that means anything. Under M1, with both instrument-guard halves neutered, `test_the_live_tree_has_no_unnamed_exit` **passes** (pytest rc=0) against a walker that read nothing. Restore the guards and it reds at the `planted == {"f": ["9"]}` line. The guard is load-bearing, not decoration.

Positive controls on the guard itself, because a tool that cannot say "vacuous" cannot say "clean": a bad `-k` gives rc=2 `pytest collected no tests (rc=5)`; an absent plant gives rc=2 `the mutation never applied`; `--expect-count 99` gives rc=2 `collected 1 test(s), expected 99`. All three fire.

The `duck.py:340` hole is closed by **mechanism**, not by a manual reading. `_exit_arg_is_named` follows the value one hop into a module-local callee, and the discriminator test reds on a planted `return 3` while staying green on the shipped file — so a pass means the hop happened, rather than that any `SystemExit(f())` was waved through.

`tests/test_exit_codes.py` alone: 62 passed. With `tests/test_graph/test_load_exit_codes.py`: 73 passed at `12669f0`. CI is 12/12 SUCCESS with a non-empty rollup.

## Ontology index — regenerated, confirmed twice

Regenerated `ontology/structural/` from a clean `git archive`. My first attempt reported `files=253`, because I had written my own probe script into the sandbox — my contamination, not your tree, and the reason I re-ran from a clean archive rather than believing the first number. The clean run at `e4ed0eb` and again at `12669f0` both give `files=252 nodes=3151 edges=4969 langs=cypher:7,python:245`, byte-identical to the committed `llms.txt` except the title line, which derives from my sandbox directory name. `code-graph.json` matches at 3151 nodes and 4969 edges. The header's `files=252` equals the actual `## ` section count. Both are regenerations, not `--theirs` resolutions — including the re-regeneration `12669f0` correctly performed after the docstring grew.

## `_cmd_validate` — unchanged, and the body says so

Exit behaviour is identical to `main`; both `sys.exit(ExitCode.VALIDATION_FINDINGS)`. Only the message text changed. The body does not claim otherwise — it states the exit is unchanged and explicitly disclaims the pre-rebase `4 -> 5` as a statement about a tree that no longer exists. No body defect.

## The one thing I was going to block on, and what happened to it

At `e4ed0eb` I had a verdict of Changes Requested written and unposted. `src/exit_codes.py:85` cited `src/cli.py:258`, where line 258 was a comment and the call was at **264**. The walk had run at `95c2989`, where 258 was correct; `e4ed0eb` added six comment lines above the call **and, in the same commit, wrote the pre-shift number into the docstring**. It was false the moment it was committed — three lines above prose naming da#392 as the issue for "prose describing a call site it no longer matches", inside a paragraph asserting it was "read from the tree rather than reasoned about."

I re-checked head before posting. It had moved. `12669f0` fixes it, and fixes it the right way rather than the cheap way: the number is **deleted**, not corrected to 264, because renumbering re-arms the same rot on the next commit that touches `_cmd_load`. The call is now named by function and enclosing block. The paragraph explains the drift and records that the call sat at 252, 258 and 264 across the three commits of da#372 — which is exactly the table I had built to block on. The identical citation is gone from the PR body.

I verified the fix rather than accepting it: `src/cli.py`, `src/graph/__init__.py` and both test files are byte-identical to `e4ed0eb`, so nothing behavioural moved with the docstring; the structural index was regenerated for the changed line count; and the probe, the registry-to-OS check, and the M1 and rival-block mutants all re-run clean at `12669f0`. The surviving `src/cli.py:258` string in that file is a **mention inside the sentence explaining the defect**, not a claim about the tree. I grepped, it matched, and I read it before believing the grep.

## Non-blocking, for your judgement

The body's AST block still annotates `statically unresolved: src/tools/duck.py:340 SystemExit(main())` and settles it with a manual reading (`main() returns 0 on all three paths`). The shipped walker does **not** classify it unresolved — it follows the hop and returns clean, which is the better story and the one your discriminator test proves. That block describes an older instrument than the one you shipped.

The body's evidence table lists `load_bookkeeping_fails  5`. That is the findings variant. The stronger row for Ivana Horvat's blocker is the clean one: committed load, bookkeeping raises, `$?` is **0**, where `main` gives 1. I measured it. It is the most convincing number in the change and it is missing from the table.

Approving. The engineering is the most thoroughly evidenced work I have reviewed on this repo, and the two-sided instrument guard is the reason I could check it instead of trusting it — a suite that cannot go red proves nothing, and this one goes red on every mutant I could think to plant.

---
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: da#392 (a prose-vs-tree drift guard) is the mechanism that would have caught the line-number defect fixed in `12669f0`; it does not exist yet, so the correctness of that paragraph currently rests on a human reading it. da#394 (the `pipeline`/`_cmd_enrich` rc=1 leak) is correctly scoped out and named in-code rather than silently carried. Separately, the only production consumer, `noorinalabs-deploy/.github/workflows/deploy-data-load.yml`, still treats any non-zero as "batch load failed", so a succeeded-with-findings load now reports `rc=5` instead of `rc=1` and still fails the job and still skips the post-load `source_corpus` read-back; per the child-repo rule that fix belongs in the deploy repo and needs its own issue. None of these three is charged against this PR.

parametrization:
Requestor: Ivana Horvat
RequestOrReplied: Approved
TechDebt: da#394 (enrich's audit write sits outside any handler, so `pipeline` still leaks rc=1 after a committed load) is knowingly deferred and correctly scoped out. The registry still cannot see a second `IntEnum` whose members carry neither an `EXIT_` prefix nor a registry member name — owned by convention, not mechanism, and documented as such. New: da#392's guard must not be grep-shaped; the retraction paragraph in `src/exit_codes.py` is a quotation, not a citation, and no lexical scan can tell them apart. Evidence posted to da#392.

Reviewed at **`12669f0`** (`12669f057472b76741b86ab7565839c25cb7ae16`). This clears my Changes-Requested at `e4ed0eb`.

## The blocker is fixed, and fixed in the stronger form

`src/exit_codes.py:85` now reads:

```
    src/cli.py   sys.exit(ExitCode.LOAD_FAILED)   # _cmd_load, in the except around load_all
```

You did not write `264`. You removed the class of fact that rots and then recorded *why*, which is more than I asked for and is the right call: the paragraph now carries its own retraction, so the next person to reach for a line number finds the argument against it instead of an invitation to update it.

I checked the retraction's own numbers rather than accepting them, since it is now the artifact making the claim:

```
203a9c3   252
95c2989   258
12669f0   264      "sat at 252, 258 and 264 across the three commits"  -> TRUE
258 -> 264 = 6     "moved it six lines down"                            -> TRUE
```

Whole-file, `95c2989..e4ed0eb` on `src/cli.py` is 11 added / 5 removed, net **+6**. Wanjiku Mwangi measured the block at 10/4, net **+6**. The counts differ by where the boundary is drawn; the net is what the sentence claims, and it is right in both readings. The docstring says the call *moved* six lines down — it never says six lines were added, so there is nothing to correct.

## Re-verified from real processes at the new head

A doc-only change can still break a tree, so I did not carry the `e4ed0eb` result forward. Fresh `git archive 12669f0` (blob check: `git hash-object src/exit_codes.py` = `de1f51a997…` = `git rev-parse 12669f0:src/exit_codes.py`), same harness, real `$?`:

```
RIGGED_CONTROL             1     <- self-test: forces the pre-fix exit through the same path
findings                   5
clean                      0
load_raises                1
audit_boom_clean           0     <- my blocker
audit_boom_findings        5     <- my blocker
post_save_manifest_boom    0     <- my blocker
```

The control still reads `1` while `findings` reads `5`, so the `5`s are a measurement and not a silence. `sys.exit` sites in `src/cli.py` are unchanged; the sole emitter of `1` is `sys.exit(ExitCode.LOAD_FAILED)` at line 264 — in `_cmd_load`, inside the `except` around `load_all`, exactly as the prose now says. 73 tests pass.

**The structural index is a real regeneration, not a hand-edit of the shifted line numbers.** I ran the generator against the exported `12669f0` tree: `files=252 nodes=3151 edges=4969`, and `diff` against the committed `llms.txt` yields **two** lines — line 1 only, where the generator titled the index after my export directory. Adding twelve lines to `exit_codes.py` moved symbol offsets, and the committed index moved with them.

## For da#392 — the guard cannot be a grep, and I can show it

Wanjiku Mwangi is right, and it is worth stating precisely because I am the reason this paragraph exists. A scan for `path.py:NNN` over `src/` at `12669f0` returns exactly one hit:

```
src/exit_codes.py:88:this paragraph wrote ``src/cli.py:258`` -- true of the tree the walk ran against,
```

That is the retraction quoting the false claim it retracts. Your claim of zero surviving citations is correct. To show the scan genuinely cannot discriminate, I planted a real citation and re-ran it:

```
src/exit_codes.py:88   ``src/cli.py:258``        <- quotation, inside the retraction
src/exit_codes.py:281  # see src/cli.py:999      <- planted citation
```

Two hits, identical shape, and nothing lexical separates them. A grep-shaped guard reds forever on the paragraph explaining why the guard exists, which leaves two exits: delete the explanation, or add an exemption. The exemption is a hand-maintained list of blessed prose — the same object as the hand-maintained reserved tuple, the hand-maintained AST node types, and the hand-maintained caller-to-exit-code mapping this module already refused three times. That is the fourth instance of the pattern, and it would be introduced by the guard meant to enforce it.

The way out is the one this module already took: assert the **structural** fact, not the prose. `sys.exit(ExitCode.LOAD_FAILED)` occurring exactly once, inside `_cmd_load`, within an `except` handler, is an AST property — checkable, and it cannot drift when a comment is rewritten. The docstring's claim then reduces to a filename and a function name, both of which an existing guard could verify by name. Nothing needs to read a line number, so nothing needs to be exempted from reading one. Written up on da#392.

## A correction of my own

My `e4ed0eb` review said `3f55c74a..e4ed0eb` spans **22 files**. It spans **32**. I ran `git diff --stat 3f55c74a e4ed0eb -- src/ tests/ docs/` and reported a pathspec-limited count as if it had no filter — the ten I dropped are `.claude/memory/*`, `.env.example`, `.pre-commit-config.yaml`, `ontology/structural/*` and `scripts/mutation_guard.sh`. Two people measured 32 and neither could reproduce me, which is the correct outcome for an unreproducible number. I have amended that comment in place rather than appending a footnote. The conclusion is unchanged and in fact stronger: the wide window sweeps in `main`'s landings, and the clean window on your contribution is `d695104..HEAD` at 7 files.

I spent the whole of this review telling you that a fact nobody can red on will rot. Mine rotted inside a day, in a paragraph about instruments, because I read a filtered output as an unfiltered one. The instrument guard belongs on the reviewer too.

## Summary

`_cmd_load` separates a load that did not happen from a load that happened and has findings, and post-load bookkeeping can no longer forge either. Every claim the registry makes about the tree is now true of the tree, and the one that could not stay true has been replaced by one that can. Approved at `12669f0`.

PR #374, opened 2026-07-09
fix(pipeline): last-loaded manifest must record what was actually loaded

### LINKED ISSUE #350: bug(loader): _cmd_load records pre-load manifest as 'last loaded' — nodes-only run poisons a later --incremental load into skipping edges
## Summary

`_cmd_load` writes the **pre-load** manifest as the **post-load** record, unconditionally — with no `nodes_only` guard. On a `load --nodes-only` run followed by a `load --incremental` run, edge-bearing Parquet files are recorded as "already loaded" and then skipped, so edges never load while the run reports success.

## Trace (`src/cli.py`, head `bf09723`)

- L177 `current_manifest = generate_manifest(data_dir)` — snapshot of **every** Parquet on disk, taken *before* any loading.
- L208-218 `load_all(..., nodes_only=nodes_only, skip_files=skipped_files if incremental else None, strict=False)`.
- L222 `save_manifest(current_manifest, data_dir / LAST_LOADED_MANIFEST_FILENAME)` — **unconditional**. Not gated on `nodes_only`, not narrowed to what `load_all` actually ingested.

So after `load --nodes-only`, `.last_loaded_manifest.json` claims the `network_edges_*.parquet` / `parallel_links.parquet` files were loaded. A later `load --incremental` computes `compare_manifests(current, previous)`, finds them `unchanged`, and passes them in `skip_files` (L193, L216). Edges are silently never loaded. `strict=False` (L213) means nothing raises.

The manifest name says `last_loaded`; the value written is `generate_manifest(data_dir)` — i.e. *last seen on disk*. Those two are only the same thing when the run was a full, non-skipping load.

## Why this is latent today, not live

Both mitigations are incidental, not designed:

1. `noorinalabs-deploy/.github/workflows/deploy-data-load.yml:309` does `rm -rf "${LOAD_DATA_DIR}"` before every run, so the containerized path has zero cross-run manifest persistence.
2. That workflow's `load_args` input only offers full nodes+edges or nodes-only — it never passes `--incremental`.

Remove either mitigation (persist the data dir for a cache, or expose `--incremental` as a `load_args` option) and the hazard becomes live. A local operator running `isnad-ingest load --nodes-only` then `isnad-ingest load --incremental` hits it today.

## Suggested fix

Record what was *loaded*, not what was *present*. Either:

- skip the `LAST_LOADED_MANIFEST_FILENAME` write entirely when `nodes_only` is set (a nodes-only run has no business claiming edge files are current), **or**
- have `load_all` return the set of files it actually ingested in `LoadSummary` and persist only those keys.

The second is the durable fix — it also covers any future partial-load mode and makes `strict=False` safe by construction.

## Provenance

Found during QA review of PR #349 (`save_manifest` best-effort on `OSError`). Pre-existing; **not** introduced or worsened by that PR — if anything #349 reduces the exposure, since a manifest that fails to write cannot poison a later run. Filed separately so #349 is not scope-crept. Re #349, re #348.

parametrization:
## Problem

`_cmd_load` wrote the manifest of **every** input file as "last loaded" regardless of which stages actually ran. A `--nodes-only` run loads no edges, yet still claimed the edge inputs had been loaded. The next `--incremental` run diffed against that claim, found nothing changed, and returned early. **The edges were never loaded, and nothing said so.**

## The defect, reproduced end-to-end against `origin/main`

`tests/test_pipeline/test_manifest_integrity.py` drives `_cmd_load(nodes_only=True)` then `_cmd_load(incremental=True)` over real Parquet fixtures and records whether `load_all` is ever invoked a second time:

```
FAILED TestNodesOnlyDoesNotPoisonIncremental::test_incremental_after_nodes_only_still_loads_edges
E   AssertionError: incremental after nodes-only short-circuited:
E     '=== Load Summary ===\n  Nodes loaded : 2\n  Edges loaded : 0\nNo changes detected. Skipping incremental load.'
E   assert 1 == 2

FAILED TestNodesOnlyDoesNotPoisonIncremental::test_nodes_only_does_not_record_a_last_loaded_manifest
E   AssertionError: a nodes-only run loaded no edges; it must not claim the edge inputs were loaded
E   assert not True

FAILED TestManifestWriteIsAtomic::test_save_manifest_leaves_no_partial_file_on_failure
E   AssertionError: assert PosixPath('.../.last_loaded_manifest.json') is None
```

`Edges loaded : 0`, then `No changes detected. Skipping incremental load.` — that is the poisoning, in the loader's own words.

## Change

The manifest is a change-detection optimisation, so its failure modes are **not symmetric**: re-loading an already-loaded file is free (every loader is an idempotent MERGE), while skipping a file that was never loaded is silent data loss. Every choice below is biased toward re-loading.

- **A partial run records nothing.** `--nodes-only` leaves the previous complete load's manifest in place, because there is no honest "fully loaded" file set for a partial run. The next incremental re-diffs against that older but truthful baseline, and the run says so on stdout instead of staying quiet.
- **`save_manifest` is atomic** — temp file plus `os.replace`. An interrupted or failing write leaves the previous manifest intact rather than a truncated one, and its temp file is cleaned up.
- **`load_manifest` fails safe.** An unparseable manifest reads as `{}`, which makes `compare_manifests` mark every input as *added* and forces a full load. Trusting a damaged file could mark an input "unchanged" and skip it forever. The corruption is logged, not swallowed.
- **A failed manifest write after a successful load warns** that the next incremental will re-load everything, rather than silently doing so.

## Guards against over-correcting

Two tests exist specifically so this fix cannot rot into "always full-load" or into a tautology:

- `test_second_full_load_with_no_changes_is_skipped` — after a *genuine complete* load, an incremental with unchanged inputs still short-circuits. The optimisation is intact.
- `test_full_load_does_record_last_loaded` — the **positive control** for the nodes-only assertion. A complete load *does* write the manifest, so its absence after a nodes-only run is a real behavioural difference rather than an assertion that could never fail.

Fixtures use realistic schema and real Arabic matn (`إنما الأعمال بالنيات…`), not placeholders.

## Verification

- `ENVIRONMENT=test uv run pytest tests/ -q` (whole tree, no `-m` filter) → **1839 passed, 7 skipped**, `PYTEST_RC=0`
- Pre-existing `tests/test_pipeline/test_manifest.py` (17 tests, incl. the da#348 read-only-mount behaviour) still passes against the atomic write
- `ruff format --check` → `FMT_RC=0`; `ruff check` → `LINT_RC=0`; `mypy` → `MYPY_RC=0`
- `python3 scripts/structural_ontology.py check --require-generator` → `CHECK_RC=0`

## Merge-order note

This touches `src/cli.py::_cmd_load` in a different region than #372 (da#354, exit codes), so they should auto-merge. #372 additionally guarantees a *raising* load never reaches this manifest write. Neither PR depends on the other to be correct; whichever lands second may want a trivial rebase.

Closes #350


parametrization:
## `main` merged in — `1db8c47`. Now genuinely mergeable, and CI actually ran.

PR#363 (da#356 + da#376) landed on `main` at `70b66ad` while this branch was open. This PR went `CONFLICTING` / `DIRTY`, and **GitHub runs no checks on an unmergeable PR** — the rollup was frozen at the old head and read as "nothing failing." That is a hard not-ready, not a green. It now reports **7 checks, all SUCCESS**, at `1db8c47`, `mergeStateStatus=CLEAN`.

Merged `main` in rather than rebasing. The only conflict was `ontology/structural/llms.txt`, which is generated: regenerated with `python3 scripts/structural_ontology.py emit` rather than hand-resolved, and `code-graph.json` regenerated alongside rather than trusted to its auto-merge. Regenerating again at this tree produces no diff, and there are no conflict markers anywhere.

I read the merged file rather than trusting the auto-merge, and checked **both** sides survived: this branch's `if nodes_only:` manifest arm at `cli.py:235`, and `main`'s `canonical_surface` fold in `identity.py` / `arabic.py`.

**Re-killed this PR's own mutant on the merged tree**, because `main` changed modules the fixtures exercise. Removing the `nodes_only` arm reds `test_nodes_only_does_not_record_a_last_loaded_manifest` and `test_incremental_after_nodes_only_still_loads_edges`; restoring it returns 8 passed. A suite that passes after a merge does not prove the guard still bites.

Full tree `1875 passed, 7 skipped`. `ruff check`, `ruff format --check`, `mypy src/` clean.

## What is deliberately NOT here yet

The da#359 seam. A quarantining full load is not `nodes_only`, so it takes the `elif` and records a last-loaded manifest for a load that refused rows — and one `--incremental` re-run then erases the only signal that they were refused. Nikolaos Papadopoulos derived that and brought it rather than closing it silently on his side.

The condition is one line and it is mine, in the block I own:

```python
if nodes_only or summary.total_refused:
    # ... do not record a last-loaded manifest
```

It needs `LoadSummary.total_refused`, which lands with da#359 (`31fa84d`). It goes in on the rebase, per the sequence da#384 → #362 → #372 → #359 → #374, together with a third test class asserting the same over a quarantining full load.

`total_refused`, not `total_malformed_ids`: `_load_hadiths` has four `continue` paths and they are two kinds — `invalid source_id` and `DoubledCorpusPrefixError` are refusals, da#191 non-canonical and da#220 matn-dedup are deliberate skips. All four increment `total_skipped`, which is why that cannot be the predicate; and a counter named for one refusal class re-arms the trap for the next one. Nikolaos took the wider version of that argument than I offered him and keyed the exit on refusal too.

parametrization:
## Architecture + data-integrity review — last-loaded manifest records what was actually loaded (closes da#350)

Reviewed at head `a19f05a`, reading code via `gh pr diff` / `gh api contents`.

**The fix is correct.** The defect: `_cmd_load` recorded the manifest of *every* input file regardless of which stages ran, so a `--nodes-only` run claimed the edge inputs were loaded; the next `--incremental` diffed against that claim, found nothing changed, and short-circuited — edges never loaded, nothing said so. The fix makes a partial (`nodes_only`) run record **nothing** and leave the previous full load's manifest in place, so the next incremental re-diffs against a truthful baseline. The bias is right: re-loading is free (idempotent MERGE), skipping a never-loaded input is silent data loss.

**The nesting concern from the brief checks out — twice over.** The rebase placed this branch's `nodes_only`/`save_manifest(...) is None` logic inside main's post-load `try/except` (da#354/#384: must not exit LOAD_FAILED after commit). I verified both guards hold:
1. `save_manifest -> Path | None` genuinely soft-fails — the write is wrapped in `try/except OSError` that logs, cleans up the temp file, and returns `None`; it does not re-raise. So the `elif save_manifest(...) is None` branch is a value check, not an exception path.
2. Even if some non-OSError escaped, the post-load block's own `except Exception:` handler at `src/cli.py` does **not** `sys.exit` — it prints a "load SUCCEEDED and is committed … do NOT read this as a load failure" WARNING and continues. A spurious LOAD_FAILED-after-commit is structurally impossible here. Both intents preserved.

**Two correctness bonuses beyond the ticket, both in the safe direction.** `save_manifest` is now atomic (write to `.tmp` sibling + `os.replace`), so an interrupted write leaves the *previous* manifest intact rather than a truncated one — a truncated manifest is worse than a stale one because leniency about it risks under-reporting changes. `load_manifest` now catches `JSONDecodeError`/`UnicodeDecodeError` and returns `{}`, forcing a full reload (every input classified *added*) rather than trusting a damaged file and skipping inputs. Both biases match the "re-load beats skip" invariant.

**Test quality.** `test_incremental_after_nodes_only_still_loads_edges` drives the whole defect end-to-end (nodes-only → incremental must still invoke `load_all` with `nodes_only=False`). `test_full_load_does_record_last_loaded` is an explicit positive control against the guard being a tautology. `test_save_manifest_leaves_no_partial_file_on_failure` monkeypatches `Path.replace` to raise and asserts the old content survives *and* no `.tmp` debris remains (`list(tmp_path.iterdir()) == [target]`). `test_corrupt_manifest_reads_as_empty_forcing_a_full_load` plants truncated JSON. The `_LoadRecorder` returns `total_edges=0` when `nodes_only` — a realistic stub. Staging fixtures use real columns + real Arabic matn. CI 7/7 green.

Sound. Approving.

---

RequestOrReplied: Approved
TechDebt: none

parametrization:
Independent second review (data-correctness + test-rigor lens). Read at head `a19f05a` via `gh pr diff` and `gh api contents`.

**The manifest write records what was actually loaded, and it is positioned post-commit.** I traced `_cmd_load`: `load_all` runs inside a `try` that exits `LOAD_FAILED` on any raise, so a load that raised reaches neither the manifest write nor the audit entry. Only after `load_all` returns (graph committed, per-batch, no spanning transaction) does the bookkeeping block run. The last-loaded manifest write for a `--nodes-only` run is now skipped entirely, leaving the previous complete load's manifest as a truthful baseline — the exact fix for the poisoning where a nodes-only run claimed the edge inputs were loaded and the next `--incremental` short-circuited on them.

**The asymmetry argument is correct.** Every loader is an idempotent MERGE, so re-loading is free while skipping a never-loaded input is silent data loss. Both fail-safe directions honor that bias: `load_manifest` treats an unparseable file as `{}` (forces a full load) and `save_manifest` is atomic (temp + `os.replace`, temp cleaned on failure), so an interrupted write leaves the prior manifest intact rather than a truncated one.

**The soft-fail cannot mask a real failure silently** — the concern I was asked to check. `save_manifest -> Path | None` returns `None` on `OSError` and logs, and the caller detects `None` and prints a visible WARNING; the outer bookkeeping `try` also prints a loud non-fatal warning on any escape. Crucially the failure direction is safe: a missing last-loaded manifest makes the next incremental re-load everything, not skip.

**Tests genuinely bite.** `test_incremental_after_nodes_only_still_loads_edges` drives the real `_cmd_load` twice over real Parquet fixtures and asserts `load_all` is invoked a second time with `nodes_only=False` — the whole defect end to end, shown red on main. `test_full_load_does_record_last_loaded` is the positive control that makes the nodes-only absence a real behavioural difference rather than a tautology; `test_second_full_load_with_no_changes_is_skipped` guards against over-correcting into always-full-load. The atomicity test monkeypatches `Path.replace` to raise and asserts the previous manifest survives with no `.tmp` debris. Fixtures use real Arabic matn, not placeholders.

CI is CLEAN at head (Test + Integration Tests + sync-drift gate green). No defect found; the fix is correct and the tests exercise it.

---

RequestOrReplied: Approved
TechDebt: none

PR #375, opened 2026-07-09
fix(graph): invalidate stale enrich metrics on load

### LINKED ISSUE #351: bug(loader): a load does not invalidate enrich-written centrality — stale betweenness_centrality survives and reads as current
## Problem

A graph **load** does not recompute — and does not invalidate — centrality properties written by a previous **enrich**. Stale values silently survive on every node that the load MERGE-matched, and they read as current to every downstream consumer.

Observed on stg after run 29035731215 (2026-07-09) loaded the post-da#337 artifact:

```
MATCH (n:Narrator) RETURN count(n), count(n.betweenness_centrality)
->  160614, 150187
```

Exactly the **old** narrator count carries a `betweenness_centrality` value (150,187), and the 10,427 newly created nodes carry none. `150,187 + 10,427 = 160,614`.

Consequence: the betweenness top-10 on stg still shows the pre-da#337 ranking, with the over-merged `أبو عبد الله` at `1.06e7`, roughly 4.6x Abū Hurayra — even though `mention_count` at graph level is now clean and that node has been correctly split into two.

## Why it matters

Nothing anywhere warns that the ranking is stale. A reviewer checking "did da#337 fix the choke-points?" reads the betweenness top-10, sees the old bad answer, and concludes the dedup failed. The inverse is more dangerous: after a *future* load, a stale-but-plausible ranking could be promoted to prod as if it had been recomputed.

The affected properties are all enrich outputs: `betweenness_centrality`, `pagerank`, `community_id`, `in_degree`, `out_degree`.

## Proposed fix

Either (preferred) have the narrator loader **clear the enrich-owned properties** on every node it creates or merges, so a loaded-but-not-enriched graph is unambiguously *absent* centrality rather than *wrong* centrality; or stamp an `enriched_at` / `enrich_source_ref` property that `enrich` sets and `load` clears, and have the API and any ranking query refuse to serve centrality when it is absent or older than the load.

Whichever is chosen, the invariant to establish is: **centrality present in the graph implies it was computed on the topology currently in the graph.**

## Related

Correct operational order is load -> prune (deploy#557) -> enrich. Today nothing enforces or records that order.

parametrization:
## Problem

`load_all` is **MERGE-only** — it never deletes — so properties an earlier `enrich` wrote onto a Narrator survive a reload untouched. A `betweenness_centrality` computed against the previous topology outlives that topology and **reads as current**. Cypher has no strict-property mode, so nothing available to a reader distinguishes a stale value from a fresh one.

## This is live on staging right now

```
MATCH (n:Narrator) RETURN
  count(n.betweenness_centrality) AS bc, count(n.pagerank) AS pr,
  count(n.community_id) AS cid, count(n.in_degree) AS ind, count(n.out_degree) AS outd;
-- 150187, 150187, 150187, 150187, 150187      (graph holds 160,614 Narrator)
```

All five metrics sit on exactly **150,187** nodes while the graph holds **160,614**. The 150,187 carry centrality from *before* the last reload; the 10,427 nodes that reload created carry none. A consumer ranking narrators by centrality cannot tell them apart. (150,187 + 10,427 = 160,614 — the `created=10,427 merged=118,807` load recorded in `reference_graph_ops_cypher_shell`.)

The property-name trap that hides this was re-confirmed on the same graph:

```
MATCH (n:Narrator) RETURN count(n.betweenness) AS misspelled_returns_zero;  -- 0, no error
```

A misspelled property returns NULL for every row and never errors. So keys were enumerated with `keys(n)`, not guessed.

## Change

The loader removes the topology-derived properties it has invalidated: `betweenness_centrality`, `community_id`, `in_degree`, `out_degree`, `pagerank` — exactly the set `src/enrich/metrics.py` writes via GDS `writeProperty:`.

### Why removal, not an enrich-generation stamp

The issue offers three primitives (invalidate / stamp / refuse to serve). Removal, because:

- **A stamp only protects consumers that opt in to checking it.** Every current consumer reads `n.betweenness_centrality` directly, so a stamp leaves them serving stale numbers — it documents the bug rather than fixing it.
- **Absence is the only value a reader cannot mistake for a measurement.** A missing property and a misspelled one both yield NULL and neither errors — the same trap above. Removal converts a silent wrong answer into a loud missing one.
- **`enrich` is a re-runnable stage.** `make enrich` recomputes.
- "Refuse to serve" lives in the API (`noorinalabs-isnad-graph`), not the loader.

The trade is deliberate and stated in the code: after a load without a follow-up enrich, centrality is **visibly absent rather than confidently wrong**. A visible gap is recoverable; a silent wrong number is not. The loader prints the count, the property names, and the remediation instead of doing it quietly.

### Scope is exactly the topology-derived set

- **Invalidated:** the five GDS metrics. Every one is a function of graph structure, so every one is falsified by a load that changes that structure. `nodes_only` invalidates too — adding nodes changes centrality just as edges do.
- **Left alone:** `topic_1…topic_3_score` on Hadith come from the **matn text**, not the graph structure. A reload changes topology; it does not change what a hadith is about. Also `trustworthiness`, `generation`, `death_year_ah` — rijal-source-derived, not GDS. Tests pin both exclusions.

### Non-termination made unrepresentable

The `WHERE` predicate and the `REMOVE` clause are generated from **one tuple**. If they could drift, a predicate matching a property the `REMOVE` did not clear would select the same rows forever and the batch loop would never terminate. `test_predicate_and_remove_clause_cannot_drift` pins it. (Same single-authoritative-source principle as `feedback_dual_detector_cross_sect_authority`.)

Batched in Python rather than `CALL { … } IN TRANSACTIONS`, which needs an implicit transaction and is rejected inside the explicit write transaction `Neo4jClient.execute_write` opens.

## Tests — shown RED against a real Neo4j

This is a new capability, so a bare import error would be a weak red. Instead the invalidation **call** was neutered (`invalidated_narrators = 0`), which is precisely what `origin/main` does, and the integration test run against a real Neo4j container:

```
FAILED tests/integration/test_graph_loading.py::TestStaleEnrichMetricsAreInvalidatedByLoad::test_stale_centrality_does_not_survive_a_reload
E   AssertionError: stale centrality survived the reload on 5 narrator(s)
E   assert 5 == 0
```

The integration test plants metrics, asserts the probe **can** see them (`before > 0` — otherwise the post-reload zero would be an inert assertion, not an observation), reloads, then asserts they are gone and enumerates `keys(n)` rather than probing a name. It also asserts the load's own data (`name_ar`, `id`) survives, so the invalidation is surgical rather than a blunt wipe.

## Verification

- `pytest tests/ -m "not integration"` → **1803 passed, 7 skipped**, `UNIT_RC=0`
- `pytest tests/integration/ -m integration` (real Neo4j via testcontainers) → `INTEG_RC=0`
- `ruff check` → `LINT_RC=0`; `ruff format --check` → `FMT_RC=0`; `mypy src/graph/ src/cli.py` → `MYPY_RC=0`
- `python3 scripts/structural_ontology.py check --require-generator` → `STRUCT_RC=0`

## Follow-up worth considering (not done here)

Once this lands, the next prod/stg load will strip centrality until `make enrich` runs. The reload runbook should sequence `load` → `enrich` explicitly. I did not touch `src/enrich/**` or the deploy workflows; happy to file that if you want it tracked.

Closes #351


parametrization:
## Architecture + data-integrity review — invalidate stale enrich metrics on load (closes da#351)

Reviewed at head `ac9bd74`, reading code via `gh pr diff` / `gh api contents`.

**The fix is correct and closes a genuine silent-staleness hole.** `load_all` is MERGE-only and never deletes, so `betweenness_centrality` (and the other GDS-written metrics) an earlier `enrich` wrote survive a reload and read as current — Cypher has no strict-property mode, so a reader cannot tell a stale value from a fresh one. Removal (not a generation stamp) is the right architecture: a stamp only protects opt-in consumers, and every existing consumer reads `n.betweenness_centrality` directly, so a stamp would leave them serving stale numbers. Absence is the only value a reader cannot mistake for a measurement. Scope is correctly limited to topology-derived properties — topic labels (matn-derived) and source-derived properties (`trustworthiness`/`generation`) are deliberately left alone.

**Termination and drift are both handled structurally.** `_STALE_PREDICATE` and `_REMOVE_CLAUSE` are generated from one `TOPOLOGY_DERIVED_NARRATOR_PROPERTIES` tuple, so a matched node is always fully cleared and therefore stops matching next batch — the "predicate matches a property REMOVE didn't clear → infinite loop" hazard is unrepresentable. The Python batch loop (`WITH n LIMIT $batch_size … REMOVE … RETURN count`) is used instead of `CALL {} IN TRANSACTIONS` because the latter needs an implicit tx and is rejected inside `execute_write`'s explicit tx — correct reasoning. `removed == 0` breaks; an empty result set also returns 0 and terminates (`test_empty_result_set_terminates`). Property names are `.isidentifier()`-validated at import before interpolation.

**Placement in `load_all` is right.** Invalidation is step 3, after nodes+edges load and before validation, unconditional of `skip_validation` (it's a data-correctness op, not a check) and run for `nodes_only` too (adding nodes shifts centrality just as edges do). It cannot run if the load raised, since control never reaches it. `LoadSummary` gains `invalidated_narrators` (defaults 0), surfaced in the CLI summary and structured logs — "an invalidation nobody is told about is a quieter version of the bug."

**Test quality is strong.** The Neo4j-testcontainer integration test plants a stale value on every Narrator, asserts `before > 0` as an explicit instrument check (so the post-reload `after == 0` is a measurement, not a silent zero), reloads, and asserts `after == 0`, `summary.invalidated_narrators == before`, that the property keys are gone via `UNWIND keys(n)` enumeration (dodging the misspelled-property-returns-NULL trap called out in `reference_graph_ops`), and that surgical data (`name_ar`, `id`) survives. `test_invalidation_is_idempotent_on_an_unenriched_graph` is the positive control. Unit tests pin scope, drift, batch summing/termination, idempotence, and batch-size passthrough. No circular import (`invalidate.py` imports only `src.utils.*`). CI 7/7 green.

Sound. Approving.

---

RequestOrReplied: Approved
TechDebt: none

parametrization:
Independent second review (data-correctness + test-rigor lens). Read at head `ac9bd74` via `gh pr diff` and `gh api contents`.

**Removal is the right primitive.** `load_all` is MERGE-only, so GDS metrics an earlier enrich wrote survive a reload and read as current — and Cypher has no strict-property mode, so a stale `betweenness_centrality` is indistinguishable from a fresh one (the same NULL-on-misspelled trap in `reference_graph_ops`). A stamp only protects opt-in consumers; absence is the only value a reader cannot mistake for a measurement. The five invalidated properties are exactly the topology-derived GDS set, with topic labels (matn-derived) and rijal-source properties correctly left alone and pinned by tests.

**Non-termination is made unrepresentable.** The `WHERE` predicate and the `REMOVE` clause are both generated from one tuple, so a predicate matching a property the REMOVE didn't clear — which would spin the batch loop forever — cannot occur. Property names are validated as identifiers before interpolation. Batched in Python (correctly, since `CALL {} IN TRANSACTIONS` is rejected inside the explicit write tx), idempotent, and it runs after the load committed and only if the load didn't raise. Placed before validation, so validation sees the invalidated state — right ordering.

**The integration test actually observes the invalidation, not just calls it** — the concern I was asked to check. It plants metrics, asserts `before > 0` first (instrument check, so the post-reload zero is a measurement and not an inert silent-zero), reloads against a real Neo4j testcontainer, asserts `after == 0` AND `summary.invalidated_narrators == before`, then enumerates `keys(n)` rather than probing a name (guarding the misspelled-property NULL trap) and asserts the load's own `name_ar`/`id` survive so the wipe is surgical. The RED-on-main demonstration neutered the call to `invalidated_narrators = 0` (what main does) and got `stale centrality survived on 5 narrator(s)` — a genuine red for the reason under test, not a bare import error.

**Unit tests cover the loop exhaustively** — scope, no-drift, safe interpolation, sum-and-stop-on-zero (3 productive + 1 terminating probe), unenriched-removes-nothing, idempotence, batch-size pass-through, and empty-result termination.

CI is CLEAN at head (Test + Integration Tests + sync-drift gate green). No defect found; the fix is correct and the tests exercise it.

---

RequestOrReplied: Approved
TechDebt: none

PR #399, opened 2026-07-10
resolve(bio_promote): re-key alias lookup to cleaned canonical form (da#389)

### LINKED ISSUE #389: bio_promote: alias rows anchor on the raw pre-clean name; 22.1% of itqan aliases (34,046/154,328) are silently dropped
Found by **Kavitha Sundaramurthy** in review of PR #383; she scoped it out of that PR correctly and asked for its own issue. Filing it so it is not lost, with her figure verified and decomposed.

`bio_promote` promotes a bio under the **cleaned** name:

```python
cleaned = clean_narrator_name(norm)   # :185
cid = make_canonical_id(cleaned)      # :190
```

but the alias pass anchors on the **raw, pre-clean** name:

```python
norm = safe_str(row.get("canonical_name_ar_normalized"))
rec = canonical_map.get(make_canonical_id(norm))   # :101 — different key
if rec is None:
    continue   # alias silently dropped
```

So whenever the cleaner changes a bio's name, the alias rows that point at that bio look up an id nobody promoted, `rec is None`, and the aliases are dropped without a counter or a log line.

Measured against `data/staging/narrator_aliases_itqan.parquet` (the only alias table today):

| outcome for an alias row | rows | share |
|---|---:|---:|
| anchor resolves | 120,282 | 77.9% |
| **anchor re-keyed by the cleaner** | **31,326** | 20.3% |
| **anchor bio dropped entirely (`clean_narrator_name` → `None`)** | **2,720** | 1.8% |
| total | 154,328 | |

Kavitha's `31,326` is the re-keyed class exactly. The `2,720` dropped-anchor rows are a second class: the bio was never promoted at all, so no key could have matched. Both end in a silently discarded alias; only the first is fixable by keying the alias pass on `clean_narrator_name(norm)` the way the promotion does.

**22.1% of itqan's alias rows never reach a canonical record**, and nothing in the pipeline says so. That is the same shape as the counters in #385: a deliberate-looking silence that is indistinguishable from "there were no aliases."

Suggested:
1. Key the alias lookup on `make_canonical_id(clean_narrator_name(norm))`, matching the promotion path, so the two agree by construction rather than by coincidence.
2. Count both classes separately on the `bio_promote` metrics — `aliases_anchor_rekeyed` and `aliases_anchor_dropped` — beside the existing `aliases_added`.
3. A test that a bio whose name the cleaner rewrites still collects its aliases. It reds today.

Not a regression from #383 — the mismatch predates it and #383 does not touch the alias pass. Related: #385.

parametrization:
## What

`bio_promote._merge_aliases` anchored each name-variant on `make_canonical_id(canonical_name_ar_normalized)` — the bio's **raw** normalized name — while `promote_bios_to_canonical` mints the canonical node under `make_canonical_id(clean_narrator_name(name))`. When the cleaner rewrote a bio's name (a benediction suffix, a `، ويقال :` variant list, a truncated head), the two ids diverged and the alias looked up an id **nobody promoted** → it attached to nothing, silently.

Same shape as da#371 / da#379, one layer over: the alias key must pass through the identical cleaner the promoter used.

## Measurement (real itqan corpus, `data/staging/narrator_aliases_itqan.parquet`, 154,328 rows)

| | count | % of aliases |
|---|---|---|
| RAW-key attaches (pre-fix) | 120,282 | 77.9% |
| **Dropped by the bug** | **34,046** | **22.1%** |
| CLEANED-key attaches (post-fix) | 151,608 | 98.2% |
| — newly recovered | 31,326 | +20.3pp |
| — residual, **pollution-dropped bio** (no node minted, explained) | 2,720 | 1.8% |
| — **unexplained** | **0** | — |

Every one of the 34,046 drops is either recovered (31,326) or is an alias whose bio the cleaner dropped as pollution (2,720) — so no canonical node was ever minted to anchor it. Both classes accounted for; zero unexplained.

## Fix (zero-inference — re-keys, does not extract)

- Re-key the alias lookup through the **same** `clean_narrator_name` the promoter applies, then gate the attach on `promoted_cids` (the set of canonical ids a bio actually promoted into this run). This mirrors the promotion decision exactly: a pollution-dropped bio mints no node, and a truncated-residue bio the da#379 guard refused to merge onto an attested narrator does **not** smuggle that narrator an alias through this pass.
- `_merge_aliases` now returns `_AliasMergeStats` — every alias row is bucketed (`added` / `unmatched_pollution` / `unmatched_no_bio` / `skipped_source` / `skipped_duplicate` / `skipped_empty`), logged on completion, with a `warning` when `unmatched_no_bio > 0`. The drop can never again be silent.

## Tests (red-first, verbatim itqan fixtures)

- `test_alias_rekeyed_to_cleaned_form_attaches` (itqan:635) — RED pre-fix (`aliases_added=0`, alias attached to nothing because raw-key ≠ cleaned-key); GREEN after.
- `test_merge_aliases_counts_recovered_and_pollution_dropped` (itqan:635 + itqan:7532) — both classes in one run: `added==1`, `unmatched_pollution==1`, `unmatched_no_bio==0`. RED pre-fix: `_AliasMergeStats` did not exist and the merge returned a bare int.

Red-first proof was run by restoring `HEAD:src/resolve/bio_promote.py` and confirming both tests fail, then restoring the fix. Full `tests/test_resolve/` suite: 471 passed, 6 skipped. mypy strict clean (100 files). Structural ontology index regenerated.

Closes #389


parametrization:
Reviewed at head `98f872c083f691f19daa9671de6e138ce95ae17b` — reading the code via `gh api contents`, not a checkout. I reviewed da#379/#383 in this module, and this is the correct next layer over that guard.

**Zero-inference — holds.** `_merge_aliases` re-keys the lookup through `clean_narrator_name(norm)` then `make_canonical_id(cleaned)`: the identical cleaner and identity function the promotion loop mints under. It gates the attach on `cid not in promoted_cids`, the exact set of ids a bio actually promoted into this run. Nothing extracts or guesses a name; no threshold appears anywhere. A truncated-residue bio the da#379 guard refused `continue`s before `promoted_cids.add(cid)`, so its alias falls into `unmatched_no_bio` and is never smuggled onto the attested narrator — the harm this fix is careful not to reintroduce.

**Both-class instrument is real, on verbatim rows.** The fixtures are verbatim itqan `name_ar`: `itqan:635` is a `، ويقال :` variant-list the cleaner truncates to its head nasab (raw id != cleaned id — the assertion the test makes explicit), and `itqan:7532` is a `وقيل :` chain the cleaner returns `None` for (no node minted). `test_merge_aliases_counts_recovered_and_pollution_dropped` runs both in one pass and asserts `added==1`, `unmatched_pollution==1`, `unmatched_no_bio==0` — recovery and the explained residual measured together, not a single number read in isolation.

**Never silent again.** `_AliasMergeStats` buckets every alias row across six exhaustive counters, all six are logged on `bio_promote_complete`, and a warning fires when `unmatched_no_bio > 0`. The 22.1% silent drop cannot recur unbucketed.

**Red-first is structurally guaranteed, not merely claimed.** Test 1 asserts its premise `make_canonical_id(raw_norm) != make_canonical_id(cleaned)`, so on HEAD's raw-key lookup the alias attaches to nothing and the `aliases == [alias]` assertion fails; test 2 imports `_AliasMergeStats`, which does not exist pre-fix, so it errors at import. The instrument demonstrably separates the two classes.

**Scope is clean.** The diff is `bio_promote.py`, its test, and the regenerated structural ontology index only. `identity.py` / `make_canonical_id` are untouched, so there is no collision with da#371 in flight.

I also confirmed `rec = canonical_map[cid]` cannot KeyError: it is reached only when `cid in promoted_cids`, and every cid entering that set is guaranteed a `canonical_map` key (added right after the node is created or merged, and the truncation-skip path `continue`s before the add). CI is green across all seven checks — Test, Type Check, fixture-realism, and the sync-drift gate included.

---

Requestee: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

parametrization:
Reviewed at head `98f872c083f691f19daa9671de6e138ce95ae17b`, reading the code and tests via `gh api` at that ref.

The fix itself is right and I want to be clear about that. The re-key routes the alias's raw `canonical_name_ar_normalized` through the same `clean_narrator_name` the promoter mints under, recomputes the id, and gates the attach on `promoted_cids` — the exact mirror of the promotion decision. Zero-inference holds: no threshold, no name guessing, and a da#379-refused truncated-residue bio never lands in `promoted_cids`, so its alias cannot smuggle onto an attested narrator. Scope stays inside `bio_promote.py`; `identity.py` / `make_canonical_id` are untouched. CI is green across all seven checks. I traced every branch of `_merge_aliases` — the six buckets are mutually exclusive and exhaustive, one bucket per row via continue-guards. The recovered-vs-pollution separation is proven red-first with verbatim itqan fixtures (itqan:635 recovered, itqan:7532 cleaner-dropped), and the re-key test is genuinely red pre-fix because raw-key and cleaned-key mint different ids.

My one blocking finding is on the measurement, which is my lens. The whole PR is justified by a guarantee — the drop can never again be silent — and the number it reports is zero unexplained. The mechanism that delivers both is the investigate bucket and the warning it gates. That bucket is only ever asserted at zero (twice), and no test drives it to a positive value. `test_alias_without_matching_bio_is_dropped` may execute the increment incidentally but asserts only the end-state aliases list, never the counter and never the log. So if the increment line were deleted or the warning branch broken, every test still passes. The instrument is shown to separate the added class from the pollution class, but never the one class whose zero the headline rests on — a control that cannot fail is not a measurement.

The ask is small and self-contained: add one red-capable test that produces a positive count in the investigate bucket — an alias whose cleaned id is absent from `promoted_cids` (a truncated-merge-skipped bio, or simply an alias with no promoted bio) — and assert both the counter equals the expected positive value and the warning fires (via caplog). That validates the never-silent guarantee end to end and lets the zero-unexplained reading be trusted, since the detector will have been shown able to register a non-zero.

---

Requestor: Ivana Horvat
Requestee: Kwesi Boateng
TechDebt: none

parametrization:
Re-reviewed the fixup at head `d5e7f9b`. The delta is test-only (+78 lines); production `bio_promote.py` is unchanged from my prior pass and still correct.

`test_unmatched_no_bio_is_counted_and_warns` resolves my finding fully. It drives the investigate bucket positive with an orphan alias whose cleaned canonical name is a real narrator that no bio promoted — cleanly distinct from the pollution case — and asserts the exact buckets by a direct call (unmatched_no_bio one, unmatched_pollution zero, added zero), then witnesses end-to-end that the warning branch actually fires with the right event and count. The instrument now demonstrably separates the class whose zero the headline rests on, so the zero-unexplained reading is trustworthy.

I checked the warning-spy reasoning the author gave. `src/utils/logging.py` configures structlog with a PrintLoggerFactory, which writes to stdout rather than through stdlib logging, so pytest's log-capture fixture would witness nothing here — the direct spy on the module logger's warning method is the correct mechanism, and because it patches the same logger object the production call resolves, it intercepts the real invocation and cannot false-pass. Red-first is structurally guaranteed as well: the three-positional-argument merge call TypeErrors against the pre-fix two-argument signature. CI is green across all seven checks.

Finding cleared. Approving.

---

Requestor: Ivana Horvat
Requestee: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

Issue #366, opened 2026-07-09
sanadset: ~122,000 of the 159,558 "No SANAD" rows carry a matn-embedded isnad — needs a two-convention splitter, NOT the NER fallback


parametrization:
## Summary

Of `sanadset`'s 165,701 rows with a null `isnad_raw_ar`, **159,558 have no `<SANAD>` tag and a `Sanad` column reading literally `No SANAD`.** That annotation is the dataset author's claim, not a measurement — and it is wrong for most of them. **≈122,000 of these rows carry an isnad embedded in the matn text.**

They are a **different population from #368** (6,143 rows whose chain sits, already segmented, in the unused `Sanad` column). #368 is deterministic recovery. This issue needs a validated isnad/matn splitter, and is therefore **not launch-gating**.

## The reconciled number, and the probe

Four numbers were proposed before this one, and each failed for a nameable reason:

| number | claim | why it failed |
|---|---|---|
| ~160,815 | recoverable | probe = "contains a transmission phrase," using `عن`/`قال`, which fire on **81.3% of pure matn**. Could not have returned low. |
| 159,558 | *legitimately chainless* | took the dataset's own `Sanad = "No SANAD"` annotation as ground truth. |
| 126,398 | recoverable | `159,558 × 0.792` — no false-positive correction, and assumes detector recall = 1. |
| ≥106,088 | recoverable | correct in kind, but the term set omitted **`ثنا`** (the abbreviated `حدثنا`) and missed an entire editorial convention. |

### There are two isnad conventions, and no single probe sees both

| corpus | isnad opens with a receipt verb | isnad chains with `عن` |
|---|---|---|
| `lk` | **98.8%** | 93.3% |
| `tusi` | **0.2%** | **98.5%** |
| `sanadset` (tagged rows) | **0.0%** | (short chains, often 1 `عن`) |

`lk` writes `حدثنا فلان، حدثنا فلان، عن فلان`. `tusi` and `sanadset`'s tagged rows write `علي بن إبراهيم عن أبيه عن ابن أبي عمير` — a bare name, then `عن`. **An opener-only probe is blind to the second convention; a `عن`-only probe is blind to nothing but is noisy.** Both prior estimates used opener-only.

Note also: `sanadset` **strips the opener from every row it successfully tags** (0/485,285 tagged isnads begin with one). The opener survives *only* in the untagged rows. That is why it is such a strong positive signal there, and why its **absence proves nothing**.

### The detector

[...]

PR #400, opened 2026-07-10
fix(parse/sanadset): recover 6,143 dropped isnads from the Sanad column (da#368)

### LINKED ISSUE #368: parse(sanadset): 6,143 real isnads discarded — parser reads the <SANAD> tag but never the Sanad column
## Summary

`src/parse/sanadset.py:465` extracts the isnad **only** from a `<SANAD>…</SANAD>` tag inside the `Hadith` markup column. The raw CSV also carries a dedicated **`Sanad` column**, which the parser never reads. For 6,143 hadiths that column holds a real narrator chain while the *opening* `<SANAD>` tag is absent — so the isnad is **silently discarded** and the hadith lands in the graph chainless.

## Mechanism: the `<SANAD>` tag is PRESENT but EMPTY

**Corrected 2026-07-09 (twice).** An earlier revision of this body — mine — said these rows carry an *orphaned `</SANAD>` closing tag with no opening tag*. **That is wrong.** Jean-Claude read the markup correctly and I did not. Measured across all 650,986 raw rows, partitioning by tag shape:

```
openTag  closeTag  Sanad column        count
  False     False   "No SANAD"       159,558    <- no tag at all  (issue #366)
   True      True   real isnad         6,052    <- tag present, EMPTY   THIS ISSUE
   True     False   real isnad            91    <- tag opened, never closed
                                       -------
                                       165,701  == staged null isnad_raw_ar, exact
```

So `_extract_tag(_SANAD_RE, full_text)` returns a **falsy empty string**, not `None`-from-absence. The guard at `:470`

```python
if sanad_text and sanad_text.lower() != "no sanad":
```

short-circuits on the empty string and the `!= "no sanad"` comparison **never executes** — which is exactly why a grep for `<SANAD>No Sanad</SANAD>` finds zero rows and why anyone grepping for a *missing* tag will call this unreproducible.

The chain was never missing. It sits in the `Sanad` column, already segmented and diacritized.

## Measurement (real `data/raw/sanadset/sanadset.csv`, 650,986 rows)

```
rows with NO opening <SANAD> tag  : 165,701   (== staged null isnad_raw_ar, exact match)
  Sanad column literally "No Sanad": 159,558
  Sanad column HAS a real isnad    :   6,143   <-- THIS ISSUE, silently discarded
    ...of which orphaned </SANAD>  :   6,052   (98.5%)
```

`<SANAD>No Sanad</SANAD>` never occurs (0 rows), so the `sanad_text.lower() != "no sanad"` guard at line 470 never fires — the drop is entirely the missing opening tag.

Samples of discarded chains (`Sanad` column, diacritized, stringified Python list):

```
Num_hadith=8   ['أَبِي', 'أَبِيهِ', 'عَلِيِّ بْنِ أَبِي طَالِبٍ']
Num_hadith=26  ['أَبُو بَحْرٍ مُحَمَّدُ بْنُ الْحَسَنِ', 'عَبْدُ اللَّهِ بْنُ جَعْفَرٍ', …]
Num_hadith=59  ['أَبُو أَحْمَدَ', 'أَبُو بَكْرِ بْنُ مَالِكٍ', 'عَبَّاسُ بْنُ حَمْدَانَ', …]
```

The format differs from the `<NAR>`-tagged markup `_extract_narrator_mentions` consumes: it is a **list literal of already-segmented names**. Recovery needs its own path, not a regex reuse.

## The other 159,558 are a different issue (#366), and they are NOT chainless

Those rows have **no `<SANAD>` tag at all** and their `Sanad` column reads literally `No SANAD`. It is tempting to call them legitimately chainless on that basis. **The dataset's own annotation is wrong.** Sampling rows whose `Sanad` column says `No SANAD`:

```
ثنا إبراهيم بن الحسين بن علي الكسائي ، قال : ثنا آدم بن أبي إياس العسقلاني ،
قال : ثنا شيبان ، عن قتادة ، عن الحسن ، في قوله : " والنازعات غرقا …
```

Four named transmitters joined by `ثنا`/`عن`. That is a chain.

Measured against a **control group** — the 485,285 rows that *do* carry a real `<SANAD>` tag, whose `<MATN>` is chain-free by construction:

| probe on matn head | control | test (`No SANAD`) | lift |
|---|---|---|---|
| loose token (`عن` / `قال`) | **85.6%** | 88.5% | ~1.0× — vacuous |
| strict opener (`حدثنا` / `أخبرنا` / `سمعت`) | 7.9% | **79.2%** | **10×** |

`عن`/`قال` saturate ordinary Arabic prose, so a detector keyed on them returns ~high on any Arabic text. On the strict opener the lift is real.

**Corrected again 2026-07-09.** An earlier revision of this section — mine — multiplied out `159,558 × 0.792 = 126,398` and called the residual `~33,160 genuinely chainless`. **Both numbers are withdrawn.** They assume the detector's *recall* is 1 and apply no false-positive correction, and neither assumption is supported. When recall was actually checked against the positive class it came back **0 / 485,285**: the source **strips the transmission opener from every row it tags**, so a tagged isnad begins with a bare narrator name and the opener survives *only* where tagging failed. That is why its presence is such strong evidence and why **its absence proves nothing**. Recall is not calibratable on this data; an independent `lk` text-match oracle yielded an overlap of 1 row.

**Superseded again 2026-07-09 (final).** `≥106,088` was *also* too low, for the opposite reason: it was an **opener-only** probe, and the corpus has **two isnad conventions**. `lk` opens with a receipt verb (98.8%); `tusi` and sanadset's own tagged rows open with a **bare narrator name and chain with `عن`** (98.5% of tusi; **0.0%** receipt verbs in sanadset's tagged rows). An opener-only probe is structurally blind to the second convention — which is the one this corpus uses.

Calibrating `D = head_opener OR (≥2 standalone عن in the first 25 tokens)` on the positive **and** negative class of **each** convention (`lk` 95.0% / 5.3%; sanadset-tagged 39.9% / 3.1%) gives, on the 159,558: opener 107,694 · ≥2 `عن` 56,839 · **D = 123,202 (77.2%)** · FPR-corrected **≈122,013**.

What is defensible: **≈122,013** rows carry a matn-embedded isnad, and that is **still a lower bound** — D's sensitivity on sanadset's own bare-name convention is only 39.9%, so short chains remain invisible to it. It is **not boundable from above** with any instrument we have. **The chainless count is not measurable** — it needs annotation, or the splitter itself.

Tracked in **#366** (≈122,013 rows, lower bound). Recovery needs a validated isnad/matn splitter and must **not** be done by re-enabling the `ner.py` fallback (#369). Note that not every hit is clean — `حدثنا : " من لا يرحم الناس … "` is a bare opener with no names — which is precisely why #366 is **not** in the launch gate and this issue is.

**This issue is the zero-inference subset**: the segmentation already exists on disk.

## Impact

- 6,143 hadiths currently enter the graph with no isnad chain. These are recoverable, fully-segmented chains — arguably *higher* quality than the tag-derived ones, since segmentation is given rather than inferred.

## Why this blocks the re-run

`parse` is upstream of `resolve` (~7.5h). Per `feedback_sweep_expensive_stage_before_launch`, a defect at or before the expensive stage costs nothing to fix now and costs the whole stage if found after launch. This is cheap, upstream, and carries **no quality risk** — which is exactly why it rides along while #366 does not.

## Acceptance

- [ ] Parser reads the `Sanad` column when the opening `<SANAD>` tag is absent, parsing the list-literal form into `isnad_raw_ar` (or directly into mentions).
- [ ] A row with neither a usable `Sanad` value nor an opening tag remains `isnad_raw_ar = None`.
- [ ] The orphaned `</SANAD>` closing tag is detected and **logged**, not silently tolerated.
- [ ] Test fixture uses the **production filename and header** (`sanadset.csv`, `Hadith,Book,Num_hadith,Matn,Sanad,Sanad_Length`) and a row copied verbatim from the real CSV — including the orphaned `</SANAD>` — so the guard cannot be inert (`feedback_fixture_makes_guard_assertion_inert`).
- [ ] Post-fix count of `isnad_raw_ar IS NOT NULL` for sanadset is `485,285 + 6,143 = 491,428`.

## Related

- Part of noorinalabs/noorinalabs-main#928 (re-run defect ledger)
- #366 — the ≈122,013 matn-embedded chains (splitter-dependent, NOT in the launch gate). Any splitter must be **validated against `lk` AND `tusi`** — one example of each convention — or it will pass on receipt-verb chains and fail silently on the bare-name majority.
- #365 — NER routing allowlist; `tusi`/`halimbahae`/`bihar`/`mis` stranded
- #352 — remove the `ner.py` matn fallback
- `src/parse/sanadset.py:500` has an adjacent `except Exception: malformed_count += 1` swallow on mention extraction

parametrization:
## Summary

`parse(sanadset)` extracted the isnad only from the inline `<SANAD>…</SANAD>`
markup. For **6,143** rows that markup is unusable and `_extract_tag(_SANAD_RE, …)`
returns falsy, so the chain was silently discarded and the hadith landed in the
graph chainless — even though the chain sits **pre-segmented in the dedicated
`Sanad` column** as a Python list literal of diacritized names.

This reads the `Sanad` column as a fallback whenever the `<SANAD>` tag path
yields nothing. **Zero-inference**: it reads an already-segmented column, not
free text — no NER / matn fallback (da#369) is invoked.

## Measured on the real 650,986-row CSV

| metric | value |
|---|---|
| `isnad_raw_ar IS NOT NULL` | 485,285 → **491,428** (+6,143) |
| recovered from `Sanad` column | **6,143** (6,052 empty tag + 91 unclosed) |
| still null (`No SANAD`, #366) | 159,558 |
| column-derived mentions | 63,163 |

The +6,143 lands exactly on the acceptance target `485,285 + 6,143 = 491,428`.

## What changed

- `_parse_sanad_column` — parse the list-literal column; returns `None` for the
  `No SANAD` sentinel, blank, free text, or any non-list literal (never invents a
  chain).
- `_mentions_from_names` — build mentions from the pre-segmented names, filtered
  through the **same `_is_narrator_like` gate** the `<NAR>` path uses. sanadset is
  a Phase-1 mention source (`resolve/ner.py`), so these flow to resolve.
- `_process_chunk` — fallback to the column when `_extract_tag(_SANAD_RE)` is
  falsy; store the raw column value verbatim in `isnad_raw_ar`; **log** a
  present-but-unusable (empty / orphaned / unclosed) `<SANAD>` tag instead of
  silently tolerating it. A usable tag still wins — the column is never consulted
  for the 485,285 good rows.

## Red-first test evidence

`TestSanadColumnRecovery` uses the **production filename + header**
(`Hadith,Book,Num_hadith,Matn,Sanad,Sanad_Length`) and two rows **copied
verbatim from the real CSV**, including the orphaned `</SANAD>`
(`feedback_fixture_makes_guard_assertion_inert`). On the pre-fix source **5/6
new tests fail**; all pass after. The instrument test asserts the not-null count
moves `0 → full population (2)` while the `No SANAD` control stays null
(`feedback_silent_zero_is_not_a_measurement`).

Full suite: **2043 passed, 10 skipped**. ruff-format/-lint, mypy strict,
structural-ontology staleness all green.

Closes #368
Part of noorinalabs/noorinalabs-main#928 (re-run defect ledger)


parametrization:
Reviewed at head `2b3c3444a2450fe07ae08b365070f7ce6b40686a` — read `src/parse/sanadset.py` and `tests/test_parse/test_sanadset_parser.py` via `gh api` contents (no checkout), traced the recovery branch against the diff rather than the PR body.

**Approve.** This lands the da#368 recovery cleanly and the QA lens finds no defect.

Fixture realism (the load-bearing check): `TestSanadColumnRecovery` writes the production filename `sanadset.csv` with the exact production header (Hadith, Book, Num_hadith, Matn, Sanad, Sanad_Length) via `csv.writer`, and the two recoverable rows are copied verbatim from the real CSV — the orphaned closing-tag row (Num_hadith 1814) and the twice-open never-closed row (2947), each keeping its populated list-literal chain column. Tracing the SANAD regex against both, neither yields a well-formed open→close capture, so `_extract_tag` returns falsy and the new column branch is the only thing that recovers them. On the pre-fix source those rows drop (and `_parse_sanad_column` does not exist), so 5 of the 6 new tests are genuinely red-first; the valid-tag test passes on both because the tag path pre-existed — the 5/6 the PR claims. No inert guard here.

Two-sided instrument confirmed: the instrument test asserts the not-null count moves 0 → the full population of 2 while the mirror `No SANAD` control row stays null, so the guard separates the two classes rather than blanket-filling. The control can genuinely fail.

Zero-inference confirmed: `_parse_sanad_column` is a pure `ast.literal_eval`→list read that rejects the `No SANAD` sentinel, blank, `None`, free text, `[]`, and any non-list literal — no NER or matn fallback (da#369 stays disabled). The `<SANAD>` tag still wins whenever usable, so the 485,285 good rows are untouched (the valid-tag control ignores a deliberately-different column value and its mention still comes from the `<NAR>` tag).

Population definition confirmed: the branch keys on a falsy `_extract_tag` result (empty / unclosed / orphaned markup), not on tag-absence. The 159,558 tag-absent rows also enter the else branch but their column is the sentinel, so they correctly stay null; recovery counts only rows carrying a real list-literal column — the 6,052 empty + 91 unclosed = 6,143. The +6,143 → 491,428 figure is an author full-corpus measurement I verified mechanistically (correct row class targeted), not by re-running the 650,986-row CSV.

CI is fully green (Lint, Type Check, Test, Integration Tests, staleness-check, fixture-realism, sync-drift gate all SUCCESS at head).

---

Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Approve** — da#368 column-fallback recovery. Reviewed at head `2b3c344` against the diff (not the PR body), with the downstream resolve path co-verified.

**Zero-inference / correctness.** `_parse_sanad_column` is a pure column read: `ast.literal_eval` of the value, rejecting the `No SANAD` sentinel (case-insensitive), blank, free text, and any non-`list` literal by returning null rather than coercing — it can never mint a chain. `_mentions_from_names` builds mentions through the exact `_is_narrator_like` gate the `<NAR>` path uses, so recovered mentions carry the same quality as tag-derived ones. The tag path strictly wins: the column is consulted only when `_extract_tag(_SANAD_RE, …)` is falsy or the tag literally reads `No SANAD`, so the 485,285 already-good rows never touch the column. The `test_valid_tag_row_ignores_the_column` control proves this with a deliberately divergent column value that is correctly ignored.

**Instrument separates classes (ML lens).** The red-first fixtures are two rows copied verbatim from the real `sanadset.csv` (Num_hadith 1814 orphaned-close, 2947 open-no-close), so on pre-fix `sanadset.py` `_SANAD_RE` finds no well-formed open→close pair and both rows drop to a null isnad — the tests genuinely fail before the fix. `test_instrument_count_moves_from_zero_to_full_population` moves the not-null count 0 → the full expected population of 2 while the mirror `No SANAD` control row stays null; it is a genuine separator, not a control-that-cannot-fail. No inert assertion.

**Downstream soundness.** sanadset is a `_PHASE1_MENTION_SOURCES` corpus, so `resolve/ner.py` loads its mentions straight from `narrator_mentions_sanadset.parquet` via the load-phase-1 path; the isnad-reading extract-from-hadiths path runs only for thaqalayn / open_hadith / fawaz / sunnah. The list-literal string stored verbatim in `isnad_raw_ar` for the 6,143 recovered rows is therefore never re-fed to the segmenter — no inference is re-introduced. The recovered mentions still pass the da#247 name-quality gate downstream, matching every other mention.

Storing the raw list-literal as `isnad_raw_ar` is the correct zero-inference choice (reconstructing Arabic prose would be inference); its cosmetic appearance as a graph property is pre-launch UI, not a data defect. CI is fully green (Lint, Type Check, Test, Integration, base-image/fixture-realism, sync-drift, structural staleness). Full suite 2043 passed.

---

Requestor: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

PR #402, opened 2026-07-10
fix(parse): fail loud on the da#355 producer gate instead of exiting 0 (da#386)

### LINKED ISSUE #386: parse_all swallows the da#355 producer gate: 're-run parse' is a no-op that exits 0
Found while reviewing PR #359 at `9c017d97dabd8558a5b33afa48da634a247ce25b`.

## The remediation text prescribes a no-op

da#355 adds a producer gate: `generate_source_id` now asserts the id grammar and raises `DoubledCorpusPrefixError` when `collection == corpus`. And `_cmd_load` (`src/cli.py:281`) now tells the operator, on a non-zero exit:

> This is a producer defect: re-run `parse` to regenerate the staging artifact, then re-load.

Following that instruction does nothing. `src/parse/__init__.py:44` wraps every adapter in `except Exception as exc: logger.error("parse_failed", ...)` and then reaches `return results`. `DoubledCorpusPrefixError` subclasses `ValueError` subclasses `Exception`, so the producer gate is swallowed, `parse` exits **0**, the stale `data/staging/hadiths_sanadset.parquet` is never regenerated, and the next `load` quarantines the same 650,986 rows and exits non-zero again.

The operator is told to perform an action that reports success and changes nothing.

## Why it matters now

At `9c017d97`, 650,986 of 853,218 staging hadith rows (76.3%) carry the doubled prefix — 100% of `sanadset`, 0% of the other eight corpora. da#355's load-side exit code is what stops that from being a silent 76.3% drop. But the *recovery path* it names is inert, so an operator who does exactly what the error says will loop.

## What this is not

This is **not** #360. That issue is `src/resolve/__init__.py: run_all` swallowing stage exceptions in the **resolve** orchestrator, and it is the reason `MissingDependencyError` and `StopAfterReached` had to subclass `BaseException`. This is `src/parse/__init__.py: parse_all` — a different module with the same shape. #360's own closing argument applies verbatim here: two workarounds for the same orchestrator behaviour is a signal the orchestrator is wrong.

## Acceptance

- `parse_all` must not swallow a grammar assertion. Either narrow the handler so `DoubledCorpusPrefixError` (or a producer-defect base class) propagates, or make `parse` exit non-zero when any adapter failed.
- A test that runs `parse` over a fixture whose producer raises the grammar assertion, and asserts a **non-zero** exit. Red before the fix.
- Once `parse` can fail, re-verify the `src/cli.py:281` remediation text end-to-end: `parse` → `load` on a doubled-id artifact must reach a state where the second `load` succeeds.

Related: #359 (the load-side gate and the remediation sentence), #382 (`collection_coverage` reports `passed=True` over the same missing 650,986 rows), #360 (the sibling swallow in `run_all`).

parametrization:
## What & why

`src.parse.run_all` wrapped every adapter in a broad `except Exception`. The da#355 producer gate (`generate_source_id` raising `DoubledCorpusPrefixError` when `collection == corpus`) subclasses `ValueError` → `Exception`, so it was **swallowed**: run_all logged `parse_failed`, set `results[name] = []`, and returned normally. `parse` therefore exited **0** with the stale `hadiths_sanadset.parquet` untouched — making da#359's `_cmd_load` remediation ("re-run `parse`") a **silent no-op**. The operator does exactly what the error says, `parse` reports success, nothing changes, and the next `load` refuses the same 650,986 rows.

## The fix

Narrow the orchestrator — **not** another `BaseException` workaround (da#360's own closing argument: two workarounds for one orchestrator behaviour means the orchestrator is wrong):

- `run_all` now catches `DoubledCorpusPrefixError` separately, **purges any parquet the failing adapter wrote this run** (a partial output set — snapshot diff), records the per-source defect, and raises `ParseProducerError` **after** the summary so the whole invocation fails loudly. Genuinely transient adapter failures are still logged and skipped — the narrowing is limited to the producer gate.
- `_cmd_parse` maps `ParseProducerError` to a new named `ExitCode.PARSE_PRODUCER_DEFECT = 10` (9 is da#369's `UNROUTED_CORPUS`). A bare `sys.exit(1)` / uncaught traceback is the anti-pattern da#394 tracks; the exit-code registry is the one owner of the space, so the pinned enumeration + ruling table are updated.

### Deliberate scope boundary
The purge is scoped to files this run touched. It does **not** re-encode the load-side `{kind}_{slug}.parquet` glob to guess a *pre-existing untouched* stale file — a hand-kept parallel filename list is the exact fragility the adapter/exit-code registries exist to abolish (cf. `narrators_bio_kaggle.parquet`, which carries no slug suffix). A legacy stale file is overwritten by the next **successful** parse of its source; until the upstream data defect (the da#353 CSV-stem `collection == corpus` fallback) is fixed, the non-zero exit is what stops it being read as fresh. This is the issue's own model: the gate makes the shape *unproducible*, forcing the upstream fix.

## Red-first evidence
Against `origin/main`, `run_all` with a producer-defect stub **returned `{'sanadset': []}` without raising** (parse would exit 0) — the swallow, reproduced. New `tests/test_parse/test_run_all_producer_defect.py` covers:
- run_all raises `ParseProducerError` on a producer defect (was: returns)
- the per-source summary still prints before the raise
- partial writes from the failing adapter are purged
- a pre-existing untouched file is left in place (documents the boundary)
- clean sources keep their output
- a transient (`RuntimeError`) failure is still swallowed — narrowing preserved
- `_cmd_parse` exits `PARSE_PRODUCER_DEFECT`

## Verification
Full local CI mirror green: ruff-format, ruff-lint, gitleaks, actionlint, cspell, mypy-strict, **pytest (2046 passed, 10 skipped)**, structural-ontology-staleness. Structural index regenerated.

Closes #386


parametrization:
Verified against the diff at HEAD `58f165c`, not the PR body.

**Producer error genuinely escapes the swallow.** `run_all` catches `DoubledCorpusPrefixError` in a clause placed *before* the broad `except Exception` — since it subclasses `ValueError`, the specific handler wins and the defect is recorded rather than logged-as-`parse_failed`. The `ParseProducerError` raise sits after the summary print, outside both the `try` and the adapter loop, so nothing in `run_all` re-catches it; `_cmd_parse` maps it to a named exit. This is the da#360 class handled correctly — no plain-`Exception` re-swallow.

**Exit code = 10, no collision.** `ExitCode.PARSE_PRODUCER_DEFECT = 10`; the pinned enumeration dict in `test_members_are_enumerated_not_grepped` carries `10`, `test_the_ruling_table_is_pinned` asserts `== 10` (with the "9 taken by da#369 UNROUTED_CORPUS" rationale), and the docstring ruling table reads 10. `@enum.unique` + `__new__`'s reserved-value guard make a duplicate inexpressible. `9` is not present in this tree (da#369 not yet rebased in), so there is no double-book.

**Purge boundary is honest.** `_purge_partial_writes` removes only staging parquet whose `(mtime_ns, size)` differs from the pre-adapter snapshot — i.e. exactly what this run touched. `test_pre_existing_untouched_file_is_left_in_place` proves a legacy `hadiths_sanadset.parquet` the adapter never touched survives; the non-zero exit, not a filename guess, is what prevents it being read as fresh. The stated refusal to re-encode the load-side `{kind}_{slug}.parquet` glob is implemented as described.

**Red-first + narrowing preserved.** The producer-defect stub makes `run_all` return `{'sanadset': []}` pre-fix and raise after; `test_transient_failure_is_still_swallowed` and `test_transient_failure_alongside_producer_defect` pin that a `RuntimeError` is still logged-and-skipped and does not suppress the producer raise. The narrowing is limited to the gate, not a blanket re-raise.

Two non-code notes, neither a correctness defect: the PR **body prose** says "= 9" while the code, pinned test, and ruling table all correctly say **10** (stale wording only); and the PR is currently `CONFLICTING` with an empty check rollup — a rebase-and-green merge gate for the orchestrator, not a defect in this diff.

---

Requestor: Ivana Horvat
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

parametrization:
Independent second review of the da#386 fix, read at HEAD `58f165c` via `gh api` (no checkout).

The substantive fix is correct on every axis I was asked to verify:

- **Fail-loud is narrow, not a blanket re-raise.** `run_all` catches only `DoubledCorpusPrefixError` (the da#355 producer gate) into a `producer_defects` map — recording it, purging this-run writes, and continuing the loop — then raises `ParseProducerError` after the parse summary. The broad `except Exception` still logs `parse_failed` and skips (`results[name] = []`). `_cmd_parse` maps the new exception to the named code and exits non-zero, so `parse` can no longer report success with a stale staging parquet in place and da#359's "re-run parse" remediation is real again.
- **No stale artifact; purge is scoped to this run.** `_purge_partial_writes` removes only staging parquet whose `(mtime_ns, size)` differs from the per-adapter pre-run snapshot. A pre-existing untouched legacy file is deliberately left (proven by `test_pre_existing_untouched_file_is_left_in_place`), and a clean earlier source keeps its output (`test_clean_sources_keep_their_output`). The non-zero exit forcing the upstream re-run is the model that holds.
- **Value 10, no duplicate on this branch.** `@enum.unique` plus the pinned `PARSE_PRODUCER_DEFECT == 10` and the ruling table agree; the comment correctly records that 9 belongs to da#369's UNROUTED_CORPUS.
- **Red-first holds.** The new tests import `ParseProducerError` (which did not exist pre-fix) and assert a raise / exit 10 that the pre-fix swallow-and-return-`{'sanadset': []}` tree cannot satisfy.

The blockers are integration, not the fix logic, and they are real:

- The branch base predates da#369. Its `exit_codes.py` carries no member 9 at all, while `main` now defines `UNROUTED_CORPUS = 9`. The PR is `mergeable_state: dirty` as a result.
- `test_members_are_enumerated_not_grepped` pins the WHOLE enum by exhaustive equality (`by_name == {...}`) with no `"UNROUTED_CORPUS": 9` entry. After the required rebase onto current main, that assertion reds unless the pinned dict is updated to include 9 alongside 10.
- The CI check rollup is empty (zero checks on this head), so a green CI cannot be certified.

Requested: rebase onto current `main` so member 9 is present, add `"UNROUTED_CORPUS": 9` to the exhaustive pin in `test_members_are_enumerated_not_grepped`, and confirm a green rollup. The 10-not-9 design is exactly right and needs no change once the base is current.

---

Requestee: Oyunbileg Batbayar
TechDebt: none

parametrization:
Re-review at head `aeeacca` after the rebase. Every item from my Changes-Requested is cleared, and the substantive da#386 fix survived the rebase intact.

- **Stale base resolved.** `exit_codes.py` now carries BOTH `UNROUTED_CORPUS = 9` (da#369) and `PARSE_PRODUCER_DEFECT = 10`; the member set is {1, 3, 4, 5, 6, 7, 8, 9, 10} — no duplicate value, and `@enum.unique` is applied, so a collision would raise at import.
- **Exhaustive pin matches the post-rebase member set.** `test_members_are_enumerated_not_grepped` now asserts equality against a dict that includes both `"UNROUTED_CORPUS": 9` and `"PARSE_PRODUCER_DEFECT": 10` — it will no longer red against current main.
- **Rollup genuinely green, not empty.** 7 checks, all SUCCESS (Lint, Type Check, Test, Integration Tests, staleness-check, base-image/fixture realism, pre-commit⇄CI sync-drift). `mergeable_state: clean`.
- **Fix logic unchanged.** `run_all` still catches only `DoubledCorpusPrefixError` → purge this-run writes → raise `ParseProducerError` after the summary; the broad `except Exception` still logs and skips; `_cmd_parse` still maps to `ExitCode.PARSE_PRODUCER_DEFECT`.

Clears my finding.

---

Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

PR #387, opened 2026-07-09
feat(cli): one module owns the exit-code space; a duplicate is inexpressible (da#384)

parametrization:
## Problem

Two open PRs each declared `= 4` in a file the other did not contain — `EXIT_MISSING_DEPENDENCY` (#362 / da#309) and `EXIT_VALIDATION_FINDINGS` (#372 / da#354). **Both suites are green, because neither can see the other's constant.** The only guard was a hand-written reserved tuple, which reddens on the collision somebody already thought of and is blind to every sibling nobody remembered to add.

`_cmd_pipeline` runs acquire → parse → resolve → load → enrich in **one process**, so the exit-code space is global. After both land, `isnad-ingest pipeline` exiting `4` would mean either *"resolve is missing its ML dependency group"* or *"load succeeded and validation has findings."* That is the conflation da#354 exists to remove, one level up.

## Change

Not a renumbering. **A duplicate becomes inexpressible.**

- **`src/exit_codes.py`** — an `@enum.unique` `IntEnum` owns every `EXIT_*`. `IntEnum` silently accepts a duplicate as an *alias*; the decorator raises `ValueError` instead.
- **`_checkpoint.py`** stops declaring `EXIT_STOPPED_AT_LIMIT` and re-exports it by import. *An import is not a declaration*, which is what keeps the registry sole.
- Module-level `EXIT_*` aliases are exported, so a migrating call site changes its **import site, not its value** — the property #362's two approvals rest on.

**When it raises, precisely:** *when the registry module is imported, and every `EXIT_*` consumer imports it.* Not "at process start" — `src/cli.py` imports `src.resolve` and `src.graph` lazily inside command functions, so absent an eager import a duplicate would surface on first use of the offending command. Pytest collection imports it, so CI catches it either way. (Oyunbileg Batbayar — "at import time" is the same imprecision this issue is about.)

| rc | constant | band |
|---|---|---|
| `0` | — | success |
| `1` | `LOAD_FAILED` | nothing ran |
| `2` | *argparse* | nothing ran |
| `3` | `STOPPED_AT_LIMIT` | frozen |
| `4` | `MISSING_DEPENDENCY` | **a stage aborted mid-pipeline; prior artifacts on disk** |
| `5` | `VALIDATION_FINDINGS` | stage completed, output suspect |
| `6` | `REFUSED_ROWS` | stage completed, output incomplete |

[...]

PR #403, opened 2026-07-10
da#369: remove NER matn fallback; hard-fail on unrouted staged corpus

### LINKED ISSUE #369: NER silently mines the matn for narrator names when isnad_raw_ar is null — 380,771 matn-derived mentions (11.7%), 38,262 narrators that exist only there
## Summary

`src/resolve/ner.py:143-145` silently falls back to mining the **entire hadith body** for narrator names when a corpus's `isnad_raw_ar` column is empty:

```python
isnad_text = safe_str(table.column(isnad_col)[i].as_py())
# Fall back to full text if isnad is missing.
if not isnad_text:
    isnad_text = safe_str(table.column(text_col)[i].as_py())
```

The matn is not an isnad. Every sentence in it is offered to the narrator segmenter. The result is **380,771 matn-derived narrator mentions — 11.7% of all 3,249,721 mentions** — and **38,262 canonical narrators that exist only because of them**, against 46,200 seen on any real isnad. That is 45% of the 84,462 mentioned narrators.

Unlike the `fawaz` case in #352, **most of this lands on hadiths we keep.** It is in the production graph right now.

## Which corpora are affected, and why it is not simply "the null ones"

`isnad_raw_ar` null profile across all nine staged corpora:

| corpus | rows | null isnad | pct |
|---|---|---|---|
| `sanadset` | 650,986 | 165,701 | 25.5% |
| `halimbahae` | 62,169 | 62,169 | **100%** |
| `fawaz` | 36,512 | 36,512 | **100%** |
| `thaqalayn` | 33,190 | 33,190 | **100%** |
| `sunnah_scraped` | 10,710 | 10,710 | **100%** |
| `mis` | 7,748 | 7,748 | **100%** |
| `bihar` | 394 | 394 | **100%** |
| `tusi` | 17,421 | 332 | 1.9% |
| `lk` | 34,088 | 16 | 0.0% |

But nullness does not decide who reaches the fallback — **a hardcoded allowlist does** (`ner.py:34-59`, iterated at `:259`, `:264`, `:269`, `:274`). `sanadset` and `lk` take a parse-time pre-extraction path and never invoke NER; `halimbahae`, `bihar`, `mis`, `tusi` are routed nowhere at all (#365). So the fallback fires for exactly three corpora:

| corpus | matn-derived mentions | share of that corpus's mentions | route |
|---|---|---|---|
| `fawaz` | 222,875 | 100% | NER over `full_text_ar` |
| `thaqalayn` | 154,114 | 100% | NER over `full_text_ar` |
| `sunnah` | 3,782 | 100% | NER over `full_text_en` |
| **total** | **380,771** | **11.7% of all mentions** | |

Name-quality corroborates the split, with no reference to any `canonical_narrator_id` (so it is stable under #356). No narrator's name exceeds five tokens:

| corpus | mentions | names > 5 tokens | narrator seen exactly once | route |
|---|---|---|---|---|
| `sanadset` | 2,666,631 | 7.8% | 0.4% | isnad (parse) |
| `lk` | 202,319 | 5.8% | 1.4% | isnad (parse) |
| `fawaz` | 222,875 | 9.6% | 5.3% | **matn (fallback)** |
| `thaqalayn` | 154,114 | **15.1%** | **10.9%** | **matn (fallback)** |
| `sunnah` | 3,782 | 7.3% | **27.0%** | **matn (fallback)** |

## Proof that the fallback itself is the defect, independent of corpus

`thaqalayn` has no reference edition to check against. So the fallback was run **in simulation against `lk`** — the one corpus where both paths are computable on the same rows (`isnad_raw_ar` and `full_text_ar` both populated).

On 3,000 sampled `lk` rows (of 33,240 eligible; sample cap stated, nothing else dropped):

| | |
|---|---|
| mentions/hadith from `isnad_raw_ar` — what the pipeline does today | **5.49** |
| mentions/hadith from `full_text_ar` — what the fallback does | **7.91** |
| inflation | **1.44×** |
| extra names introduced | 8,606 (2.87 per hadith) |
| …substring of the **matn** | **53.7%** |
| …substring of the **isnad** | **0.1%** |
| …neither (segmentation-boundary variants) | 46.1% |

*Instrument check:* were the fallback harmless, inflation would be ~1.00× and extras ~0.

The corresponding measurement on `fawaz` against its `lk` twin (30,748 text-identical pairs): of 32,838 extra mentions, **84.5% are substrings of `lk`'s matn and 0.4% of its isnad**. *Instrument check:* the same substring probe finds 155,994/188,344 (82.8%) of `lk`'s own extracted names inside `lk`'s own `isnad_raw_ar`, so it can return a hit; the 0.4% is real.

What this produces, concretely — the highest-`mention_count` "narrators" attested only via the fallback:

```
25  مالك الامر عندنا انه لا يتوضا من رعاف ولا من دم ولا من قيح يسيل من الجسد …
19  الدباء والمزفت
17  هذا من رسول الله صلى الله عليه وسلم
16  ارايت اذ اوينا الى الصخره، فاني نسيت الحوت …
```

These are sentences from hadith bodies, indexed as people.

## Why a silent fallback is the wrong shape

A corpus with no isnad column is a **data-completeness fact**. Responding to it by extracting from a different field is a silent substitution of one thing for another, and it degrades quietly: nothing logs it, nothing fails, and the damage only surfaces as strange narrators in the UI 380,771 mentions later.

This is the same class as #359 — *stop silently repairing bad input; start asserting.* The `_collapse_double_corpus` rewrite there and this fallback here both **guess at intent when an invariant is violated, and cannot tell a legitimate case from a defect.** Related instances this week: `restore.sh` verifying zero checksums and reporting success, and `dedup` writing an empty parquet on `ImportError`.

## Suggested fix

1. **Delete the fallback.** A corpus whose `isnad_raw_ar` is null yields **no narrator mentions**, loudly.
2. **Replace the allowlist with an explicit per-corpus contract.** Each staged corpus declares its extraction route and whether its `full_text` embeds an isnad. A staged corpus that declares nothing is a **hard error**, not a silent skip (#365).
3. **Give the corpora that genuinely embed an isnad a parse-stage splitter** so they populate `isnad_raw_ar` properly, and flow through the correct path: `fawaz` (`src/parse/fawaz.py:104-108`, `:141-145` hardcode `isnad_raw_ar: None`), `thaqalayn`, `sunnah_scraped`. `halimbahae` (99.6% of rows carry a head-position transmission phrase) and `bihar` (75.1%) qualify too, but are currently unrouted and **must not be routed before the splitter exists**, or we import this defect wholesale.

Note that fixing #365 by naively adding the unrouted corpora to `_ARABIC_SOURCES` would take `halimbahae`'s 62,169 rows and `bihar`'s 394 straight through this fallback. **#365 must not land before this issue does.**

## Scope and sequencing

This is **upstream of `resolve`, in `parse`** — `resolve` cannot recover an isnad/matn boundary that `parse` destroyed, and `load` can only choose whether to keep matn fragments. **It therefore blocks the launch re-run.** It is the same upstream fix #317 requires.

#352 must land before the resolve re-run, and its effect must be measured **after** #356 lands — #356 changes the narrator node set, this changes the mention set feeding it, and measuring them together would attribute one bug's residue to the other.

## Provenance

Found while investigating #352; scope corrected from "fawaz-only" after the drop-set probe used there was shown to be incapable of surfacing non-orphaning instances. Full analysis: the revised design note on #352. Companion issues: #364, #365, #366.

Tracking: noorinalabs-main#928.

All figures above are computed from raw parquet, per-corpus routing sets, and text substring tests — **stable under #356**, which changes only `canonical_narrator_id`.

parametrization:
## da#369 — remove the NER matn fallback; hard-fail on an unrouted staged corpus

Closes #369 · wave-24 (#928 defect-sweep) · **quality risk: REDUCES risk** (removes an inference source, adds none)

### What changed
- **Removed the full_text (matn) fallback** in `src/resolve/ner.py::_extract_from_hadiths` — the path that mined `full_text_ar` for narrator names when `isnad_raw_ar` was null. In prod it produced **380,771 matn-derived pseudo-narrator mentions (11.7%)** and **38,262 narrators that existed ONLY in matn text**. NER now reads only the dedicated isnad column; a null isnad is skip-and-counted.
- **NER extraction is per-corpus opt-in + hard-fails on an unrouted staged corpus.** `run()` discovers every staged corpus by reading its `source_corpus` column (filename-independent) and raises `UnroutedCorpusError` **before any mention output is written**.
- `UnroutedCorpusError` subclasses **`BaseException`** — exactly like `MissingDependencyError` / `StopAfterReached` — so it escapes `run_all`'s per-stage `except Exception` (a plain `Exception` is swallowed as "step failed" and the pipeline reports success). The CLI maps it to a new dedicated exit code **`EXIT_UNROUTED_CORPUS` (`ExitCode.UNROUTED_CORPUS = 9`)**, owned by `exit_codes.py` (registry member + alias + pinned tests updated).

### Sequencing (critical)
Routing is **not** fixed here — this PR only removes the blunt fallback and installs the loud hard-fail. **da#365 lands AFTER** and layers the correct per-corpus routing allowlist on top of this guard. `fawaz` stays *routed* to Arabic (routing untouched) but, lacking a real isnad column, now yields 0 mentions until da#365 gives it a surgical isnad-splitting route. The da#271 fawaz-via-full_text tests are repurposed to encode the new contract with da#365 breadcrumbs.

### Red-first evidence
Against the **pre-fix tree** (fallback restored, guard removed), these fail; against the fix they pass:
- `TestArabicExtraction::test_does_not_mine_full_text_when_isnad_null` — null-isnad + populated full_text_ar → `[]`
- `TestFawazArabicExtraction::test_fawaz_full_text_no_longer_mined` — fawaz full_text no longer mined → `[]`
- `TestUnroutedCorpusHardFail::test_unrouted_staged_corpus_raises` — a staged unrouted corpus (`bihar`) → `UnroutedCorpusError`, no output written

### Local CI mirror (all green)
`ruff format --check` · `ruff check` · `mypy src/` (100 files) · `pytest` (2039 passed, 10 skipped) · structural-ontology index regenerated · pre-commit + pre-push hooks passed.

Suggested reviewers: **Alejandra Reyes-Fuentes**, **Kwesi Boateng**.


parametrization:
Reviewed at head `511c4f8e7b180e833e07dca8514ad4ac07cef0b3`, reading each file at HEAD via `gh api` (no checkout). Verified against the diff, not the PR body.

**1. The matn mine is fully gone.** `_extract_from_hadiths` reads only the dedicated isnad column (`isnad_raw_ar`/`isnad_raw_en`); a null isnad is `null_isnad_count += 1; continue` — skip-and-counted, not silently dropped. No path reads `full_text_ar` for names anywhere in the module. The prod pollution source (380,771 matn-derived mentions / 38,262 matn-only narrators) has no code path left.

**2. The hard-fail escapes the swallow (my da#360 domain).** `UnroutedCorpusError` subclasses `BaseException`, not `Exception` — confirmed both in the class def and pinned by `test_unrouted_error_is_base_exception` (`issubclass(..., BaseException)` and `not issubclass(..., Exception)`). The NER step in `run_all` is wrapped in `except Exception:  # noqa: BLE001`, so a plain `Exception` here would be exactly the da#360 swallow — logged as a failed step while the pipeline marches on and reports success. Because it is a `BaseException`, it propagates through that guard out of `run_all` to `_cmd_resolve`'s `except UnroutedCorpusError` → `sys.exit(EXIT_UNROUTED_CORPUS)`. Same posture `StopAfterReached`/`MissingDependencyError` already rely on. It is raised at the TOP of `run()` (right after the start log), before any `all_rows` accumulation or `write_parquet`; `test_unrouted_staged_corpus_raises` asserts `narrator_mentions_resolved.parquet` does not exist after the raise. No partial artifact.

**3. Exit-code integration.** `ExitCode.UNROUTED_CORPUS = 9` is a real enum member (next free after 8; no collision, and the 4/5 space is untouched), with the `EXIT_UNROUTED_CORPUS` alias, both added to `__all__`, re-exported through `src/resolve/__init__.py`. `@enum.unique` + `__new__`'s reserved-value guard keep a duplicate inexpressible per the #387 contract. The pinned tests are updated in both places that enumerate the ruling table (`test_members_are_enumerated_not_grepped` and `test_the_ruling_table_is_pinned`).

**4. Red-first.** All three fail on the pre-fix tree and pass after: `test_does_not_mine_full_text_when_isnad_null` (was `assert len(rows) > 0`, now `assert rows == []`); `test_fawaz_full_text_no_longer_mined` (was asserting Arabic names present, now `assert rows == []`); `test_unrouted_staged_corpus_raises` (pre-fix `run()` only iterated the fixed route sets so a staged `bihar` was ignored with no error — `pytest.raises(UnroutedCorpusError)` would have failed).

**5. Sequencing.** Routing is untouched — `fawaz` stays in `_ARABIC_SOURCES` (`test_fawaz_still_routed_to_arabic`), so it is a routed-but-silent corpus, not an unrouted hard-fail. The loud guard is left in place for da#365 to layer per-corpus routing on top. fawaz/halimbahae going to 0 mentions post-fix is the intended, documented contract (they need da#366's validated splitter), not a regression.

**6. CI.** `gh pr view 403 --json statusCheckRollup` — all 7 checks SUCCESS (Lint, Type Check, Test, Integration Tests, staleness-check, base-image/fixture realism, sync-drift gate).

No defects. The BaseException escape mechanism is correct and is precisely the shape da#360 exists to prevent.

---

RequestOrReplied: Approved
TechDebt: none

parametrization:
Reviewed at head `511c4f8e7b180e833e07dca8514ad4ac07cef0b3`, reading the diff and files via `gh api`.

**Reduces risk, adds none.** The change deletes the `full_text_ar`/`full_text_en` fallback from `_extract_from_hadiths`; a null isnad column now takes the `null_isnad_count += 1; continue` skip-and-count path. There is no new extraction, transliteration, or guessing anywhere in the diff — `_discover_staged_corpora` only reads the `source_corpus` column to decide routing, it never touches matn text. A null isnad is skipped and counted, never matn-mined.

**Hard-fail is correct and total.** `run()` computes the staged corpora and raises `UnroutedCorpusError` at the very top, before `all_rows` is built or any parquet is written — `test_unrouted_staged_corpus_raises` confirms `narrator_mentions_resolved.parquet` does not exist on the failure path. Discovery reads each staged file's `source_corpus` column (filename-independent), so a mis-named file cannot hide an unrouted corpus. `UnroutedCorpusError` subclasses `BaseException` (not `Exception`) exactly like `MissingDependencyError`/`StopAfterReached`, so it genuinely escapes `run_all`'s per-stage `except Exception` (NER step handler) rather than being swallowed and reported as a successful resolve; the CLI's `_cmd_resolve` catches it and exits the new dedicated `EXIT_UNROUTED_CORPUS` (9), which `exit_codes.py` registers under `@enum.unique` with the pinned-table tests updated.

**The fawaz consequence is understood and intended.** fawaz stays routed to Arabic (routing is untouched — da#365's domain) so it does not hard-fail, but with a null isnad column and matn-mining removed it now yields 0 mentions. Real-isnad recovery correctly waits for a validated per-corpus splitter, not a re-enabled matn mine — consistent with #928. The da#271 fawaz tests are repurposed, not dropped: `test_fawaz_still_routed_to_arabic` pins the preserved routing and `test_fawaz_full_text_no_longer_mined` encodes the new zero-mention contract with da#365 breadcrumbs.

**Red-first holds by construction.** `test_does_not_mine_full_text_when_isnad_null` and `test_fawaz_full_text_no_longer_mined` assert `rows == []` where the pre-fix fallback would have mined the populated `full_text_ar`; `test_unrouted_staged_corpus_raises` asserts a raise the pre-fix `run()` (which only iterated fixed route sets) could not produce.

CI is clean — all seven checks SUCCESS (Lint, Type Check, Test, Integration, fixture-realism, sync-drift gate, structural staleness), 2039 passed.

---

RequestOrReplied: Approved
TechDebt: none

PR #404, opened 2026-07-10
resolve: discriminated stage result — non-zero exit on raised stage + detector provenance (da#360, da#378)

### LINKED ISSUE #360: resolve: run_all swallows every stage exception and still reports resolve_pipeline_complete (exit 0)

### LINKED ISSUE #378: resolve: make "zero-row output indistinguishable from a true negative" unrepresentable — discriminated stage result (sequence with #360)
## Summary

Three subsystems produced the same bug in one week:

| Site | Failure | Reported as |
|---|---|---|
| `dedup.py` (#309, fixed in PR #362) | declared ML dep absent | zero-row `parallel_links.parquet` = "no parallels found" |
| `run_all` (#360) | any stage raises | `resolve_pipeline_complete`, exit 0 |
| `ner.py:143-145` (found by Jean-Claude) | isnad column null | mines the matn, reports success |

One shape: **a failure condition handled by returning a success-shaped value.** The caller cannot distinguish "ran and found nothing" from "could not run".

A lint can flag the syntactic family. It cannot decide the semantic question — and that limit is the point of this issue.

## Why a lint is the wrong tier here

`ruff`'s `BLE001` covers only the `except Exception` subset. It would **never** have caught #309, whose handlers were *specific* (`except ImportError`). (`BLE001` is also currently disabled and pre-suppressed at 33 sites — that is #377, a separate, narrower problem.)

More fundamentally, "success-shaped" is not statically decidable. An empty list is the *correct* answer sometimes. `src/resolve/parallels.py:353` already proves it:

```python
log = logger.warning if indexed and not records else logger.info
```

It returns an empty result and **discriminates**: "had candidates, found none" is a warning; "had no candidates" is info. No lint can tell that apart from `dedup.py`'s old `return _write_empty_output(staging_dir)` on `ImportError` — both are "a function returning an empty container."

`org feedback_enforcement_hierarchy` says prefer hook > skill > charter. This is the case where even a **hook is the wrong tier**, because the property a hook would check is undecidable from syntax. The right tier is **a type**: make the illegal state unrepresentable rather than detectable.

### 1. A discriminated stage result

Replace the bare `list[Path]` / `Path` / `None` returns from resolve stages with an explicit sum type:

```python
Ok(rows: int, outputs: list[Path])       # ran to completion
Empty(reason: TrueNegative)              # ran; genuinely nothing to emit
Degraded(reason: str, missing: list[str])# could not run as configured (explicit opt-in)
# and: raise for "could not run and nobody said that was OK"
```

Then:

- A caller *cannot* read a `Degraded` as an `Empty` — it must handle both arms.
- `run_all` cannot report `resolve_pipeline_complete` while holding a `Degraded`, because the type forces the branch.
- The CLI's exit status becomes a function of the result type, not of whether an exception escaped.
- `mypy --strict` (already enforced in CI) makes an unhandled arm a **build failure**, not a review comment.

PR #362 already gestures at this — `MissingDependencyError`, plus distinct `dedup_no_hadith_files` / `dedup_no_texts` / `dedup_skipped_degraded` log events — but a log event is a **convention**, and conventions decay. The whole point of #377 is that a suppression outlived (indeed preceded) its rule; the same rot will reach these log events.

### 2. Scope it against #360

#360's fix must change how `run_all` consumes stage results anyway (it currently swallows every exception and continues). That is the moment to change what stages *return*. Doing #360 first with bare returns and retrofitting a result type later means touching all nine stages twice.

**Sequencing: #360 and this issue are one change, not two.** This issue supplies the return type; #360 supplies the orchestrator that consumes it.

### 3. Keep a cheap syntactic net underneath

Not a substitute for the type — a backstop while it is rolled out, and a guard for code outside `src/resolve/`:

- An `ast-grep` rule (this repo prefers `ast-grep`; see org `docs/TOOLCHAIN.md`) matching an `except` handler whose body `return`s and contains no `raise`, plus early-return guards immediately preceded by `logger.error(...)`.
- High signal, needs an allowlist for the legitimate `parallels.py:353` shape.
- Runs over `src/**`, not just `src/resolve/`, so `ner.py` and future `src/parse/` instances are covered.

## Acceptance

- Resolve stages return a discriminated result; `Degraded` and `Empty` are distinct types, not both "an empty list".
- `mypy --strict` fails on an unhandled arm at every call site.
- `run_all` cannot log `resolve_pipeline_complete` while any stage returned `Degraded` or raised.
- The CLI exit status is derived from the result type.
- An `ast-grep` gate covers `src/**` for the syntactic family, with an explicit, commented allowlist.
- `parallels.py:353`'s existing discrimination is preserved (it is the reference implementation, not a violation).

## Related

- #309 — the instance (fixed, PR #362)
- #360 — the orchestrator swallow; **sequence this issue with it**
- #361 — missing required input -> empty success, 11 sites; same family
- #377 — `BLE001` pre-suppression; the narrow lint problem, distinct from this
- `ner.py:143-145` — third subsystem, same shape
- Part of noorinalabs/noorinalabs-main#928

parametrization:
## Summary

Makes the resolve stage result a **discriminated type** so two silent-success conflations become unrepresentable (per #928), not merely detected after the fact.

### da#360 — a raised stage can no longer exit 0
`run_all` wrapped every stage in `except Exception`, logged-and-continued, reached `resolve_pipeline_complete`, returned a dict, and the CLI exited **0** — a 7.5-hour run could report success having skipped a stage.

- `run_all` now returns `dict[str, StageOutcome]` where each value is `StageRan` / `StageSkipped` / `StageErrored`. A stage that produced no file is `StageRan(files=[])` — the honest "ran, no output" — never the `[]` that also meant "raised".
- Any `StageErrored` makes `run_all` raise `ResolveStageError` after the best-effort, dependency-aware sweep (aggregating **every** failed stage). `_cmd_resolve` maps it to the new `ExitCode.RESOLVE_STAGE_FAILED` (9). BaseExceptions (`StopAfterReached`, `MissingDependencyError`) still propagate unchanged.

### da#378 — a degraded artifact is distinguishable from a true negative
The composed `parallel_links.parquet` carried no detector provenance, so a `DEDUP_REQUIRE_ML=false` degraded artifact was **byte-identical** to a true negative (both a plain `PARALLEL_LINKS_SCHEMA` empty table).

- New `src/resolve/_provenance.py`: `DetectorStatus` (StrEnum) + `DetectorProvenance`, and `write_parallel_links(table, path, provenance)` — the single sanctioned writer, with `provenance` a **required** argument (missing-provenance is unrepresentable through it).
- `dedup` stamps its semantic status (`RAN` / `DEGRADED_NO_ML` / `NO_INPUT` / `NO_TEXTS`) into its artifact's parquet metadata; `run_all` reads it back and stamps the **composed** artifact with the combined provenance. `read_provenance` is the discriminator: a true negative reads back `RAN`, a degraded run `DEGRADED_NO_ML`.
- The graph loader reads `parallel_links.parquet` by column and is unaffected by the added metadata (`schema.equals` ignores metadata by default).

## Red-first evidence
`tests/test_resolve/test_discriminated_stage_result.py` (8 tests). Confirmed against `origin/main`:
- **da#360 swallow**: `origin/main:src/resolve/__init__.py` has the per-stage `except Exception: logger.error(...)` sweep and `return results` at the run_all tail -> exit 0.
- **da#378 byte-identity**: `origin`'s `_write_empty_output` and the normal 0-pair write both emit `PARALLEL_LINKS_SCHEMA.empty_table()` via plain `pq.write_table` -> identical bytes. A test pins this equality, then asserts the stamped writes separate the two classes (instrument verified on **both**: degraded->`DEGRADED_NO_ML`, true-neg->`RAN`) and that bytes now differ.

## Verification
- Full suite: **2045 passed, 10 skipped**
- ruff-format / ruff-lint / gitleaks / mypy-strict / pytest / structural-ontology-staleness: all green (pre-commit + pre-push)
- Structural ontology index regenerated for the new module

Closes #360
Closes #378


parametrization:
Renumbered `ExitCode.RESOLVE_STAGE_FAILED` **9 → 11** (central deconfliction of the three-way 9-collision: da#369 keeps 9, da#386 takes 10, this takes 11). Verified 11 is unclaimed in `exit_codes.py`; updated the pinned enum test + ruling table. Full local gate green (ruff/gitleaks/mypy-strict/pytest/structural-staleness), pushed at head `ee13d6d`. (PR body edit is blocked by the projects-classic GraphQL deprecation on this repo — the renumber and its rationale are in the commit message.)

parametrization:
I verified the PR at head `70462f5` by reading the source at HEAD via `gh api` (no checkout), and CI is 7/7 green. Four of the five requested checks hold cleanly, but the compose step reintroduces the exact da#360 swallow this PR exists to close, so I cannot approve.

What holds:

- da#378 provenance. `write_parallel_links` takes provenance as a required positional arg, so a provenance-less write is unrepresentable through the sanctioned writer; `read_provenance` returns `RAN` for a true negative and `DEGRADED_NO_ML` for a degraded run, and the test proves the two artifacts are byte-different (`degraded.read_bytes() != true_negative.read_bytes()`) while the un-stamped baseline is pinned byte-identical. The two classes separate.
- Exit-code coexistence. `RESOLVE_STAGE_FAILED = 11`, `UNROUTED_CORPUS = 9`, `PARSE_PRODUCER_DEFECT = 10` all live in the one `@enum.unique` `ExitCode`, and the exhaustive `by_name` pin lists all three. No collision, no rebase marker residue.
- Semantic-merge correctness. `UnroutedCorpusError` subclasses `BaseException` (ner.py line 72), the CLI catches it before `ResolveStageError`, and the per-stage `except Exception` in `run_all` cannot swallow it. The completion `total` uses `sum(len(v.files) for v in results.values() if isinstance(v, StageRan))` — it is StageOutcome-aware, not da#369's `len(v)`, which would have raised on the new return type.
- The base da#360 mechanism is sound for a stage that raises with no downstream compose overwrite: `StageErrored` -> `ResolveStageError` -> `_cmd_resolve` exit 11, proven red-first.

The defect — the `parallels` raise is masked, and `run_all` can still exit 0:

In `run_all` (`src/resolve/__init__.py`), when `parallels.run(...)` raises, the handler records `outcomes["parallels"] = StageErrored(...)` and leaves `deterministic_links = None`. Execution then falls through to the compose block at lines 786-799:

```
if _do("dedup") or _do("parallels"):
    provenance = DetectorProvenance(...)
    composed = _compose_parallel_links(staging_dir, semantic_links, deterministic_links, provenance)
    if composed is not None:
        outcomes["parallels"] = StageRan("parallels", [composed])
```

`_compose_parallel_links` returns a non-None path whenever `semantic is not None` — i.e. whenever `dedup` ran (or was skipped past with a prior artifact on disk), because `semantic_links` is then a table (even an empty one). So on the common path where `dedup` succeeded and only `parallels` raised, `composed is not None`, and line 799 overwrites the `StageErrored` outcome with a `StageRan`. The `errored` scan at line 835 then finds nothing, `run_all` returns its map, and `_cmd_resolve` exits 0 — a raised stage reporting success, which is the precise thing da#360 forbids.

This is production-reachable, not theoretical: under `DEDUP_REQUIRE_ML=false`, `dedup` writes an empty table on purpose (so `semantic_links` is a non-None empty table), and if the `parallels` anchor scan then raises mid-run, the 7.5-hour run exits 0 having lost the deterministic detector entirely. The `--from-step parallels` resume path hits it too (dedup skipped, `semantic_links` read from the prior artifact, parallels raises, error masked). No test covers "parallels raises while dedup produced links" — the run_all tests stub `dedup.run` to succeed but never make `parallels.run` raise — which is why CI is green over the hole.

Suggested fix: guard the overwrite so it never clears a recorded failure, e.g. `if composed is not None and not isinstance(outcomes.get("parallels"), StageErrored):`. Writing the composed artifact best-effort is fine; relabelling a raised stage as ran is not. Please add a red-first test that makes `parallels.run` raise with `dedup` having produced links and asserts `run_all` raises `ResolveStageError` / the CLI exits 11.

---

TechDebt: none

parametrization:
Independent second review (QA / red-first lens), read at HEAD `70462f5` via `gh api` — no checkout.

**da#360 — a raised stage can no longer exit 0.** Red-first is genuine on the actual diff. `test_run_all_raises_when_a_stage_raises` asserts `run_all` RAISES `ResolveStageError` when a stage raises; the docstring names the origin behaviour it fails against (origin wrapped every stage in `except Exception`, logged-and-continued, returned a dict → CLI printed "Resolution complete." exit 0). `test_sweep_continues_but_still_exits_non_zero` proves the sweep is dependency-aware (bio_promote still runs after NER fails) yet the run still refuses success and aggregates only the stage that actually RAISED (disambiguate is a `dependency_unmet` skip, not an error). The conflation is UNREPRESENTABLE, not merely detected: the outcome is a discriminated `StageRan | StageSkipped | StageErrored` union — "ran, produced no file" is `StageRan(files=[])`, structurally distinct from a raise's `StageErrored`; there is no fourth "empty list that might mean either" state. CLI maps `ResolveStageError` → `EXIT_STAGE_FAILED` and `test_cli_resolve_exits_stage_failed_code` pins a non-zero code; the monkeypatch of `src.resolve.run_all` is valid because `_cmd_resolve` re-imports `run_all` at call time.

**da#378 — a degraded artifact is byte-distinct from a true negative.** The instrument is verified on BOTH classes, which is the load-bearing check. `test_unstamped_empty_writes_are_byte_identical` first pins the pre-fix conflation (two plain `empty_table()` writes ARE byte-identical, `read_provenance` returns `None`) — the honest control. Then `test_discriminator_separates_the_two_classes` drives the real degraded branch to `DEGRADED_NO_ML` AND writes a `RAN` zero-row true negative, and asserts the discriminator labels each correctly, that the two statuses differ, and that the bytes now differ. Not a one-sided assertion, not a silent zero. The conflation is closed at WRITE time: `write_parallel_links` takes `provenance` as a REQUIRED positional arg (no default), and a repo-wide check confirms every `parallel_links.parquet` write in `src/` (dedup success + all three empty branches, parallels, run_all compose) routes through it — no bare `pq.write_table` survives, so a provenance-less artifact is unrepresentable through the sanctioned writer.

**Exit code 11.** Coexistence with 9/10 is pinned exhaustively: `test_members_are_enumerated_not_grepped` compares the full member set by exact dict-equality (9=UNROUTED_CORPUS, 10=PARSE_PRODUCER_DEFECT, 11=RESOLVE_STAGE_FAILED); `@enum.unique` + the `__new__` reserved-value guard make a duplicate or a 0/2 claim raise at import. 11 appears exactly once.

**Two updated pre-existing tests are legit contract changes, not weakenings.** `test_dedup.py::test_creates_empty_parquet` keeps every original assertion (exists, 0 rows, `schema.equals` — unaffected because `equals` ignores metadata) and ADDS provenance assertions; the signature change is the new required status arg. `test_run_all.py::test_run_all_promoted_bio_only_narrator_survives` stubs dedup to a no-op — legitimate scoping: the test's subject is bio_promote ordering + the deterministic cross-sect edge, both still fully asserted. Pre-PR, dedup's raise here (missing-ML-deps env fragility) was silently swallowed so its output was dropped anyway; the stub makes that explicit now that da#360 surfaces the raise, rather than hiding a product defect — and the compose path is independently exercised by `test_discriminated_stage_result`.

CI 7/7 green; mergeable, mergeStateStatus CLEAN.

---

Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

parametrization:
Re-reviewed the delta at head `05493de`. The compose-overwrite masking I flagged is fixed, and I confirmed the sweep is complete. Flipping to approved.

- The guard is exactly the fix. The compose block now reads `if composed is not None and not isinstance(outcomes.get("parallels"), StageErrored):`, so the composed artifact is still written (its provenance already carries `deterministic=ERRORED`) but a raised `parallels` stage is no longer relabelled as ran. The `errored` scan one block down now sees the `StageErrored`, `run_all` raises `ResolveStageError`, and `_cmd_resolve` exits 11 — the da#360 guarantee holds on the common path (dedup produced links, parallels raised).
- The red-first test lands the exact path. `test_parallels_raise_survives_compose_overwrite` stubs `dedup.run` to write a real `parallel_links.parquet` (so `semantic_links` is non-None and `_compose_parallel_links` returns a path — the precise trigger) and `parallels.run` to raise, then asserts `run_all` raises `ResolveStageError` with `errored == ["parallels"]`. That is the masking path I described, pinned. It reds with the guard reverted (exit 0) and greens restored.
- The sweep is complete. I enumerated every `outcomes[...]` assignment: each step writes its key in exactly one mutually-exclusive branch of its own try/except/else. The compose block writing `"parallels"` is the only place a later success could second-write a key set in an earlier branch, and it was the only `StageErrored`-clobbering path. No other stage has a cross-branch second-writer.

CI is 7/7 green at this head. The four checks that passed at the prior head are unchanged. Cleared.

---

RequestOrReplied: Approved
TechDebt: none

PR #407, opened 2026-07-11
fix(resolve): remap merge_log canonical_ids through fuzzy-cluster collapse (da#313)

### LINKED ISSUE #313: merge_log.parquet has 211,619 pre-existing orphaned canonical_id refs (append-only stale ids)
## Problem
`data/curated/merge_log.parquet` carries **211,619 orphaned `canonical_id` references** (of 2,936,192 rows) that do not resolve to any row in `narrators_canonical.parquet`. Discovered while validating the da#311 scrub (#311/#312): the count is identical against both the original run-5 canonical AND the da#311-scrubbed canonical, so it is fully **pre-existing** — the da#311 scrub introduced 0 new orphans.

## Root cause (hypothesis)
`merge_log` is an **append-only history** of merge decisions across three stages (`exact` 1,257,286 / `crossref` 939,768 / `fuzzy` 739,138). A `canonical_id` logged at an early stage can later be reassigned or consolidated by a subsequent disambiguate / bio_promote / cluster pass, at which point the earlier-stage id no longer exists in the *final* canonical table. So the orphans are stale historical ids, not corruption — but any consumer that joins `merge_log.canonical_id → narrators_canonical` (e.g. a merge-provenance / audit view) will silently drop ~7% of rows.

## Scope / options (for triage)
1. **Leave as-is + document** — treat `merge_log` as append-only provenance where stale ids are expected; document that consumers must not assume referential integrity to the final canonical.
2. **Post-process compaction** — rewrite each logged `canonical_id` to its final surviving id (follow the merge chain) so the log stays joinable, or drop terminal-orphan rows.
3. **Emit a `final_canonical_id` column** at write time so the raw decision id and the resolved id are both preserved.

Not urgent and explicitly **out of scope for da#311** (owner directive on #312: file a follow-up, don't fix inline). Filing for triage/prioritization.

## Evidence
```
merge_log orphans vs ORIGINAL run-5 canonical:      211,619 / 2,936,192
merge_log orphans vs da#311-scrubbed canonical:     211,619 / 2,936,036  (0 new from scrub)
merge_stage distribution: exact 1,257,286 | crossref 939,768 | fuzzy 739,138
```

parametrization:
## What

`fuzzy_cluster.cluster_canonical_narrators` collapses cross-source name variants into a survivor record — it rewrites `narrators_canonical.parquet` (dropping the absorbed rows) and remaps the **mention** rows so NARRATED edges follow the merge. It never applied that same absorbed→survivor remap to **`merge_log.parquet`**, so every merge_log row keyed on an absorbed `canonical_id` (da#356 records one row per bio-matched mention) was left referencing a `nar:` node the loader no longer creates — an **orphaned canonical_id**.

This is the da#313 orphan mechanism, re-measured at HEAD. The issue's original count (211,619) predates da#356/#376/#371 re-minting every `nar:` id, so the number is superseded; the **mechanism** is what's fixed here.

## Root cause

`cluster_canonical_narrators` builds `remap: absorbed_id → survivor_id`, calls `_remap_mention_canonical_ids(mentions_path, remap)`, but had no equivalent for the merge_log. Absorbed ids vanish from the rewritten canonical table → merge_log dangles.

## Fix

- `_remap_merge_log_canonical_ids` — mirrors `_remap_mention_canonical_ids` exactly: streams row-group by row-group, routes each absorbed `canonical_id` in merge_log to its cluster survivor, preserving the on-disk non-nullable `canonical_id` field. **Identity routing, not a filter — no row is dropped.**
- `cluster_canonical_narrators` gains optional `merge_log_path`; `run_all` passes `merge_log.parquet` when present.
- `ClusterMetrics.merge_log_remapped` reports the count.

## Red-first evidence

On a fixture where `محمد بن اسماعيل` is absorbed into `محمد بن اسماعيل البخاري`:

| | merge_log canonical_ids | orphans vs canonical |
|---|---|---|
| current code (no `merge_log_path`) | `[absorbed, survivor]` | **1** |
| after fix (`merge_log_path` passed) | `[survivor, survivor]` | **0** (`merge_log_remapped=1`, 2 rows preserved) |

## Tests (`tests/test_resolve/test_fuzzy_cluster.py`)

- `test_merge_log_orphaned_without_remap` — red witness / control: clustering that omits `merge_log_path` leaves the absorbed id dangling.
- `test_merge_log_remapped_to_survivor` — absorbed→survivor; control row already on the survivor untouched; both rows preserved; zero orphans.
- `test_merge_log_remap_is_idempotent` — a second pass over the collapsed set remaps nothing.

Full `tests/test_resolve/` green (486 passed, 6 skipped); ruff + mypy clean.

## Note on the downstream scrub

`scripts/scrub_matn_canonical.py` already *drops* merge_log rows whose canonical_id no longer survives (da#311/#312) — a post-hoc mitigation. This PR fixes the root cause upstream, so the scrub finds nothing to drop.

Closes #313

### DIFF
--- a/src/resolve/__init__.py
+++ b/src/resolve/__init__.py
@@ -565,9 +565,11 @@ def run_all(
             logger.info("resolve_step", step="cluster", status="running")
             canonical_path = output_dir / "narrators_canonical.parquet"
             mentions_path = output_dir / "narrator_mentions_resolved.parquet"
+            merge_log_path = output_dir / "merge_log.parquet"
             cluster_metrics = fuzzy_cluster.cluster_canonical_narrators(
                 canonical_path,
                 mentions_path=mentions_path if mentions_path.exists() else None,
+                merge_log_path=merge_log_path if merge_log_path.exists() else None,
                 staging_dir=staging_dir,
                 resume=resume,
                 stop_after=stop_after,
--- a/src/resolve/fuzzy_cluster.py
+++ b/src/resolve/fuzzy_cluster.py
@@ -275,6 +275,9 @@ class ClusterMetrics:
     multi_member_clusters: int
     cross_source_clusters: int
     mentions_remapped: int = 0
+    # da#313: merge_log rows whose absorbed canonical_id was routed to its cluster
+    # survivor, keeping the audit log join-consistent with narrators_canonical.
+    merge_log_remapped: int = 0
     # da#376: records that produced no blocking key and were therefore never offered to
     # `_can_merge`. Reported so "did not merge" can be told apart from "was never scored".
     unblockable_records: int = 0
@@ -286,11 +289,16 @@ class ClusterMetrics:
             if self.unblockable_records
             else ""
         )
+        merge_log = (
+            f", {self.merge_log_remapped} merge_log rows remapped"
+            if self.merge_log_remapped
+            else ""
+        )
         return (
             f"fuzzy-cluster: {self.input_records} → {self.output_records} canonical "
             f"({self.merged_records} merged into {self.multi_member_clusters} clusters, "
             f"{self.cross_source_clusters} cross-source); "
-            f"{self.mentions_remapped} mentions remapped{unblockable}"
+            f"{self.mentions_remapped} mentions remapped{merge_log}{unblockable}"
         )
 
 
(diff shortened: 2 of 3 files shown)


parametrization:
Independent second review (QA / red-first lens), da#313.

Verified at HEAD `32ed30ab`:

1. Instrument is real and separates. `_merge_log_orphans` genuinely counts merge_log `canonical_id` values absent from the live `narrators_canonical` set — the exact thing the loader would dangle. `test_merge_log_orphaned_without_remap` runs `cluster_canonical_narrators(canonical)` WITHOUT `merge_log_path` and gets orphan count 1 (`== [absorbed_id]`); the remapped test passes `merge_log_path` and gets orphan count 0 with `merge_log_remapped == 1`. Red-first is genuine — the red path is the current production call site behaviour, not a strawman.

2. No silent filter / no row dropped. `test_merge_log_remapped_to_survivor` asserts `len(rows) == 2` (both preserved), the absorbed row's `canonical_id` becomes the survivor, and the control row already on the survivor is untouched. `_remap_merge_log_canonical_ids` is identity routing (`table.set_column`, not a filter): count in == count out, only the value changes. The zero-orphan outcome is achieved by remapping, not by deleting.

3. Survivor-election direction is exercised, not assumed. Fixture makes البخاري (mention_count 10) the survivor over the mention_count-2 variant; test asserts m1 lands on `make_canonical_id(البخاري)`. A reversed election would fail the test — so the assertion pins direction, and CI green confirms it.

4. No number carried blind. PR body explicitly supersedes the issue's 211,619 (predates da#356/#376/#371 re-key) and proves the mechanism on a fixture. No hard-coded count in code or tests.

5. Scrub relationship is subset-no-op, not a fight. Upstream remap makes every merge_log canonical_id resolve to a surviving node, so `scripts/scrub_matn_canonical.py`'s drop-orphans pass finds nothing to drop — the post-hoc mitigation becomes inert, correct ordering.

6. Faithful mirror with one correct divergence. New fn matches `_remap_mention_canonical_ids` (streamed, bounded-memory, `target != cid` guard, idempotent). It passes `schema.field(id_idx)` to `set_column` rather than a bare name because `_MERGE_LOG_SCHEMA.canonical_id` is `nullable=False` (disambiguate.py:936) — a bare name would relax nullability and reject the write. Deliberate, documented, correct. Fixture uses the real `_MERGE_LOG_SCHEMA` verbatim.

Idempotence: second pass over the collapsed single-record set yields empty remap → early-return 0. Correct.

CI: all 6 checks SUCCESS (incl. fixture-realism gate); mergeable CLEAN.

---

Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

parametrization:
Approve. da#313 root cause confirmed and the fix is symmetric with the mention remap: `cluster_canonical_narrators` builds one absorbed→survivor `remap` dict and hands the SAME dict to both `_remap_mention_canonical_ids` and the new `_remap_merge_log_canonical_ids`, so the audit log follows the canonical collapse identically — no divergence.

Zero-inference verified: it is identity routing, not a filter. Every row-group is streamed back out; an absorbed-keyed row is rewritten to its survivor and no row is dropped. The non-nullable `canonical_id` field is preserved by reusing `schema.field(id_idx)` (a bare column name would relax it to nullable and reject the write), matching the production `_MERGE_LOG_SCHEMA` (`canonical_id`, `nullable=False`).

Red-first + control hold. The fixture imports the production `_MERGE_LOG_SCHEMA` and collapses `محمد بن اسماعيل` into `...البخاري`. `test_merge_log_orphaned_without_remap` bakes the red witness — clustering without `merge_log_path` leaves `orphans == [absorbed_id]`, proving the merge genuinely occurs and orphans the absorbed id — while `test_merge_log_remapped_to_survivor` shows orphan=0, `merge_log_remapped == 1`, BOTH rows preserved (`len(rows) == 2`), the absorbed row routed to the survivor, and the already-on-survivor row untouched. The control targets the real orphan class, not a blanket drop.

No conflict with the post-hoc `scrub_matn_canonical.py`: that scrub drops merge_log rows referencing pollution-dropped canonical ids (unrecoverable matn) — an orthogonal class from cluster-absorbed ids. After this upstream remap there are no cluster-orphan rows left for it to find; the scrub still independently prunes its own pollution-drop orphans. Complementary, no double-handling. CI is fully green (Lint, Type Check, Test, Integration Tests, Base-image pin + fixture realism, sync-drift gate).

---

Requestor: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

Issue #391, opened 2026-07-10
narrators_canonical: alias 'عاءشه زوج النبي' (the Prophet's wife) is attached to 'عاءشه بنت سعد', a different narrator


parametrization:
Found while verifying a provenance claim in PR #383. It is a data defect, not a defect in that PR — #383 does not touch the alias pass and its `src/` is byte-identical across the last three heads.

## What is on disk

In the **live** `data/curated/narrators_canonical.parquet` (129,234 rows), the string `عاءشه زوج النبي` — "ʿĀʾisha, wife of the Prophet" — appears:

- **0 times** as a canonical name (`name_ar` / `name_ar_normalized`)
- **1 time** inside the `aliases` list of `nar:489a30cf-d2e6-5cfc-8dce-07f7a52e3a67`, whose name is **`عاءشه بنت سعد`** (ʿĀʾisha bint Saʿd), `mention_count=143`

ʿĀʾisha bint Saʿd b. Abī Waqqāṣ is a tābiʿiyya. She is not the Prophet's wife. **An alias identifying one woman is attached to a different woman's canonical node.**

The staging side disagrees with the curated side. In `data/staging/narrator_aliases_itqan.parquet`:

```
bio_id                       'itqan:342'
canonical_name_ar_normalized 'عاءشه بنت ابي بكر الصديق'     <- ʿĀʾisha bint Abī Bakr
alias                        'عائشة زوج النبي'
alias_normalized             'عاءشه زوج النبي'
```

So the alias table anchors it to ʿĀʾisha bint Abī Bakr, and the canonical artifact hangs it on ʿĀʾisha bint Saʿd.

## What I established, and what I did not

**Established.** `make_canonical_id('عاءشه بنت ابي بكر الصديق')` is `nar:38b102fc-4690-554a-aa92-179145f1cb71`, and **that id is not a canonical node**. `clean_narrator_name` is a no-op on that anchor, so the raw and cleaned keys agree — this is therefore **not** the #389 alias-anchor mismatch. The alias cannot have arrived through `bio_promote`'s alias pass, because the node that pass would have looked up does not exist.

**Not established.** How it got there. The candidates are the alias union in `fuzzy_cluster._merge_cluster` and the alias accumulation in `disambiguate._upsert_canonical`. I have not read either closely enough to say, and I am not going to guess at a mechanism in an issue title.

[...]

PR #408, opened 2026-07-11
fix(resolve): scrub bare relational-pronoun deixis from the alias list (da#391)

parametrization:
## Summary

`Closes #391`. Zero-inference containment for the **relational-deixis-in-aliases** sub-defect of the ʿĀʾisha chimera (`nar:489a30cf` on the pre-fix artifact): the alias whose own form is `امتاه`/`خالته`/`عمتي` — an unnamed-relation reference the pipeline already drops from a NAME — survives as a searchable alias of a real narrator.

Root cause: `clean_narrator_name` drops a bare relational-pronoun from a name at NER/bio time, but the three alias-union sites append alias surfaces **raw**. The `aliases` `list<string>` is the blind spot the cleaner never reaches.

## Change (owner-approved scope)

- Expose `is_mubham_relational()` in `name_quality` — the existing step-6 whole-name drop, reusing the same `_MUBHAM_RELATIONAL` lexicon (single source of truth).
- Apply it at **all three** alias-union sites:
  - `disambiguate._upsert_canonical`
  - `fuzzy_cluster._merge_cluster`
  - `bio_promote._merge_aliases` — the da#94 itqan alias-table feed the issue traced; accounted as a new `skipped_deixis` bucket so the module's sum-invariant still holds (no silent drop).
- Extend the lexicon with the closed **1st-person-possessive (`-ي`) and vocative (`-اه`)** kinship forms the 3rd-person-only set missed (`عمتي`/`امتاه`/`جدي`/`عمي`/…). Closed, kinship-only — no name adjudication.

## Measurement + instrument control (#928 doctrine)

Measured on the **stale** curated artifact (129,234 nodes; it predates da#356/#376/#371, so these are the pre-fix defect shape, not a HEAD count):

- **Red-first:** 1,112 canonical nodes carry ≥1 bare-relational alias across 28 distinct tokens (1,593 alias-strings dropped) — a **class**, not a singleton, riding the top-mention nodes (`ابن عبس` mc=50,962; `عبد الله` mc=19,412; `مالك` mc=17,187 carry `ابيه`/`ابي`).
- **Separation control (proves the instrument):** the predicate strikes **only** whole single-token deixis — **0** multi-token hits — so a real kunya like `ابي اسحاق` is never touched.
- **Scope control:** the scrub is alias-only; the 9 nodes whose *own name* is a deixis token (a separate da#247 **NAME-level** residual) are left untouched — not created or worsened here.
- **Target:** the ʿĀʾisha cluster's 27 aliases → drops exactly `['امتاه','خالته','عمتي']`, keeps 24.

## Explicitly NOT in scope (deferred, per owner)

The other two sub-defects of #391 — **over-merge** (distinct persons folded onto one node; corpus-wide, e.g. `ابن عبس` absorbing `ابن عباس`) and **cross-gender absorption** (`هشام`) — are node-**identity** decisions (da#356/#376 crossref/fuzzy), not alias-list content. They cannot be un-merged by an alias scrub and must not be threshold-tuned. Their status is da#356/#376 **efficacy**, confirmable only by the gated re-run — deferred to that verification.

## Tests

Red-first unit tests at each of the 4 surfaces (pre-fix appends the deixis raw):
- `TestIsMubhamRelational` (predicate + precision boundary)
- `TestUpsertAliasDeixisScrub` (disambiguate)
- `test_merge_cluster_drops_bare_relational_alias` (fuzzy_cluster)
- `test_merge_aliases_drops_bare_relational_deixis` (bio_promote + sum-invariant)

Full affected suites (429), `ruff format --check`, `ruff check`, and `mypy --strict` all green; pre-push mypy + full pytest pass. Rebased onto current main (da#405 untracked the structural index, so it is no longer part of this diff).

Suggested reviewers: **Kavitha Sundaramurthy**, **Jean-Claude Habimana**.

### DIFF
--- a/src/parse/name_quality.py
+++ b/src/parse/name_quality.py
@@ -156,6 +156,7 @@ __all__ = [
     "clean_narrator_name_display",
     "asserted_name_tokens",
     "cleaner_removed_content",
+    "is_mubham_relational",
     "split_compound_narrators",
     "strip_markup",
 ]
@@ -326,6 +327,28 @@ _MUBHAM_RELATIONAL = frozenset(
         "زوجته",  # his wife          "زوجها" his/her spouse
         "زوجها",
         "عنه",  # "from him" — transmission particle, never a name
+        # da#391: the 1st-person-possessive (``-ي``) and vocative (``-اه``) kinship
+        # forms. The set above is exhaustively 3rd-person ("his/her …") plus the bare
+        # 1st-masc ``ابي``; a mention/alias phrased in the narrator's OWN voice
+        # ("امتاه" = "O mother!", "عمتي" = "my aunt", "جدي" = "my grandfather") is the
+        # same unnamed-relation reference and equally never a proper ism. Closed,
+        # kinship-only — no name adjudication. Attested as surviving aliases on the
+        # curated set (``عمتي``/``امتاه`` on the ʿĀʾisha chimera; ``جدي``×91, ``عمي``×32).
+        "امي",  # my mother
+        "ابني",  # my son            "ابنتي"/"بنتي" my daughter
+        "ابنتي",
+        "بنتي",
+        "جدي",  # my grandfather     "جدتي" my grandmother
+        "جدتي",
+        "اخي",  # my brother         "اختي" my sister
+        "اختي",
+        "عمي",  # my paternal uncle  "عمتي" my paternal aunt
+        "عمتي",
+        "خالي",  # my maternal uncle "خالتي" my maternal aunt
+        "خالتي",
+        "امتاه",  # O mother! (vocative) "اماه" variant
+        "اماه",
+        "ابتاه",  # O father! (vocative)
     }
 )
 
@@ -1367,6 +1390,27 @@ def _tokenize_and_strip_affixes(text: str) -> tuple[list[str], list[str], bool]:
             tokens.pop(0)
             raw_tokens.pop(0)
     return raw_tokens, tokens, stripped_number
+
+
+def is_mubham_relational(name_normalized: str | None) -> bool:
+    """True when the WHOLE name is a single bare relational-pronoun (mubham) token.
+
+    "his father", "my aunt", "O mother!" — an unnamed-relation reference, never a
+    person. This is exactly the drop :func:`clean_narrator_name` step 6 already applies
+    at NER/bio time; it is exposed so the resolve **alias-union** sites can apply the
+    SAME drop to the ``aliases`` ``list<string>`` — the field ``clean_narrator_name``
+    never reaches, and the da#391 blind spot (a relational deixis surviving as an alias
+    of a real narrator, so any alias lookup resolves the wrong person).
+
+    Normalizes internally (``normalize_arabic`` is idempotent) so a raw or already-
+    normalized alias both resolve. The single-token guard is the precision boundary:
+    a real multi-token kunya (``"ابي اسحاق"`` = Abū Isḥāq) returns ``False``, and the
+    lexicon is kinship-only, so no real single-token ism is ever matched.
+    """
+    if not name_normalized:
+        return False
+    tokens = normalize_arabic(name_normalized).split()
+    return len(tokens) == 1 and tokens[0] in _MUBHAM_RELATIONAL
 
 
 def clean_narrator_name(name_normalized: str | None) -> str | None:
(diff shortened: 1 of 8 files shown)


parametrization:
Independent second-review of da#391 (PR #408) — scrub relational-deixis from the alias list. I authored da#389's `_merge_aliases` bucketing, so I verified this PR preserves that partition.

**Sum-invariant holds (my da#389 contract).** The new `skipped_deixis` arm is inserted as a `continue`-terminated branch between `skipped_empty` and `unmatched_pollution`, keyed on `variant` (the alias itself). It is a clean partition addition: every alias row still lands in EXACTLY one of the seven buckets — added / skipped_duplicate / skipped_empty / skipped_source / skipped_deixis / unmatched_pollution / unmatched_no_bio. The deixis skip is counted, never a silent drop, and surfaces in `promote_bios_to_canonical` as `aliases_skipped_deixis`. `skipped_deixis` is a no-default field on the `_AliasMergeStats` NamedTuple, so any un-updated construction would hard-fail — CI Test green confirms all sites carry it. The `unmatched_no_bio` WARN witness is untouched: the deixis arm only intercepts deixis variants, so non-deixis aliases still reach that path, and its test is unchanged. `test_merge_aliases_drops_bare_relational_deixis` asserts the full 7-term sum equals 2.

**Zero-inference + separation.** `is_mubham_relational` requires `len(tokens) == 1`, so multi-token surfaces (a real kunya `ابي اسحاق`) are never struck — 0 multi-token hits by construction, matching the PR's separation control. The lexicon extension is a closed 1st-person-possessive/vocative kinship set (`عمتي`/`امتاه`/`جدي`/…); none is a standalone proper ism.

**Red-first at all four surfaces.** Each test is RED on pre-fix code (the field/guard does not exist): bio_promote (`skipped_deixis` absent, variant appended raw), disambiguate `_upsert_canonical`, fuzzy_cluster `_merge_cluster`, and the predicate itself.

**Scope.** Alias-only; `clean_narrator_name` and name-level paths untouched. All three alias-union append sites are guarded — I grepped every `resolve/*.py` and found no fourth raw-append site. Over-merge (a) and cross-gender (c) are correctly NOT in this diff (node-identity, deferred to re-run verification).

CI: all six checks SUCCESS.

---

Requestor: Kwesi Boateng
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

parametrization:
Reviewed at head `0be856e` against the diff (via `gh api .../pulls/408/files`), not the PR body.

**Zero-inference — confirmed.** `is_mubham_relational()` reuses the existing step-6 `_MUBHAM_RELATIONAL` frozenset as the single source of truth (used at line 1413 by the new predicate and line 1521 by `clean_narrator_name` step 6 — verified at HEAD). The lexicon extension is a closed, kinship-only set of 1st-person-possessive (`-ي`) and vocative (`-اه`) forms — no name adjudication, no threshold. Notably the shared frozenset also tightens NAME-level cleaning for the added forms; that is correct and consistent, guarded by the same single-token boundary, and only ever improves (never worsens) the separate da#247 NAME-level residual on a re-run.

**Separation control — the load-bearing check, verified.** The predicate strikes only whole single-token deixis (`len(tokens) == 1`), and the control runs over real names: `test_real_name_or_empty_is_not_mubham` holds the multi-token kunya `ابي اسحاق` (Abū Isḥāq), real single-token isms `عاءشه`/`هشام`/`سعد`, and empty/None all False. It proves both that a real multi-token name is untouched AND that no real single-token ism is stripped — not a blind fix.

**All three alias-union sites, red-first — confirmed.** `disambiguate._upsert_canonical` (guard added to the append condition; test asserts `خالته` dropped, real multi-token `bio_alias` attaches), `fuzzy_cluster._merge_cluster` (both the member own-spelling and existing-alias-carry paths guarded; test drops `خالته` from a member's alias list, keeps the real variant), and `bio_promote._merge_aliases` (new `skipped_deixis` bucket, no silent drop — the sum-invariant is asserted over all 7 buckets == 2). None is a silent drop; each test fails on pre-fix code (raw append, and `skipped_deixis` did not exist).

**Scope discipline — confirmed.** Alias-only; nothing touches identity/threshold/crossref logic, so the deferred over-merge (a) and cross-gender (c) sub-defects are genuinely absent from this diff. The 9 NAME-level-deixis nodes are not created or worsened here.

**CI green** — Lint, Type Check, Test, Integration Tests, Base-image/fixture-realism, and the Pre-commit⇄CI sync-drift gate all SUCCESS at head.

---

Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

Issue #282, opened 2026-07-03
Chain identity collapsed: multi-isnad hadiths flattened into one sequence, fabricating TRANSMITTED_TO adjacencies


parametrization:
## Bug (found during the main#920 graph re-architecture exploration — Finding B)

**Chain identity is collapsed in the loaded graph.** Evidence from reading the actual loaders (see the full report on noorinalabs-main#920):

- `TRANSMITTED_TO` edges key on `hadith_id` only; `narrator_mentions_resolved` has no `chain_id` column; adjacency pairs are built by grouping ALL of a hadith's mentions and sorting by position.
- Consequence: a hadith with MULTIPLE isnads gets its chains flattened into one position-sorted sequence — **fabricating narrator adjacencies that exist in no actual chain** and destroying the bundle shape (single-strand vs bundled/mutawatir) that common-link / isnad-convergence analysis depends on.
- The `Chain` node model exists in `src/models/` but is never wired into edges.

## Impact

- Graph-correctness: some `TRANSMITTED_TO` edges are artifacts of flattening, not attested transmission.
- Blocks the isnad-convergence / common-link class of analyses (main#920 question 3) entirely.

## Fix direction

Reify chain identity end-to-end: carry a `chain_id` (or chain ordinal per hadith) from parse → NER mention rows → resolved mentions → the graph loader, and build `TRANSMITTED_TO` adjacencies per-chain, never per-hadith. Wire the existing Chain node model in (or explicitly retire it). Needs a data-shape check per source parser (which parsers even segment multiple isnads today?) — that sub-question overlaps the da#279 corpora-shape audit; coordinate.

## Scope note

Pre-existing defect (not a wave-23 regression); filed for visibility and scope reconciliation at next wave planning. Fixing implies a pipeline re-run to regenerate mentions with chain identity.

Part of the noorinalabs-main#920 follow-up set.

parametrization:
## Cross-link: da#282 is **not** the same fix as da#352 — and its evidence base is being silently discarded

Design note posted on da#352 → https://github.com/noorinalabs/noorinalabs-data-acquisition/issues/352#issuecomment-4929331652

[...]

PR #410, opened 2026-07-11
Reify isnad-chain identity: build TRANSMITTED_TO per-chain, not per-hadith (da#282)

parametrization:
## What

Reify **isnad-chain identity** so `TRANSMITTED_TO` (and `NARRATED`, and the `Chain` node) is built **within a chain**, never flattened across a hadith's multiple chains. `Closes #282`.

## The bug (measured)

`TRANSMITTED_TO` grouped mentions by `hadith_id` and sorted by `position_in_chain`. When one `hadith_id` carries **more than one transmission sequence**, sorting interleaves them and MERGEs narrator pairs that exist in no actual chain.

**Data-shape audit (the issue's `which parsers segment multiple isnads?`):**
- The Arabic/English segmenter **linearizes** one isnad string — no parser emits genuinely-distinct parallel isnads as separate position sequences today.
- The one real present-day case of `>1` position-0 sequence under one `hadith_id` is **`lk`**: it emits an Arabic isnad **and** an English isnad for the same hadith, each numbered from `position 0`. Interleaved → fabricated cross-language adjacencies (length-mismatched chains make this resolution-independent).
- `sanadset`, thaqalayn/ner isnad-column: one sequence per hadith → unaffected.

## The fix (zero-inference)

Add **`chain_index`** (per-hadith isnad ordinal) — matches existing `Chain.chain_index` / `chain_node_id(hadith_id, chain_index)`:
- `NARRATOR_MENTION_SCHEMA` + `NARRATOR_MENTIONS_RESOLVED_SCHEMA`, `RESOLVE_SCHEMA_VERSION` 1→2. Nullable → coalesces to 0.
- Producers: **lk `ar=0`/`en=1`** (preserving the upstream distinction, not inventing one); `sanadset=0`; ner isnad-column `=0`; ner phase1 loader carries the parse column through; muhaddithat link `=0`.
- Consumers grouped by `(hadith_id, chain_index)`: `_load_transmitted_to`, `_load_narrated`, `load_nodes._load_chains` (one Chain node per chain now, not hardcoded 0).
- Cast-passthrough consumers (`disambiguate`, `narrator_split`, `fuzzy_cluster`) + row-filter scrubs preserve the column unchanged.

`chain_index` is **never all-zero on real data** (lk emits 1) — not an inert schema add.

## Red-first evidence

`test_two_chains_one_hadith_no_cross_chain_edge`: hadith with chain 0 = `A→B`, chain 1 = `C→D` (both start at position 0, the lk shape).
- **Pre-fix (per-hadith flatten, reproduced by neutralizing grouping):** pairs = `{(A,C),(C,B),(B,D)}` — all fabricated; real `(A,B)`/`(C,D)` absent.
- **Post-fix:** pairs = `{(A,B),(C,D)}` exactly.

Pinned unaffected: single-isnad (all `chain_index=0`) and legacy **null** `chain_index` (coalesce→0). NARRATED: both chain openers emit, MERGE-dedupe to one edge when they share a canonical (lk ar/en). lk producer test asserts `ar→0`,`en→1` under one `source_hadith_id`.

## Behavioral note for reviewers

For lk hadiths with both isnads this creates **one Chain node per language** (`chn:<h>-0` ar, `chn:<h>-1` en) instead of one flattened node; edges MERGE-dedupe so counts don't inflate. Correct consequence of preserving the two sequences — flag if the owner prefers same-isnad-across-languages deduped to one chain (separate concern).

## Out of scope (flagged follow-ups)

`disambiguate._adjacent_death_years` and `narrator_split` estimate death years from `(hadith_id, position±1)` neighbors — still hadith-scoped, so a multi-isnad hadith can pick a cross-chain neighbor. Pre-existing; not broken here; `chain_index` makes the fix trivial. Touches da#376 death-year territory — surfacing for an owner/architect call. Wiring `Chain` nodes into edges (the issue's other half) also remains open.

## Green
`pytest` (2084 passed, 10 skipped), `ruff check`, `ruff format --check`, `mypy --strict` clean locally; pre-commit + pre-push hooks passed.

Suggested reviewers: Jean-Claude Habimana, Kavitha Sundaramurthy.

### DIFF
--- a/src/graph/load_edges.py
+++ b/src/graph/load_edges.py
@@ -126,6 +126,21 @@ def _val(row: dict[str, Any], key: str, default: Any = None) -> Any:
     return default if v is None else v
 
 
+def _chain_index(row: dict[str, Any]) -> int:
+    """Per-hadith isnad-chain ordinal for a mention row (da#282).
+
+    Chain edges are grouped by ``(hadith_id, chain_index)`` — never by hadith
+    alone — so a hadith carrying more than one transmission sequence (today: lk's
+    Arabic + English isnads, each numbered from position 0) does not have its
+    chains flattened into one position-sorted list, which fabricated narrator
+    adjacencies that exist in no actual chain. A missing/None ``chain_index``
+    (legacy pre-da#282 file, or a single-isnad producer that left it unset)
+    coalesces to 0, so single-chain hadiths are unaffected.
+    """
+    v = row.get("chain_index")
+    return int(v) if v is not None else 0
+
+
 # ---------------------------------------------------------------------------
 # 1. TRANSMITTED_TO — consecutive narrator pairs in each chain
 # ---------------------------------------------------------------------------
@@ -247,7 +262,9 @@ def _load_transmitted_to(
         logger.warning("transmitted_to_files_missing", dir=str(staging_dir))
         return EdgeLoadResult("TRANSMITTED_TO", 0, 0, 0)
 
-    # Group mentions by hadith.
+    # Group mentions by (hadith_id, chain_index) so each isnad-chain is sequenced
+    # on its own (da#282). Keying on hadith alone flattened every chain of a
+    # multi-isnad hadith into one position-sorted list — see ``_chain_index``.
     #
     # Curated mention-links (provenance-bearing rows from the muhaddithat orphan
     # producer, da#228) are NARRATED-ONLY by contract and MUST NOT enter chain-pair
@@ -263,7 +280,7 @@ def _load_transmitted_to(
     # (the correct edge) while never fabricating a transmission pair. A normal
     # chain mention has no ``provenance`` column (row.get → None), so this only
     # ever excludes curated links.
-    by_hadith: dict[str, list[dict[str, Any]]] = {}
+    by_chain: dict[tuple[str, int], list[dict[str, Any]]] = {}
     dropped_noncanonical = 0
     for fp in files:
         rows = _read_parquet_rows(fp)
@@ -285,7 +302,7 @@ def _load_transmitted_to(
                 continue
             if row.get("provenance"):
                 continue  # curated NARRATED-only link — never a transmission pair
-            by_hadith.setdefault(hid, []).append(row)
+            by_chain.setdefault((hid, _chain_index(row)), []).append(row)
 
     # Build all chain pairs. Validate the group's hadith id ONCE up front: every
     # pair _build_chain_pairs emits re-derives it, so a malformed id would raise
@@ -294,7 +311,7 @@ def _load_transmitted_to(
     # leaves batches 1..N-1 already written to Neo4j.
     all_pairs: list[dict[str, Any]] = []
     malformed_ids = 0
-    for hid, mentions in by_hadith.items():
+    for (hid, _cidx), mentions in by_chain.items():
         try:
             hadith_node_id(hid)
         except DoubledCorpusPrefixError:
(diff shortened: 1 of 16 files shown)


parametrization:
Architect review — da#282, isnad-chain identity (`chain_index`), head `6612eff`.

**Root mechanism + fix — confirmed at all three consumers.** `TRANSMITTED_TO` (`_load_transmitted_to`), `NARRATED` (`_load_narrated`), and Chain-node synthesis (`load_nodes._load_chains`) each moved their group key from `hadith_id` to `(hadith_id, chain_index)`, resolved through the shared `_chain_index`/`row.get("chain_index") or 0` coalesce. The bug is exactly as measured: keying on hadith alone flattens a hadith's multiple sequences into one position-sorted list, and the real present-day case (lk emitting an Arabic AND an English isnad, each renumbered from position 0) then interleaves into fabricated cross-language adjacencies. Grouping-per-chain is the correct and minimal fix.

**Producers set it correctly.** lk `ar=0`/`en=1` (preserving the upstream ar/en distinction rather than inventing one), sanadset `=0` on both mention paths, ner isnad-column `=0`, ner phase1 loader carries the parse column through defensively (`has_chain_index` guard + `run()` array-assembly adds the column), muhaddithat curated link `=0`. No producer leaves it unset where a genuine second sequence exists — the only >1 case is lk, and it emits 1.

**Not an inert schema add (#928).** `chain_index` is non-zero on real data — the lk parser test (`test_arabic_and_english_isnads_are_separate_chains`) asserts `ar→0, en→1` on actual parse output, and the edge fixture writes a real `int32` chain 1 that the loader reads back and acts on. The column changes behavior; the fixture does not silently drop it (conftest writes it as a real column, so this avoids the #927 inert-fixture trap). Legacy null → coalesce-0 is the correct back-compat path and is pinned by `test_absent_chain_index_defaults_to_single_chain`.

**Red-first is genuine.** `test_two_chains_one_hadith_no_cross_chain_edge` uses the exact lk shape (chain 0 A→B, chain 1 C→D, both from position 0) and asserts `pairs == {(A,B),(C,D)}` while explicitly excluding the pre-fix fabricated set `{(A,C),(C,B),(B,D)}`. Reverting the grouping change reproduces the interleaved sort `[A(0),C(0),B(1),D(1)] → (A,C),(C,B),(B,D)`, which fails the equality — so the test distinguishes the fix, it does not merely assert current output. Single-isnad and NARRATED-dedupe (openers sharing a canonical MERGE to one edge) are pinned too. No edge inflation.

**Follow-ups are genuinely separable — the core is fully fixed.** da#282's core defect is the fabricated `TRANSMITTED_TO` adjacencies (plus the parallel NARRATED/Chain flattening); all three are closed here. The two flagged items are not this fix hiding behind `Closes`: (a) `_adjacent_death_years`/`narrator_split` estimating death years from `(hadith_id, position±1)` is a distinct heuristic consumer, pre-existing, that produces an estimate rather than a graph edge — it is da#376 death-year territory, not the adjacency bug; (b) wiring Chain nodes into edges is the issue's separately-scoped other half. Neither is debt this PR incurs, so `TechDebt: none`. I do recommend the death-year cross-chain-neighbor case be explicitly folded into da#376 (now trivial given `chain_index`) so it does not fall between issues — but that is a follow-up, not a blocker.

CI rollup is green (Lint, Type Check, Test, Integration Tests, fixture-realism, sync-drift gate all SUCCESS); PR is BEHIND and will rebase before merge — code correctness is sound now.

---

RequestOrReplied: Approved
TechDebt: none

parametrization:
**Approve.** Independent second review (read at head `6612eff`, no checkout).

I traced the `chain_index` schema end-to-end and it propagates unbroken through every stage.

**Zero-inference.** `chain_index` records a distinction the sources already carry — `lk` emits an Arabic and an English isnad for one hadith, each independently numbered from position 0, and the per-hadith segmenter yields a single linear chain. The field is a structural ordinal set at the producer (ar 0 / en 1 for lk; 0 everywhere else), never a heuristic and never a threshold. It stops the loader from COLLAPSING two already-separate sequences into one; it infers nothing.

**Propagation.** Every producer of the superset schema sets the field — `lk_corpus` (ar/en), `sanadset` (both the NAR-tag and names paths), `ner._extract_from_hadiths` and `_load_phase1_mentions` (the latter reading defensively so a legacy parse file with no such column still loads as chain 0), `muhaddithat_links`, and `ner.run()` which writes it under the resolved schema. The only two constructors of the parse superset schema are `lk_corpus` and `sanadset`, both updated, so no producer breaks on the added column. The three cast-passthrough resolve stages — disambiguate, narrator_split, fuzzy_cluster — read the full row-group, cast to the resolved schema, replace only the canonical id (and confidence), and rewrite the whole table, so the new column rides through unchanged rather than being dropped. Every graph-truth grouper is rekeyed off hadith alone and onto the chain: TRANSMITTED_TO groups on the chain bucket, NARRATED takes the position-0 narrator of each chain, and Chain nodes key on the chain with the id carrying the real ordinal.

**Red-first and the instrument separates.** The two-chains-one-hadith test builds A to B on chain 0 and C to D on chain 1, both opening at position 0 — exactly the lk shape — and asserts BOTH that the real adjacencies are present AND that the three fabricated ones the old per-hadith flatten produced are absent, pinned by an exact set equality. Under the pre-fix grouping the first assertion fails, so the test genuinely separates fixed from broken.

**Back-compat and no inflation.** A null ordinal coalesces to chain 0 (covered), single-isnad hadiths are unchanged (covered), the schema-version constant moves to 2, and the NARRATED and Chain paths both MERGE, so lk's two same-opener language chains dedupe to one edge rather than inflating — asserted directly.

One out-of-scope note for a possible follow-up, not a blocker and not a regression introduced here: the death-date neighbour estimator in narrator_split still groups by hadith alone when it looks at position plus-or-minus-one neighbours, so for a multi-chain lk hadith it could treat a cross-language narrator as a chain neighbour. Its behaviour is unchanged by this PR, it sits outside the graph-adjacency scope of this change, and the impact is small since lk's two lanes are parallel translations that resolve to the same canonicals. Worth filing separately now that the ordinal exists.

CI is behind and will be rebased before merge; assessed on code correctness at head.

---

RequestOrReplied: Approved
TechDebt: none

PR #412, opened 2026-07-11
fix(resolve): fail loud on a missing required input instead of a success-shaped empty (da#361)

### LINKED ISSUE #361: resolve: missing required input silently yields an empty success across 11 sites (same class as #309)

parametrization:
## What & why

`resolve` returned a **success-shaped empty** (`[]` / `None` / a zero-row parquet) when a **required input artifact was absent** — the same defect class as #309 (a missing *dependency*) one layer up (a missing *input*), and as #360 (a swallowed raise). A missing input means an upstream stage did not run (or ran against the wrong dir): an upstream defect, indistinguishable to every caller and to the exit status from an input that ran and legitimately produced nothing.

Landable now because **#360** makes `run_all` surface a raise instead of swallowing it.

## Mechanism

New shared guard `src/resolve/_inputs.py` mirroring `_deps.py`:

- `require_input(present=...)` raises `MissingInputError` on an absent required input.
- `MissingInputError` subclasses **`Exception`** (not `BaseException` like `MissingDependencyError`) **on purpose** — so `run_all` **catches** it as a `StageErrored` and reuses #360's machinery → exit `EXIT_STAGE_FAILED` (11). No new exit code, no CLI change.
- Fires **only on absence** — each site keeps its honest empty return for the present-but-zero-rows true negative. The two stay distinguishable (not a blanket "empty == error").

## Site enumeration (all 11 from the issue, measured at HEAD)

**Made fatal (absent required input → raise):**
| Site | Input | Notes |
|---|---|---|
| dedup `run_dedup` | `hadiths_*.parquet` | the `NO_INPUT` case #309 stamped and explicitly deferred as "tracked separately" — this is that follow-up. `NO_TEXTS` stays an honest empty. |
| ner `_load_phase1_mentions` | phase1 mentions of a **staged** corpus | staged-aware |
| ner `_extract_from_hadiths` | hadiths of a **staged** corpus | staged-aware |
| disambiguate `run` | `narrators_bio_*.parquet` (candidates) | present-but-0-rows → honest `[]` |
| bio_promote | `narrators_bio_*.parquet` | present-but-all-filtered → honest success |
| narrator_split | `narrators_canonical.parquet` | empty canonical → honest no-op |
| disambiguate `_load_mentions` | mentions | currently uncalled; kept honest for future wire-up |

**Left honest (NOT this defect class — a measurement finding):** NER legitimately produces zero mentions for a bio-only corpus (e.g. `muhaddithat`, which NER skips) — "an input that ran and produced nothing", which #361 explicitly excludes. So these stay honest no-ops:
- disambiguate `_count_mentions()==0`
- narrator_split absent `narrator_mentions_resolved.parquet`

The **`muhaddithat` live integration test** is the oracle that proved raising on these two was a false positive on a legitimate bio-only resolve.

**Optional, confirmed & declared (no raise):** `date_reconcile`, `tabaqa_dates` — pure date-enrichers with no narrators to date when the canonical table is absent.

**Staged-aware ner:** `run` loops over *configured* sources; a source merely absent from a partial staging subset is a legitimate skip. Only a corpus that IS staged but missing its expected file raises — so empty/partial-staging runs (and `test_empty_staging_dir`) keep working.

## Red-first evidence

- `test_missing_input_fail_loud.py` (new, 17 tests): ner + disambiguate raises, the `MissingInputError` contract (`Exception` not `BaseException`), each raise paired with a present-but-empty honest-success test, `run_all` → `ResolveStageError` integration, and the two optional/carve-out no-ops.
- Flipped in place (were asserting the old empty success): `test_dedup.py::TestEmptyInput` (`@ml`), `test_bio_promote.py::test_no_bio_files_*`, `test_narrator_split.py::test_stage_no_canonical_*`.
- Three `run_all` spy harnesses now also stub `narrator_split` (their "replace every stage" contract previously omitted it — harmless only while it silently no-op'd).

Full suite green under `uv run` (`2047 passed`, integration deselected); the `muhaddithat` live suite passes (`3 passed`) with Docker; ruff-format/lint + mypy-strict clean; pre-commit + pre-push hooks passed.

Closes #361

### DIFF
--- a/src/resolve/__init__.py
+++ b/src/resolve/__init__.py
@@ -29,6 +29,7 @@ from typing import TYPE_CHECKING
 from src.exit_codes import EXIT_STAGE_FAILED
 from src.resolve._checkpoint import EXIT_STOPPED_AT_LIMIT, StopAfterReached
 from src.resolve._deps import EXIT_MISSING_DEPENDENCY, MissingDependencyError
+from src.resolve._inputs import MissingInputError
 from src.resolve._provenance import DetectorProvenance, DetectorStatus, read_provenance
 from src.resolve.ner import EXIT_UNROUTED_CORPUS, UnroutedCorpusError
 from src.utils.logging import get_logger
@@ -46,6 +47,7 @@ __all__ = [
     "RESOLVE_STEP_ORDER",
     "RESUMABLE_STEPS",
     "MissingDependencyError",
+    "MissingInputError",
     "ResolveMetrics",
     "ResolveStageError",
     "StageErrored",
--- a/src/resolve/_inputs.py
+++ b/src/resolve/_inputs.py
@@ -1,0 +1,107 @@
+"""Required-input guards for resolve stages (da#361).
+
+Companion to :mod:`src.resolve._deps` (da#309), one layer up. ``_deps`` guards a
+missing *dependency* (an unimportable module); this guards a missing *input*
+artifact (a parquet an upstream stage should have produced).
+
+The defect class (da#361)
+-------------------------
+Several ``src/resolve`` stages returned a success-shaped empty result — ``[]`` /
+``None`` / a zero-row parquet — when a **required** input artifact was absent. A
+missing input means an upstream stage did not run (or ran against the wrong
+directory): an upstream defect. It is NOT the same thing as an input that ran and
+legitimately produced nothing, yet the two were indistinguishable to every caller
+and to the process exit status — the same silent success #309 fixed for
+dependencies and #360 fixed for a swallowed raise.
+
+Distinguishing the two is the whole point, so this guard fires **only on
+absence**, never on a present-but-empty input. Each call site keeps its honest
+empty return for the true-negative case (input present, genuinely zero rows) and
+delegates only the *absence* decision here.
+
+Why ``Exception``, not ``BaseException``
+----------------------------------------
+:class:`~src.resolve._deps.MissingDependencyError` subclasses ``BaseException`` so
+it **sails past** ``run_all``'s per-stage ``except Exception`` and maps to its own
+exit code (``4``). :class:`MissingInputError` makes the opposite choice **on
+purpose**: a missing input is exactly the "a stage could not do its job"
+condition da#360's machinery already models. A plain ``Exception`` is *caught* by
+``run_all`` and recorded as a :class:`~src.resolve.StageErrored`, which makes
+``run_all`` raise :class:`~src.resolve.ResolveStageError` at the end → exit
+:data:`~src.exit_codes.EXIT_STAGE_FAILED` (11). So a stage that raises on a
+missing input is aggregated with every other failed stage rather than aborting
+the best-effort sweep, and needs no new exit code.
+
+da#360 is the reason raising is meaningful at all here: before it, ``run_all``
+swallowed the raise, logged ``resolve_step_failed`` and marched on to
+``resolve_pipeline_complete``, exiting 0. Raising a missing-input error into that
+old orchestrator would have been swallowed exactly like the empty return it
+replaces — which is precisely why da#361's "Proposed direction" made this depend
+on da#360.
+"""
+
+from __future__ import annotations
+
+__all__ = ["MissingInputError", "require_input"]
+
+
+class MissingInputError(Exception):
+    """An enabled resolve stage's **required** input artifact is absent.
+
+    Distinct from a genuinely empty input (input present, zero rows), which each
+    stage still reports as an honest success. The message names *which* stage,
+    *which* input, *which* upstream stage should have produced it, and *what to
+    run* — so an operator can act without reading the traceback.
+
+    A plain ``Exception`` (not ``BaseException``) so ``run_all`` catches it as a
+    :class:`~src.resolve.StageErrored`; see this module's docstring for why.
+    """
+
+    def __init__(
+        self,
+        *,
+        stage: str,
+        input_desc: str,
+        produced_by: str,
+        remediation: str,
+    ) -> None:
+        self.stage = stage
+        self.input_desc = input_desc
+        self.produced_by = produced_by
+        self.remediation = remediation
+        super().__init__(self._render())
+
+    def _render(self) -> str:
+        return (
+            f"resolve stage {self.stage!r} is missing its required input "
+            f"({self.input_desc}), normally produced by {self.produced_by}. This is an "
+            f"upstream defect, not an empty result: continuing would write a "
+            f"success-shaped empty output indistinguishable from a genuine negative. "
+            f"Remediation: {self.remediation}"
+        )
+
+
+def require_input(
+    *,
+    stage: str,
+    present: bool,
+    input_desc: str,
+    produced_by: str,
+    remediation: str,
+) -> None:
+    """Raise :class:`MissingInputError` when a required input is absent.
+
+    ``present`` is the caller's own presence check — ``path.exists()`` for a
+    single artifact, ``bool(list(dir.glob(pattern)))`` for a globbed set. Passing
+    the boolean (rather than a path) keeps this one helper serving both the
+    single-file and glob sites, and — critically — lets the caller draw the line
+    between ABSENT (raise here) and PRESENT-BUT-EMPTY (the caller's own honest
+    empty return), which is the distinction da#361 exists to preserve.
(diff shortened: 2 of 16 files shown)


parametrization:
Approve. Independent second review (QA / red-first lens), verified at head 1c46985 via gh api.

The false-positive class is the danger here, and the fix draws the missing-input vs true-empty line correctly at every site. Each of the 7 fatal sites raises only on genuine absence and keeps its honest empty return for the present-but-zero-rows negative: dedup Case A NO_INPUT raises while Case B NO_TEXTS stays an honest empty; disambiguate raises on absent bio shards but returns an honest empty on zero-row shards; bio_promote raises on absent shards but returns None when all rows filter out; ner raises only for a corpus that IS staged yet missing its expected file, and skips unstaged configured sources. Every raise is paired with a present-but-empty honest-success test.

The two excluded sites are the trap and they are correctly left honest: disambiguate _count_mentions()==0 returns an honest empty (never require_input), and narrator_split with an absent mentions file returns None. The muhaddithat LIVE integration test is a genuine oracle for this, not a decorative one: it drives the real parse to resolve run_all path on a bio-only corpus (no isnad text, NER skips it) that traverses BOTH carve-out sites, then asserts the canonical table exists with 4 rows. That assertion is only reachable if run_all completed without raising; a require_input raise at either carve-out would be caught as a StageErrored, aggregated into ResolveStageError, and propagate out of the unguarded run_all call, failing the test. Kavitha over-raised initially and this test caught it. The oracle holds.

The MissingInputError contract is pinned both ways: it subclasses Exception (not BaseException like MissingDependencyError), the contract test asserts both directions, and the run_all integration test proves a stage raising it is caught as StageErrored and surfaces as ResolveStageError (exit 11), reusing da#360 machinery with no new exit code. The rebase composition is coherent: require_input is imported at every use site, bio_promote retains both the da#391 alias handling and the new guard, dedup keeps MissingDependencyError alongside require_input, and no guard is bypassed. CI is fully green including the muhaddithat live Integration Tests job, and the PR is MERGEABLE.

---

Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

parametrization:
Reviewed at head `1c46985` via `gh api` (no checkout). da#361 — a missing required input now fails loud across the resolve stages instead of writing a success-shaped empty. This is the last parse/resolve launch-gate row and it is correct.

The design distinction holds. `MissingInputError` subclasses `Exception` (not `BaseException`), verified in `_inputs.py` and pinned by `test_is_an_exception_not_a_baseexception`. That is the deliberate opposite of da#369's `UnroutedCorpusError`: a missing input is a stage failure that `run_all`'s per-stage `except Exception` catches as `StageErrored`, aggregating to `ResolveStageError` → exit `EXIT_STAGE_FAILED` (11) via da#360's machinery, with no new exit code — while the unrouted-corpus config error sails past to the CLI. `test_run_all_surfaces_missing_input_as_stage_error` exercises the caught path end-to-end.

Absence vs present-but-empty is preserved at every one of the 7 fixed sites. `require_input(present=...)` delegates only the absence decision; each caller keeps its honest empty for a genuine zero. dedup Case A (no hadith files) raises, Case B (files present, no English matn) stays `NO_TEXTS` honest empty. Both ner helpers are staged-aware (`present=not required`): a staged corpus with a missing file raises, an unstaged configured source is a legitimate skip. disambiguate and bio_promote raise on absent `narrators_bio_*.parquet` but return honest `[]`/`None` on present-but-zero-candidate shards. narrator_split raises on an absent canonical table. The dead `_load_mentions` helper raises too, kept honest for future re-wiring; being uncalled it cannot false-positive.

The two excluded sites correctly do NOT raise — this is the load-bearing carve-out. disambiguate's `_count_mentions(output_dir) == 0` returns an honest empty (`disambiguate_no_mentions`), and narrator_split's absent `narrator_mentions_resolved.parquet` returns an honest `None` (`narrator_split_no_mentions`). NER legitimately yields zero mentions for a bio-only corpus (muhaddithat), so a raise here would break every bio-only resolve. Kavitha's initial over-raise on these was caught by the muhaddithat lightup test and reverted; the guards are now pinned by `test_bios_present_mentions_absent_is_honest_empty` and `test_stage_canonical_present_no_mentions_is_honest_noop`, and the unmodified `tests/integration/test_muhaddithat_lightup.py` passes green as the live regression backstop.

Composition order in `ner.run` is correct. The da#369 unrouted guard runs first, then per-helper `require_input` fires on `if not path.exists()` BEFORE da#282's `chain_index` read (line 1157, inside the branch that only executes when the file is present) — you never read `chain_index` off a missing file. da#365 routing, da#282 chain_index, and the new require_input coexist cleanly.

Red-first is honored: 17 tests, each raise paired with a present-but-empty honest-success test; the flipped dedup/bio_promote/narrator_split suites are legitimate contract changes. Full CI green — Test, Integration Tests, Lint, Type Check, sync-drift gate, base-image/fixture-realism all pass.

---

RequestOrReplied: Approved
TechDebt: none

PR #414, opened 2026-07-11
feat(graph): prune-narrators subcommand — canonical-set SHRINK (da#413)

### LINKED ISSUE #413: Add prune-narrators CLI subcommand: DETACH DELETE Narrator ids not in narrators_canonical.parquet (canonical-set shrink; deploy#557 invokes it)
Discovered during deploy#557 (Aisha Idrissi, measure-first). The canonical-set SHRINK cannot be done in pure cypher-shell: the only correct orphan definition is `Narrator.id NOT IN narrators_canonical.parquet`, and that set exists only in the parquet — unreadable by a cypher-only graph-ops job.

**Evidence (Aisha's graph↔parquet join):** graph has 160,614 Narrator ids vs 129,234 canonical ids.
- orphan (in-graph, not-in-canonical) = 31,380, of which **18,518 HAVE edges** → a zero-degree/"unreferenced" predicate misses them.
- 57,993 zero-degree canonical nodes are **legitimate** (owner da#352: a narrator attested only in a deduped edition IS a real narrator) → a zero-degree prune would wrongly delete them.
- No in-graph stamp separates orphan from legit; only canonical-set membership does.

**This issue = the da side of the split:** add a `prune-narrators` CLI subcommand, symmetric with `load`/`migrate`/`enrich`:
- Reads `narrators_canonical.parquet`, computes the `canonical_id` set.
- Batched `DETACH DELETE` of Narrator nodes whose `id` ∉ that set (removes the 31,380 orphans, edges and all).
- **GUARD (load-bearing): refuse — non-zero exit, ZERO deletion — if the canonical set is empty or the parquet is unreadable.** A bad read must never wipe the graph. (This is the da#309/da#361 fail-loud-on-missing-input discipline applied to a destructive op.)
- `--dry-run`: report the count that WOULD be deleted, delete nothing.
- Typed + unit-tested (Neo4j testcontainer); exit codes from the existing `src/exit_codes.py` space.

**deploy#557** (Aisha) builds the destructive-op WORKFLOW that mounts the curated parquet (rclone-pull → /app/data, like deploy-data-load.yml) + invokes this subcommand + graph-side verify (pre/post counts, a known-canonical narrator survives, a known-collapsed id is gone, canonical count preserved) + prod environment approval + `dry_run=true` default.

parametrization:
## What

Adds a `prune-narrators` CLI subcommand (symmetric with `load`/`migrate`/`enrich`) that performs the canonical-set **SHRINK**: DETACH DELETE every `Narrator` node whose `id` is absent from `narrators_canonical.parquet`.

`deploy#557` (Aisha Idrissi) builds the destructive-op workflow that invokes this — the subcommand contract below is fixed and this PR confirms it.

## Why this can't be pure cypher (deploy#557 measure-first)

The only correct orphan predicate is `Narrator.id NOT IN narrators_canonical.parquet`, and that set exists only in the parquet:

- **18,518 of the 31,380 orphans HAVE edges** → a zero-degree/"unreferenced" predicate misses them.
- **57,993 zero-degree canonical narrators are legitimate** (owner da#352) → a zero-degree prune wrongly deletes them.

So this reads the keep-set in Python (the same shape `src.graph.migrate` uses), computes the graph complement, and DETACH-DELETEs it in batches. The complement is a **set difference, not a degree test** — that is the whole point.

## The guard (load-bearing, red-first)

`read_canonical_ids` refuses — `EmptyCanonicalSetError`, **ZERO deletion**, before the graph is read or written — when the keep-set is **empty**, **missing**, or **unreadable/malformed**. A bad read must never wipe the graph (da#309/da#361 fail-loud-on-missing-input applied to a destructive op).

## Exit-code semantics for deploy#557 to map

| Condition | Exit | Meaning |
|-----------|------|---------|
| success (real run or dry-run) | `0` | prune ran; graph in intended state |
| canonical set empty / missing / unreadable | `12` `ExitCode.EMPTY_CANONICAL_SET` (**new**) | **refused, graph EXACTLY as before** — fix the parquet and retry |
| Neo4j unreachable (pre-flight) | `8` `ExitCode.DB_UNREACHABLE` | reused via `_check_neo4j()`, like `load`/`migrate` |

**New code 12, not `LOAD_FAILED`:** a prune is not a load (the registry is emphatic about not naming a stage you're not in), and the safe "deleted nothing" state must be **unambiguous** so the workflow can tell "refused, safe to retry" apart from any partial-deletion failure. Named for the concept (empty/missing/unreadable all fold to "no usable keep-set"), per the `REFUSED_ROWS` precedent. `--dry-run` reports the count that would be deleted (plus a sample) and deletes nothing.

## Acceptance (green, red-first + controls)

**Neo4j testcontainer** (`tests/integration/test_prune_narrators.py`) proves on a real graph:
1. an orphan (`id ∉ canonical`) **with an edge** is DETACH DELETEd, edge and all;
2. a **legitimate zero-degree canonical narrator SURVIVES** (the 57,993-class control — the whole reason pure-cypher failed);
3. the guard **REFUSES with zero deletion** on empty and on missing canonical (graph verified unchanged by reading it back);
4. `--dry-run` deletes nothing.

Unit tests (`tests/test_graph/test_prune.py`, MockNeo4jClient) cover the complement computation, guard-touches-nothing (zero client calls), batching, and dry-run. Full suite green; pre-push mypy + pytest passed.

Closes #413

### DIFF
--- a/src/cli.py
+++ b/src/cli.py
@@ -525,7 +525,12 @@ def _cmd_migrate_transmitted_hadith_ids(*, batch_size: int = 1000) -> None:
         print("\nMigration complete.")
 
 
-def _cmd_prune_narrators(*, canonical: str, dry_run: bool = False) -> None:
+def _cmd_prune_narrators(
+    *,
+    canonical: str,
+    dry_run: bool = False,
+    max_missing_fraction: float | None = None,
+) -> None:
     """DETACH DELETE Narrator nodes whose id is not in narrators_canonical.parquet (da#413).
 
     The canonical-set SHRINK: after a re-resolve collapses ids, the graph still holds
@@ -536,13 +541,28 @@ def _cmd_prune_narrators(*, canonical: str, dry_run: bool = False) -> None:
     nothing.
 
     Refuses — non-zero exit :attr:`ExitCode.EMPTY_CANONICAL_SET`, ZERO deletion — when
-    the canonical set is empty or its parquet is unreadable/missing. A bad read must
-    never wipe the graph.
+    the keep-set cannot be trusted: it is empty / unreadable / missing
+    (:class:`EmptyCanonicalSetError`), or it is well-formed but is not **this graph's**
+    (:class:`ForeignCanonicalSetError` — a stale ``--canonical``, caught by the
+    ``missing`` provenance signal). A bad read must never wipe the graph, and neither
+    must the wrong parquet.
     """
     from pathlib import Path
 
-    from src.graph.prune import EmptyCanonicalSetError, prune_narrators, summary_line
+    from src.graph.prune import (
+        DEFAULT_MAX_MISSING_FRACTION,
+        UnusableCanonicalSetError,
+        prune_narrators,
+        summary_line,
+    )
     from src.utils.neo4j_client import Neo4jClient
+
+    # `None` (not the literal) is the CLI default, so the ceiling has exactly one home:
+    # src.graph.prune. A hardcoded default here — the shape of migrate's `--batch-size`
+    # — is a second copy of a number that must not drift, and cli.py lazy-imports the
+    # module precisely so `--help` need not pay for pyarrow + the neo4j driver.
+    if max_missing_fraction is None:
+        max_missing_fraction = DEFAULT_MAX_MISSING_FRACTION
 
     # This command reads the canonical parquet and writes ONLY to the graph and
     # stdout. It never writes back into the parquet's directory (no manifest, no audit
@@ -551,21 +571,30 @@ def _cmd_prune_narrators(*, canonical: str, dry_run: bool = False) -> None:
 
     canonical_path = Path(canonical)
 
-    # The guard runs inside prune_narrators, BEFORE the graph is touched. Catch it
-    # here to exit the named code rather than crash — nothing was deleted.
+    # Both guards run inside prune_narrators, BEFORE any graph WRITE. Catch their common
+    # supertype here to exit the named code rather than crash — nothing was deleted.
     try:
-        # The guard reads the parquet, not the DB. Check connectivity first (like
+        # The first guard reads the parquet, not the DB. Check connectivity first (like
         # load/migrate) so an unreachable DB is DB_UNREACHABLE, not a later crash.
         _check_neo4j()
         with Neo4jClient() as client:
-            result = prune_narrators(client, canonical_path, dry_run=dry_run)
-    except EmptyCanonicalSetError as guard:
+            result = prune_narrators(
+                client,
+                canonical_path,
+                dry_run=dry_run,
+                max_missing_fraction=max_missing_fraction,
+            )
+    except UnusableCanonicalSetError as guard:
         print(f"\nERROR: {guard}", file=sys.stderr)
         sys.exit(ExitCode.EMPTY_CANONICAL_SET)
 
     print("=== prune-narrators (canonical-set shrink, da#413) ===")
     print(f"  Canonical ids (keep-set)   : {result.canonical_ids_seen}")
     print(f"  Orphans (id ∉ canonical)   : {result.orphans_identified}")
+    # The keep-set provenance reading. ~0 says the parquet is this graph's; a large value
+    # says it is a stale/foreign one — the one thing no post-op gate can check, because
+    # every one of them is derived from this same keep-set.
+    print(f"  Missing (canonical ∉ graph): {result.missing}")
     if result.sample:
         shown = ", ".join(result.sample)
         more = result.orphans_identified - len(result.sample)
@@ -772,6 +801,22 @@ def _cmd_pipeline() -> None:
     print("=== Full Pipeline Complete ===")
 
 
+def _unit_fraction(raw: str) -> float:
+    """argparse type for a fraction in [0.0, 1.0].
+
+    An out-of-range ceiling is operator error, so it must be argparse's usage exit (2),
+    not a traceback and not a silently clamped value that would quietly weaken a
+    destructive-op guard.
+    """
+    try:
+        value = float(raw)
+    except ValueError:
+        raise argparse.ArgumentTypeError(f"{raw!r} is not a number") from None
+    if not 0.0 <= value <= 1.0:
+        raise argparse.ArgumentTypeError(f"must be in [0.0, 1.0], got {value}")
+    return value
+
+
 def main() -> None:
     """Run the isnad-ingest CLI."""
     from src.graph.validate import _DEFAULT_VALIDATION_TIMEOUT_SECONDS
@@ -939,13 +984,29 @@ def main() -> None:
         help=(
             "Path to narrators_canonical.parquet. Its canonical_id set is the "
             "authoritative keep-set: every Narrator whose id is absent from it is "
-            "DETACH DELETEd. Refuses (exit 12, zero deletion) if empty/unreadable."
+            "DETACH DELETEd. Refuses (exit 12, zero deletion) if it is empty, missing "
+            "or unreadable — or if it is not THIS graph's keep-set (see "
+            "--max-missing-fraction). Point it at the parquet the live graph was "
+            "loaded from: a stale one deletes the narrators it fails to mention."
         ),
     )
     prune_parser.add_argument(
         "--dry-run",
         action="store_true",
         help="Report the count that would be deleted (and a sample); delete nothing.",
+    )
+    prune_parser.add_argument(
+        "--max-missing-fraction",
+        type=_unit_fraction,
+        default=None,
+        metavar="FRACTION",
+        help=(
+            "Ceiling on 'missing' (canonical ids the graph has never seen), as a "
+            "fraction of the keep-set. Above it the keep-set is refused as belonging to "
+            "a DIFFERENT graph (exit 12, zero deletion) — the stale --canonical guard. "
+            "Default: half the keep-set. Raise it (up to 1.0, which disables the "
+            "ceiling) only when a genuinely large partial load is intended."
+        ),
     )
 
     audit_parser = subparsers.add_parser("audit", help="View recent pipeline audit entries")
@@ -1028,7 +1089,11 @@ def main() -> None:
     elif args.command == "migrate-transmitted-hadith-ids":
         _cmd_migrate_transmitted_hadith_ids(batch_size=args.batch_size)
     elif args.command == "prune-narrators":
-        _cmd_prune_narrators(canonical=args.canonical, dry_run=args.dry_run)
+        _cmd_prune_narrators(
+            canonical=args.canonical,
+            dry_run=args.dry_run,
+            max_missing_fraction=args.max_missing_fraction,
+        )
     elif args.command == "audit":
         _cmd_audit(last_n=args.last)
     elif args.command == "pipeline":
--- a/src/cli.py
+++ b/src/cli.py
@@ -525,6 +525,62 @@ def _cmd_migrate_transmitted_hadith_ids(*, batch_size: int = 1000) -> None:
         print("\nMigration complete.")
 
 
+def _cmd_prune_narrators(*, canonical: str, dry_run: bool = False) -> None:
+    """DETACH DELETE Narrator nodes whose id is not in narrators_canonical.parquet (da#413).
+
+    The canonical-set SHRINK: after a re-resolve collapses ids, the graph still holds
+    the orphaned ``Narrator`` nodes (many WITH edges — a zero-degree predicate misses
+    them). The only correct orphan predicate is canonical-set membership, which lives
+    in the parquet, so this is a da-side subcommand rather than a cypher-only job
+    (deploy#557). ``--dry-run`` reports the count that would be deleted and deletes
+    nothing.
+
+    Refuses — non-zero exit :attr:`ExitCode.EMPTY_CANONICAL_SET`, ZERO deletion — when
+    the canonical set is empty or its parquet is unreadable/missing. A bad read must
+    never wipe the graph.
+    """
+    from pathlib import Path
+
+    from src.graph.prune import EmptyCanonicalSetError, prune_narrators
+    from src.utils.neo4j_client import Neo4jClient
+
+    canonical_path = Path(canonical)
+
+    # The guard runs inside prune_narrators, BEFORE the graph is touched. Catch it
+    # here to exit the named code rather than crash — nothing was deleted.
+    try:
+        # _check_neo4j only after the guard would have a chance to fire? No: the guard
+        # reads the parquet, not the DB, and reads nothing destructive. Check
+        # connectivity first (like load/migrate) so an unreachable DB is DB_UNREACHABLE.
+        _check_neo4j()
+        with Neo4jClient() as client:
+            result = prune_narrators(client, canonical_path, dry_run=dry_run)
+    except EmptyCanonicalSetError as guard:
+        print(f"\nERROR: {guard}", file=sys.stderr)
+        sys.exit(ExitCode.EMPTY_CANONICAL_SET)
+
+    print("=== prune-narrators (canonical-set shrink, da#413) ===")
+    print(f"  Canonical ids (keep-set)   : {result.canonical_ids_seen}")
+    print(f"  Narrator nodes in graph    : {result.graph_narrators_seen}")
+    print(f"  Orphans (id ∉ canonical)   : {result.orphans_identified}")
+    if result.sample:
+        shown = ", ".join(result.sample)
+        more = result.orphans_identified - len(result.sample)
+        suffix = f" (+{more} more)" if more > 0 else ""
+        print(f"  Sample orphan ids          : {shown}{suffix}")
+    if dry_run:
+        print(
+            f"\nDRY RUN: {result.orphans_identified} Narrator node(s) WOULD be "
+            "DETACH DELETEd. Nothing was deleted."
+        )
+    else:
+        print(f"  Narrator nodes deleted     : {result.deleted}")
+        if result.orphans_identified == 0:
+            print("\nNo orphans — every Narrator in the graph is in the canonical set.")
+        else:
+            print("\nPrune complete.")
+
+
 def _cmd_enrich(
     *,
     only: list[str] | None = None,
@@ -858,6 +914,30 @@ def main() -> None:
         help="Edges rewritten per write batch (default: 1000)",
     )
 
+    prune_parser = subparsers.add_parser(
+        "prune-narrators",
+        help=(
+            "DETACH DELETE Narrator nodes whose id is not in narrators_canonical.parquet "
+            "(canonical-set shrink; da#413)"
+        ),
+    )
+    prune_parser.add_argument(
+        "--canonical",
+        type=str,
+        required=True,
+        metavar="PARQUET",
+        help=(
+            "Path to narrators_canonical.parquet. Its canonical_id set is the "
+            "authoritative keep-set: every Narrator whose id is absent from it is "
+            "DETACH DELETEd. Refuses (exit 12, zero deletion) if empty/unreadable."
+        ),
+    )
+    prune_parser.add_argument(
+        "--dry-run",
+        action="store_true",
+        help="Report the count that would be deleted (and a sample); delete nothing.",
+    )
+
     audit_parser = subparsers.add_parser("audit", help="View recent pipeline audit entries")
     audit_parser.add_argument(
         "--last",
@@ -937,6 +1017,8 @@ def main() -> None:
             sys.exit(ExitCode.VALIDATION_FINDINGS)
     elif args.command == "migrate-transmitted-hadith-ids":
         _cmd_migrate_transmitted_hadith_ids(batch_size=args.batch_size)
+    elif args.command == "prune-narrators":
+        _cmd_prune_narrators(canonical=args.canonical, dry_run=args.dry_run)
     elif args.command == "audit":
         _cmd_audit(last_n=args.last)
     elif args.command == "pipeline":
--- a/src/cli.py
+++ b/src/cli.py
@@ -541,17 +541,21 @@ def _cmd_prune_narrators(*, canonical: str, dry_run: bool = False) -> None:
     """
     from pathlib import Path
 
-    from src.graph.prune import EmptyCanonicalSetError, prune_narrators
+    from src.graph.prune import EmptyCanonicalSetError, prune_narrators, summary_line
     from src.utils.neo4j_client import Neo4jClient
+
+    # This command reads the canonical parquet and writes ONLY to the graph and
+    # stdout. It never writes back into the parquet's directory (no manifest, no audit
+    # entry, unlike `load`) — deploy#557 mounts that dir read-only, so a stray write
+    # would trip the :ro mount (da#348).
 
     canonical_path = Path(canonical)
 
     # The guard runs inside prune_narrators, BEFORE the graph is touched. Catch it
     # here to exit the named code rather than crash — nothing was deleted.
     try:
-        # _check_neo4j only after the guard would have a chance to fire? No: the guard
-        # reads the parquet, not the DB, and reads nothing destructive. Check
-        # connectivity first (like load/migrate) so an unreachable DB is DB_UNREACHABLE.
+        # The guard reads the parquet, not the DB. Check connectivity first (like
+        # load/migrate) so an unreachable DB is DB_UNREACHABLE, not a later crash.
         _check_neo4j()
         with Neo4jClient() as client:
             result = prune_narrators(client, canonical_path, dry_run=dry_run)
@@ -561,7 +565,6 @@ def _cmd_prune_narrators(*, canonical: str, dry_run: bool = False) -> None:
 
     print("=== prune-narrators (canonical-set shrink, da#413) ===")
     print(f"  Canonical ids (keep-set)   : {result.canonical_ids_seen}")
-    print(f"  Narrator nodes in graph    : {result.graph_narrators_seen}")
     print(f"  Orphans (id ∉ canonical)   : {result.orphans_identified}")
     if result.sample:
         shown = ", ".join(result.sample)
@@ -569,16 +572,23 @@ def _cmd_prune_narrators(*, canonical: str, dry_run: bool = False) -> None:
         suffix = f" (+{more} more)" if more > 0 else ""
         print(f"  Sample orphan ids          : {shown}{suffix}")
     if dry_run:
+        print(f"  Narrator nodes in graph    : {result.graph_total}")
         print(
             f"\nDRY RUN: {result.orphans_identified} Narrator node(s) WOULD be "
             "DETACH DELETEd. Nothing was deleted."
         )
     else:
         print(f"  Narrator nodes deleted     : {result.deleted}")
+        print(f"  Narrator nodes remaining   : {result.graph_total}")
+        print(f"  Orphans remaining (post)   : {result.orphans}")
         if result.orphans_identified == 0:
             print("\nNo orphans — every Narrator in the graph is in the canonical set.")
         else:
             print("\nPrune complete.")
+
+    # The load-bearing machine-readable line deploy#557's verify greps. Emitted on
+    # BOTH modes and last, so it is unmissable on stdout.
+    print(summary_line(result))
 
 
 def _cmd_enrich(
--- a/src/cli.py
+++ b/src/cli.py
@@ -530,6 +530,7 @@ def _cmd_prune_narrators(
     canonical: str,
     dry_run: bool = False,
     max_missing_fraction: float | None = None,
+    max_orphan_fraction: float | None = None,
 ) -> None:
     """DETACH DELETE Narrator nodes whose id is not in narrators_canonical.parquet (da#413).
 
@@ -540,29 +541,38 @@ def _cmd_prune_narrators(
     (deploy#557). ``--dry-run`` reports the count that would be deleted and deletes
     nothing.
 
-    Refuses — non-zero exit :attr:`ExitCode.EMPTY_CANONICAL_SET`, ZERO deletion — when
-    the keep-set cannot be trusted: it is empty / unreadable / missing
-    (:class:`EmptyCanonicalSetError`), or it is well-formed but is not **this graph's**
-    (:class:`ForeignCanonicalSetError` — a stale ``--canonical``, caught by the
-    ``missing`` provenance signal). A bad read must never wipe the graph, and neither
-    must the wrong parquet.
+    Refuses — non-zero exit :attr:`ExitCode.UNUSABLE_CANONICAL_SET`, ZERO deletion — when
+    the keep-set cannot be trusted, on any of three axes:
+
+    * it is empty / unreadable / missing (:class:`EmptyCanonicalSetError`);
+    * it is well-formed but is not **this graph's** — a stale ``--canonical``, caught by
+      the ``missing`` provenance signal (:class:`ForeignCanonicalSetError`);
+    * it *is* this graph's but keeps so little of it that the prune is a wipe — a
+      TRUNCATED parquet, caught by the delete magnitude (:class:`ExcessiveDeletionError`).
+      Note this one has ``missing == 0``: every id-provenance check passes on it.
+
+    A bad read must never wipe the graph — and neither must the wrong parquet, nor half
+    of the right one.
     """
     from pathlib import Path
 
     from src.graph.prune import (
         DEFAULT_MAX_MISSING_FRACTION,
+        DEFAULT_MAX_ORPHAN_FRACTION,
         UnusableCanonicalSetError,
         prune_narrators,
         summary_line,
     )
     from src.utils.neo4j_client import Neo4jClient
 
-    # `None` (not the literal) is the CLI default, so the ceiling has exactly one home:
+    # `None` (not the literal) is the CLI default, so each ceiling has exactly one home:
     # src.graph.prune. A hardcoded default here — the shape of migrate's `--batch-size`
     # — is a second copy of a number that must not drift, and cli.py lazy-imports the
     # module precisely so `--help` need not pay for pyarrow + the neo4j driver.
     if max_missing_fraction is None:
         max_missing_fraction = DEFAULT_MAX_MISSING_FRACTION
+    if max_orphan_fraction is None:
+        max_orphan_fraction = DEFAULT_MAX_ORPHAN_FRACTION
 
     # This command reads the canonical parquet and writes ONLY to the graph and
     # stdout. It never writes back into the parquet's directory (no manifest, no audit
@@ -583,10 +593,11 @@ def _cmd_prune_narrators(
                 canonical_path,
                 dry_run=dry_run,
                 max_missing_fraction=max_missing_fraction,
+                max_orphan_fraction=max_orphan_fraction,
             )
     except UnusableCanonicalSetError as guard:
         print(f"\nERROR: {guard}", file=sys.stderr)
-        sys.exit(ExitCode.EMPTY_CANONICAL_SET)
+        sys.exit(ExitCode.UNUSABLE_CANONICAL_SET)
 
     print("=== prune-narrators (canonical-set shrink, da#413) ===")
     print(f"  Canonical ids (keep-set)   : {result.canonical_ids_seen}")
@@ -1008,6 +1019,21 @@ def main() -> None:
             "ceiling) only when a genuinely large partial load is intended."
         ),
     )
+    prune_parser.add_argument(
+        "--max-orphan-fraction",
+        type=_unit_fraction,
+        default=None,
+        metavar="FRACTION",
+        help=(
+            "Ceiling on the share of the Narrator graph this prune may DELETE. Above it "
+            "the prune is refused (exit 12, zero deletion) — the TRUNCATED --canonical "
+            "guard, which --max-missing-fraction cannot catch because a truncated "
+            "keep-set's ids are all genuinely this graph's ('missing' is 0). Default: "
+            "0.9 — a legitimate post-re-mint cleanup can orphan ~55%% of an accumulated "
+            "graph, so a tighter ceiling would refuse real work; a truncated keep-set "
+            "deletes 97%%+. Raise it (1.0 disables) only when a wipe is genuinely meant."
+        ),
+    )
 
     audit_parser = subparsers.add_parser("audit", help="View recent pipeline audit entries")
     audit_parser.add_argument(
@@ -1093,6 +1119,7 @@ def main() -> None:
             canonical=args.canonical,
             dry_run=args.dry_run,
             max_missing_fraction=args.max_missing_fraction,
+            max_orphan_fraction=args.max_orphan_fraction,
         )
     elif args.command == "audit":
         _cmd_audit(last_n=args.last)
--- a/src/exit_codes.py
+++ b/src/exit_codes.py
@@ -47,6 +47,7 @@ code    name                   what is on disk when it fires
 ``9``   ``UNROUTED_CORPUS``    nothing resolve would have written (NER Step-1 abort)
 ``10``  ``PARSE_PRODUCER_DEFECT`` the clean sources' staging; the defective source's is absent
 ``11``  ``RESOLVE_STAGE_FAILED`` prior stages' artifacts; a stage raised (da#360)
+``12``  ``EMPTY_CANONICAL_SET`` the graph, untouched; ``prune-narrators`` refused (da#413)
 ======  =====================  ===================================================
 
 da#360, da#369 and da#386 all branched from a main whose next free code was ``9``
@@ -161,6 +162,7 @@ import enum
 
 __all__ = [
     "EXIT_DB_UNREACHABLE",
+    "EXIT_EMPTY_CANONICAL_SET",
     "EXIT_ENRICH_FAILED",
     "EXIT_LOAD_FAILED",
     "EXIT_MISSING_DEPENDENCY",
@@ -333,6 +335,32 @@ class ExitCode(enum.IntEnum):
     fails the whole invocation rather than being logged and skipped.
     """
 
+    EMPTY_CANONICAL_SET = 12
+    """``prune-narrators`` refused: its authoritative keep-set was unusable (da#413).
+
+    What is on disk when it fires: **the graph, exactly as it was** — zero nodes
+    deleted, zero edges touched. This is the load-bearing guarantee of a destructive
+    op: a bad read must never wipe the graph.
+
+    ``prune-narrators`` DETACH-DELETEs every ``Narrator`` whose ``id`` is not in
+    ``narrators_canonical.parquet``. That keep-set is the ONLY thing standing between
+    the command and deleting a node, so the command refuses — before it reads or
+    writes the graph at all — if the keep-set cannot be trusted to name the survivors.
+
+    Named for the worst of the three doors, not for one instance. The guard raises on
+    an **empty** keep-set (which would delete every narrator), a **missing** parquet,
+    and an **unreadable/malformed** parquet alike: each yields no usable set, and the
+    exception message names which door fired. Naming the member after only the empty
+    case would re-arm the trap for the missing/unreadable cases, the same mistake
+    :attr:`REFUSED_ROWS` documents for the malformed-id-vs-blank-id classes.
+
+    This is the da#309 / da#361 fail-loud-on-missing-input discipline applied to a
+    *destructive* op. It gets its own code — not :attr:`LOAD_FAILED` — because a
+    prune is not a load, and because the safe "deleted nothing" state must be
+    unambiguous: a distinct code lets deploy#557's workflow tell "refused, graph
+    safe, fix the parquet and retry" apart from any partial-deletion failure.
+    """
+
 
 # Module-level aliases. Consumers may import either the enum or these names; the
 # alias form keeps a migrating call site to an *import-site* change rather than a
@@ -348,3 +376,4 @@ EXIT_REFUSED_ROWS = ExitCode.REFUSED_ROWS
 EXIT_UNROUTED_CORPUS = ExitCode.UNROUTED_CORPUS
 EXIT_PARSE_PRODUCER_DEFECT = ExitCode.PARSE_PRODUCER_DEFECT
 EXIT_STAGE_FAILED = ExitCode.RESOLVE_STAGE_FAILED
+EXIT_EMPTY_CANONICAL_SET = ExitCode.EMPTY_CANONICAL_SET
--- a/src/exit_codes.py
+++ b/src/exit_codes.py
@@ -47,7 +47,7 @@ code    name                   what is on disk when it fires
 ``9``   ``UNROUTED_CORPUS``    nothing resolve would have written (NER Step-1 abort)
 ``10``  ``PARSE_PRODUCER_DEFECT`` the clean sources' staging; the defective source's is absent
 ``11``  ``RESOLVE_STAGE_FAILED`` prior stages' artifacts; a stage raised (da#360)
-``12``  ``EMPTY_CANONICAL_SET`` the graph, untouched; ``prune-narrators`` refused (da#413)
+``12``  ``UNUSABLE_CANONICAL_SET`` the graph, untouched; ``prune-narrators`` refused (da#413)
 ======  =====================  ===================================================
 
 da#360, da#369 and da#386 all branched from a main whose next free code was ``9``
@@ -162,7 +162,7 @@ import enum
 
 __all__ = [
     "EXIT_DB_UNREACHABLE",
-    "EXIT_EMPTY_CANONICAL_SET",
+    "EXIT_UNUSABLE_CANONICAL_SET",
     "EXIT_ENRICH_FAILED",
     "EXIT_LOAD_FAILED",
     "EXIT_MISSING_DEPENDENCY",
@@ -335,7 +335,7 @@ class ExitCode(enum.IntEnum):
     fails the whole invocation rather than being logged and skipped.
     """
 
-    EMPTY_CANONICAL_SET = 12
+    UNUSABLE_CANONICAL_SET = 12
     """``prune-narrators`` refused: its authoritative keep-set was untrustworthy (da#413).
 
     What is on disk when it fires: **the graph, exactly as it was** — zero nodes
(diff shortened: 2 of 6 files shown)


parametrization:
## Review — canonical-set SHRINK (da#413)

Reviewed the artifact at head `42a4b11` via the contents API, not the PR body. This is a `DETACH DELETE` against the production knowledge graph, so I went at the guard adversarially and tried to find an input that reaches a delete. **The guard holds.** Approving.

### 1. The empty/unreadable keep-set guard — verified, and NOT inert

`read_canonical_ids` (`src/graph/prune.py:143-156`) has three doors: `exists()` → missing; a broad `except` around `pq.read_table` → unreadable/malformed; a falsy-filtered set comprehension → empty. It is called on the **first line** of `prune_narrators` (`src/graph/prune.py:180`), before the graph read at `:182`.

The part I care about most, given main#927 (an assertion no fixture can trigger proves nothing): **the fixtures genuinely fire each door.**

- `TestGuardTouchesNothing` (`tests/test_graph/test_prune.py:105-121`) parametrizes all three doors and asserts `client.calls == []` — not "no writes", but *no client call at all*. The guard demonstrably never reaches Neo4j.
- The unreadable fixture is `path.write_text("garbage")` (`:113`) — a real non-parquet, so `pq.read_table` really raises.
- The empty fixture is `write_narrators_canonical(dir, [])`. I checked the helper (`tests/test_graph/conftest.py:88-89`): it builds the arrow schema from column literals independent of `rows`, so an empty list yields a **valid parquet with a zero-row `canonical_id` column**. The *empty* door genuinely fires rather than silently collapsing into the *unreadable* door — which is the failure mode I was looking for.
- Integration (`tests/integration/test_prune_narrators.py:118-141`) proves refusal on a **real** graph by reading the node set and edge count back afterward, not by trusting a return value.
- `test_blank_ids_do_not_count_as_a_keep_set` (`:94-99`) covers rows-present-but-all-blank. That is the subtle one and it is covered.

I also chased the id-space question, because a mismatch there would make the complement *everything*: `Narrator.id` is loaded from the parquet's `canonical_id` column (`src/graph/load_nodes.py:267-274`, `MERGE (n:Narrator {id: row.id})` at `:216`), and `read_canonical_ids` reads that same column — the same shape `src/resolve/muhaddithat_links.py:188` already uses. Same id space. A parquet with a *wrong* column name raises inside `read_table` and folds into the unreadable refusal. `--canonical` is `required=True` (`src/cli.py:937`), so there is no default path to prune against by accident.

### 2. Nothing prunes by degree — the 57,993 class is safe

Confirmed at the source: the only graph read is `MATCH (n:Narrator) RETURN n.id AS id` (`src/graph/prune.py:67`) and the orphan set is a pure Python set difference (`:183`). There is no `degree()`, `size()`, or `count()` predicate anywhere in the module. The integration test seeds a canonical zero-degree narrator and asserts it survives with an explicit message (`tests/integration/test_prune_narrators.py:92`). The owner ruling in da#352 is respected.

### 3. Batching / partial failure

One `execute_write` per 1,000-id chunk (`src/graph/prune.py:203-205`), each its own transaction. A mid-run failure therefore leaves earlier batches committed. That is **safe** — only orphans ever enter a batch, and a re-run is idempotent — but the exit code is imprecise. Filed as tech debt, see below; it does not put a canonical narrator at risk.

Delete lookup is index-backed (`REQUIRE n.id IS UNIQUE` in `src/utils/neo4j_client.py:130` creates the backing index), so the per-id `MATCH` is not a scan.

### 4. `--dry-run` genuinely does not write

`prune_narrators` returns at `src/graph/prune.py:186-201`, before the write loop is even reached. Asserted twice: zero write calls against the mock (`tests/test_graph/test_prune.py:233`) and a graph re-read on a live container (`tests/integration/test_prune_narrators.py:115-116`).

### 5. The cross-repo summary contract matches deploy#557

I read the consumer rather than taking the PR body's word for it. `summary_line` (`src/graph/prune.py:243-247`) emits the four keys that `graph-prune-narrators.yml` extracts with its `summary_field` sed helper, and every gate lines up:

- dry-run `graph_total` is the **pre**-count, which is what the workflow compares against its own narrator count (yml `:414`);
- real-run `deleted` is pre-minus-post readback, which the workflow checks against the dry-run prediction (Gate C, yml `:480`);
- real-run `orphans` is the **post** count, which the workflow gates to zero (Gate D, yml `:486`).

Producer and consumer agree in both modes. Exit `12` is uniquely held (`@enum.unique` plus the `__new__` reserved-value check), and the workflow propagates a non-zero rc rather than interpreting it, so the new code costs deploy nothing.

The readback discipline throughout — measuring the post-state by re-reading the graph instead of trusting a driver-side counter — is the right instinct for a destructive op, and it is what lets me trust the numbers rather than the narration.

### Tech debt (non-blocking, filed)

- **#416** — the guard bounds empty/missing/unreadable but not a **truncated-but-non-empty** keep-set. A resolve run that crashed after writing 500 of 129,234 rows produces a parquet that is present, readable and non-empty, and would prune essentially the whole narrator graph. Production is protected today, but *only* by the deploy-side bound (yml `:419`, refuse when the orphan count is at or above the graph total); a direct CLI invocation has no such bound. Worth a da-side sanity bound so the op is safe on its own and not merely inside its envelope.
- **#417** — a mid-batch failure exits `1`, which the registry defines as "the load raised" (the very reuse da#413 rightly rejects when minting its own code); and the `orphans` key in the summary line flips meaning between dry-run and real-run. Both consumers are correct today, but it is a trap for the next one, and changing the line is a coordinated change with deploy#557.

Neither blocks. The destructive path does what it claims, refuses when it cannot be trusted, and proves it against a real graph.

---
RequestOrReplied: Approved
TechDebt: #416, #417

parametrization:
## Review — architecture & cross-repo contract lens (Jean-Claude Habimana)

Reviewed the artifact at head `42a4b11d537587028c48dfaa554c0e28c95267ca` via the contents API, and co-verified against the consumer at `noorinalabs-deploy` PR #574 head `eef7ea7cb95aed0f8ba45ef59cde86c311ee9aa4` (`.github/workflows/graph-prune-narrators.yml`). Kavitha is taking the destructive-op-guard lens; I took the layering and the producer/consumer interface, so this is deliberately a different read.

Let me start with what is right, because most of this is very good — and then with the one thing that I think blocks.

### The cross-repo contract co-verifies clean

I checked every point of the interface against the actual deploy#574 source rather than either PR's framing. It matches, including the part I expected to find broken.

The summary shape is exactly congruent. `summary_line` (`src/graph/prune.py:242-246`) emits the four keys `canonical`, `graph_total`, `orphans`, `deleted`; the workflow greps exactly those four (`graph-prune-narrators.yml:404-406` for the dry run, `452-453` for the real run) and hard-fails when any is unparseable (`407-409`, `454-456`). `src/cli.py:591` prints it unconditionally and last, in both modes, so the consumer's "a missing instrument reading is a stop" stance is actually satisfiable.

The subtlest and most breakable part of the contract is the **mode-dependent meaning of `orphans`**, and it lines up precisely. On a dry run it is the would-delete count, and the workflow consumes it as the *prediction* (`ORPHANS_K`, line 406, carried into `EXPECTED_POST` at 424). On a real run it is the post-delete remaining count, and the workflow consumes it as the *must-be-zero* assertion (`ORPHANS_POST_SUB`, line 453, gated at 486). The producer honours both meanings (`prune.py:196` dry, `prune.py:226` real). A mismatch here would have been invisible to both repos' CI, and there isn't one.

The invocation surface matches too — subcommand name (`cli.py:928`), `--canonical` required (`934-937`), `--dry-run` store_true (`945-948`) against the workflow's two invocations at lines 398 and 446. The in-container path resolves: the workflow pulls to `${PRUNE_DATA_DIR}/curated/narrators_canonical.parquet` (yml:307-314) and mounts that dir at `/app/data`, so `CANON=/app/data/curated/narrators_canonical.parquet` (yml:333) is real. And `cli.py:547-550` is right that the command must never write back into that directory — the mount is `:ro` and it doesn't.

Exit codes: the consumer does not in fact *map* codes (the PR body says it does) — it propagates the rc verbatim (yml:402, 450) and requires zero. That is weaker than advertised but it is not a mismatch: `12` and `8` are both non-zero and both stop the workflow. No action needed on this PR; I've noted it for the deploy reviewer.

Idempotency is sound. A second real run recomputes an empty complement, skips the loop entirely, reads back, and reports zero deleted and zero orphans at rc 0 — a true no-op. deploy#574 additionally short-circuits at `ORPHANS_K -eq 0` (yml:436-439) before it even invokes the delete. Two independent layers agree.

And the design instincts I most wanted to see are all here: the guard runs before *any* graph access so a bad read cannot reach Neo4j (`prune.py:180`); the outcome is measured by reading the graph **back** rather than trusting the driver's own deleted-counter (`prune.py:211-213`); the integration test keeps a zero-degree canonical narrator alive, which is the actual control that the degree-predicate alternative would have failed; and `EMPTY_CANONICAL_SET` is a new member rather than an overload of `LOAD_FAILED`, so "refused, deleted nothing" is unambiguous on the wire.

### Is a compensating prune the right layer at all?

Yes — with one condition, and the condition is not met.

I looked for the cleaner invariant before accepting the compensating one. A loader-owned canonical set would be strictly more coherent (one process, one parquet read, one transaction boundary), but it is a much larger change to a MERGE-only loader that is on the #928 critical path, and it does not help the graphs that are *already* drifted, which is the whole problem. A generation/epoch marker on the node is the other candidate — and it does not exist today: the `generation` property that `load_nodes.py:222` sets is the ṭabaqa domain field, not a load epoch. So a post-hoc set-difference really is the only correct instrument available, and reading the keep-set in Python because it lives in a parquet is not a workaround, it's the only place the predicate exists. The module docstring argues this well and the 18,518-with-edges / 57,993-zero-degree-legitimate numbers make the case airtight against every graph-side proxy.

**The condition is that a compensating action must be bound to the generation it is compensating for. This one is not, and that is my blocker.**

### Blocking — nothing binds the keep-set to the graph it is pruning

`parquet_ref` in deploy#574 is a free-text `workflow_dispatch` input (yml:91 region). Whatever parquet an operator names becomes the authoritative keep-set, and every `Narrator` outside it is DETACH DELETEd. Nothing in either repo checks that the parquet supplied is the parquet the live graph was loaded from.

Now look at what the four post-op gates actually assert (yml:467-493). Gate A says the graph was not emptied. Gate B says post equals pre minus predicted-orphans. Gate C says the delete count equals the dry-run prediction. Gate D says no orphan remains and the survivors do not outnumber the canonical ids. Every one of those is a **self-consistency** check: they verify that the prune faithfully obeyed the parquet it was handed. Not one of them can observe whether that parquet was the right parquet. They all pass for *any* parquet.

The only magnitude guard is yml:419 — refuse if `orphans >= pre`, i.e. it fires only when the keep-set would delete literally the entire graph.

So consider a stale `parquet_ref` — the previous resolve run, which is the single most plausible operator error. The premise of this whole PR is that a re-resolve collapses ids *en masse* (da#356, da#376). So the previous run's canonical set disagrees with the current graph's node set by a large margin, and the ids the current run minted all fall outside it. The prune would delete tens of thousands of perfectly legitimate narrators, edges and all — and every gate would pass, and the workflow would print "post-op verification PASSED". That is the exact shape of the trap our own `feedback_silent_zero_is_not_a_measurement` describes: the detector cannot separate the two classes it exists to separate, so the reading it produces is not a measurement.

What makes me block rather than file this is the recovery story. Gate A's own failure message tells the operator to "restore from backup" — and per deploy#558/#559/#560, backups have **never run** on stg or prod and restore has never been rehearsed. A wrong-parquet prune is therefore irreversible in practice. Combined with `feedback_security_guard_inline_not_followup` (when the threat model needs a runtime guard, a follow-up issue is tracking, not a fix), this needs the guard inline.

The good news is that the instrument is nearly free and it lives naturally in *this* PR, not in deploy. `prune_narrators` already holds both sets in memory. It computes `graph - canonical` and never looks at the reverse. **`canonical - graph`** — ids in the keep-set that the graph has never seen — is the provenance signal, at zero extra queries:

- For the correct parquet it should be ~0. Your own da#352 join says so: 71,241 + 57,993 = 129,234, exactly the canonical count, so today every canonical id *is* present in the graph.
- For a stale or wrong-run parquet it is large, and it is large *precisely in proportion to how wrong the parquet is*.

One important caveat, so this doesn't turn into a guard that false-fails: **do not assert it to zero.** A legitimately partial load leaves canonical ids un-loaded by design — `REFUSED_ROWS` (6), `STOPPED_AT_LIMIT` (3), and a `--nodes-only` run all produce a graph that correctly lacks some canonical ids. A hard `== 0` would refuse a valid prune after any of those. What I'm asking for is that the quantity be **computed, surfaced in `PRUNE_NARRATORS_SUMMARY` as a fifth key, and refused on at an obviously-wrong magnitude** — so that the operator's mandatory dry-run shows it, and so deploy#574 has something to gate on. That is the difference between gates that prove the prune obeyed the parquet and gates that prove the parquet was the graph's own.

Adding a fifth key is backward-compatible with deploy#574 as written — its `summary_field` greps by key and ignores unknown ones — so this does not desynchronise the in-flight consumer. Please pick a key name that does not *end* in `canonical`, `orphans`, `graph_total` or `deleted`, for the reason in #419 below.

I'd also accept a different mechanism if you prefer one — `src/pipeline/manifest.py` already records the md5 of every input file a load consumed, so a manifest-based binding is conceivable — but it is not plumbed to where the prune runs (deploy#574 pulls the one parquet into a scratch dir; the load's manifest isn't there), so the reverse-complement is the cheap correct move today.

### Two non-blocking findings, both filed

Failure semantics on a partially-completed batched delete — **#418**. The batch loop at `prune.py:203-205` has no handler and `_cmd_prune_narrators` catches only the guard exception, so a mid-loop driver error escapes to CPython's default handler and exits `1`. Per your own registry that is `LOAD_FAILED`, documented as "nothing the load would have written" — while in fact batches `0..k-1` are committed and an unknown number of nodes and edges are gone. Note the asymmetry: you were careful to mint `12` so that "refused, deleted nothing" is unambiguous, then let "partially deleted" collapse onto the code that means "nothing happened". I am **not** blocking on it, because the complement is computed before any write and every batch is a subset of it — a partial run can only ever delete orphans, never a canonical narrator, and a re-run converges. The damage is to observability, not to the data. But since you'll be in `prune.py` for the blocker anyway, this is a cheap thing to fix in the same pass if you want it gone.

Summary-line key namespace — **#419**. The consumer's `summary_field` (yml:359-361) matches keys with a greedy, unanchored `.*$2=`. Nothing collides today, but the line is designed to grow, and the most natural next addition breaks it: add `edges_deleted=` to a `DETACH DELETE` summary and `summary_field ... deleted` greedily matches *that*, feeding an edge count into gate C's node-count comparison. The consumer-side anchoring belongs on deploy#574 and I'm raising it there; the producer-side half (treat the key set as an append-only namespace, pin it with a unit test over `summary_line`) is filed here.

### Two smaller observations, no action needed

`graph_ids_pre` filters on `if row.get("id")` (`prune.py:182`), so a null-id `Narrator` would be excluded from `graph_total`, while deploy's pre-count is a plain `count(n)` over all `Narrator`. The two would disagree and yml:414 would refuse with "instrument/graph disagreement". That fails *safe* — it refuses rather than deletes — so I'm flagging it only so nobody later "fixes" the divergence in the wrong direction.

`deleted = len(graph_ids_pre) - len(graph_ids_post)` is a difference of two totals measured at different times, so a concurrent `deploy-data-load` run would corrupt the arithmetic. The workflow's concurrency group only serialises prunes against prunes, not against loads. It also fails safe (gates B and C would red), so this is a note rather than a request.

### Summary

Excellent work on the parts that are genuinely hard — the layering argument is correct and well-evidenced, the guard-before-graph-access ordering is right, measuring the outcome by readback instead of the driver's self-report is right, and the mode-dependent `orphans` semantics co-verify exactly against the consumer, which is the one place I most expected to find a silent cross-repo break.

I'm requesting changes on one thing only: the keep-set's authority is the load-bearing premise of the entire destructive op, and it is the one proposition none of the gates can check. Compute `canonical - graph`, surface it in the summary, and refuse on an implausible magnitude — and I'll approve on the re-read.

---
TechDebt: #418 (partial-delete exits LOAD_FAILED), #419 (summary key namespace / unanchored consumer regex)

parametrization:
## Blocker addressed — the keep-set is now bound to the graph it prunes

New head: **`616dbc3ccfef6503f16b1bc0a3600d9ca72b6027`**. Both prior verdicts were given at `42a4b11` and do not carry across the new commit, so I'm asking **Kavitha Sundaramurthy** and **Jean-Claude Habimana** to re-read at this sha.

Jean-Claude — you were right, and the framing of the blocker did the work: the gates all prove the prune *obeyed* the parquet, and every one of them is derived from the keep-set they'd be validating, so none can ask whether it was the *right* parquet. I implemented the instrument you specified, and deliberately not the policy.

### What landed

**`missing` = `|canonical − graph|`**, computed from the two sets already in memory — zero extra queries — and emitted as the fifth key, appended **last**:

```
PRUNE_NARRATORS_SUMMARY canonical=<n> graph_total=<n> orphans=<n> deleted=<n> missing=<n>
```

It is **reported, not asserted to zero** — exactly as you argued. `REFUSED_ROWS`, `STOPPED_AT_LIMIT` and a nodes-only load each leave canonical ids legitimately unloaded, so a hard zero-assert here would refuse a valid partial-load prune. deploy#574 owns the refuse-or-proceed policy.

It **is** refused on implausible magnitude, before any write: `ForeignCanonicalSetError` above `--max-missing-fraction` (default `0.5`, overridable, `1.0` disables). The justification for one half, from the da#352 numbers you cited: for the parquet the graph was loaded from, `missing` is 0 — 71,241 edged + 57,993 zero-degree = 129,234 = the whole canonical count. A partial load pushes it up but the keep-set still describes *this* graph. At one half it no longer does — more of the keep-set is unknown to the graph than known, so it has lost majority authority over the thing it is being used to prune. That is not a partial load; that is a different graph. It is the "obviously wrong" line, not a tuned threshold pretending to precision.

I've been explicit in the module docstring about what this does **not** buy: a *subtly* stale parquet (one generation back, most ids still present) slips under any plausible ceiling. Catching that one is deploy#574's `missing == 0` gate. The two layers are complementary and neither alone is sufficient — I'd rather say so than let the ceiling read as complete cover.

### The measured separation

Same graph (12 Narrators: 10 live-generation canonical + 2 real orphans), same command, two keep-sets — both readable, both non-empty, both 10 ids:

| keep-set | `missing` | outcome |
|---|---|---|
| correct (the parquet the graph was loaded from) | 0/10 = **0%** | proceeds, deletes the 2 orphans |
| correct, but a **partial** load (2 canonical never loaded) | 2/10 = **20%** | **reported, not refused** — deletes exactly the 2 orphans |
| **stale** (previous run; da#337 splitter re-keyed 8 of 10) | 8/10 = **80%** | **refused, zero rows deleted** |

With the ceiling lifted, that stale keep-set is measured deleting **10 of 12 nodes, 8 of them legitimate live narrators** — while `graph_total`, `orphans` and `deleted` all stay perfectly consistent and every post-op gate greens. That run is pinned as a test, so the hazard is demonstrated rather than asserted.

### On the `#419` coupling — verified, not taken on trust

You told me to check it myself, so I ran deploy#574's actual `summary_field` sed (`graph-prune-narrators.yml:359-361` at `eef7ea7`) against the real five-key line, rather than paraphrasing it in Python `re`.

**`missing=` does not collide.** All five keys extract their own value. But a clean read from a detector I hadn't proven *live* would have been worth nothing, so the test first feeds the same sed a line it **must** misread — `edges_deleted=999` alongside `deleted=4` — and requires it to return `999`. It does. Your predicted hazard reproduces exactly; the harness is therefore known to be capable of detecting a collision, and only then does the clean read on `missing=` mean anything. Both are permanent tests, and `SUMMARY_KEYS` now carries a mechanical suffix-invariant so whoever adds the sixth key can't quietly re-arm it.

### Exit code — I reused `12` rather than minting `13`

Worth flagging since you may disagree. All four doors leave **the graph exactly as it was**, name the same remedy, and are propagated verbatim by a consumer that only asks whether the rc is zero. `src/exit_codes.py`'s own rule is that a code is defined by what is on disk when it fires, and that is identical here — so a fifth member would encode a distinction nothing acts on, while re-opening the three-way collision hazard that module exists to record. The distinction lives where it's actually consumed: a distinct exception subclass in-process, and a diagnostic on stderr. I broadened the member docstring to name the fourth door explicitly. Happy to mint `13` instead if you read the registry differently.

### Scope

I left **#418** (partial-delete collapsing onto `LOAD_FAILED`) to its own issue rather than folding it in — it needs its own exit code, which is its own collision question, and I didn't want to bundle that into a blocker fix. Your call if you'd rather it rode along.

The cross-repo contract you co-verified is untouched: the four existing keys, their order, and the mode-dependent meaning of `orphans=` are all exactly as they were.

### Checks

All six green at `616dbc3` — Lint, Type Check, Test, **Integration Tests**, base-image/fixture-realism, and the sync-drift gate. The two new real-Neo4j tests (stale keep-set refused with the graph and its edges intact; partial load reported and still pruned) run in the Integration Tests job — my local Docker daemon is down, so CI is where they were actually exercised.

---
Requestee: Kavitha Sundaramurthy, Jean-Claude Habimana
RequestOrReplied: Reply

parametrization:
## Re-review at `616dbc3c` — ChangesRequested

Second pass, reviewing the artifact at the new head via the contents API. The stale-keep-set work is genuinely strong and I want to be specific about that before the blocker, because the blocker is a *sibling* of the hole you just closed, not a rejection of the fix.

### What I verified, and what holds

**The wrong-but-valid fixture is real, and the guard is not inert.** This was my first question and it is answered properly. `_stale_generation_ids()` (`tests/test_graph/test_prune.py:442-447`) builds the stale keep-set from the **production minting functions** — `make_discriminated_canonical_id` (da#337 same-name split) against `make_canonical_id` — over ten real corpus narrators, written as full parquet rows. `test_the_stale_keep_set_passes_every_pre_existing_guard` (`:482-499`) then proves the fixture actually *reaches* the bad state: `read_canonical_ids` does not raise, the keep-set is ten valid `nar:` ids, eight of which the live graph has never seen. That is the control that stops this being theatre, and it is exactly the check `feedback_fixture_makes_guard_assertion_inert` demands.

**The damage is measured, not asserted.** `test_without_the_ceiling_the_stale_keep_set_deletes_the_live_generation` (`:501-531`) reproduces Nino's finding in-repo: ten of twelve nodes deleted, eight of them live edge-bearing narrators, with gates A, C and D all still green and only `missing` separating it from a correct run. Then `test_stale_keep_set_is_refused_before_any_write` (`:533-549`) pins the green with `client.deleted_batches == []`. **Refusal precedes any write**: the raise sits at `src/graph/prune.py:348`, the write loop at `:379`. It fires on `--dry-run` too. Question 3, answered.

**The dual is pinned.** `test_correct_keep_set_partial_load_is_reported_and_proceeds` (`:594-616`) shows a legitimate `REFUSED_ROWS`-style partial load reporting a non-zero `missing` and still pruning only the orphans, and `test_the_two_classes_are_separated_by_missing` (`:618-647`) runs both classes against the same graph and requires the signal to separate them. That is the right shape — a number for one class alone proves nothing. Question 2, answered.

**The collision check was actually done, not assumed.** This is the part I'd single out. `_consumer_extract` (`:333-342`) shells out to the **real POSIX `sed`** with deploy#574's expression copied verbatim from `graph-prune-narrators.yml` at head `eef7ea7`, rather than paraphrasing it into Python `re` — and `test_the_harness_can_actually_detect_a_collision` (`:357-369`) feeds it `edges_deleted=999` and *requires* the harness to be hijacked, so the clean five-key read below it means something. I independently confirmed the invariant: hijack requires one key to be a suffix of another, and `missing` is a suffix of none of `canonical` / `graph_total` / `orphans` / `deleted`, nor they of it. `SUMMARY_KEYS` pins it mechanically at `:299-305`. Question 4, answered — and answered better than asked.

---

### The blocker: `missing` bounds the direction that does not delete

`missing = |canonical − graph|` is the right *provenance* signal and I have no argument with it. But the magnitude ceiling built on it (`src/graph/prune.py:348`) bounds the one quantity that **cannot cause a deletion**, and the deleting quantity — `orphans = |graph − canonical|`, already computed on the very next line at `:357` — is left with **no bound at all**. I grepped for one; there is none.

That asymmetry leaves a keep-set that is wrong in the *other* direction wide open: a **truncated but same-generation** parquet. A resolve run stopped early (`STOPPED_AT_LIMIT` — which your own module docstring names) or one whose canonical write was partial produces a parquet with, say, 500 rows of *real, current-generation* ids. Every id in it **is** in the graph. So:

| check | value | outcome |
|---|---|---|
| Guard 1 (empty/missing/unreadable) | readable, 500 ids | passes |
| Guard 2 — `missing > 0.5 × canonical` | `0 > 250` | **passes** |
| deploy#574 `:419` — `orphans >= graph_pre` | `160,114 >= 160,614` | **passes** |
| deploy#574's planned `missing == 0` policy | `missing` is 0 | **passes — actively blesses it** |
| **result** | | **DETACH DELETE 160,114 of 160,614 = 99.7% of the graph** |

And every post-op gate then goes green exactly as before: the graph is not empty (500 remain), `deleted` equals the dry-run prediction, `orphans_post` is 0, survivors ⊆ canonical. The workflow prints `verification PASSED`.

This is the *same* failure Jean-Claude and Nino identified — every gate derived from the keep-set being validated — reached through a different door. The `missing == 0` policy in deploy#574 does not merely miss it; because `missing` is exactly 0 here, it certifies it. So this cannot be deferred to the consumer: the policy layer as designed is blind to it by construction.

**The fix is nearly free and symmetric to the one you already wrote.** You hold `graph_ids_pre` and `orphans` in memory at `:357`. A ceiling on the delete magnitude — refuse when `len(orphans) / len(graph_ids_pre)` exceeds some plausible fraction, with the same `--max-...-fraction` escape hatch and the same `UnusableCanonicalSetError` supertype so the CLI catch and exit code need no change — closes it. Deleting >50% of the narrator graph is never a routine prune, and if it genuinely is, the operator can say so explicitly, exactly as they can today for a large partial load.

I filed this as tech debt (#416) after my first pass and, given that the deploy-side bound turns out not to catch it either and the planned policy gate blesses it, I no longer think a tracking issue is the right instrument. Per `feedback_security_guard_inline_not_followup`, a threat model that needs a runtime guard gets the guard inline, not a follow-up. Please fold the delete-side ceiling into this PR and I will re-review promptly — everything else here is ready.

### Still-open tech debt (non-blocking, unchanged)

**#417** — a mid-batch failure still exits `1` (`LOAD_FAILED`, the reuse this PR rightly rejects when minting its own code), and the `orphans` key still flips meaning between dry-run and real-run. Both consumers are correct today; neither needs to change for this PR.

Reusing exit `12` for `ForeignCanonicalSetError` via the `UnusableCanonicalSetError` supertype is the right call — the operator-visible fact ("refused, graph untouched, fix the parquet") is identical, and `src/exit_codes.py:356-358` documents the fourth door rather than leaving the member's name to imply a narrower contract. No objection there.

---
TechDebt: #417

parametrization:
## Re-review at `616dbc3c` — Jean-Claude Habimana

Re-read the artifact at head `616dbc3ccfef6503f16b1bc0a3600d9ca72b6027`, and re-verified the contract against deploy#574 at `eef7ea7cb95aed0f8ba45ef59cde86c311ee9aa4` — which is **unchanged** since my first pass, so the consumer-side policy half is not landed yet and every backward-compatibility question below is a live one, not a hypothetical.

The instrument I asked for is here and it is built correctly. I have to hold ChangesRequested anyway, on a second door into the same catastrophe that I under-specified last time — and I want to be clear that this is a gap in my ask, not in your execution.

### The four things I was asked to confirm — all four clear

**The instrument is genuine, and free.** `graph_id_set` at `src/graph/prune.py:342` makes membership O(1), and `missing = sum(1 for cid in canonical_ids if cid not in graph_id_set)` at `:347` is a true reverse difference over the two sets already in memory. No extra query, no second graph read. That is exactly what I asked for and not a weaker proxy for it.

**You avoided the trap I named.** `missing` is reported, never asserted to zero. The only refusal is on magnitude — `missing > max_missing_fraction * len(canonical_ids)` at `:348`, defaulting to 0.5 — so a legitimately partial load (`REFUSED_ROWS`, `STOPPED_AT_LIMIT`, nodes-only) that leaves some canonical ids unloaded still prunes. The ordering is right too: `ForeignCanonicalSetError` is raised at `:349`, after the graph read it necessarily depends on but strictly before the delete loop at `:379`, and it fires in dry-run mode as well, which is the mode where a wrong parquet actually needs to surface. The comment at `:344-346` — that `missing` is mode-invariant because a deleted orphan is by definition not canonical, so the delete cannot move the number — is a sharper observation than the one I made.

**#419 was verified, not assumed.** `SUMMARY_KEYS` at `:418` declares the namespace, `test_no_key_is_a_suffix_of_another` pins the invariant, and `test_missing_does_not_collide_with_any_existing_key` re-implements the consumer's actual sed and runs `missing` through it. Most importantly, `test_the_harness_can_actually_detect_a_collision` feeds the harness a line that *must* be hijacked and asserts that it is — a positive control proving the collision detector can detect a collision at all. A suffix test that passes against a harness incapable of failing would have been worth nothing, and you built the control rather than trusting the green. That is the right instinct and it is the thing I would have asked for next.

**The contract still holds with the fifth key.** `missing=` is appended last (`:438-440`); none of `canonical`, `graph_total`, `orphans`, `deleted` is a suffix of it, nor it of them, so the consumer's four existing greedy extractions at deploy#574 `graph-prune-narrators.yml:404-406` and `:452-453` parse identically against the new line. The mode-dependent meaning of `orphans` — the subtlest axis, and the one I most expected to break — survives intact (`:371` dry, `:403` real). The new refusal shares rc `12`, and the consumer only asks whether the rc is zero, so an unchanged deploy#574 stays safe against this producer. Nothing desynchronises.

On sharing rc `12` rather than minting a thirteenth: I think that call is right, and the docstring defends it on the registry's own terms — a code is defined by what is on disk when it fires, and all four doors leave the graph untouched and name the same remedy. One inconsistency worth a moment of your time, and it is not a blocker: that same docstring says *"naming the member after only the empty case would re-arm the trap for the other three"* — while the member is still named `EMPTY_CANONICAL_SET`. The value should stay `12`; the name wants to be `UNUSABLE_CANONICAL_SET`, which is what your own exception supertype is already called. Zero behavioural change, and it makes the member say what it now means.

### Blocking — a second door to the same 97%, and the guard is blind to it

I ran the real `prune_narrators` from this head against three keep-set classes, with a fake client serving the da#352 measured graph (160,614 narrators, 129,234 canonical). Not a reconstruction of the logic — the actual function, actual parquet, actual guard:

```
ceiling = missing > 0.5 * |canonical|

A correct parquet          PROCEEDED  missing=0       deleted=31380   (19.5% of graph)  survivors=129234
B foreign stale parquet    REFUSED (ForeignCanonicalSetError)  missing=129234/129234
C truncated 5,000-id set   PROCEEDED  missing=0       deleted=155614  (96.9% of graph)  survivors=5000
```

**A** is the happy path and it is not false-refused — good, the ceiling is not too tight. **B** is the class the new guard was built for and it is caught cleanly. **C** is a keep-set of 5,000 ids that are all *genuinely this graph's* — a truncated or partially-written parquet, a partial upload, a null-heavy `canonical_id` column, of the **right** resolve run. And `missing` is **0**, because every id in the keep-set really is in the graph. The set isn't foreign. It's just *incomplete*.

So `ForeignCanonicalSetError` stays silent. deploy#574's planned `missing == 0` gate would *also* stay silent — `missing` is 0, that gate passes. And the workflow's magnitude guard at `yml:419` only fires at `orphans >= pre`, i.e. at 100%; 96.9% sails under it. I walked the four post-op gates against C and every one passes: post (5,000) equals pre (160,614) minus predicted orphans (155,614); deleted equals predicted; orphans-remaining is 0; survivors do not exceed the canonical count. The workflow prints **verification PASSED** having irreversibly deleted 155,614 legitimate narrators and every edge attached to them, with no backup to restore from.

That is the same sentence I wrote last time — the gates prove the prune obeyed the parquet, never that it was the right parquet — arriving through the other door. `missing` closes *foreign*. It does not close *truncated*, and truncated is not the exotic case: it is what a half-written parquet or an interrupted resolve publishes.

I want to name my own share of this clearly. In my first review I flagged the null-heavy-parquet variant and called it *"partially covered by finding #1's instrument."* That was wrong — it is not covered at all, and it is not covered precisely because `missing` is the wrong axis for it. You implemented the instrument I specified, faithfully and better than I specified it. The specification was short a guard.

**What I think closes it.** A delete-side magnitude ceiling, symmetric with the one you already built: refuse when `len(orphans)` is an implausible fraction of `len(graph_ids_pre)`, raised in the same place, before any write, overridable deliberately. It mirrors `max_missing_fraction` exactly, so the shape is already in your hands.

One caution, because I do not want to hand you a guard that false-refuses the way I warned about last time: **do not set that ceiling at 0.5.** The graph is the union of every load ever run against it, so after a genuine large re-resolve the legitimate orphan set can exceed half the graph — I get about 55% for a graph holding an old 160k load plus a heavily re-minted 129k one. A ceiling around 0.9 with a `--max-orphan-fraction` escape hatch catches C at 96.9% while leaving a real large cleanup room to run, and it puts the operator's override on the record when they genuinely mean it.

### Standing tech debt

**#418** (a partially-completed batched delete still exits `1` = `LOAD_FAILED`, documented as "nothing written", while batches `0..k-1` are committed) is untouched at this head and still stands. Still not a blocker for the reason I gave — the complement is fixed before any write, so a partial run can only ever delete orphans and a re-run converges.

**#419** is half-closed. The producer side is genuinely done here — append-only namespace, suffix test, verified control. The consumer-side anchor in deploy#574 is not landed, so the issue stays open. **To the sequencing question: it does not need to land before this merges.** `missing` is proven non-colliding by a test with a working control, so the fifth key is safe against today's consumer. The unanchored regex is a trap for the *sixth* key, not this one — it should land before anyone adds another, not before this.

### Where this leaves us

Everything I blocked on last time is properly addressed, and the collision control you built goes beyond what I asked. The remaining item is one more guard of a shape you have already written once, and I would clear it on sight. I am sorry to be sending this back for a door I should have seen the first time.

---
TechDebt: #418 (partial-delete exits LOAD_FAILED), #419 (consumer-side anchor for the summary key regex; producer half done here)

parametrization:
## Delete-side ceiling added — re-review at `df8a17da2df2bec504bebb8e227018b645984854`

New head invalidates both verdicts again. **Kavitha Sundaramurthy** and **Jean-Claude Habimana**, please re-read at that sha.

Kavitha, Jean-Claude — you both found the same door independently, from different angles, and you are right. `missing` bounds `canonical − graph`, a set whose elements are *not in the graph* and therefore **cannot be deleted**. It is structurally incapable of bounding `orphans = graph − canonical`, which is the set that actually deletes. I implemented the instrument that was specified and never asked what bounded the other direction. That is on me, not on the spec.

And the sharpest part of both reviews is the same observation: the planned `missing == 0` policy in deploy#574 does not merely *miss* a truncated keep-set — because `missing` is exactly 0 on it, that gate **certifies** it. A guard that greenlights the catastrophe it was built to prevent is worse than no guard.

### What landed

`ExcessiveDeletionError`: refuse when `|orphans| / |graph|` exceeds `--max-orphan-fraction`. Same shape as the ceiling I already built — before any write, in dry-run too, same `UnusableCanonicalSetError` supertype, so the CLI catch and exit code are unchanged.

### The ceiling is 0.9, not 0.5 — and that is deliberate

This is the one place I did **not** follow the guidance I was given. Kavitha, you wrote that deleting >50% is never a routine prune; I was separately advised to use 33–50%. Jean-Claude, you warned specifically against 0.5. I ran it, and Jean-Claude is right, so I want to show the working rather than just assert it.

I ran the **real** `prune_narrators` from this head against five keep-set classes, at da#352 scale (160,614 narrators, 129,234 canonical) — the actual function, actual parquet, actual guards:

| class | `missing` | `orphan%` | verdict |
|---|---|---|---|
| **A** correct parquet | 0 | 19.5% | PROCEED — deletes 31,380 |
| **A2** accumulated graph, full re-mint | 0 | **55.4%** | **PROCEED** — deletes 160,614 |
| **B** foreign / stale parquet | 129,234 | — | REFUSE (`ForeignCanonicalSetError`) |
| **C** truncated 5,000-id keep-set | **0** | 96.9% | REFUSE (`ExcessiveDeletionError`) |
| **D** truncated 500-id keep-set | **0** | 99.7% | REFUSE (`ExcessiveDeletionError`) |

**A2 is why 0.5 is wrong.** The graph is the union of every load ever run against it, so the mass id re-mint this subcommand *exists to clean up after* (da#356/da#376) legitimately orphans the entire previous generation: an old 160,614-node generation plus a freshly loaded, fully re-minted 129,234-node one is 289,848 nodes, of which 160,614 are legitimate orphans — **55.4%, with `missing == 0`, and it is a correct prune.** A 0.5 ceiling refuses this command's own primary use case. The operator overrides; the override becomes reflex; the guard is dead. That is the same failure mode as a hard `missing == 0` refusing a legitimate partial load — the trap Jean-Claude warned me off last round, and it would have been hypocritical to walk into it one guard later.

The legitimate band tops out around 55% and the threat band starts at 97%. The gap between them is empty, and 0.9 sits in it with margin on both sides. It is the "no legitimate prune looks like this" line, not a tuned parameter.

Kavitha — if you still want 0.5 after seeing A2, say so and I will change it; but I think A2 is a prune we will actually need to run.

### Tests — the same bar, applied to the new door

- **The control first.** `test_the_truncated_keep_set_passes_every_pre_existing_guard` proves the fixture reaches the bad state through *everything that already exists*: the empty/unreadable guard passes, `missing` is **exactly 0** (not merely small), the consumer's whole-graph-wipe check passes, and — Weronika's proposed invariant — `EXPECTED_POST == CANON_N` passes too. Every guard we have ever designed says this keep-set is fine, and it would delete 95%+ of the graph. Without that control the new ceiling would be theatre, which is precisely the point you made, Kavitha.
- **The red.** With the ceiling lifted, the truncated set is *measured* deleting 192 of 202 nodes while every post-op gate greens.
- **The green.** Default ceiling: refused, `deleted_batches == []`, graph byte-identical. Also on a real Neo4j — node count *and* edge count unchanged.
- **The dual.** A2 proceeds under the default, and a counterfactual test pins that a **0.5 ceiling would have refused it** — so nobody can "tighten" the ceiling later without a red test explaining why not.

Fixtures are minted by the production `make_canonical_id` from real name morphology (base name + nisba), not hand-faked ids.

### Also folded in

- **#416 closed inline**, per your note, Kavitha — a threat model needing a runtime guard gets the guard, not a tracking issue.
- **`ExitCode.EMPTY_CANONICAL_SET` → `UNUSABLE_CANONICAL_SET`** (value stays `12`), Jean-Claude's catch. Its own docstring argued that "naming the member after only the empty case would re-arm the trap for the other doors" while doing exactly that — and the truncated door is the one the narrow name most obscured. Zero behavioural change; the consumer propagates the rc verbatim and never names the member.

I did **not** add a sixth summary key for the orphan fraction — deploy#574 can derive it from `orphans` and `graph_total`, and per your read, Jean-Claude, the unanchored regex is a trap for the sixth key. Not walking into it.

**#417 / #418 / #419** remain untouched non-blocking tech debt.

### Checks

All six green at `df8a17d` — Lint, Type Check, Test, **Integration Tests**, base-image/fixture-realism, sync-drift gate.

---
Requestee: Kavitha Sundaramurthy, Jean-Claude Habimana
RequestOrReplied: Reply

parametrization:
## Addendum at `3c30856da925a25ff09ff47a3fedc6f1101b149c` — deriving the ceiling turned up a hole it does *not* close

Same guard, no behaviour change. I was asked to **derive** 0.9 rather than take it on faith — show why the gap is wide enough to be meaningful, and what would narrow it. Doing that properly surfaced something I had not said out loud, and it changes what this PR should be understood to guarantee.

**Kavitha, Jean-Claude — please read this before approving.** It does not undo the fix, but it materially narrows the claim.

### The band closes from both ends

**From below — skipping the prune.** Each unpruned re-mint leaves another stale generation resident, so a *legitimate* cleanup's orphan fraction is `k/(k+1)`: 50% at `k=1`, 75% at `k=3`, and it **reaches 0.9 at `k=9`**. The 0.9 ceiling is only sound while the prune runs regularly. deploy#557/#574 runs it per load (`k=1`), so this is fine today — but a graph left unpruned for many cycles would start false-refusing its own cleanup.

**From above — a mild truncation. This one does not close, and no ceiling closes it.** The delete-side ceiling catches a keep-set truncated to a *remnant*. It cannot catch one truncated to a *fraction*. With the keep-set retaining share `f` of the canonical set, the prune deletes `1 − f × 129,234/160,614`:

| `f` | deletes | verdict |
|---|---|---|
| 0.10 | 92.0% | REFUSED |
| 0.124 | 90.0% | REFUSED (the ceiling) |
| **0.20** | **83.9%** | **PASSES** — 134,768 legitimate narrators gone |
| **0.40** | **67.8%** | **PASSES** |
| **0.60** | **51.7%** | **PASSES** |

**A parquet that lost a fifth of its rows deletes 84% of the narrator graph and is refused by nothing here.** And lowering the ceiling cannot fix it, because **83.9% (truncated) and 55.4% (the legitimate full re-mint, class A2) are the same *kind* of number.** The bands genuinely overlap. A magnitude cannot separate them.

So the honest guarantee — now stated in the module docstring instead of implied away:

> A **wrong** keep-set (foreign) is refused. A **remnant** keep-set is refused. A **short** keep-set is **not** — it is only reported, and the operator's dry run is the sole thing standing in front of it.

### What actually closes it, and why I didn't do it here

Only a **completeness** signal can — an expected row count, or the `src/pipeline/manifest.py` md5 of the resolve output — so a short parquet is caught by being **short**, not by being **drastic**. That is a different axis from any magnitude ceiling. Jean-Claude, this is your manifest-binding suggestion from the first review, and I now think you were pointing at the real cure rather than an alternative to it.

It is not plumbed to where the prune runs (deploy#557 pulls the one parquet into a scratch dir; the load's manifest is not there), and I am not going to smuggle a cross-repo plumbing change into a review fix. **I recommend a follow-up issue for keep-set completeness binding** — happy to take it — and I would treat it as a real gap rather than a nice-to-have, given there is still no restore path (deploy#558/#559/#560).

### Pinned, not just documented

`TestKnownBlindSpotMildTruncation` asserts the **hole**: a keep-set missing 60% of its rows deletes 120 of 200 legitimate narrators with `missing == 0` and no refusal. It exists so nobody — us included, in three months — reads the two ceilings as closing truncation.

It should start **failing** the day a completeness binding lands. Deleting it then is correct. "Fixing" it now by lowering the ceiling until it goes green would destroy the legitimate A2 prune the ceiling was raised to protect — which is precisely the trap this whole review chain has been walking us out of.

All six checks green at `3c30856`.

---
Requestee: Kavitha Sundaramurthy, Jean-Claude Habimana
RequestOrReplied: Reply

parametrization:
## Re-review at `3c30856d` — Approved. Clearing my ChangesRequested.

**On the threshold: you were right and I was wrong. Keep 0.9.** You asked me directly, so let me answer directly rather than defer.

I did not take the A2 row on faith — I re-derived the whole band against the da#352 numbers before reading your argument. It reproduces exactly, and it carries a consequence I had not seen when I wrote "deleting >50% is never a routine prune":

| | orphan fraction | |
|---|---|---|
| legitimate full re-mint (A2), `missing == 0` | **55.4%** | must proceed |
| keep-set truncated to 20% of its rows, `missing == 0` | **83.9%** | must refuse |
| keep-set truncated to 60% of its rows, `missing == 0` | **51.7%** | must refuse |

**A2 sits *inside* the truncation band.** That is the finding. It means no ceiling anywhere separates legitimate from short — the classes are the same kind of number — so the choice is not "0.5 is safer than 0.9". The choice is: refuse this command's own primary use case and get overridden by reflex, or place the line above the worst legitimate case and be honest that it is a backstop on the catastrophic tail rather than a completeness check. You took the second, which is correct, and it is the same trap Jean-Claude steered you off one round earlier with a hard zero on `missing`. My 0.5 would have re-armed it on the other axis.

Two things convinced me the placement is principled rather than convenient. First, `test_a_ceiling_at_one_half_would_have_refused_that` (`tests/test_graph/test_prune.py:861-874`) pins the counterfactual, so 0.9 is defended by a test and not by a comment — nobody "tightens" it later without going red. Second, you volunteered the ceiling's own failure mode from below: the `k/(k+1)` creep of unpruned generations, which I checked and which reaches exactly 0.9 at `k=9`. So 0.9 is sound while deploy#557/#574 runs the prune per load (`k=1` → 50%), and you said so in the docstring instead of leaving me to find it. A guard whose limits are documented by its author is worth more than one whose author only knows its strengths.

### A2 is a genuine scenario, not a constructed one

I checked this rather than assuming, because the whole argument rests on it. The loader is `MERGE (n:Narrator {id: row.id})` (`src/graph/load_nodes.py:216`) — MERGE-only, so the graph is the union of every load ever run. A re-resolve that re-mints ids en masse therefore leaves the previous generation resident in full, and the new canonical set orphans all of it with `missing == 0`. That is not a hypothetical: it is precisely the da#356 de-keying and da#376 fold this subcommand was built to clean up after, and a change to `make_canonical_id` / `normalize_arabic` re-mints everything. A2 is the *worst legitimate case*, and bounding a backstop by the worst legitimate case is the right construction.

### The control holds — all four instruments, verified

`test_the_truncated_keep_set_passes_every_pre_existing_guard` (`:732-768`) is what I asked for and it delivers: the empty/missing/unreadable guard passes (ten real ids), `missing` is asserted **exactly 0** (not merely small — so deploy#574's planned policy does not just miss this, it certifies it), deploy's whole-graph-wipe check passes, and Weronika's `EXPECTED_POST == CANON_N` invariant passes. Every guard we have ever designed says the keep-set is fine, and it would delete 95%+ of the graph. Without that the new ceiling would be theatre; with it, it is load-bearing.

The rest checks out: refusal precedes any write (raise at `src/graph/prune.py:559`, write loop at `:590`), `deleted_batches == []` on refusal, it fires on `--dry-run`, the empty-graph case cannot divide by zero, the boundary is strictly-greater and pinned, and both classes are proven against a real Neo4j (`test_truncated_keep_set_refused_against_a_real_graph`). The dual proceeds: `test_full_regeneration_orphans_the_majority_and_still_proceeds` deletes 57.1% without refusal. CI is 6/6 with a non-empty rollup.

### The one hole that remains — filed, not blocked

`TestKnownBlindSpotMildTruncation` (`:877-917`) asserts a **hole** rather than a fix, and pins the condition for its own deletion. That is the most honest thing in this PR. A keep-set that lost a fifth of its rows still deletes 83.9% and is refused by nothing here.

I am not blocking on it, and I want to be explicit about why the call differs from last round, because it is not a softening. Last round the fix was three lines using data already sitting in memory — that is a guard, and per our own rule a guard goes inline. This one cannot be fixed in this module at all: it needs a **completeness** signal (expected row count, or the resolve manifest's md5) that is not among the command's inputs, because deploy pulls a lone parquet from a free-text `parquet_ref` and the producing run's manifest does not travel with it. Lowering the ceiling does not substitute for it — it just kills A2. Filed as **#425** with the full analysis and the cross-repo plumbing it needs; **#416** is closed as superseded, since its premise ("no bound at all") is no longer true.

Until #425 lands, the guarantee is exactly the one your docstring states — foreign refused, remnant refused, short only *reported* — and the operator's dry run is what stands in front of the last case. Shipping that with the limit written down beats shipping a ceiling that pretends to more.

One prose nit, no action needed on my account: the summary at `src/graph/prune.py:128` calls the 55%–97% band "empty", which the mild-truncation section twenty lines later correctly shows it is not. The correction is in the same docstring under a heading telling the reader to read it, so it self-repairs — but if you touch this file again, softening the earlier line would stop a skimmer stopping at the reassuring half.

Excellent work. This is a destructive op I would now be comfortable running against prod.

### Still-open tech debt (non-blocking, unchanged)

**#417** — mid-batch failure still exits `1` (`LOAD_FAILED`), and the `orphans` key still flips meaning between dry-run and real-run. Neither needs to change for this PR.

---
RequestOrReplied: Approved
TechDebt: #417, #425

parametrization:
## Re-review at `3c30856d` — Jean-Claude Habimana

Note the head has moved past `df8a17da`: current head is `3c30856dda925a25ff09ff47a3fedc6f1101b149c` ("docs(graph): pin what the two ceilings do NOT catch — mild truncation"), 6/6 green. That docs commit turns out to be the most important one in the series, so I reviewed there.

**Approving.** The delete-side ceiling is right, the control is genuine, and the contract still holds. But I want to answer the question I was asked plainly, because the answer is not the one the commit message implies.

### The band is not empty — and your own docstring is what proves it

I was asked whether anything lives between 55% and 97%. It does, and you found it before I did.

I ran the real `prune_narrators` at this head against the full spectrum — a graph of 160,614 narrators, a 129,234-id canonical set, varying the share `f` of canonical rows the parquet actually retains:

```
A  single-gen (da#352 measured)    PROCEED missing=0   deletes  31380 = 19.5%
A2 accumulated, full re-mint       PROCEED missing=0   deletes 160614 = 55.4%
B  foreign / stale parquet         REFUSE  ForeignCanonicalSetError
C  truncated to 5,000              REFUSE  ExcessiveDeletionError
D  truncated to 500                REFUSE  ExcessiveDeletionError

   mild truncation f=0.124         REFUSE  ExcessiveDeletionError   (the ceiling)
   mild truncation f=0.20          PROCEED missing=0   deletes 134768 = 83.9%
   mild truncation f=0.40          PROCEED missing=0   deletes 108921 = 67.8%
   mild truncation f=0.60          PROCEED missing=0   deletes  83074 = 51.7%
   mild truncation f=0.80          PROCEED missing=0   deletes  57227 = 35.6%
```

Every number in your table reproduces exactly, including A2 at 55.4% — which is the figure I derived independently, so it is a reproduction and not a reverse-engineering of my estimate. The arithmetic is sound: 160,614 legitimate orphans in a 289,848-node accumulated graph, and it is a *correct* prune. Your case against 0.5 is made, and it is made with the command's own primary use case, which is the strongest possible way to make it.

But that means the band is populated, and the docstring says so itself. Lines 138-142 give the `k/(k+1)` law — the legitimate orphan fraction reaches 0.9 at `k=9` stale generations. Lines 144-165 show mild truncation overlapping the legitimate band outright: a parquet retaining a fifth of its rows deletes 83.9%, and a legitimate full re-mint deletes 55.4%, and those are the same kind of number. So the sentence ten lines above them — that 0.9 *"sits in the empty band between ~55% and ~97%"* and *"is the 'no legitimate prune looks like this' line, not a tuned parameter"* — is not true, and the material immediately below it is the refutation.

This does not change my verdict, because **0.9 is the right number anyway**, and for a better reason than the one given. It is a tuned backstop whose tuning is defensible on asymmetry: a false refusal costs an operator a dry run and a deliberate override, while a false acceptance costs the graph irreversibly and there is no backup to restore from. Choosing a threshold under that asymmetry is exactly what a backstop is. It just isn't a law, and it shouldn't be described as one — in a module whose sibling `src/exit_codes.py` makes a point of *recording* the false claims it once made about itself rather than quietly correcting them, a claim standing ten lines above its own refutation is the one thing that will mislead whoever next re-tunes this. Please soften it to what the tables actually support. I've written the correction into #426 so it is not lost; it does not need to precede the merge.

### The control is exactly what was claimed

`test_the_truncated_keep_set_passes_every_pre_existing_guard` does what it says. It asserts the keep-set is readable and non-empty (guard 1 passes), that `missing` is **exactly 0** rather than merely small — so the foreign-set guard doesn't just miss the fixture, it blesses it — that the consumer's whole-graph-wipe check passes, and that Weronika's `EXPECTED_POST == CANON_N` invariant passes on it too. That last assertion is the one that matters most: it demonstrates the truncated keep-set defeats **all three** designs the three of us independently proposed. Without it the new ceiling could have been guarding a state nothing can reach, and we would all have believed it. This is the right lesson from `feedback_fixture_makes_guard_assertion_inert`, and it is applied to a guard I asked for, which is the harder direction to apply it in.

### Contract, once more

`SUMMARY_KEYS` is unchanged at five — `ExcessiveDeletionError` adds a failure *mode*, not a key — so deploy#574's four extractions at its unchanged head `eef7ea7c` are untouched, and the mode-dependent `orphans=` semantics still hold. The new error shares rc `12` with its siblings, and the consumer only asks whether the rc is zero, so an unmerged, unmodified deploy#574 remains safe against this producer. Nothing desynchronises.

The rename to `UNUSABLE_CANONICAL_SET` is in. The member now says what it means, and it agrees with the exception supertype.

### What I am approving, and what must not follow it unguarded

The one thing neither ceiling catches is a **valid-but-short** canonical parquet, and you are right that no ceiling can — the bands overlap, so a fraction cannot separate them, and there is no completeness signal in this command's inputs. The cure is to bind the keep-set to the run that produced it (an expected row count, or the `src/pipeline/manifest.py` md5), so a short parquet is caught by being *short* rather than by being *drastic*. Filed as **#426** with the measurements above.

I am not blocking on it, and I want to be explicit about why, because it is a different case from my two earlier blocks. Both of those asked for a guard buildable from data already in hand, which is why they were fair to demand inline. This one is not: it requires a new contract with the producer, and demanding that inside a PR scoped to the prune subcommand would be scope creep dressed up as rigour.

But the exposure is real, so it needs a sequencing answer rather than a shrug. **This subcommand alone creates no exposure** — nothing invokes it automatically. The exposure is created by the *workflow*, which takes an operator-supplied `parquet_ref` and runs the destructive op against a prod graph whose backups have never run. So the honest line is: **da#414 is safe to merge; deploy#574 must not merge until #426 lands** (or until it carries an equivalent floor supplied by the workflow). I have said the same to the orchestrator so the sequencing is on the record and not just in this comment.

### Closing

Three passes, and each one closed the door I named and then went one step further than I asked — the collision control on #419, the `k/(k+1)` analysis, and now a docstring that documents the hole its own guards do not cover, which is the opposite of what most people do when they have just shipped a guard. The residual is the one an honest instrument can't reach without a new input, and it is written down rather than papered over. Good work, and thank you for the patience across three rounds.

---
RequestOrReplied: Approved
TechDebt: #418 (partial-delete exits LOAD_FAILED), #419 (consumer-side anchor for the summary key regex), #426 (completeness binding — gates deploy#574, not this PR)

PR #420, opened 2026-07-11
fix(deps): bump soupsieve 2.8.3 → 2.8.4 to clear CVE-2026-49476/49477 (#415)

### LINKED ISSUE #415: ci(publish): GHCR publish red on main — soupsieve 2.8.3 HIGH advisories block the load image (CVE-2026-49476/49477)
## Summary

`Publish to GHCR` is **red on `main`** ([run 29134247000](https://github.com/noorinalabs/noorinalabs-data-acquisition/actions/runs/29134247000), from the merge of #412). Trivy fails the build on two HIGH advisories in `soupsieve`:

| Package | CVE | Severity | Installed | Fixed in |
|---|---|---|---|---|
| `soupsieve` | CVE-2026-49476 | HIGH | 2.8.3 | **2.8.4** |
| `soupsieve` | CVE-2026-49477 | HIGH | 2.8.3 | **2.8.4** |

(Memory exhaustion via large comma-separated selectors; ReDoS.)

## Why this matters now

The `load` image is the artifact the #928 re-run-from-`parse` pipeline consumes. While this publish is red, **`stg-latest` is not being refreshed** — so the graph-load side of the pipeline is pinned to a stale image. This blocks the reload, not just the badge.

## Root cause

`soupsieve` is a **transitive** dependency, pulled in by `beautifulsoup4` (declared directly: `beautifulsoup4>=4.12` in `pyproject.toml`). `uv.lock` pins `soupsieve==2.8.3`. Nothing in our code constrains it — the lock simply predates the 2.8.4 release.

Not the base-image-drift class: this is a Python dependency with an available fix, and Trivy is right to fail on it.

## Fix

Bump the transitive pin:

```
uv lock --upgrade-package soupsieve
```

Expect `soupsieve` to move 2.8.3 → 2.8.4 in `uv.lock` with no other lock movement. No source change; no direct dependency added (`beautifulsoup4` already floors correctly and 2.8.4 is within its range).

## Acceptance

- [ ] `uv.lock` pins `soupsieve` at >= 2.8.4; diff touches only that package's block.
- [ ] Full local check-set green (the repo's pre-commit + pre-push mirror of CI).
- [ ] `Publish to GHCR` green on the PR **and** confirmed green on `main` post-merge — a green PR is not proof for a post-merge-only workflow (charter / `verify_deployable_merge`).
- [ ] Trivy reports 0 HIGH/CRITICAL on the published `load` image digest.

## Note for the implementer

Do **not** widen this into the base-image CVE work. The five/ten OS-package CVEs seen on the sibling ingest-platform image are a **different defect** — tracked as noorinalabs-main#947 (the buildx cache defeats the in-image `apt-get upgrade`). This issue is the Python-dependency half only, and it stands alone.

parametrization:
## Summary

`Publish to GHCR` has been red on `main` since the merge of #412. The Trivy gate (`severity: CRITICAL,HIGH`, `ignore-unfixed: true`, `exit-code: 1`) fails the `load` image on two HIGH advisories in **soupsieve 2.8.3**, both fixed in **2.8.4**:

| CVE | Severity | Installed | Fixed in |
|---|---|---|---|
| CVE-2026-49476 — memory exhaustion via large comma-separated selector lists | HIGH | 2.8.3 | 2.8.4 |
| CVE-2026-49477 — ReDoS via the selector parser | HIGH | 2.8.3 | 2.8.4 |

`soupsieve` is **transitive** via `beautifulsoup4` (declared `beautifulsoup4>=4.12`). Nothing in our code constrains it — the lock simply predated the 2.8.4 release. The fix is a lock bump, no source change, no new direct dependency:

```
uv lock --upgrade-package soupsieve
```

**Why this is on the critical path:** the `load` image is the artifact the #928 re-run-from-`parse` pipeline consumes. While the publish is red, `stg-latest` is not refreshed and the graph-load side of the reload stays pinned to a stale image.

## The diff

Only the `soupsieve` block moves — version, sdist, and the single wheel. Nothing else in the lock changed, and `uv sync --frozen` (the CI install step) resolves cleanly.

```diff
 [[package]]
 name = "soupsieve"
-version = "2.8.3"
+version = "2.8.4"
```

## Verification

Trivy 0.72.0, invoked with the publish workflow's own settings (`--severity HIGH,CRITICAL --ignore-unfixed`):

- **Control** — `trivy fs` over the *pre-fix* lock at `origin/main` reproduces **exactly the two soupsieve HIGHs**. The detector demonstrably separates the two classes, so the post-fix zero below is a measurement rather than a silent zero.
- **Treatment** — `trivy fs` over this branch's lock reports **0 findings and exits 0** (the gate uses `exit-code: 1`). It is clean at *all* severities, including unfixed.
- **OS layer** — a remote scan of the digest-pinned base (`python:3.14-slim@sha256:44dd0449…`, debian 13.5) reports **0 fixed HIGH/CRITICAL** *before* the Dockerfile's `apt-get -y upgrade`, which can only improve on it.
- **Full local check-set green** — commit stage (ruff-format, ruff-lint, gitleaks, actionlint, cspell, dockerfile-base-pin, fixture-realism) and push stage (mypy strict, pytest), plus the sync-drift gate and its 50 unit tests. No `--no-verify`.

## What this does NOT prove

`ghcr-publish.yml` triggers on **`push: [main]`, tags, and `workflow_dispatch` only — there is no `pull_request` trigger**, so the `Publish to GHCR` / Trivy check **cannot appear on this PR's status rollup**. A green PR here is therefore not proof the `main` publish goes green; that is only confirmed post-merge (charter / `verify_deployable_merge`). The evidence above is the strongest available pre-merge substitute.

One residual gap: the built image also carries `gcc` + `libpq-dev` (apt-installed on top of the base). Scanning those requires an actual image build, and this box has no Docker daemon, so that delta is unproven. The base and Python layers — where the reported CVEs live — are both proven clean.

## Not in scope

The base-image OS-package CVE class on the sibling ingest-platform image (buildx cache defeating the in-image `apt-get upgrade`) is a **separate defect**, tracked as noorinalabs-main#947. This is the Python-dependency half and it stands alone.

Closes #415

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1949,11 +1949,11 @@ wheels = [
 
 [[package]]
 name = "soupsieve"
-version = "2.8.3"
+version = "2.8.4"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/7b/ae/2d9c981590ed9999a0d91755b47fc74f74de286b0f5cee14c9269041e6c4/soupsieve-2.8.3.tar.gz", hash = "sha256:3267f1eeea4251fb42728b6dfb746edc9acaffc4a45b27e19450b676586e8349", size = 118627, upload-time = "2026-01-20T04:27:02.457Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/47/2c/0a5f6f8ee0d5589e48c7640213ed5175d52cf540a06725b628cc1a45d6ce/soupsieve-2.8.4.tar.gz", hash = "sha256:e121fd02e975c695e4e9e8774a5ee35d74714b59307868dcc5319ad2d9e3328e", size = 121110, upload-time = "2026-05-24T13:55:57.154Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/46/2c/1462b1d0a634697ae9e55b3cecdcb64788e8b7d63f54d923fcd0bb140aed/soupsieve-2.8.3-py3-none-any.whl", hash = "sha256:ed64f2ba4eebeab06cc4962affce381647455978ffc1e36bb79a545b91f45a95", size = 37016, upload-time = "2026-01-20T04:27:01.012Z" },
+    { url = "https://files.pythonhosted.org/packages/5e/f5/0c41cb68dcae6b7de4fac4188a3a9589e21fb31df21ea3a2e888db95e6c9/soupsieve-2.8.4-py3-none-any.whl", hash = "sha256:e7e6b0769c8f51ed59acab6e994b00621096cfb1c640a7509295987388fbaf65", size = 37304, upload-time = "2026-05-24T13:55:55.406Z" },
 ]
 
 [[package]]


parametrization:
## Review — soupsieve 2.8.3 → 2.8.4 (closes #415)

Approved. Clean, minimal, and correct. I verified the three things worth verifying, and I have a concrete answer on the image-scan question that I think settles it without a pre-merge dispatch.

### 1. The lock diff really is soupsieve-only

Confirmed against the diff at head `bf6857bb`. One file, 3 insertions / 3 deletions, every changed line inside the single soupsieve `[[package]]` block — the `version`, the `sdist` line, and the one wheel line. No sibling package moved, no `[package.metadata]` churn, no resolution-wide reshuffle. A `uv lock --upgrade-package` can absolutely drag siblings; this one did not.

### 2. 2.8.4 fixes both CVEs, and beautifulsoup4 admits it

Both advisories in the GitHub advisory DB name a vulnerable range of `<= 2.8.3` and a first-patched version of exactly `2.8.4` — GHSA-2wc2-fm75-p42x (memory exhaustion via large comma-separated selector lists) and GHSA-836r-79rf-4m37 (ReDoS in the selector parser), both rated high. So the bump lands precisely on the patched floor, not short of it.

`beautifulsoup4` 4.14.3 is the sole dependent — soupsieve appears exactly twice in the lock, once as bs4's dependency edge and once as its own package block. Nothing else in the tree pulls it, and bs4's unpinned constraint takes 2.8.4 without complaint.

### 3. The image-scan gap — I do not think we need the dispatch

Tarek's framing of the gap is honest and correct, and I want to reinforce the part he got right before I close it: `ghcr-publish.yml` triggers only on push-to-main, `v*` tags, and manual dispatch. There is no pull_request trigger, so the publish job and its Trivy gate are **structurally absent** from this PR's rollup. Green here is genuinely not proof that the main publish goes green. That reasoning is sound and I would not have accepted a hand-wave over it.

But the counterfactual he could not run has, in fact, already been run — by `main` itself.

I pulled the job log from the most recent failing publish on main (run `29134247000`, head `5ae1443a`). That run built the **real** image from `Dockerfile.load` — digest-pinned python:3.14-slim base, `apt-get -y upgrade`, and the `gcc` + `libpq-dev` apt layer on top — pushed it, and ran Trivy against the pushed digest with the workflow's own settings. What it found:

- Debian 13.5 OS layer, 130 packages — **0** findings. That layer *is* the gcc + libpq-dev delta Tarek could not scan for lack of a Docker daemon. It has already been scanned, on the real built image, by the real gate, and it contributes nothing.
- Every python-pkg row — **0** — with exactly one exception.
- `soupsieve-2.8.3.dist-info/METADATA` — **2**.

Those two are the **only** vulnerability identifiers anywhere in the 2,085-line job log, and Trivy's own table names the fixed version as 2.8.4 — the version this PR pins. So the publish gate on main is failing on exactly one cause, this PR removes exactly that cause, and every other surface of the image already scans clean under the identical settings.

A `workflow_dispatch` on the branch would rebuild that same image with one package changed to the version Trivy itself designates as the fix. I do not think it buys information we do not already have. And there is a second reason it buys less than it appears to: the merge build re-runs `apt-get -y upgrade` against a fresh Debian index with a fresh Trivy DB, so a newly-published base-image CVE could still take the publish red for reasons that have nothing to do with this change — and a dispatch *today* would not prevent that either, since it is equally a point-in-time snapshot. The dispatch cannot de-risk the only thing that stays risky. That residual is the familiar org-wide Trivy base-image drift, not a defect in this PR.

Recommendation — **merge without the pre-merge dispatch**. If the publish does go red on main for some unrelated new base-image CVE, that is a separate finding and a separate fix, and we will see it immediately since `uv.lock` is in the publish path filter (so this merge does fire the workflow).

### Tech debt filed

The trigger gap itself deserves to outlive this PR, so I filed it rather than leaving it in a review thread: da#421 (`tech-debt`, wave-24) — add a pull_request-scoped build-and-scan job to `ghcr-publish.yml`, build-only with no registry write, reusing the existing Trivy settings. #415 is the proof it matters: a fully green PR put a HIGH advisory into main and took the publish red for 8+ consecutive runs, and verifying *this* PR pre-merge required me to go read main's failing-run logs instead of just reading a check on the PR. The sibling user-service repo closed this same gap in us#87; data-acquisition never did.

Nice work being straight about the gap rather than papering over it — it is what made it cheap to close.

---
Requestee: Tarek Mansour
RequestOrReplied: Approved
TechDebt: da#421

parametrization:
## Review — verification / evidence axis (Oyunbileg Batbayar)

Approving, and recommending **merge now — no `workflow_dispatch` first**. My reasoning is below, because the recommendation rests on a piece of evidence stronger than the one this PR argues from, and I want it on the record rather than in my head.

### 1. The rollup is missing the check this PR exists to fix — confirmed

I verified the `on:` block myself at head `bf6857b` rather than taking it from the brief. It is `push.branches: [main]`, `push.tags: [v*]`, and `workflow_dispatch`. There is no `pull_request` trigger. So `Publish to GHCR` and its Trivy gate **cannot appear on this PR's rollup**, and the 6/6 green here — Lint, Type Check, Test, Integration Tests, base-image pin, sync-drift gate — does **not** include the one gate this PR is cleaning up after.

An absent check is not a passing check. Tarek says this plainly in the PR body under "What this does NOT prove", which is the right instinct and I want to credit it: he did not let the green badge stand in for the claim. I have filed the structural gap as tech debt (#422), mirroring the sibling fix in noorinalabs-user-service#87 (merged, closed us#61) — build with `push: false` / `load: true`, scan the locally-loaded image, move no registry pointer.

### 2. Grading the substitute evidence

**Methodology — correct, and the part I most want to endorse.** He did not read a bare zero off an uncalibrated instrument. He ran the detector on *both* classes: the pre-fix lock at `origin/main` reproduces exactly the two soupsieve HIGHs, his lock returns none. A detector that demonstrably separates the two classes is the precondition for the zero meaning anything at all, and skipping that step is how a silent zero gets mistaken for a clean bill of health.

**Was it the right control?** Yes. He invoked Trivy with `--severity HIGH,CRITICAL --ignore-unfixed`, which is exactly what the workflow's step passes, and the workflow gates on `exit-code: 1`, which is the condition he reports on. The settings match; the control is the right control.

**The `trivy fs` gap — he names one, and there are two more worth stating.** A filesystem scan of the lock cannot see:

- the **apt delta** — the image apt-installs `gcc` + `libpq-dev` on top of the base. He flags this himself. It is *not* a theoretical worry: `libpq-dev` pulls `libpq5`, and binutils arrives via `gcc`; both have real fixed-HIGH history. It also is not discarded by a builder stage — `runtime` is `FROM base`, and `base` is the stage that runs the apt install, so those packages **ship in the scanned runtime image**. Correctly identified as the live gap.
- the **base image's own preinstalled Python packages** (`pip`, `setuptools` in `python:3.14-slim`) — these are `python-pkg` targets inside the image but appear in no lock file, so `trivy fs` is blind to them. His remote scan of the digest-pinned base does cover them; the gap is closed but he did not name it.
- in the other direction, and in his favour: the venv is built `uv sync --frozen --no-dev`, so the image's Python set is a strict **subset** of the full lock he scanned. His treatment scan is therefore *over*-inclusive, not under-inclusive. A zero over the superset is a stronger result than the image's Python target would give, not a weaker one.

### 3. The gap he calls unproven is already proven — by his own red run

This is the finding I bring. The currently-failing publish on `main` (run 29134247000, `5ae1443a`) is **not** a lock scan. It is a full `trivy image` scan by digest of the real image, built from this same Dockerfile, same pinned base, same apt lines. Its report enumerates the complete finding set across every target:

```
INFO  Detected OS  family="debian" version="13.5"
INFO  [debian] Detecting vulnerabilities...  os_version="13" pkg_num=130

Report Summary
  ...-data-acquisition-load@sha256:67e2470...  debian      0 vulnerabilities
  app/.venv/.../soupsieve-2.8.3.dist-info/METADATA   python-pkg   2

Total: 2 (HIGH: 2, CRITICAL: 0)
  soupsieve  CVE-2026-49476  HIGH  fixed  2.8.3 -> 2.8.4
  soupsieve  CVE-2026-49477  HIGH  fixed  2.8.3 -> 2.8.4
```

130 debian packages scanned — that count **includes `gcc`, `libpq-dev`, and their transitive apt deps** — and the debian target reports zero at the workflow's own settings. The two soupsieve HIGHs are the *entire* finding set of the real image. So the apt delta, the exact residual the PR body concedes is unproven, is empirically clean, measured in the actual artifact, by the actual gate. The PR is more thoroughly evidenced than it claims to be.

The delta between that scanned image and the post-merge image is precisely one thing: soupsieve 2.8.3 to 2.8.4, which is the fix version Trivy itself names. Both findings clear, nothing else was ever flagged, so post-merge Trivy goes to zero and exits 0.

### 4. Merge, not dispatch

A `workflow_dispatch` on the branch would buy an image scan — but we already have an image scan of this image, with an unscanned-surface count of zero. It would re-derive a result we can already read, and it would cost a round-trip on the artifact that is currently blocking the reload. Merge.

One honest residual, which I do not consider blocking and which is not this PR's to carry: `apt-get -y upgrade` is unpinned, so a debian HIGH *published since* 01:14Z today could in principle surface on the post-merge build. In practice the `base` stage sits upstream of the `COPY pyproject.toml uv.lock`, so a lock-only change leaves it cache-hit under `cache-from: type=gha,scope=load` and the OS layer will be bit-identical to the one just measured at zero. That standing class is the org-wide base-image CVE risk, tracked separately at noorinalabs-main#947 — not a defect in a 3-line lock bump.

Nothing else to raise. The bump is correct, the lock delta is confined to the soupsieve block, and the evidence behind it is better than average for a dependency PR.

---
Requestor: Oyunbileg Batbayar
Requestee: Tarek Mansour
RequestOrReplied: Approved
TechDebt: noorinalabs-data-acquisition#422 — add `pull_request` trigger to `ghcr-publish.yml` so the Trivy gate runs pre-merge (mirrors us#87)

parametrization:
## Pre-merge proof obtained — the publish gate this PR fixes was dispatched against a real image

Tarek correctly flagged that `ghcr-publish.yml` has no `pull_request` trigger, so `Publish to GHCR` (and its Trivy gate) is **structurally absent** from this PR's 6/6 rollup — the check this PR exists to fix could not run on it. His local `trivy fs` substitute was methodologically sound (he calibrated the detector on the pre-fix lock first, reproducing exactly the two CVEs, rather than reading a bare zero), but it could not scan the built image's `gcc` + `libpq-dev` layer.

I dispatched the real workflow on this branch to close that gap. Safe to do here because da gates `stg-*` on `push && refs/heads/main` (unlike the sibling ingest-platform, whose dispatch *does* move `stg-latest`), so this emitted only an immutable `sha-<short>` tag and moved no pointer.

**Run [29138030830](https://github.com/noorinalabs/noorinalabs-data-acquisition/actions/runs/29138030830) — `success`. Every step green, including `Run Trivy vulnerability scan`.**

The fix is confirmed present in the built image, not just in the lock:

```
#14 4.936  + soupsieve==2.8.4
```

So the residual gap Tarek honestly declined to claim closed is now closed: the image builds, the Trivy gate passes on the real digest, and the two HIGH advisories are gone. Reviewers can treat the publish as **proven**, not merely likely.

## One thing this run surfaced that is NOT this PR's problem

The same log shows da's apt layer served from cache:

```
#11 [base 2/2] RUN apt-get update && apt-get -y upgrade && apt-get install -y --no-install-recommends gcc libpq-dev && rm -rf /var/lib/apt/lists/*
#11 CACHED
```

That is **noorinalabs-main#947** — the buildx GHA cache defeating the in-image `apt-get -y upgrade`, so the org's within-tag-drift defense never executes. This run is now direct measured confirmation that da carries it too, not just ingest-platform (which is red from it today). da has simply not been bitten yet, because no OS-package advisory has landed against its frozen layer.

**Explicitly out of scope for this PR** — do not widen it. Recorded on #947.

PR #429, opened 2026-07-11
feat(resolve): declare the canonical tally at the WRITE, not at publish (da#428)

### LINKED ISSUE #428: publish_parquet: emit curated/_manifest.txt declaring canonical_ids + md5 — the keep-set has no provenance and no ceiling can supply it
## Summary

`prune-narrators` can delete **83.9% of the production graph** against a valid parquet, with `missing=0` and every derived gate green (da#426). The fix is a **completeness binding**, and it needs a producer half that does not exist: **nothing published to B2 declares how many canonical ids the resolve run actually wrote.**

The consumer half is implemented in deploy#574.

> **⚠️ Correction (2026-07-11). This issue previously claimed deploy#574 is "fail-closed: no manifest, no prune." That was FALSE — I wrote it, and the workflow says the opposite.** At `a141f672`, `graph-prune-narrators.yml` pulls `curated/_manifest.txt` *if it exists* but explicitly does **not require it**: *"`expected_canonical_ids` is the required declaration, and the manifest, once it exists, replaces the human as the declarer."*
>
> So the prune is **runnable today**, on a human-supplied count. And per the finding below, **that human has no trustworthy source for the number** — the pipeline's only count is a read-back of the very parquet being pruned against. A required input whose only available value is circular is not a gate; it is a foot-gun wearing the costume of one.
>
> **deploy#574 must make the manifest REQUIRED** (no manifest → refuse), which is what this issue always assumed it did. That makes the human-declarer path — which cannot be sourced correctly — impossible rather than merely discouraged, and it makes #574 safe to merge ahead of this issue instead of dangerous.

## Why a manifest, and why nothing else works

Every gate in the prune path — da's magnitude ceiling, and all four of deploy#574's post-op gates — is derived from the keep-set it is meant to validate. They prove the prune obeyed the parquet. **None can ask whether it was the right parquet.**

`missing` (da#414) closes the *stale*-parquet door. It does **not** close the *short*-parquet door, and no ceiling can:

| `f` (canonical rows retained) | deletes | verdict |
|---|---|---|
| 0.124 | 90.0% | REFUSED (0.9 ceiling) |
| **0.20** | **83.9%** | **PROCEEDS** — 134,768 legitimate narrators gone |
| 0.40 | 67.8% | PROCEEDS |
| 0.60 | 51.7% | PROCEEDS |

`missing` is **0** in every row — on a truncated-but-same-generation parquet every id in the keep-set genuinely *is* in the graph, so the signal does not merely miss the case, it **certifies** it. And the ceiling cannot be tightened, because a **legitimate** full re-mint prune deletes **55.4%**. The legitimate and catastrophic bands overlap. Magnitude is the wrong axis.

Two alternatives were ruled out before filing, so nobody re-proposes them:

- **The parquet's own footer `num_rows`.** A short-but-valid parquet is *self-consistent* — its footer honestly reports the short count. The declaration must be **external to the file**.
- **B2's per-object hash / `rclone hashsum`.** Binds the file to itself. A truncated-but-intact upload has a perfectly good hash. That catches transfer corruption, not a short *write*.

## Current state, verified at the artifact

```
$ rclone lsf --recursive pipeline:noorinalabs-pipeline/staged/narrator-resolve/2026-07-09-dcb6205/
curated/narrator_mentions_resolved.parquet
curated/narrator_mentions_resolved_muhaddithat.parquet
curated/narrators_canonical.parquet
staging/... (23 more parquets)
```

Filtered for `manifest|.json|.md5|.sha|checksum|_SUCCESS|count` → **zero hits.**

`scripts/publish_parquet.py` uploads a fixed list of parquets and its `verify_remote()` checks only that the object *names* appear in the listing. `scripts/narrator_graph/run.py` prints counts to stdout and persists none of them. The `write_manifest()` helper that does exist (sha256 + sizes, `src/acquire/base.py`) lives in the **acquire** layer and never runs in the resolve/publish path.

## ⛔⛔ CORRECTION 3 (2026-07-11) — SPLIT THE ARTIFACT. `publish_parquet.py` CANNOT honestly declare the count.

**Weronika Zielinska, blocking deploy#574 — and this supersedes the single-`_manifest.txt` design below.** She read `scripts/publish_parquet.py` instead of reasoning about it:

> It is a **separate process that runs after resolve**, takes `data/curated/` **from disk**, and rclone-copies it. It has **no pyarrow import, no counting of any kind**, and — structurally — **no access to resolve's in-memory tally, because resolve exited before it started.**

**So when da#428 adds a manifest to that script, the only way it can produce `canonical_ids` is by reading the parquet it is about to upload. Which is a read-back. Which agrees perfectly with a short parquet.**

> **The circularity has not been eliminated. It has been RELOCATED from the operator to the publisher** — and this time it will be **invisible**, because it will be wearing the words "producer-signed provenance" and a green md5. **The count a truncated parquet needs in order to sail through every gate is exactly the count the most obvious implementation of this issue will compute for it.**

### THE REQUIRED SHAPE: two artifacts, each declaring only what its author can honestly attest

| artifact | written by | declares | why only this author can |
|---|---|---|---|
| **`curated/_resolve_run.txt`** | **resolve** — the process that *wrote* the parquet (its LAST writer; see Correction 2) | `canonical_ids` (in-memory tally, before the write), `run_status=complete`, per-stage `StageOutcome`, producing git sha, run id | The tally exists **only while that process is alive**. Nothing downstream can reconstruct it. |
| **`curated/_manifest.txt`** | **`publish_parquet.py`** | `md5`, `bytes`, file list | **Transfer integrity is the only thing a publisher is in a position to know.** It must NOT declare a count. |

**The consumer requires BOTH objects and cross-checks them.** Publish **copies the tally forward, never recomputes it.**

**This makes a read-back implementation of this issue STRUCTURALLY IMPOSSIBLE rather than merely discouraged** — which is the same argument used to justify making the prune workflow inert, applied one level up. **This is the last cheap moment to define the contract**, and it costs ~10 lines while the workflow cannot run anyway.

Also required, and currently unenforceable: the **run-completion assertion** needs a *parser slot*. deploy#574's existing `manifest_field` regex is `[0-9a-f]+` — **it cannot parse a word value at all**, so `run_status=complete` could not be read even if it were emitted. A second parser is needed. *A comment addressed to a future PR is not a gate.*

## The ask

`scripts/publish_parquet.py` writes **`curated/_manifest.txt`**, and uploads it **LAST** — so its presence is itself the evidence that the publish ran to completion.

```
CANONICAL_MANIFEST file=narrators_canonical.parquet rows=<n> canonical_ids=<n> md5=<32hex> bytes=<n>
```

One line per curated object. deploy#574 reads the `narrators_canonical.parquet` line and refuses unless **both** hold:

- `md5` equals the md5 of the bytes it pulled → the artifact is the one that was published (transfer integrity)
- `canonical_ids` equals the `canonical=<n>` that `prune-narrators --dry-run` reports → **the keep-set is the whole keep-set** (completeness)

### ⛔ The count MUST be an in-memory tally declared BEFORE the write — NOT a read-back at publish time

**This section supersedes the naive reading of the code snippet below, and it is the whole point of the issue.** It came out of deploy#574 review (Nino Kavtaradze), who traced the resolve pipeline's entire reporting surface and found that **the "independent" count does not exist today.**

Both counts resolve emits (`resolve_metrics_summary`, `resolve_pipeline_complete`) are fed by one field, set in `src/resolve/__init__.py::_collect_disambig_metrics`:

```python
table = pq.read_table(canonical_path)
metrics.canonical_narrator_count = table.num_rows      # a read-back of the file it just wrote
except Exception:                                       # ...and a failed read leaves it at its DEFAULT
    logger.warning("canonical_metrics_read_failed")     #    — a silent zero in the number the gate hangs on
```

There is no independent tally anywhere in the pipeline, and `publish_parquet.py` emits no count at all. **So an operator who follows deploy#574's instruction to the letter — "never take the count from the parquet you are about to prune against" — has nothing else to take it from. The instruction is un-followable as written.**

If the manifest's `canonical_ids` is produced by **reading the parquet at publish time** (the obvious implementation, and what `publish_parquet.py` is positioned to do), **the manifest inherits this hole and closes nothing the md5 doesn't.** Against the three failure modes:

| failure mode | read-back manifest | in-memory-tally manifest |
|---|---|---|
| truncated **upload** (local complete, B2 short) | CAUGHT | CAUGHT |
| partial **write** (local file itself short) | **NOT caught** — declared == actual == short | **CAUGHT** |
| resolve **stopped early** (short but self-consistent) | **NOT caught** | **NOT caught** — see below |

A read-back manifest defends **1 of 3**, and that one is already covered by the md5. It would be a gate that certifies coverage it cannot see.

> ### ⛔⛔ CORRECTION 2 (2026-07-11) — "THE producing stage" DOES NOT EXIST. `narrators_canonical.parquet` has **FOUR** writers.
>
> **This is the single most important thing in this issue and it invalidates the naive reading of the paragraph below.** Found by Jean-Claude Habimana (who named two writers); verified at `origin/main`, where there are **at least four**, all in `RESOLVE_STEP_ORDER`:
>
> | stage | what it does to `narrators_canonical.parquet` |
> |---|---|
> | `disambiguate` | **OVERWRITES** it |
> | `bio_promote` | **MERGEs** into it — and this is not marginal: **43,109 of 129,234** canonical ids are bio-only `mention_count = 0` reference entries (da#352 ruled these real narrators) |
> | `cluster` (`fuzzy_cluster.py`) | fuzzy-clusters it **in place**, *dropping* merged ids |
> | `narrator_split` | **rewrites** it with peeled rows *appended* |
>
> `src/resolve/__init__.py:135` says so explicitly: *"The order is load-bearing: `disambiguate` OVERWRITES `narrators_canonical.parquet` while `bio_promote` MERGEs into it, so nothing may reorder them (da#99/da#117)."*
>
> **Consequence if you implement the paragraph below literally.** A tally taken at `disambiguate` declares **~86k** while the published file honestly holds **~129k**. The completeness equality then **false-FAILs on every healthy run** — and this issue *already warns* exactly where that leads: *"a false-FAIL on a destructive op sends an operator to 'restore from backup' after a prune that did exactly the right thing — and the next person to see it will be tempted to weaken the gate."* The defect arrives through a door the original spec never checked.
>
> **Compounding it — the resume machinery reaches the vacuous manifest legitimately.** With `--from-step` (da#268/da#272), the tally-minting stage can be `StageSkipped`, in which case **no in-memory tally exists at all** and the only available number is a read-back. The circular manifest, reached by a legitimate operation.
>
> **THE REQUIREMENT:**
> 1. The tally is taken by the **LAST writer of `narrators_canonical.parquet` that actually ran**, and **recomputed by every writer** — not by a stage chosen by name. Two (four) producers of one composed artifact must derive the field from the same authoritative source.
> 2. **`publish_parquet.py` MUST REFUSE to publish if the tally-minting stage was skipped** — never fall back to a read-back. A skipped stage is not a smaller number; it is *no number*.
> 3. A test must pin the four-writer composition: a fixture where `bio_promote` adds rows after `disambiguate` must show the manifest declaring the **post-`bio_promote`** count. **A test that only exercises `disambiguate` would pass while being wrong** — [[feedback_fixture_makes_guard_assertion_inert]].

**Required:** the count is threaded from the **producing stage's in-memory tally** — `len({cid for cid in canonical_map if cid})` at the moment the rows are built, *before* `pq.write_table` — and carried to `publish_parquet.py` as data. The producer declares what it *intended to write*; the consumer checks what *arrived*. If those two numbers come from the same read, there is no binding.

**And the third mode is not a count problem at all.** If resolve stops early, the tally and the file are both short and perfectly consistent — **no count equality of any kind catches it.** That needs a separate **run-completion assertion** at publish time: `publish_parquet.py` must refuse to publish unless the resolve run reached its terminal stage (e.g. a `resolve_pipeline_complete` sentinel written by the last stage, checked at publish). Ship it in this issue; a manifest without it leaves the door it looks like it closed still open.

### `canonical_ids` must be DISTINCT NON-EMPTY, not a naive row count — this is the part that will bite

`read_canonical_ids()` (`src/graph/prune.py`) returns a **`set`** and filters falsy ids:

```python
ids = {cid for cid in table.column("canonical_id").to_pylist() if cid}
```

So `canonical=` in `PRUNE_NARRATORS_SUMMARY` is a **distinct, non-empty** count. If the manifest declares a raw `len(df)` instead, the consumer's equality gate **false-FAILS** on any duplicate or null `canonical_id` in the parquet.

That is not a cosmetic mismatch. A false-FAIL on a *destructive* op sends an operator to "restore from backup" after a prune that did exactly the right thing — and the next person to see it will be tempted to weaken the gate. **The producer must compute `canonical_ids` with the same definition the consumer observes:**

```python
# At the PRODUCING stage, over the in-memory rows — NOT over a re-read of the written file.
canonical_ids = len({cid for cid in canonical_ids_in_memory if cid})
```

(The definition must match `read_canonical_ids()` exactly; the *source* must not be the parquet. Both halves are load-bearing — same definition, different derivation.)

Emit `rows` too (it is cheap and useful for the other curated files), but the gate keys on `canonical_ids`.

### Format: key=value, not JSON

Deliberate. The consumer parses this **on the VPS**, inside an `ssh-action` script. `jq` is used only on the GH runner in noorinalabs-deploy, never VPS-side, and no VPS-side script block uses `python3` — a JSON manifest would add a parsing dependency to the production box. The `key=value` instrument line is the same shape as `PRUNE_NARRATORS_SUMMARY`, which both repos already parse with an anchored whole-token extraction.

## Acceptance

- `publish_parquet.py` emits `curated/_manifest.txt` with a `CANONICAL_MANIFEST` line per curated object, uploaded **after** all parquets.
- `canonical_ids` is computed as **distinct non-empty** `canonical_id` values — with a test pinning it against `read_canonical_ids()` on the same fixture, so producer and consumer provably count the same thing. (A test that merely asserts `canonical_ids == len(df)` would pass while being wrong.)
- A **red-first** test: a parquet with a duplicate and a null `canonical_id` must produce `canonical_ids < rows`, and the two must not be interchangeable.
- **`canonical_ids` is threaded from the producing stage's in-memory tally, not re-read from the parquet.** A test must pin this: mutate the written parquet on disk (truncate it) *after* the tally is taken and assert the manifest still declares the ORIGINAL count — i.e. that producer and file can disagree. **If that test cannot be written, the implementation is a read-back and the issue is not done.**
- **The manifest carries RUN PROVENANCE, not just counts** — producing git sha, resolve run id, and the per-stage `StageOutcome` summary. Rationale (Jean-Claude Habimana): the manifest is a plain text file in a mutable bucket, and **nothing in a counts-only manifest distinguishes an honest producer-tallied one from one an operator hand-authored in a single line from the parquet itself.** #574 cannot tell them apart *by construction* — they are the same five fields. The workflow's "there is no break-glass" is therefore a statement of *intent*, not a property of the artifact, and the pressure to hand-mint is **maximal at exactly the moment the ordering hazard below bites.** Provenance makes hand-minting an act of forgery rather than an act of convenience (~10 lines in the producer). Do **not** sign/HMAC it — pre-launch, no adversary, not worth the key management.
- **`publish_parquet.py` refuses to publish unless the resolve run reached its terminal stage** (run-completion assertion) — the "resolve stopped early" mode that no count equality can see.
- `verify_remote()` treats the manifest as required, as it does the curated parquets.
- Republishing an existing ref is out of scope; deploy#574 correctly refuses refs that predate this.

Consumer: noorinalabs-deploy#574. Residual this closes: da#426 (and its duplicate da#425).

parametrization:
## What

The **producer half** of the prune completeness binding. deploy#574 is merged and **inert until this lands** — it requires provenance objects in B2 and refuses every ref published today.

`prune-narrators` can delete **83.9% of the production graph** against a *valid but truncated* parquet, with `missing == 0` and every derived gate green (da#426). There are no backups yet. Every gate in the prune path is derived from the keep-set it validates, so all of them prove the prune obeyed the parquet and **none can ask whether it was the WHOLE parquet**. Only the process that *wrote* the file can supply that.

Two artifacts, each declaring only what its author can honestly attest:

| artifact | written by | declares |
|---|---|---|
| `curated/_resolve_run.txt` | **resolve** (`src/resolve/_run_record.py`) | `canonical_ids` (in-memory tally, **before** the write), `run_status`, git sha, per-stage outcomes |
| `curated/_manifest.txt` | `publish_parquet.py` | **md5 + bytes ONLY** — forbidden from declaring a count |

Publish **copies the tally forward and never recomputes it.**

## Three things in the issue were wrong. Verified at the artifact.

**1. There are SIX writers of `narrators_canonical.parquet`, not four.** The issue names `disambiguate`, `bio_promote`, `cluster`, `narrator_split`. It misses **`reconcile`** (`date_reconcile.py:527`) and **`tabaqa_dates`** (`tabaqa_dates.py:227`) — both of which run *after* `narrator_split` in `RESOLVE_STEP_ORDER` and rewrite the file (row set preserved, **bytes changed**). So the issue's own prescription — "the tally is taken by the LAST writer" — names the wrong stage, and a tally pinned there would ship a **stale md5**.

The cure is to stop naming stages. `write_canonical()` is the single sanctioned writer; it re-mints the tally on every call, so whichever writer runs last is automatically the one that stands. A seventh writer inherits the guarantee, and an AST gate fails CI if one ever bypasses it (a guard is pinned by the branch that calls it).

**2. A null `canonical_id` is unrepresentable.** The issue asks for a red-first test with "a duplicate and a null". `canonical_id` is **non-nullable** in `NARRATORS_CANONICAL_SCHEMA` — the cast in `write_parquet` rejects it. The representable hazards are a **duplicate and an empty string**, and those are what the write-path test uses. The count is DISTINCT NON-EMPTY either way, matching `read_canonical_ids()` exactly.

**3. "Refuse if the tally-minting stage was skipped" is a proxy for the real property**, and a leaky one — it does not catch `scripts/scrub_matn_canonical.py` / `scrub_relational_pollution.py`, which rewrite the canonical parquet *outside resolve entirely*. What publish actually enforces is the property itself: **is this file the one the tally was taken over, from a run that finished?** The record carries the md5 of the bytes the tally was minted against; publish refuses if it does not match. That is strictly stronger and catches the scrub-script hole too. The md5 binds tally to bytes — it cannot *produce* a count, so it introduces no circularity.

## Fail-closed by construction

Every `write_canonical` stamps `run_status=incomplete` — at the moment of a write the run has not finished. Only the successful end of `run_all` upgrades it. A killed process, a raised stage, a `--stop-after` probe all leave `incomplete` on disk and publish refuses; **nothing has to remember to mark failure**. And `finalize_run` refuses to upgrade a record it did not mint, so a `--from-step dedup` resume cannot bless an earlier *partial* run's tally as `complete` — a false PASS reachable by an entirely legitimate operation.

## The falsifier, and why the obvious version of it is worthless

The acceptance criterion is: truncate the parquet after the tally is taken and the record must still declare the original count. **I wrote that test first and a read-back mutant sailed straight through it.** A read-back taken at write time has already read the *complete* file by then, so it reports 100 too — the test passes under both implementations while appearing to prove the point.

The shape that actually separates them is the **partial write**: `write_parquet` is made lossy, the writer intends 100 rows, the disk receives 20. An in-memory tally declares 100 and disagrees with the file — that disagreement *is* the completeness signal. A read-back declares 20 and agrees with it perfectly.

Every guard here was mutation-tested: each mutant was verified to actually reach the forbidden state, and each was killed **by its own guard, naming its own reason** (a non-zero exit is not evidence the check under test fired).

## Verified against the real merged consumer

Ran deploy's actual `scripts/verify_prune_provenance.sh` against artifacts this producer generates:

- **accepts** the healthy artifact -> `PRUNE_PROVENANCE canonical_ids=129234 run_status=complete md5=e00466d1...`
- the completeness equality **holds on a healthy run** (129,234 declared == 129,234 read) — no false-FAIL, which was the whole hazard of correction 2
- **refuses** each defect naming the right reason: no record; `run_status=incomplete`; a manifest declaring a count; changed bytes; wrong operator expectation

(My first run of this was **vacuous** — rc=127, the gate script was not in the local deploy checkout, so every "refusal" was just bash failing to find the file. The known-positive control is what caught it.)

## Also structural

`publish_parquet.py` **provably never imports pyarrow** — asserted by executing it in a subprocess and checking `sys.modules`, with the detector verified on a known-positive first. It therefore *cannot* read the parquet back and count it. The read-back is impossible, not discouraged.

## Testing

- `mypy src/` clean (CI's exact command). The 8 pre-existing `scripts/` mypy errors are identical on `origin/main` and outside CI's scope.
- ruff format + lint clean.
- CI unit job green: 2,230 passed, 10 skipped.
- Integration job needs Docker (down on this box; runs in CI). No test asserts exact curated-dir contents, so the new sidecar cannot break one.

Closes #428.

Technical debt introduced by this change: none.

### DIFF
--- a/scripts/publish_parquet.py
+++ b/scripts/publish_parquet.py
@@ -11,6 +11,8 @@ box one-offs, da#342).
 
 CONTRACT — objects published under ``noorinalabs-pipeline/<parquet_ref>/``:
 
+    curated/_resolve_run.txt              (written by RESOLVE; copied forward, da#428)
+    curated/_manifest.txt                 (written HERE, uploaded LAST; md5 + bytes ONLY)
     curated/narrators_canonical.parquet
     curated/narrator_mentions_resolved.parquet
     curated/narrator_mentions_resolved_muhaddithat.parquet
@@ -45,6 +47,7 @@ stderr.
 from __future__ import annotations
 
 import argparse
+import hashlib
 import os
 import shutil
 import subprocess
@@ -81,6 +84,51 @@ STAGING_FULL_ONLY_GLOBS: tuple[str, ...] = (
 )
 
 
+CANONICAL_PARQUET_NAME = "narrators_canonical.parquet"
+
+# --- The da#428 provenance contract -----------------------------------------
+# Two artifacts, each declaring ONLY what its author can honestly attest.
+#
+#   curated/_resolve_run.txt   Written by RESOLVE (src/resolve/_run_record.py) — the
+#                              process that WROTE the parquet. It alone holds an
+#                              in-memory tally taken BEFORE the write, so its count
+#                              does NOT agree with a short file, and it alone knows
+#                              whether the run reached its terminal stage.
+#
+#                                RESOLVE_RUN canonical_ids=<n> run_status=complete git_sha=<sha>
+#
+#   curated/_manifest.txt      Written HERE. md5 + bytes ONLY — and it is FORBIDDEN
+#                              from declaring a count.
+#
+#                                CANONICAL_MANIFEST file=<name> md5=<32hex> bytes=<n>
+#
+# WHY THIS SCRIPT MUST NOT DECLARE A COUNT, AND WHY IT CANNOT IMPORT THE PRODUCER
+#
+# This is a SEPARATE PROCESS. It starts after resolve has exited, reads data/curated/
+# off disk and uploads it. Resolve's in-memory tally does not exist any more. So the
+# only way this script could produce `canonical_ids` is by READING BACK the parquet it
+# is about to upload — and a read-back agrees perfectly with a truncated file. It would
+# be a gate certifying coverage it cannot see, wearing the words "producer-signed
+# provenance" and a green md5. da#426 measured that shape deleting 83.9% of the graph.
+#
+# That is why this module imports NO pyarrow and does NOT import src.resolve (which
+# would pull pyarrow in transitively and hand it the ability to count). The inability
+# is STRUCTURAL, not a convention — `tests/test_scripts/test_publish_provenance.py`
+# executes this module in a subprocess and asserts pyarrow never enters sys.modules.
+# The consumer enforces the other side: noorinalabs-deploy's verify_prune_provenance.sh
+# REFUSES outright a manifest that carries `canonical_ids=`.
+#
+# So publish COPIES THE TALLY FORWARD, byte for byte, and never recomputes it.
+RESOLVE_RUN_FILENAME = "_resolve_run.txt"
+MANIFEST_FILENAME = "_manifest.txt"
+
+# Both provenance objects are REQUIRED — the consumer refuses the prune without either.
+PROVENANCE_REQUIRED: tuple[str, ...] = (RESOLVE_RUN_FILENAME, MANIFEST_FILENAME)
+
+_RUN_KEYWORD = "RESOLVE_RUN"
+_MANIFEST_KEYWORD = "CANONICAL_MANIFEST"
+
+
 class PublishError(RuntimeError):
     """A publish precondition or post-publish verification failed."""
 
@@ -91,6 +139,122 @@ class PlannedUpload:
 
     local: Path
     remote_subpath: str  # e.g. "curated/narrators_canonical.parquet"
+
+
+@dataclass(frozen=True)
+class ResolveRunDeclaration:
+    """What resolve declared about the canonical parquet — parsed, never recomputed.
+
+    ``line`` is kept verbatim so the record is copied forward exactly as the producer
+    wrote it. This script re-derives none of these fields.
+    """
+
+    canonical_ids: int
+    run_status: str
+    parquet_md5: str
+    line: str
+
+
+def parse_resolve_run(text: str) -> ResolveRunDeclaration | None:
+    """Parse the ``RESOLVE_RUN`` line of ``curated/_resolve_run.txt``.
+
+    ``None`` for absent/malformed — which :func:`assert_run_publishable` turns into a
+    hard refusal. A missing instrument reading is a stop, never a silent zero.
+
+    Whole-token extraction (``split`` on spaces, exact ``key=``), matching the
+    consumer's ``token_value``: no key can be hijacked by a longer one ending in its
+    name (``parquet_md5=`` can never answer for ``md5=``).
+    """
+    head = next((ln for ln in text.splitlines() if ln.startswith(f"{_RUN_KEYWORD} ")), None)
+    if head is None:
+        return None
+    tokens: dict[str, str] = {}
+    for tok in head.split(" ")[1:]:
+        key, sep, value = tok.partition("=")
+        if sep and key:
+            tokens[key] = value
+    try:
+        return ResolveRunDeclaration(
+            canonical_ids=int(tokens["canonical_ids"]),
+            run_status=tokens["run_status"],
+            parquet_md5=tokens.get("parquet_md5", ""),
+            line=head,
+        )
+    except (KeyError, ValueError):
+        return None
+
+
+def assert_run_publishable(decl: ResolveRunDeclaration | None, canonical_md5: str) -> None:
+    """Refuse to publish a keep-set that carries no honest, matching tally.
+
+    Each door names a distinct failure, and none of them can be answered by reading
+    the parquet — which is the point.
+    """
+    if decl is None:
+        raise PublishError(
+            f"curated/{RESOLVE_RUN_FILENAME} is absent or malformed. The keep-set carries no "
+            "declaration from the process that WROTE it, and this publisher holds no tally of "
+            "its own — it is a separate process and any count it computed would be a read-back "
+            "of the parquet it is uploading, which agrees with a truncated file. Re-run resolve "
+            "so it declares the tally (da#428); do NOT hand-author this file."
+        )
+    if decl.run_status != "complete":
+        raise PublishError(
+            f"the resolve run that produced this keep-set did NOT complete "
+            f"(run_status={decl.run_status!r}). Its canonical set is partial BY THE PRODUCER'S "
+            "OWN ACCOUNT — every id in it is real, so 'missing' would be 0 and every derived "
+            "gate would go green while the prune deleted the narrators the run never got to "
+            "name. No count equality of any kind can see this; only the producer can attest it."
+        )
+    if decl.canonical_ids <= 0:
+        raise PublishError(
+            f"resolve declares canonical_ids={decl.canonical_ids} — a prune against a zero-id "
+            "keep-set would delete every narrator."
+        )
+    if decl.parquet_md5 != canonical_md5:
+        raise PublishError(
+            f"curated/{CANONICAL_PARQUET_NAME} is NOT the file resolve took its tally over.\n"
+            f"  tally was taken over md5: {decl.parquet_md5}\n"
+            f"  file on disk has     md5: {canonical_md5}\n"
+            "The declared count therefore describes different bytes than the ones about to be "
+            "uploaded, so it cannot bind them. This is what a resume that skipped every "
+            "canonical-writing stage looks like, and what a scrub script that rewrote the "
+            "parquet outside the sanctioned writer looks like. A skipped tally is not a smaller "
+            "number — it is NO number. Re-run resolve through a canonical writer; the publisher "
+            "must never fall back to counting the file itself."
+        )
+
+
+def file_md5(path: Path) -> str:
+    """md5 of ``path``'s bytes, streamed. Transfer integrity — NOT a count."""
+    digest = hashlib.md5(usedforsecurity=False)
+    with path.open("rb") as handle:
+        while chunk := handle.read(1024 * 1024):
+            digest.update(chunk)
+    return digest.hexdigest()
+
+
+def manifest_line(name: str, md5: str, size_bytes: int) -> str:
+    """One ``CANONICAL_MANIFEST`` line. md5 + bytes ONLY — never a count."""
+    return f"{_MANIFEST_KEYWORD} file={name} md5={md5} bytes={size_bytes}"
+
+
+def render_manifest(uploads: list[PlannedUpload]) -> str:
+    """The ``_manifest.txt`` body: one line per curated object.
+
+    Transfer integrity is the only thing a publisher is in a position to know, so it
+    is the only thing this declares. It carries no ``canonical_ids`` and no ``rows``:
+    the consumer REFUSES a manifest that declares a count, because a publisher that
+    thinks it knows the count has read the file back.
+    """
+    return (
+        "\n".join(
+            manifest_line(u.local.name, file_md5(u.local), u.local.stat().st_size)
+            for u in uploads
+            if u.remote_subpath.startswith("curated/")
+        )
+        + "\n"
+    )
 
 
 # --- Pure logic (unit-tested; no I/O) ---------------------------------------
@@ -179,6 +343,12 @@ def verify_remote(listing: set[str], planned: list[PlannedUpload]) -> None:
     ``listing`` is the set of bucket-relative keys under ``<parquet_ref>/`` as
     returned by ``rclone lsf --recursive`` (e.g. ``curated/narrators_canonical.parquet``).
     Raises :class:`PublishError` loudly on any absence.
+
+    The two provenance objects are required exactly as the curated parquets are
+    (da#428). A ref whose parquets landed but whose ``_resolve_run.txt`` did not is a
+    ref the consumer will refuse to prune against — and it must fail HERE, loudly, at
+    publish time, rather than surfacing as a mysterious refusal on the production box
+    at the moment someone is trying to run a destructive op.
     """
     uploaded_missing = [pu.remote_subpath for pu in planned if pu.remote_subpath not in listing]
     if uploaded_missing:
@@ -186,9 +356,8 @@ def verify_remote(listing: set[str], planned: list[PlannedUpload]) -> None:
             "post-publish verification failed — object(s) absent from B2 listing: "
             f"{', '.join(sorted(uploaded_missing))}"
         )
-    curated_missing = [
-        f"curated/{name}" for name in CURATED_REQUIRED if f"curated/{name}" not in listing
-    ]
+    required = [f"curated/{name}" for name in (*CURATED_REQUIRED, *PROVENANCE_REQUIRED)]
+    curated_missing = [key for key in required if key not in listing]
     if curated_missing:
         raise PublishError(
             "post-publish verification failed — required curated object(s) missing: "
@@ -338,11 +507,38 @@ def _run(argv: list[str] | None) -> int:
     log(f"[publish-parquet] mode        = {mode}")
     log(f"[publish-parquet] curated dir = {curated_dir}")
     log(f"[publish-parquet] staging dir = {staging_dir}")
-    log(f"[publish-parquet] {len(plan)} object(s) to publish under {BUCKET}/{ref}/")
+
+    # --- The completeness gate (da#428) --------------------------------------
+    # BEFORE anything is uploaded. Resolve's record is the only number in the whole
+    # prune path not derived from the artifact being validated; if it is absent, if
+    # the run did not finish, or if it describes different bytes than the ones about
+    # to be shipped, we refuse — we never fall back to counting the parquet.
+    canonical_local = curated_dir / CANONICAL_PARQUET_NAME
+    canonical_md5 = file_md5(canonical_local)
+    run_record = curated_dir / RESOLVE_RUN_FILENAME
+    decl = (
+        parse_resolve_run(run_record.read_text(encoding="utf-8")) if run_record.is_file() else None
+    )
+    assert_run_publishable(decl, canonical_md5)
+    assert decl is not None  # narrowed by assert_run_publishable
+    log(
+        f"[publish-parquet] resolve declared canonical_ids={decl.canonical_ids} "
+        f"run_status={decl.run_status} (tally minted in-memory by the writing process; "
+        "this publisher computes no count)"
+    )
+
+    # Written from the same bytes the gate just approved. md5 + bytes ONLY.
+    manifest_local = curated_dir / MANIFEST_FILENAME
+    provenance = [
+        PlannedUpload(run_record, f"curated/{RESOLVE_RUN_FILENAME}"),
+        PlannedUpload(manifest_local, f"curated/{MANIFEST_FILENAME}"),
+    ]
+    total = len(plan) + len(provenance)
+    log(f"[publish-parquet] {total} object(s) to publish under {BUCKET}/{ref}/")
 
     if args.dry_run:
         log("[publish-parquet] --dry-run: the following WOULD be published (nothing uploaded):")
-        for upload in plan:
+        for upload in [*plan, *provenance]:
             log(f"  {upload.local} -> {remote_base(ref)}/{upload.remote_subpath}")
         print(ref)
         return 0
@@ -350,14 +546,23 @@ def _run(argv: list[str] | None) -> int:
     if shutil.which(args.rclone_bin) is None:
         raise PublishError(f"rclone binary {args.rclone_bin!r} not found on PATH")
 
+    manifest_local.write_text(render_manifest(plan), encoding="utf-8")
+
     env = build_rclone_env()
     for upload in plan:
         rclone_copy_file(upload, ref, env, rclone_bin=args.rclone_bin)
 
+    # Provenance LAST, manifest last of all: its presence is itself the evidence that
+    # the publish ran to completion. _resolve_run.txt is copied forward verbatim —
+    # the publisher never re-derives what only the producer could honestly attest.
+    for upload in provenance:
+        rclone_copy_file(upload, ref, env, rclone_bin=args.rclone_bin)
+
+    full_plan = [*plan, *provenance]
     log("[publish-parquet] verifying published objects (rclone lsf)...")
     listing = rclone_list(ref, env, rclone_bin=args.rclone_bin)
-    verify_remote(listing, plan)
-    log(f"[publish-parquet] OK - {len(plan)} object(s) verified present under {BUCKET}/{ref}/")
+    verify_remote(listing, full_plan)
+    log(f"[publish-parquet] OK - {len(full_plan)} object(s) verified present under {BUCKET}/{ref}/")
 
     # Final stdout line: the ref for the operator to paste into deploy-data-load.yml.
     print(ref)
--- a/src/resolve/__init__.py
+++ b/src/resolve/__init__.py
@@ -31,6 +31,7 @@ from src.resolve._checkpoint import EXIT_STOPPED_AT_LIMIT, StopAfterReached
 from src.resolve._deps import EXIT_MISSING_DEPENDENCY, MissingDependencyError
 from src.resolve._inputs import MissingInputError
 from src.resolve._provenance import DetectorProvenance, DetectorStatus, read_provenance
+from src.resolve._run_record import finalize_run
 from src.resolve.ner import EXIT_UNROUTED_CORPUS, UnroutedCorpusError
 from src.utils.logging import get_logger
 
@@ -127,6 +128,33 @@ class ResolveStageError(Exception):
 def _outcome_files(outcome: StageOutcome | None) -> list[Path]:
     """Files a stage produced, or ``[]`` for a skipped/errored/absent stage."""
     return outcome.files if isinstance(outcome, StageRan) else []
+
+
+def _stage_provenance(outcomes: dict[str, StageOutcome]) -> tuple[str, ...]:
+    """Render each stage's outcome as a ``RESOLVE_STAGE`` body for the run record.
+
+    Forensics, not a gate — the consumer keys only on ``canonical_ids`` and
+    ``run_status``. But a counts-only record cannot be told apart from one an
+    operator hand-authored in a single line straight off the parquet, and the
+    pressure to hand-mint is maximal exactly when the gate bites. Recording which
+    stages ran, skipped and why makes hand-minting an act of forgery rather than
+    one of convenience (da#428, Jean-Claude Habimana). Deliberately unsigned:
+    pre-launch, no adversary, not worth the key management.
+    """
+    rendered: list[str] = []
+    for step in RESOLVE_STEP_ORDER:
+        outcome = outcomes.get(step)
+        match outcome:
+            case StageRan():
+                detail = f"outcome=ran files={len(outcome.files)}"
+            case StageSkipped():
+                detail = f"outcome=skipped reason={outcome.reason}"
+            case StageErrored():
+                detail = f"outcome=errored exc={outcome.exc_type}"
+            case _:
+                detail = "outcome=absent"
+        rendered.append(f"step={step} {detail}")
+    return tuple(rendered)
 
 
 # Canonical execution order of the resolve steps. ``run_all(from_step=...)`` skips
@@ -846,6 +874,16 @@ def run_all(
     # StageErrored now forces a non-zero exit via the CLI's ResolveStageError
     # handler, aggregating every failed stage from this one (possibly 7.5-hour) run.
     errored = [o for o in outcomes.values() if isinstance(o, StageErrored)]
+
+    # da#428: stamp the run's terminal status onto curated/_resolve_run.txt. This is
+    # the ONLY place run_status may be upgraded to `complete`; every write_canonical
+    # stamps `incomplete`, so a killed process, a raised stage or a --stop-after probe
+    # all leave `incomplete` on disk and the publish refuses. `complete` is exactly
+    # "reached this line with no stage in error" — the one fact no count can supply,
+    # because a run that halts after writing a coherent partial file is internally
+    # consistent and its tally matches what it managed to write.
+    finalize_run(output_dir, _stage_provenance(outcomes), complete=not errored)
+
     if errored:
         for e in errored:
             logger.error("resolve_stage_errored", step=e.step, exc_type=e.exc_type)
--- a/src/resolve/_run_record.py
+++ b/src/resolve/_run_record.py
@@ -1,0 +1,382 @@
+"""The resolve run's own record of what it wrote — ``curated/_resolve_run.txt`` (da#428).
+
+``prune-narrators`` DETACH-DELETEs every ``Narrator`` absent from the canonical
+keep-set. da#426 measured a *valid but truncated* ``narrators_canonical.parquet``
+deleting **83.9% of the graph** with ``missing == 0`` and every derived gate green:
+every gate in the prune path is derived from the keep-set it is meant to validate,
+so all of them prove the prune obeyed the parquet and none can ask whether it was
+the *right* parquet — nor whether it was the *whole* parquet.
+
+The missing signal is **completeness**, and it cannot be recovered downstream:
+
+* The parquet's own footer ``num_rows`` is self-consistent on a short file.
+* B2's per-object md5 binds the file to itself; a truncated-but-intact upload has
+  a perfectly good hash.
+* ``publish_parquet.py`` is a **separate process** that starts after resolve has
+  exited and reads ``data/curated/`` off disk. It holds no tally, so any count it
+  computed could only be a READ-BACK of the parquet it is about to upload — which
+  agrees with a short file by construction.
+
+**The tally exists only while the writing process is alive.** So the process that
+*wrote* the parquet must declare, before the write, how many canonical ids it
+intended to write; that declaration is this module's record, and it is the only
+number in the whole prune path not derived from the artifact being validated.
+
+The consumer contract (noorinalabs-deploy ``scripts/verify_prune_provenance.sh``)::
+
+    RESOLVE_RUN canonical_ids=<n> run_status=complete git_sha=<sha>
+
+Two artifacts, each declaring only what its author can honestly attest: this one
+carries the count and the run's terminal status; ``publish_parquet.py``'s
+``_manifest.txt`` carries md5 + bytes and is FORBIDDEN from carrying a count. The
+gate refuses a manifest that declares ``canonical_ids`` at all — a publisher that
+thinks it knows the count has read the file back.
+
+Why the tally is bound to the WRITE and not to a named stage
+------------------------------------------------------------
+``narrators_canonical.parquet`` does not have "a producing stage". It has **six**
+writers, every one of them in :data:`~src.resolve.RESOLVE_STEP_ORDER`:
+
+===============  ==========================================================
+stage            effect on ``narrators_canonical.parquet``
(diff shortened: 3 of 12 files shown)


parametrization:
## QA review — da#429 (head `3e3a1f3`)

I did not reason about these tests; I executed them. I built a detached worktree at the PR head, wrote **eight mutants**, and for each one **proved the mutant was live in the source pytest imports** (by grep, and where possible by observing the forbidden state actually occur) before reading any pass/fail number. A green suite under a mutant means *either the guard held OR the mutation never landed* — and twice in this review the mutation had silently not landed, so this was not a formality.

**Six of eight mutants were killed, each by its own guard, naming its own reason. Two survived.** The implementation at HEAD is correct — every door works, and I could not make it publish or bless a bad artifact. What is missing is that **neither of this PR's two integration seams is pinned by any test**, and both are one-line-deletable with all 2,229 tests still green.

### The falsifier works, and its history is the strongest thing in this PR

I wrote the read-back implementation myself — tally taken *after* `write_parquet`, by counting the file — and ran your suite against it.

- `test_tally_is_the_in_memory_count_not_a_read_back_of_a_short_write` goes **RED**: `assert 20 == 100`, failing on the property assertion, not on an incidental one. Your in-test known-positive control (`canonical_id_tally(pq.read_table(path)) == 20`) passed, so the mutant genuinely reached the forbidden state.
- `test_record_is_not_re_derived_when_the_parquet_changes_underneath_it` — **the orchestrator's gating acceptance criterion, verbatim** — **PASSES under the read-back mutant.** Under the whole read-back, exactly one test out of 26 died.

So your account is exactly right, and it is worth stating plainly: **had you implemented the acceptance criterion as written, this PR would have shipped a read-back behind a green falsifier.** The partial-write shape is the only thing that separates the two implementations. That is the single most important line of code in this changeset and it does what it claims.

### The kills (mutant proven live in each case)

- **tally as naive row count** (`return table.num_rows`) — killed by the definition test *and* by `test_producer_tally_equals_consumer_read_canonical_ids`, which pins the count against the real `read_canonical_ids` on a fixture carrying a duplicate and an empty string. The distinct-non-empty definition matches the consumer, so the equality does not false-FAIL on a healthy run. That was my item 4 and it holds.
- **`write_canonical` stamps `complete`** — killed by the fail-closed test.
- **`finalize_run` drops the `run_id` guard** — killed by `test_finalize_must_not_upgrade_a_record_it_did_not_mint`. That is the resume path (item 3): a `--from-step` run that touched no canonical writer cannot bless an earlier partial run's tally.
- **md5 not bound to the written bytes** — killed by the last-writer test.
- **publisher imports pyarrow** — killed by the structural subprocess test (whose own detector is verified on a known-positive first).
- **a real seventh writer bypassing `write_canonical`** (I added a live `write_parquet(..., schema=NARRATORS_CANONICAL_SCHEMA)` call to `tabaqa_dates.py`) — the AST gate fires, naming the file and line.

I also confirmed against the **merged** consumer on deploy `origin/main` (`ca14a65`, fetched via `git show` — the local checkout is stale, which is the rc=127 vacuity you hit) that the workflow enforces the completeness equality at line 649 (`CANON_N -ne PROV_CANON_IDS`). That matters for what follows.

### What survived — the two unpinned seams

Both are the failure this PR itself names in `test_no_resolve_module_writes_the_canonical_master_directly`: *"a guard is pinned by the BRANCH THAT CALLS IT, not by the existence of the function."* You applied that principle rigorously to `write_canonical` and then missed it for your own two seams.

**1. `_run()` → `assert_run_publishable` is not pinned.** I removed the call from `publish_parquet.py` (grep-verified: zero call sites remaining). **All 38 publish tests passed.** Every refusal test calls `assert_run_publishable(...)` *directly* as a unit; not one goes through `main()`. The mutant was not merely unkilled, it was *live and dangerous*: `rclone copyto ... /curated/narrators_canonical.parquet` actually fired — an unbacked keep-set reached the upload.

This is **not** a graph-deletion hole, because the consumer's completeness equality catches a short parquet independently. But it defeats the intent your own `verify_remote` docstring states: fail *here, loudly, at publish*, rather than "surfacing as a mysterious refusal on the production box at the moment someone is trying to run a destructive op." A false-FAIL mid-destructive-op is, by this PR's own reasoning, how a good gate gets weakened.

**2. `run_all` → `finalize_run` is not pinned.** I replaced the call with `pass`. **The entire unit suite — 2,229 tests — stayed green.** Nothing asserts that a successful `run_all` produces `run_status=complete`. If that seam ever breaks, `run_status` stays `incomplete` forever and **publish refuses every healthy run** — fail-closed, so nothing is at risk, but the false-FAIL lands at the end of the 7.5-hour re-run, on the launch path.

Tellingly, `_write_run_record` in `test_publish_parquet.py` already takes `status` and `canonical_ids` parameters that **nothing ever varies**. The fixture for the missing test is already sitting there.

### What I am asking for

Two tests. I wrote and validated the first one; take it as-is if it is useful.

```python
@pytest.mark.parametrize(
    ("mutate", "reason"),
    [
        ("incomplete", "did NOT complete"),
        ("md5",        "NOT the file resolve took its tally over"),
        ("absent",     "absent or malformed"),
    ],
)
def test_run_refuses_and_uploads_NOTHING(tmp_path, monkeypatch, capsys, mutate, reason):
    ...  # _fake_run raises AssertionError if rclone is ever invoked
    rc = main(["--parquet-ref", "...", "--curated-dir", str(curated), "--staging-dir", str(staging)])
    assert rc != 0, "publish ACCEPTED an artifact the consumer will refuse"
    assert reason in capsys.readouterr().err, "refused, but NOT by the door under test"
    assert calls == [], "an unbacked keep-set was uploaded to B2"
```

It asserts the **reason**, not merely a non-zero exit, so adding a guard earlier cannot silently un-test the door behind it. Verified: **RED on all three doors** under the disarmed gate (proven live), **GREEN at HEAD**.

The second: drive `run_all` to a successful finish over a small fixture and assert the record on disk declares `run_status=complete` with the tally preserved — and, in the errored case, `incomplete`. A one-sided assertion here is half a control; please pin both directions.

Neither is a behaviour fix. Both are the difference between a gate that works today and a gate that stays working. Given this is the last gate before a 7.5-hour re-run and a destructive op on 83.9% of a graph with no backups, and given the fix is a test I have already written and proven can go red, I would rather it land inline than as a follow-up issue.

Everything else is excellent — the two-artifact split, the fail-closed default, the AST gate, the verified-instrument-before-the-reading discipline throughout. CI is green across all six checks.

---
Requestor: Oyunbileg Batbayar
TechDebt: none

parametrization:
## Architect review — da#429 @ `3e3a1f3`

I verified this at the artifact, executed against the **merged** consumer, and mutation-tested every guard. I took nothing in the PR body on trust. **Approved.**

The headline: **the acceptance criterion I helped specify was worthless, Nikolaos caught that himself, and he was right to overrule it.** I proved that rather than assuming it.

### 1. The falsifier — I wrote the read-back mutant myself

I reimplemented `write_canonical` to derive the tally from the file after the write (`canonical_id_tally(pq.read_table(canonical_path))`) and ran both candidate criteria against it:

| criterion | result vs. the read-back mutant |
|---|---|
| **Truncate the parquet AFTER the tally** (the criterion in the issue) | **PASSED — inert.** Catches nothing. |
| **Lossy write** — writer intends 100 rows, disk receives 20 (Nikolaos's replacement) | **RED.** Killed it. |

The mutant's own log line is the proof: `canonical_ids=20 rows=20` against a writer holding 100. A read-back taken at write time has already read the complete file, so the truncate-after criterion reports 100 under *both* implementations. Nikolaos wrote that version first, watched a mutant sail through it, diagnosed why, and replaced it. That is the discipline working — and it is the thirteenth instance today of a guard that could not see what it existed to catch, found by the implementer on his own work.

### 2. The writer set — six, and the tally is bound to the write, not to a stage

I enumerated them from `RESOLVE_STEP_ORDER` and the source, not from the brief. `disambiguate` (overwrites), `bio_promote` (merges), `cluster` (drops merged ids), `narrator_split` (appends peeled rows), `reconcile` and `tabaqa_dates` (row set preserved, **bytes change**). The last two run *after* `narrator_split`, so the issue's "pin the tally to the last writer" would have shipped a **stale md5**. `write_canonical` re-mints on every call, so whichever writer runs last is automatically the one that stands, and a seventh inherits it. No false-FAIL on a healthy run — the count genuinely moves across writers (`[3, 4, 3, 4, 4, 4]`), so that assertion is not vacuously true of a constant.

### 3. The resume path — refuses, and the refusal is not vacuous

`StopAfterReached` subclasses **`BaseException`** (verified at `_checkpoint.py:209`), so a `--stop-after` probe propagates past every per-stage `except Exception` and never reaches `finalize_run`. It leaves `incomplete` on disk and publish refuses. A `--from-step` resume past every canonical writer leaves a *foreign* record, and the `run_id` guard refuses to bless it — I deleted that guard and `test_finalize_must_not_upgrade_a_record_it_did_not_mint` went red.

Replacing "refuse if the tally-minting stage was skipped" with the **md5 binding** is strictly stronger, and I confirmed it catches the hole the issue's version misses: a `pq.write_table` rewrite outside resolve (a scrub script) is **refused** — `is NOT the file resolve took its tally over`. A skipped tally is no number, and this says so.

### 4. The count definition — same definition, different derivation

The consumer's `read_canonical_ids()` is `{cid for cid in ... if cid}`; `canonical_id_tally` is `len({cid for cid in ... if cid})`. Distinct and non-empty on both sides, pinned against the **real** consumer import rather than a restatement of it. On a real artifact with a duplicate and an empty string, 5 rows yields 3 ids on both sides — no false-FAIL on the destructive op.

### Executed against the merged consumer, with a known-positive control

deploy#574's `verify_prune_provenance.sh` at `origin/main`, fed artifacts from the real producer (rc=127 checked first, so this is not the vacuous run Nikolaos owns up to):

- healthy run → **rc=0**, `PRUNE_PROVENANCE canonical_ids=129 run_status=complete md5=93a86fd1…` (the control — without it every refusal below is bash failing to find a file)
- run never completed → rc=1 · manifest declaring a count → rc=1 (`declares canonical_ids — it must not`) · manifest absent → rc=1 · record absent → rc=1
- **the da#426 shape** — valid-but-truncated parquet, 16/100 rows → **refused at publish**

Six-mutant matrix, each verified to actually apply before its result was counted, each **killed by its own test naming its own reason**: `run_id` guard, run-status door, md5 door, absent-record door, zero-id door, manifest-count ban. `publish_parquet.py` provably never imports pyarrow — detector fired on a known-positive first, so the read-back is impossible, not merely discouraged.

`mypy src/` clean (a cold-vs-warm cache disagreement made me re-run it cache-free and verify with a deliberate canary that mypy fires at all), ruff clean, 2,230 passed, CI rollup **6/6 SUCCESS** at `3e3a1f3`.

### One finding — real, but not a blocker

The AST bypass detector matches only the bare-name `write_parquet(…, NARRATORS_CANONICAL_SCHEMA)` form. I probed it and it is **blind** to `pq.write_table(t, canonical_path)` — the form the scrub scripts actually use and the most natural way a future writer would reach for — and to dotted calls and aliased schemas. That is our own recorded lesson about a gate needing to match every syntactic form.

I did **not** block on it, and I want the reasoning on the record rather than the conclusion: the md5 binding is form-agnostic — it hashes bytes and does not care how they arrived — so I checked whether the blind spot opens any path to a false PASS. It does not. A bypassing writer running *after* the last `write_canonical` mismatches the md5 and publish refuses (I ran exactly that). One running *before* is re-minted over the final table. Every path fails closed. The blindness costs a clear early CI signal, not the safety property. Widen the detector to the call forms it currently cannot see.

---
RequestOrReplied: Approved
TechDebt: `_bypassing_writers` (tests/test_resolve/test_run_record.py) matches only the bare-name `write_parquet(...)` call form — blind to `pq.write_table(...)`, dotted `base.write_parquet(...)`, and schemas passed via an alias variable. Not a safety hole (the md5 binding catches every bypass fail-closed, verified), but the early-warning gate is narrower than it reads. Widen to all call forms.

parametrization:
## Both seams pinned, detector widened — `e7ec4d8`

Oyunbileg: you were right, and the finding is the same rule I wrote into this PR's own AST gate and then broke one layer up. I applied "a guard is pinned by the branch that calls it" rigorously to `write_canonical` and not at all to my own two call sites. Taking the correction as given, not with caveats.

**Seam 1 — `_run()` → `assert_run_publishable`.** Pinned by a parametrized test over all three doors (incomplete run, md5 mismatch, absent record) driven through `main()`. It asserts the **reason** rather than a non-zero exit — an earlier guard could otherwise supply the exit and silently un-test the door behind it — and asserts rclone was never invoked. Deleting the call site now fails with the mutant's own evidence:

```
E  AssertionError: rclone must never be invoked for a refused keep-set:
   ['rclone', 'copyto', '.../curated/narrators_canonical.parquet',
    'PIPELINE:noorinalabs-pipeline/staged/.../curated/narrators_canonical.parquet']
```

That is your live upload, reproduced. I added the known-negative you'd expect too: `main()` must still **publish the healthy artifact**, so a publisher that refuses everything cannot pass the three refusal tests.

**Seam 2 — `run_all` → `finalize_run`.** Pinned in **both** directions over the real `run_all` (reusing the `_install_benign_spies` harness from da#360, with a `disambiguate` spy that genuinely mints a tally through the sanctioned writer): `complete` on a successful run with the tally preserved and the per-stage forensics present, `incomplete` when a stage raises. Replacing the call with `pass` now fails with `run_all never finalized the record`. You were right that a one-sided assertion is half a control.

**Jean-Claude — the detector blind spot, fixed inline rather than deferred.** You were right that it costs the early signal and not the safety property, and right not to block. But "a gate must match every syntactic form of the thing it forbids" is a lesson this org has already paid for, and the blind form was the one the scrub scripts actually use. It now matches bare **and** dotted call forms and keys on `canonical_path` as well as the schema, with four positive and four negative forms as an explicit separation test. Verified against a **real** seventh writer using the form the old gate could not see:

```
E  AssertionError: these write narrators_canonical.parquet without minting a tally:
   {'tabaqa_dates.py': [227]}
```

The old detector was silent on exactly that line. Residual limit is stated in the docstring rather than hidden: a schema reached through an alias variable with an inline-built path would still slip past — narrower early warning, not a safety hole, since the md5 binding is form-agnostic.

All three mutants were verified **live in source** before any number was read, and each is killed by its own test naming its own reason. CI is 6/6 SUCCESS at `e7ec4d8`; unit count 2,230 → 2,243.

One thing worth keeping beyond this PR, since it is the second time today the same shape bit: **the acceptance criterion is itself an instrument and needs a control.** The truncate-after-the-tally criterion was specified as the gate, and it passes under a full read-back — Oyunbileg and Jean-Claude both executed that independently. Writing the bad implementation and requiring the criterion to kill it *before* it becomes a gate is cheap, and it is the only thing that would have caught it.

parametrization:
## Architect re-review — da#429 @ `e7ec4d8`

Re-approved. The delta since my `3e3a1f3` approval is **tests only — zero production-source change** (`git diff --name-only 3e3a1f3 e7ec4d8 -- src/ scripts/` is empty), so everything I verified at the artifact last round stands bit-for-bit: the six-writer routing, the md5 binding, the four publish doors, the `BaseException` stop-after path. I re-verified only what moved.

### Oyunbileg's two seams — I mutated both, and both are now pinned

Each mutant was confirmed to actually apply before its result was counted, and each was killed **by its own test**:

| mutation | before | now |
|---|---|---|
| delete `assert_run_publishable(...)` from `_run()` | **38 publish tests green, and `rclone copyto` fired** | **killed** by `test_main_refuses_a_bad_keep_set_and_uploads_NOTHING`, on all three defect classes |
| delete `finalize_run(...)` from `run_all()` | **2,229 tests green** | **killed** by `test_run_all_finalizes_a_successful_run_to_complete` |
| `finalize_run(..., complete=True)` **unconditionally** | — | **killed** by `test_run_all_leaves_an_errored_run_incomplete` |

The seam-1 test is the right shape: it monkeypatches `subprocess.run` to hard-fail if rclone is invoked at all, so it proves **nothing was uploaded** rather than merely observing a non-zero exit — the distinction that a guard added *earlier* can silently un-test every guard behind it. And the reason is asserted per defect, not just the refusal.

The third row is the one I want to single out. That negative is Nikolaos guarding his **own new test** against being vacuous — `finalize_run(complete=True)` unconditionally would have satisfied the positive test on its own. Writing the mutant that would defeat your own fresh guard, in the same commit you add it, is the habit this whole PR is about.

### My TechDebt, taken inline — verified, and it is a real widening

I re-ran the exact probe matrix that produced the finding. Both forms I named now fire:

- `pq.write_table(t, canonical_path)` — the scrub-script form — **FIRES** (was blind)
- `base.write_parquet(t, canonical_path, schema=...)` — dotted — **FIRES** (was blind)
- `write_parquet(t, canonical_path, schema=S)` — aliased schema — **FIRES** (was blind)
- `write_canonical(...)` and unrelated-artifact writes — **stay silent**, so it did not buy coverage with false positives

### da#430 stays OPEN — the detector does not yet cover *all* forms

You asked me to close it if the detector now covers every form. It does not, so I am not closing it. One reachable form is still blind, and it is the one a new writer is most likely to reach for:

```
pq.write_table(t, output_dir / CANONICAL_PARQUET_NAME)   -> BLIND
pq.write_table(t, out_path)                              -> BLIND
```

`_CANONICAL_MARKERS` keys on the *variable name* `canonical_path` and on the schema symbol. Every current resolve module happens to name it `canonical_path` (I checked all six), so today's code is fully covered — but a seventh writer that uses the module's own exported `CANONICAL_PARQUET_NAME` constant, or just names its variable `out_path`, slips past. Adding `CANONICAL_PARQUET_NAME` to `_CANONICAL_MARKERS` is a one-token fix that closes the first of those.

**This is not a block and I am not asking for a re-push.** The reasoning is the same one that made it debt rather than a blocker last round, and it has not changed: the md5 binding is form-agnostic — it hashes bytes and does not care how they arrived — so every bypass, blind form included, still fails closed at publish. I verified that by execution, not by inspection. The detector is early warning; the md5 is the safety property.

What I do want is for the residual to live in **da#430**, not in this comment. The docstring now states the limit honestly rather than hiding it, which is the right standard — a declared gap is a gap someone can close, and an undeclared one is a gap that gets discovered by a prune.

`mypy` clean (cache-free, canary-verified), ruff clean, **2,243 passed**, CI rollup **6/6 SUCCESS** at `e7ec4d8`.

The caveat I carried last round is unchanged and I will keep carrying it: this hardens the **publish** gate. It does not create backups. The prod wipe still depends on deploy#559/#583/#584/#585.

---
RequestOrReplied: Approved
TechDebt: da#430 stays open (narrowed, not closed). The widened bypass detector now catches the dotted, `pq.write_table`, and aliased-schema forms I filed, but `_CANONICAL_MARKERS` still keys on the variable name `canonical_path`, so `pq.write_table(t, output_dir / CANONICAL_PARQUET_NAME)` and any writer naming its path variable something else remain blind. Not a safety hole — the md5 binding catches every bypass fail-closed, verified by execution — but the early-warning gate is narrower than it reads. Add `CANONICAL_PARQUET_NAME` to the markers.

parametrization:
## QA re-review — da#429 at `e7ec4d8`

Both seams are pinned. **I re-ran my two original mutants and they now die, and I ran six more to check the new tests aren't vacuous in a new way.** Eight mutants, each grep-proven live in the source before I read any number. Full suite green 3/3 (2,242 passed, +13). Everything below was executed, not reasoned about.

### My two mutants now die

- **Remove `assert_run_publishable(...)` from `_run()`** — kills `test_main_refuses_a_bad_keep_set_and_uploads_NOTHING` on **all three** parametrizations. Previously all 38 publish tests stayed green while `rclone copyto` actually fired. Fixed.
- **Replace `finalize_run(...)` with `pass` in `run_all`** — kills `test_run_all_finalizes_a_successful_run_to_complete`. Previously the entire 2,229-test suite stayed green. Fixed.

### The new tests are two-sided, and I confirmed it by mutation

- **Gate refuses EVERYTHING** (`assert_run_publishable` raises unconditionally) — killed by `test_main_publishes_the_healthy_artifact`. So the three refusals cannot be satisfied by a publisher that refuses the honest artifact too.
- **`finalize_run(..., complete=True)` unconditionally** — killed by `test_run_all_leaves_an_errored_run_incomplete`. Your own docstring predicted exactly this, and it holds. You guarded your own new test against the failure mode you were fixing, which is the thing I most wanted to see.

### No door is shadowed — I checked this specifically

This was the sharpest risk: adding the `main()`-level guards **upstream** of the existing unit refusals could have silently un-tested them. So I disarmed each door of `assert_run_publishable` individually:

| door disarmed | what died |
|---|---|
| absent-record | `test_refuses_when_the_record_is_absent` **and** the `absent` e2e parametrization — nothing else |
| `run_status` | `test_refuses_when_the_run_did_not_complete` **and** the `incomplete` e2e parametrization — nothing else |
| md5 binding | `test_refuses_when_the_tally_describes_different_bytes` **and** the `md5` e2e parametrization — nothing else |

Each door kills exactly its own pre-existing unit test **and** its own new e2e parametrization, one-to-one, and nothing else. The new layer sits alongside the old one rather than in front of it, and every refusal still fails for its own reason. The reasons are asserted, not just the non-zero exit.

### The widened detector

Narrowing it back to bare-name-only kills exactly the three dotted forms (`pq.write_table`, `base.write_parquet`, `pyarrow.parquet.write_table`) and nothing else — so the widening is real and `pq.write_table(t, canonical_path)`, the form the scrub scripts actually use, is genuinely covered now. Taking da#430 inline rather than as a follow-up was the right call.

I then probed it on forms **outside** your parametrize list. It fires correctly on the kwargs form (`pq.write_table(table=t, where=canonical_path)`) and produces no false positives. It misses four:

```
pq.write_table(t, out / "narrators_canonical.parquet")     # you DECLARE this one
pq.write_table(t, output_dir / CANONICAL_PARQUET_NAME)     # path via your own constant
wt(t, canonical_path)                                      # function-name alias
write_parquet(t, p, schema=SCHEMA_ALIAS)                   # schema alias
```

**Your safety argument for the residual is correct, and I executed it rather than take it on the docstring's word.** I built a real seventh writer in the exact form the detector cannot see — path inline, no schema reference — dropped 80 of 100 narrators, and ran the real publish gate over the result:

```
sanctioned run   : canonical_ids=100 run_status=complete
after bypass     : file now holds 20 rows (record still declares 100)
PUBLISH REFUSED  -> narrators_canonical.parquet is NOT the file resolve took its tally over.
```

The md5 binding is form-agnostic — it hashes bytes and does not care how they arrived — so a bypass the detector cannot see still **fails closed at publish**. The detector is an early-warning signal, and its blind spots are early-warning gaps, not safety holes. That is the right architecture: the gate that must not be bypassable is the one that does not depend on recognising the bypass.

Worth folding the three undeclared forms into the stated residual limit at some point, since the `CANONICAL_PARQUET_NAME` one uses your own constant and is a plausible way a future writer reaches for the path. Not blocking, and not worth another round on this PR.

### Housekeeping

I hit an 11-test failure on one run and chased it down before trusting it — my worktree had been pruned out from under the running suite. Rebuilt and re-measured: green 3/3, and the base commit is green cold too. **There is no regression; disregard that reading if you saw it.** CI is green on all six checks at `e7ec4d8`.

Nothing further from me. The falsifier holds, both seams are pinned, the new tests are two-sided, no door is shadowed, and the one residual is provably not a safety hole.

---
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

PR #435, opened 2026-07-12
fix(validate): collection_coverage must FAIL on NULL expected_count (#382)

### LINKED ISSUE #382: collection_coverage passes a collection that loaded 0 of 650,986 rows: NULL expected_count is skipped, then reported as 'all within threshold'
`collection_coverage` is the **sole completeness check** on the load, and it reports `passed=True, details='all within threshold'` for a collection that loaded **0 of 650,986** hadiths.

## Mechanism

`deviation_pct` is computed as NULL when `expected_count IS NULL`, and `_classify_collection_coverage` skips NULL rows. **Exactly 1 of 68 staged collection rows carries an `expected_count`** — and `sanadset`, the largest corpus at 650,986 of 853,218 staging hadiths (76.3%), is not the one.

So for 67 of 68 collections the check has no expectation to compare against, produces NULL, is skipped, and contributes nothing. It then reports the aggregate as passing. **A validator that skips every row it cannot measure and then declares the set healthy is not reporting a pass — it is reporting that it ran.**

Demonstrated by Oyunbileg Batbayar while reviewing da#359 (PR #359, comment `4930193629`): fed the post-quarantine shape, the classifier returns `passed=True, details='all within threshold'` over a collection with zero loaded rows. He ran it; this is not a reading of the code.

## Why this is worse than an inert validator

This is the check the **staging gate** rests on. The re-run plan on `noorinalabs/noorinalabs-main#928` is: load stg → verify → promote to prod. If `collection_coverage` cannot distinguish "loaded everything" from "loaded nothing" for the corpus that *is* three quarters of the data, then step 5 verifies nothing and the stg gate is decorative for exactly the failure mode it exists to catch.

It was also the third of three blind gates in the da#359 chain — producer gate swallowed by `parse_all`'s `except Exception`, load gate folding quarantines into an unattributable `total_skipped`, and this. Parse exits 0, load exits 0, validation passes, 76.3% of the corpus is absent.

## Family

Eleventh member of the silent-fallback family tracked on `noorinalabs/noorinalabs-main#928`, and structurally identical to `deploy#565` — the `BackupFailure` alert whose `> 0` guard makes a never-succeeded backup read as healthy. Both are named for the thing they do not detect. Both advertise coverage they do not have.

Direct instance of the standing rule: **a silent zero is not a measurement**, and **a guard you cannot make fire is not a guard**.

## Acceptance

- A collection with **no** `expected_count` must not be silently skipped. Either it fails the check, or it is reported as *unmeasured* in a count the classifier surfaces — never absorbed into "all within threshold."
- The classifier must distinguish three outcomes: **passed**, **failed**, **not measurable**. `details` must state how many collections fell in each.
- `expected_count` must be populated for `sanadset` — and, separately, an audit of how many of the 68 have one.
- **Prove the guard red first.** Construct the exact post-quarantine shape Oyunbileg used (sanadset at 0 rows loaded) and show the current classifier returning `passed=True` before the fix, and a non-pass after. A fixture that cannot produce the failing state proves nothing (`feedback_fixture_makes_guard_assertion_inert`).
- The fixture must be built from the **real** staged collection shape, not a synthetic 2-row table.

## Stage position

Runs at **validation, after `load`** — downstream of the `resolve` choke point. **Not a launch-gate item**; it may land while the ~7.5h re-run is in flight. But it **must be merged and proven before the staging load is verified**, or the stg gate cannot be trusted and prod promotion has no evidence behind it.

Part of #359

parametrization:
## Summary

`collection_coverage` validation reported a **false PASS** for a collection that loaded 0 of 650,986 rows. `_classify_collection_coverage` skipped every row whose `deviation_pct` was NULL — the cypher's `CASE WHEN c.expected_count IS NOT NULL ... ELSE null` arm — and then reported "all within threshold". So a collection with a missing `expected_count` that loaded nothing was silently rolled into a green summary, hiding a total load failure.

This is the exact blind spot already documented in `tests/test_graph/test_malformed_id_exit.py`: *"collection_coverage cannot measure a collection with a NULL expected_count, which is 67 of 68 (da#382)."*

## Fix

Switch to **positive classification**: only a numeric deviation within threshold PASSes. A NULL/missing `expected_count` is un-evaluable and now **fails the gate** with an explicit `expected_count missing — coverage un-evaluable (actual=N)` detail. A collection that DOES have a real `expected_count` within threshold still passes unchanged.

## Red → green evidence

QA red-first. Two tests were written to fail on current `main` and pass after the fix:

- `test_null_expected_is_fail` (replaces the old `test_null_expected_is_pass`, which asserted the buggy behavior) — NULL `expected_count` + 0 loaded rows → **FAIL**.
- `test_null_expected_not_masked_by_healthy_sibling` — control: a within-threshold collection alongside a NULL-expected one; the whole run must still FAIL and name the un-evaluable collection.

Before the fix (against `main` code):
```
FAILED ...::test_null_expected_is_fail
FAILED ...::test_null_expected_not_masked_by_healthy_sibling
2 failed
  -> ValidationResult(passed=True, details='all within threshold', ...)
```
After the fix: full `tests/test_graph/` suite **285 passed**; ruff format/lint, mypy strict all clean (pre-commit + pre-push green).

## Files changed
- `src/graph/validate.py` — `_classify_collection_coverage`: NULL/non-numeric `deviation_pct` now contributes a failure instead of being skipped.
- `tests/test_graph/test_validate.py` — red-first tests.

Closes #382

### DIFF
--- a/src/graph/validate.py
+++ b/src/graph/validate.py
@@ -301,13 +301,33 @@ def _classify_collection_coverage(
     rows: list[dict[str, object]],
     deviation_threshold: float,
 ) -> ValidationResult:
+    """Compare expected vs actual hadith counts per collection.
+
+    A collection fails the gate two ways:
+
+    * ``deviation_pct`` exceeds ``deviation_threshold`` — the loaded count
+      strayed too far from the collection's ``expected_count``.
+    * ``deviation_pct`` is NULL — the collection carries no ``expected_count``
+      (the cypher's ``CASE ... ELSE null`` arm), so coverage is **un-evaluable**.
+      The predecessor *skipped* those rows and then reported "all within
+      threshold", so a collection that loaded 0 of its 650,986 rows was rolled
+      into a green summary — a false PASS hiding a total load failure (da#382).
+      An un-evaluable collection is not a healthy one: it fails the gate. Only a
+      numeric deviation within threshold is a positive PASS.
+    """
     count = len(rows)
     failures: list[str] = []
     for row in rows:
+        cid = row.get("collection_id", "?")
         dev = row.get("deviation_pct")
-        if dev is not None and isinstance(dev, int | float) and dev > deviation_threshold:
-            cid = row.get("collection_id", "?")
-            failures.append(f"{cid}: {dev:.1f}% deviation")
+        if isinstance(dev, int | float):
+            if dev > deviation_threshold:
+                failures.append(f"{cid}: {dev:.1f}% deviation")
+        else:
+            actual = row.get("actual", "?")
+            failures.append(
+                f"{cid}: expected_count missing — coverage un-evaluable (actual={actual})"
+            )
     passed = len(failures) == 0
     details = "all within threshold" if passed else "; ".join(failures)
     return ValidationResult(query_name, passed, details, count)
--- a/tests/test_graph/test_validate.py
+++ b/tests/test_graph/test_validate.py
@@ -181,12 +181,39 @@ class TestCollectionCoverageClassification:
         result = _classify("collection_coverage", rows)
         assert result.passed is False
 
-    def test_null_expected_is_pass(self) -> None:
+    def test_null_expected_is_fail(self) -> None:
+        """da#382: a NULL/missing ``expected_count`` makes coverage un-evaluable.
+
+        The cypher returns ``deviation_pct = null`` for a collection whose
+        ``expected_count`` is NULL. The old classifier *skipped* that row and
+        reported "all within threshold" — so a collection that loaded 0 of its
+        650,986 rows was silently reported healthy. An un-evaluable collection
+        must FAIL the gate, not silently pass.
+        """
         rows: list[dict[str, object]] = [
-            {"collection_id": "col:unknown", "expected": None, "actual": 42, "deviation_pct": None},
+            {"collection_id": "col:unknown", "expected": None, "actual": 0, "deviation_pct": None},
         ]
         result = _classify("collection_coverage", rows)
-        assert result.passed is True
+        assert result.passed is False
+        assert result.is_fatal is True
+        assert "col:unknown" in result.details
+
+    def test_null_expected_not_masked_by_healthy_sibling(self) -> None:
+        """The un-evaluable collection must fail the whole run even when another
+        collection is genuinely within threshold — the failure cannot be rolled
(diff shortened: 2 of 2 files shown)


parametrization:
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

Verified against the diff at HEAD (1c8df85), not the PR body. The rewritten _classify_collection_coverage now routes any non-numeric/None deviation_pct into the else arm and appends an un-evaluable failure, so a collection whose expected_count is NULL FAILs instead of being skipped and rolled into a false 'all within threshold' PASS. A numeric deviation within threshold still appends nothing, so genuinely-healthy collections keep passing — the change is not over-broad. is_fatal resolves True because ValidationResult.warning defaults False (not a timeout/warning downgrade), matching the new assertion. Both new tests are red-first: test_null_expected_is_fail inverts the old True->False expectation (fixture actual 42->0), and test_null_expected_not_masked_by_healthy_sibling pairs a within-threshold col:bukhari with a NULL col:empty and asserts the run fails — both would pass under the old skip-None logic and now correctly fail. Full CI green (Lint/Type/Test/Integration/fixture-realism/sync-drift).

parametrization:
Requestor: Kwesi Boateng
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

Verified against the code at head 1c8df85, not the PR body.

Behavior is correct. In _classify_collection_coverage the branch is now isinstance(dev, int | float): a numeric deviation within threshold is the only positive PASS, over-threshold is a FAIL, and the else arm makes a NULL or non-numeric deviation_pct an un-evaluable FAIL with a clear per-collection detail. This closes the da#382 hole where the predecessor skipped the NULL row and still reported all within threshold, rolling a 0-of-650,986 load into a green summary.

is_fatal holds without a code change: the property is not passed and not warning, and the classifier constructs ValidationResult with warning defaulting to False, so a NULL-expected FAIL is fatal as test_null_expected_is_fail asserts.

Both tests are genuinely red-first against the old skip-NULL logic: test_null_expected_is_fail (old code returned passed True, new asserts False plus is_fatal True) and test_null_expected_not_masked_by_healthy_sibling (col:empty NULL alongside a within-threshold col:bukhari; old code passed, new fails and names col:empty). No over-broad regression: the real within-threshold PASS path is unchanged and the change is scoped to the two files.

PR #437, opened 2026-07-12
resolve: key narrator-split chain index on (hadith_id, chain_index) (#411)

### LINKED ISSUE #411: narrator_split death-date neighbour window groups by hadith_id alone — a multi-chain hadith picks a cross-chain neighbour (chain_index now available)
Discovered during da#282 review (Alejandra Reyes-Fuentes + both reviewers, PR#410).

`narrator_split._build_chain_index` (death-date estimation) groups `position±1` neighbours by `hadith_id` alone. For a hadith carrying more than one transmission sequence (the real present-day case is `lk`, which emits a separately-numbered Arabic AND English isnad per hadith), a narrator can be assigned a "neighbour" from the OTHER sequence — a cross-chain neighbour that was never adjacent in any real chain.

**Now trivially fixable:** da#282 (merged, PR#410) added `chain_index` to the resolved-mentions schema. The fix is to group the death-date neighbour window by `(hadith_id, chain_index)` instead of `hadith_id` alone — the exact same rekeying da#282 applied to `TRANSMITTED_TO`/`NARRATED`/Chain-nodes.

**Impact: LOW, NON-GATE.** For `lk` the Arabic and English isnads are parallel translations that resolve to the same canonical narrators, so the cross-language neighbour is usually the same person's translation. Behaviour is unchanged by da#282 (pre-existing). This does NOT gate the re-run.

**Note:** touches da#376 death-year territory — the death-date estimation heuristic is owner/architect-sensitive, so scope the fix to the grouping-key change only (zero-inference, no heuristic change) unless the owner wants more.

parametrization:
## Summary

PR-A (issue #411) — the hard prerequisite for the batched narrator-split re-run.

`src/resolve/narrator_split.py::_build_chain_index` keyed its `chains` structure on **`hadith_id` alone**, while the graph loaders key chains on the composite **`(hadith_id, chain_index)`** (`src/graph/load_edges.py::_chain_index`, da#282). For a **multi-isnad hadith** (same `hadith_id`, multiple `chain_index` values — today lk's Arabic + English isnads, each numbered from position 0), dropping `chain_index` flattened both isnads into one position-sorted list. `_collect_datable`'s pos±1 neighbour lookup then fabricated cross-chain teacher/student neighbours that exist in no real chain, corrupting the death-date neighbour window the split decision (and the #337/#346 context signal built on top of it) rests on.

## Fix (mechanical key-alignment — not the #337/#346 discriminator, which is PR-B)

- `_build_chain_index` now reads `chain_index`, keys `chains` on `(hadith_id, chain_index)`, and carries the ordinal through `candidate_mentions`. New `_coalesce_chain_index` helper mirrors `load_edges._chain_index` (None → 0, so single-isnad hadiths and legacy pre-da#282 files are unaffected). The da#337-PR-3a OOM bound is preserved and slightly tightened — pass 2 restricts to candidate-touching `(hadith_id, chain_index)` chains instead of whole hadiths.
- `_collect_datable`'s ±1 neighbour lookup keys on `(hadith_id, chain_index)`, so a neighbour is only ever within the same chain.

## Test evidence (RED → GREEN)

New `tests/test_resolve/test_narrator_split.py::test_multi_isnad_no_cross_chain_neighbour`: one `hadith_id`, two `chain_index` values (0 and 1), distinct attested narrators per chain, candidate at position 1 in both. Asserts each candidate mention's ±1 anchors come **only** from its own chain.

- **RED** on the pre-fix (hadith_id-only) code — chain-0 mention's anchors leaked in chain-1's narrators (d.250, d.320):
  ```
  E  AssertionError: assert {...320'} == {...170'}
  E    Extra items in the left set: 'راو مءرخ رقم 320', 'راو مءرخ رقم 250'
  ```
- **GREEN** after the composite-key fix. Full module: 17 passed. Full `tests/test_resolve/`: 539 passed, 6 skipped.

## Files changed
- `src/resolve/narrator_split.py` — `_coalesce_chain_index` (new), `_build_chain_index` (composite key + 4-tuple candidate_mentions), `_collect_datable` (composite-key neighbour lookup), algorithm docstring line updated.
- `tests/test_resolve/test_narrator_split.py` — `_mention_row` gains an optional `chain_index`; new `test_multi_isnad_no_cross_chain_neighbour`.

## Scope
Exactly the key-alignment + its neighbour-window consequence + the test. The #337/#346 isnad-neighbour clustering (PR-B) depends on this landing first and is **not** implemented here.

Closes #411

### DIFF
--- a/src/resolve/narrator_split.py
+++ b/src/resolve/narrator_split.py
@@ -32,7 +32,9 @@ Algorithm (per eligible canonical id ``C``, name ``N``)
 -------------------------------------------------------
 1. Gather every mention with ``canonical_narrator_id == C``.
 2. **Isnad-adjacency evidence** — for each such mention, look at the chain neighbours
-   at ``position_in_chain`` ±1 in the same ``hadith_id`` and map them to their
+   at ``position_in_chain`` ±1 in the same ``(hadith_id, chain_index)`` chain (da#411 —
+   the composite key the graph loaders use; NOT ``hadith_id`` alone, which would
+   fabricate cross-chain neighbours in a multi-isnad hadith) and map them to their
    *attested* ``death_year_ah`` (precision ≠ ``tabaqa_estimate`` — an *estimated*
    window is never used as evidence, which also keeps the pass cascade-free on
    re-run). A teacher (neighbour at position−1) dies a transmission gap *earlier*;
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

Verified at head ce0da9d. Both the chain builder and the neighbour window now key on the composite hadith_id + chain_index, and _coalesce_chain_index is byte-identical to the mirror it cites in load_edges._chain_index — int(v) if v is not None else 0 — so single-isnad hadiths keep their old grouping and multi-isnad hadiths stay separated. _collect_datable confines each mention's pos-1/pos+1 lookup to chains.get on its own chain key, so a neighbour can never cross a chain_index boundary within one hadith. I grepped the whole module at head: the production caller and every other consumer pass the new shapes consistently, with no lingering hadith_id-only key.

The red-first test is load-bearing, not a trivial pass. It builds one hadith with two isnads, the split candidate at position 1 in each, and attested anchors far apart per chain — chain 0 at d.100/d.170, chain 1 at d.250/d.320. Under the pre-fix hadith_id-only keying the pos-1/pos+1 lookup collects all four anchors for each candidate mention, so the equality assertion and both non-intersection assertions fail; under the composite key each mention draws only its own chain's two anchors. It drives _build_chain_index and _collect_datable directly, the tightest surface for this fix. The OOM bound from da#337 is preserved — footprint is now restricted to candidate-touching chains, a subset of candidate-touching hadiths. Scope is limited to the key alignment, neighbour window, and test; chain_index is a real nullable int32 column in the resolved schema. CI is green across Lint, Type Check, Test, Integration, base-image/fixture-realism, and the sync-drift gate.

parametrization:
RequestOrReplied: Approved
TechDebt: #439

Reviewed at head_sha ce0da9d against the loader source at the same ref.

1. Key alignment is correct and exact. _coalesce_chain_index (int(value) if value is not None else 0) is byte-identical to load_edges._chain_index. chain_index is a first-class nullable int32 in NARRATOR_MENTIONS_RESOLVED_SCHEMA (da#282), so both passes read it safely on production data — no crash risk. No off-by-one: position_in_chain (within-chain ordinal) and chain_index (which isnad) are orthogonal, and the code keeps them distinct throughout.

2. Neighbour window is sound. _collect_datable looks up chains.get((hadith_id, chain_index), ()) and gates on npos == position ±1. There is no path that reaches the ±1 comparison without first selecting the composite-key bucket, so a ±1 neighbour can never cross a chain_index boundary within one hadith_id. The candidate_mentions 4-tuple correctly carries chain_index from pass 1 into the lookup.

3. Test is genuinely red-first. On the pre-fix hadith_id-only key, h1's two isnads flatten to one position-sorted list; the pos-1 candidate's ±1 window then picks up positions 0 and 2 from BOTH isnads, leaking chain-1's d.250/d.320 into chain-0's anchor set — exactly the reported AssertionError. The composite key confines each mention to its own chain (GREEN). Behaviour change is limited to the key, as claimed; the OOM bound is preserved and tightened to candidate-touching chains rather than whole hadiths.

4. Clean signal for PR-B, with one carry-forward. This removes the cross-chain fabrication PR-B's context clustering must not see — the blocking prerequisite is met. Non-blocking, and NOT introduced by this PR: the splitter's exact-position ±1 window is not identical to the loader's TRANSMITTED_TO adjacency, which pairs consecutive RESOLVED mentions across position gaps (_build_chain_pairs). On chains with a gap (an intermediate unresolved/null-canonical mention), the loader creates an edge the splitter's ±1 window misses — an under-count, not fabrication. Filed as #439 so PR-B settles the adjacency definition deliberately (match consecutive-resolved, or pin exact-±1 with a gappy-fixture test) rather than inheriting the divergence silently.

Approving #437 on its stated scope: the key-alignment, its neighbour-window consequence, and the red-first test are all correct and complete.

Issue #398, opened 2026-07-10
bio_promote truncated-merge guard over-refuses 792 full-nasab isnad-tail rows (recall cost + doc-contract precision)


parametrization:
## Summary

da#383's `bio_promote` truncated-merge guard refuses **1,690** bio→canonical merges. **792 of them are full-nasab rows that lost only an isnad tail** — the asserted name is complete; only the chain continuation (`عن …`, `حدثنا …`) was cut. Refusing them is **safe** (see below) and the merge that shipped is correct, but it carries a real **recall cost**, and `cleaner_removed_content`'s contract overstates its precision for this class.

Not a data-corruption defect. Does **not** gate the `resolve` re-run — refusing is non-corrupting in both directions. This is a recall + documentation follow-up.

## Measured (Kavitha Sundaramurthy, against the local dev corpus, read-only, using the branch's own functions at `2dd4946`)

Base `main` had **no guard** at merge-base `70b66ad7` — I verified this independently: `git show 70b66ad7:src/resolve/bio_promote.py | grep -c 'cleaner_removed_content|truncat'` → **0**, vs **10** on `origin/main`. So da#383 owns the entire 1,690-row behaviour change; pre-383, all 1,690 back-filled.

```
1,690 refusals
├─ 427  bare-ism        (the عبیده class) — refusing is correct, unambiguous
└─ 909  nasab
    ├─ 792  lost ONLY an isnad tail (removed suffix starts عن / حدثنا / …)
    │       → the NAME IS COMPLETE; only the chain continuation was cut
    └─ 117  other

grade distribution of the 909 nasab:
  504 daif · 113 thiqa · 102 saduq · 84 unknown · 62 kadhdhab · 44 matruk
```

**This corrects the record.** An earlier review characterized the 909 as *"`لا يعرف`/`مجهول` homonyms keyed by teacher."* Only **84/909** are unknown; the rest are graded narrators, and the "keyed by teacher" pattern is the isnad tail, not a disambiguator.

## Why the shipped merge is still correct (why this is a follow-up, not a regression)

[...]

Issue #397, opened 2026-07-10
bio_promote: the truncated-merge guard refuses 773 gloss-tail rows whose residue is the complete asserted name


parametrization:
## Summary

`cleaner_removed_content` (da#379, PR #383) correctly separates the cleaner's **affix normalizations** from its **truncations**. But within "truncation" it does not separate two categorically different cuts, and the `bio_promote` guard refuses both.

Measured at `2dd49463d88af4094398174d9abd646b10ad55a1` over the real staging corpus (`narrators_bio_{itqan,kaggle,muhaddithat}.parquet`, 140,174 named rows) against `data/curated/narrators_canonical.parquet`:

- `cleaner_removed_content` is True for **13,233** rows.
- Of those, **1,690** target a canonical narrator with `mention_count > 0` and are therefore refused (`skipped_truncated_merge`).

Splitting the 1,690 refusals by what the cut actually removed:

| Class | Cut removed | Count |
|---|---|---|
| **A — gloss-tail** | tail begins `عن`, residue is a complete nasab (>= 3 tokens, contains `بن`) | **773** |
| **B — name-cut** | bare ism/kunya residue, or a non-`عن` cut (`قيل`, `قال`, …) | 917 |

## Class B is exactly what the guard is for

```
raw     : ابو سلمه ، قيل : هو الحكم بن عبد الله بن خطاف ، وقيل : عبد الله بن سعد
residue : ابو سلمه          target mention_count = 9072
```

An entry about an **unidentified** Abū Salama whose identity the compiler disputes. The residue is a bare kunya; the canonical bearer at 9,072 mentions is Abū Salama b. ʿAbd al-Raḥmān. Merging would back-fill a disputed identity onto a major tābiʿī. **Refusing is correct, and this is a real corruption that `main` currently performs.**

## Class A is a precision gap

```
raw     : اسحاق بن مره عن انس
residue : اسحاق بن مره      target mention_count = 18
```

`عن انس` is a **teacher key** — the rijāl compiler's disambiguator for a narrator known only through one chain. It is not name material. The residue *is* the complete name the source asserted. The canonical bearer's 18 isnad mentions are, on the face of it, the same person.

The guard refuses these anyway. Class-A targets sum **48,638 mentions**; **110** of them carry >= 100 mentions. Each loses the grade / death-year enrichment that `bio_promote` exists to deliver — and which `main` currently delivers.

## Why this is not simply "merge them"

[...]

PR #441, opened 2026-07-12
bio_promote: count accepted truncated-residue residual + correct guard docs (#385/#397/#398)

### LINKED ISSUE #385: bio_promote: the accepted truncated-residue residual (13,090 merges + 6,756 new nodes) is uncounted
Found while reviewing PR #383 (da#379).

PR #383 scopes its guard to bio rows whose truncated residue would claim an **attested** canonical record (`mention_count > 0`). That boundary is right, and the PR states the residual openly: a truncated residue landing on a zero-mention catalog node still merges two catalog entries under a mangled key.

What is missing is the residual's **size**. `skipped_truncated_merge` counts only the drops. Measured against the current staging bios and `data/curated/narrators_canonical.parquet`:

| outcome for a truncated bio row | itqan | kaggle | muhaddithat | total |
|---|---|---|---|---|
| dropped (target attested) — counted | 1,787 | 0 | 2 | 1,789 |
| merged onto a zero-mention node — **uncounted** | 7,606 | 5,484 | 0 | 13,090 |
| minted a **new** canonical node keyed on the residue — **uncounted** | 4,538 | 2,218 | 0 | 6,756 |

So the accepted residual is 13,090 merges plus 6,756 new nodes named by a truncation, and the pipeline emits no signal for either. This is the same shape as the `cluster_records_unblockable` counter added in PR #363: an outcome that is deliberate should still be observable, or "we accepted it" becomes indistinguishable from "it never happened."

Suggested: two counters on the `bio_promote` metrics alongside `skipped_truncated_merge` — one for a truncated residue merging onto an unattested node, one for a truncated residue minting a new node — plus a single aggregate `logger.warning` naming the deferred extraction work as the owner. No behaviour change.

Related: the deferred name/prose extraction and the kaggle corpus (#299).

parametrization:
## Summary

Observability + documentation only. The da#379 truncated-merge guard's **refuse-line is unchanged** — no behavioral loosening. A blast-radius measurement (below) showed the proposed loosening would join **distinct persons**, so the behavioral recovery stays deferred; this PR ships only the safe half of PR-CD.

## What changed

- **#385 — count the accepted truncated-residue residual.** Two counters on `bio_promote` (`truncated_residue_minted`, `truncated_residue_onto_unattested`) + one aggregate `logger.warning` naming the deferred name/prose extraction (da#299) as owner. The guard refuses only a truncated residue claiming an *attested* node; every other truncated residue is accepted (mints a new node, or fuses onto a zero-mention node) and was previously **silent** — "we accepted it" was indistinguishable from "it never happened" (`feedback_silent_zero_is_not_a_measurement`). Live-corpus residual: **~6,983 rows minted new nodes (4,199 distinct cids) + ~4,765 merged onto zero-mention nodes.**
- **#398 — correct `cleaner_removed_content`'s contract.** A `True` means "removed more than an affix", **not** "cut into the name". The docstring now separates a bare-ism *cut into the name* from a *tail-after-a-complete-name* removal (`اسحاق بن مره عن انس` → `اسحاق بن مره`), and states plainly that deciding whether such a residue is the same person is a homonymy question a token diff cannot answer.
- **#397 — correct the misleading inline guard comment.** It justified refusal as "always safe" over "bare ism" residues; in fact ~40% of the mention-bearing refusals are **full nasabs**, and refusal is non-corrupting-but-withholds-enrichment, not symmetrically "safe". The comment now documents the deliberate class-2 over-refusal pending disambiguation.

## Why the behavioral loosening is NOT in this PR (the measurement)

Measured with the branch's own cleaner against the live staging corpus (`narrators_bio_{itqan,kaggle,muhaddithat}`, 138,707 named rows) + `data/curated/narrators_canonical.parquet` (107,951 canonical):

- A refined guard would newly accept **460** merges onto attested narrators (240 distinct target names, Σ 41,997 mentions).
- **Bidirectional, unweighted, Arabic-reader lens:** the newly-accepted set joins distinct persons. **45 of 240 target name-keys carry conflicting jarḥ grades** across bio entries — `محمد بن الأزهر الجوزجاني` (mc=4) carries all of daif/kadhdhab/matruk/saduq/thiqa/unknown on one key; `جابر بن يزيد` (mc=53) daif+kadhdhab+unknown. One name-key graded both *thiqa* and *kadhdhab* is provably >1 person, keyword-independent. A further ≥106/460 rows carry an explicit disputed-identity marker (مجهول / لا يعرف / وقيل اسمه) or a death-year/generation conflict.
- Instrument verified on both classes: dispute-marker base rate 4.4% across all itqan bios vs 17.8% on the teacher-tail set (4× separation).

Loosening the guard would back-fill *forger*/*abandoned* verdicts onto attested narrators by bare name-match — the da#423 direction (`feedback_drop_gate_bidirectional_ab`). The behavioral recall recovery (class-A merge / same-person recovery) is **deferred** behind biographical homonym disambiguation (jarḥ-grade / death-year separation) + the full bidirectional-unweighted A/B + Arabic rijāl gate. #397 and #398 stay open as the deferred trackers.

## Tests

2 new positive-witnessed counter tests (`test_truncated_residue_minting_a_new_node_is_counted_and_warns`, `test_truncated_residue_onto_zero_mention_node_is_counted`). Full suite green; ruff format+lint, mypy strict, pytest all pass (pre-commit + pre-push).

Closes #385
Re #397
Re #398

### DIFF
--- a/src/parse/name_quality.py
+++ b/src/parse/name_quality.py
@@ -1834,7 +1834,7 @@ def asserted_name_tokens(name_normalized: str | None) -> tuple[str, ...]:
 
 
 def cleaner_removed_content(name_normalized: str | None, cleaned: str | None) -> bool:
-    """True when :func:`clean_narrator_name` cut into the name the source asserted (da#379).
+    """True when :func:`clean_narrator_name` removed name-span material, not just an affix (da#379).
 
     ``clean_narrator_name`` does two categorically different things that its return
     value alone does not distinguish:
@@ -1842,18 +1842,32 @@ def cleaner_removed_content(name_normalized: str | None, cleaned: str | None) ->
     * **Normalization** — markup, honorifics, editorial connectives and a bracketed
       catalog entry number are affixes *around* a name. Stripping them leaves the
       name the source asserted, whole: ``"( 4875 ) عبد الله بن موسى"`` →
-      ``"عبد الله بن موسى"``, still that narrator's full name.
-    * **Truncation** — an isnad/matn tail or a colon-joined body is *part of the same
-      span* as the name, and cutting it keeps only a head. The residue is not a name
-      the source asserted, and it is frequently a bare ism:
-      ``"عبيدة مولى رسول الله ذكره بن شاهين …"`` (an obscure client of the Prophet)
-      reduces to ``"عبيده"`` — the name of a different, heavily-attested narrator.
+      ``"عبد الله بن موسى"``, still that narrator's full name. This returns ``False``.
+    * **Truncation** — an isnad/matn tail or a colon-joined body is cut off, keeping
+      only a head. This returns ``True``. Note it does NOT imply the cleaner "cut into
+      the name" (da#398): truncation splits into two sub-cases the return value again
+      does not separate —
 
-    A caller that keys identity on the cleaner's output must tell these apart. An
-    affix is not a tail. Comparing against :func:`asserted_name_tokens` — rather than
-    against a re-tokenization of the raw input — is what makes that distinction; a
-    re-tokenization sees the entry number, the connective and the honorific as
-    "content removed" and refuses a merge the source fully supports.
+        - *cut into the name*: the residue is a fragment, frequently a bare ism —
+          ``"عبيدة مولى رسول الله ذكره بن شاهين …"`` (an obscure client of the Prophet)
+          reduces to ``"عبيده"``, the name of a different, heavily-attested narrator.
+        - *tail after a complete name*: the asserted name is INTACT and only a chain
+          continuation after it was removed — ``"اسحاق بن مره عن انس"`` → ``"اسحاق بن
+          مره"`` (the full nasab; only the teacher-key ``عن انس`` cut). The name was
+          not cut *into*; a tail *after* it was removed. Roughly 40% of the corpus's
+          truncations are this class (da#397/da#398).
+
+    So a ``True`` means "removed more than an affix", NOT "the residue is not a name".
+    A caller keying identity on the cleaner's output must not over-read it (da#398):
+    it separates affix-normalization from name-material removal, and no finer. Deciding
+    whether a tail-after-complete-name residue is the SAME person as an existing record
+    is a homonymy question this function cannot answer — it needs biographical
+    disambiguation, not a token diff.
+
+    Comparing against :func:`asserted_name_tokens` — rather than a re-tokenization of
+    the raw input — is what draws the affix/removal line; a re-tokenization sees the
+    entry number, the connective and the honorific as "content removed" and refuses a
+    merge the source fully supports.
     """
     if not name_normalized or not cleaned:
         return False
--- a/src/resolve/bio_promote.py
+++ b/src/resolve/bio_promote.py
@@ -242,6 +242,14 @@ def promote_bios_to_canonical(
     skipped_source = 0
     skipped_pollution = 0
     skipped_truncated_merge = 0
+    # The ACCEPTED truncated-residue residual (da#385). The guard only refuses a
+    # truncated residue that would claim an *attested* node (mention_count > 0);
+    # every other truncated residue is accepted — it either lands on a zero-mention
+    # catalog node or mints a new node keyed on the truncation. Both are deliberate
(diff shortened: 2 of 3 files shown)


parametrization:
Requestor: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

Reviewed at head 1c74083 against the code, not the PR body. All three required points hold; no behavioral change slipped in.

Guard refuse-line is byte-unchanged. The refusal — existing = canonical_map.get(cid); if truncated and existing is not None and (existing.get('mention_count') or 0) > 0: skipped_truncated_merge += 1; continue — sits as unchanged context, as does truncated = cleaner_removed_content(norm, cleaned). The diff around it is comments plus two pure counter increments; no condition, threshold, continue, or return was added or moved on the guard path, so promotion outcomes are identical. The change is observability-only.

The two da#385 counters are correct and count only ACCEPTED (post-guard) residues. truncated_residue_minted increments inside the if cid not in canonical_map mint branch, only when truncated. truncated_residue_onto_unattested increments inside the else branch, only when truncated — and because the guard already refused any truncated-onto-attested (mention_count > 0) target above, everything reaching that else is a zero-mention node, so the name matches the count. existing is not None iff cid in canonical_map, so the two branches partition cleanly with no double-count. Both surface in bio_promote_complete and are summed once into the bio_promote_truncated_residue_accepted WARN, which fires only when the total is nonzero. The two new tests are red-first by construction (the complete-event keys do not exist without this change) and exercise the mint path and the zero-mention-merge path respectively, the latter also confirming the guard's mention_count == 0 exemption still accepts.

The doc/contract corrections are accurate. cleaner_removed_content is literally tuple(cleaned.split()) != asserted_name_tokens(name_normalized), so True means the cleaner dropped tokens the source asserted as name-span, beyond affixes — which is exactly what the rewritten docstring now says. The prior 'cut into the name' wording was the over-read #398 targets: it conflated the bare-ism cut-into case with the tail-after-a-complete-name case (ishaq bin mirra 'an anas to ishaq bin mirra), both of which return True. The inline guard comment now describes both classes and correctly frames refusing class 2 as a conservative non-corrupting default (homonymy is not folded by make_canonical_id), consistent with the unchanged guard. The behavioral loosening being deferred rather than shipped matches the reduced scope.

parametrization:
RequestOrReplied: Approved
TechDebt: none

Reviewed at head 1c74083 via gh pr diff / contents API — no checkout. Reduced-scope PR-CD (counters + doc corrections); confirmed no behavioral loosening slipped in.

1. Guard refuse-line unchanged. The da#379 refuse condition is byte-identical to main (`if truncated and existing is not None and (existing.get("mention_count") or 0) > 0:` → `skipped_truncated_merge += 1; continue`). The measured-unsafe behavioral loosening is correctly still deferred, same bar as #347b/#431. The guard stays scoped to attested (mention_count > 0) targets only.

2. Both da#385 counters are correct and observation-only. `truncated_residue_minted` increments in the `cid not in canonical_map` mint branch, `truncated_residue_onto_unattested` in the `else` merge branch; both gated on `truncated`. Reaching the else with truncated=True is provably a zero-mention target — a mention_count > 0 node was already refused by the guard above — so the onto-unattested classification is sound. Neither counter alters control flow, promotion, or refusal; they only feed the `bio_promote_complete` info fields and the new aggregate WARN. The two new tests pin each path exactly (mint-new against an empty curated dir → minted=1; zero-mention target via `_write_attested_narrator(..., mentions=0)` → onto_unattested=1), and use the verbatim itqan prose fixture so the guard assertion is not inert.

3. da#397/da#398 doc corrections are accurate. `cleaner_removed_content`'s executable body is unchanged (`tuple(cleaned.split()) != asserted_name_tokens(...)`); the docstring rewrite correctly narrows the claim from "cut into the name" to "removed name-span material beyond an affix", which is what the code actually returns — True covers BOTH the cut-into-name (bare-ism) and the tail-after-complete-name classes, and a caller must not over-read it as "the residue is not a name". The bio_promote guard comment's two-class framing and the conflicting-jarḥ homonymy rationale for the conservative default are consistent with the da#423 direction.

All six CI checks green, mergeable. No behavioral change; docs and counters only.

PR #442, opened 2026-07-12
Tag & keep bio-only narrators: attestation property (#370)

### LINKED ISSUE #370: OWNER DECISION: are the 42,770 bio-only narrators (no isnad attestation) members of the graph? Split P2 out of #352
> **This issue is an owner decision, not an implementation task.** It asks what a *narrator* is when the only surviving attestation is biographical. Nothing should be built here until that is answered.

## The decision, in one line

**`queries/validation/orphan_narrators.cypher`'s `// Expected: 0` contract is correct for one population of zero-degree narrators and wrong for another.** We have been treating both as one bug (#352), which is why it has stalled.

## First: the expensive-sounding ruling is empirically free

The owner was asked whether a narrator attested only inside a deliberately-deduplicated edition is a real narrator, and answered *"I feel like it is"* — a soft yes, with acknowledged uncertainty about the cost.

**The cost is approximately zero, and the question turned out not to arise.** `is_canonical_hadith` (da#191) drops `fawaz`'s six-books copies as duplicates of the `lk` spine. Re-keying on Arabic letters + whitespace only (dropping the punctuation that the current dedup key is sensitive to), **30,790 of 34,532 fawaz six-books rows — 89.2% — are an exact full-text match to an `lk` row.** Same book, same isnad, same matn, different orthography.

Of the 3,359 rows with no exact twin, pushed through a 5-gram shingle index over `lk`:

| best shingle overlap vs `lk` | rows |
|---|---|
| ≥ 0.9 | 445 (13.2%) |
| ≥ 0.5 | 2,612 (77.8%) |
| ≥ 0.2 | 3,264 (97.2%) |
| **< 0.05 — no shared 5-gram, plausibly a tradition `lk` lacks** | **4** |

*Instrument check:* the same probe scores mean **1.000** on 200 known exact twins. It can return a high number; the low numbers are real.

**So honouring the ruling on the dropped-edition question costs at most ~4 traditions and zero attested narrators.** The dropped editions are duplicates and were correctly dropped. That question is settled and cheap.

**The question that actually costs something is the one below**, and it was hiding behind the first.

## The two populations

Simulating the loader's degree sources against today's artifact — `NARRATED` for each kept hadith's position-0 narrator, `TRANSMITTED_TO` for each non-self-loop adjacent pair on a gate-passing hadith, `STUDIED_UNDER` for name-resolved network endpoints:

| | count |
|---|---|
| `narrators_canonical` | 129,234 |
| with degree | 71,501 |
| **zero-degree** | **57,733** |

Decomposed:

| | count | what it is | treatment |
|---|---|---|---|
| **P1** attested only on dropped (`fawaz`) hadiths | 14,963 | matn-sentence fragments — **not narrators** | **data defect.** Fix at `parse` (#369). No ruling needed. |
| **P2** never mentioned in any isnad | **42,770** | bio-only records from `itqan` / `kaggle` / `muhaddithat`, no studentship edge | **← this issue. Owner's call.** |
| **P3** mentioned on a kept hadith yet still zero-degree | 0 | — | — |

(#352's title claims 70,855 in "three distinct populations." From today's artifact there are **two**, and P3 is empty. The delta is artifact-version skew: `narrator_mentions_resolved.parquet` was rewritten after the measured graph was loaded.)

**P1 is garbage** — 78.2% have `mention_count == 1`, 23.9% of the names exceed five tokens, and the highest-count ones are literal matn sentences. Deleting them needs no ruling.

**P2 is a real question.** These are people with names, dates, generations, trustworthiness gradings and sect affiliations, sourced from biographical dictionaries. We simply have no surviving isnad in our corpus in which they transmit, and no studentship edge that names them. Of the 22,908 `STUDIED_UNDER` endpoint ids in `network_edges_*`, only 11,852 resolve to a real canonical node — so name-keyed studentship rescues far fewer than one might hope.

## What is being asked

**Is a narrator with a biography but no attested transmission a member of the isnad graph?**

Three defensible answers, each with a different consequence:

1. **Yes — load them.** They are real historical people; the absence of an isnad is our corpus's incompleteness, not theirs. `orphan_narrators.cypher` must then be relaxed for P2 (and *only* for P2). Consequence: ~42,770 zero-degree nodes, reachable by search and narrator pages, contributing nothing to centrality, PageRank or graph traversal. They will look like dead ends in the UI.
2. **No — do not load a Narrator node without at least one edge.** Consequence: we discard sourced biographical data we deliberately acquired (the `itqan` source exists for this purpose — see the licensing decision recorded for it), and the narrator search index shrinks by a third.
3. **Yes, but mark them.** Load with an explicit property (e.g. `attestation: "biographical_only"`), let the UI present them as biography-only entries, and scope `orphan_narrators.cypher` to `attestation <> "biographical_only"`. Consequence: a schema change and a UI decision, but the validation contract becomes meaningful again and no sourced data is lost.

I would recommend (3), but this is genuinely a product question about what the graph is *for*, not a data-architecture question, and it should not be decided by whoever happens to be fixing #352.

## Why this must be split out of #352

#352 currently conflates P1 and P2 under a single failing validation query. As long as it does:

- `orphan_narrators.cypher` stays red regardless of how much real pollution we fix, so our one contract over narrator connectivity is uninformative and **the next genuine orphan regression is invisible**;
- any attempt to make the query pass will be tempted to relax it globally, which would silently bless P1's 14,963 matn fragments;
- and the owner's ruling gets applied to a population it was never asked about.

`// Expected: 0` is right for P1 and wrong for P2. Splitting them is the whole fix to the validation contract.

## Provenance

Analysis in the revised design note on #352 (tracking noorinalabs-main#928). Companion issues: #364, #365, #366, #369.

**Stability:** the 89.2% / 4-unique-traditions figures are computed from raw text and are **stable under #356**. The P1/P2 sizes (14,963 / 42,770 / 57,733) are keyed on `canonical_narrator_id` and **will move** when #356 re-keys the narrator set — the *shape* of the decision will not, but re-measure before acting on the counts.

parametrization:
## What

Implements the owner's **"tag & keep"** ruling on #370: a narrator attested only by a biographical source, with no surviving isnad transmission, is a real member of the graph and is kept — now **marked** so it's distinguishable from a genuine orphan and presentable by the UI as a biography-only entry.

Adds an `attestation` tag to the canonical narrator:
- `biographical_only` when `mention_count == 0` (a bio-promoted record — `bio_promote` already stamps `mention_count = 0`, "a bio is provenance, not a chain hit")
- `isnad_attested` otherwise

## Design — one shared helper, re-derived at every build site

`derive_attestation` (new `src/resolve/attestation.py`) is a pure function of `mention_count`, re-derived from each row's **final** `mention_count` at all three canonical-table build sites: `disambiguate` (mint), `bio_promote` (merge), and `fuzzy_cluster._merge_records` (id-collapse, which **sums** `mention_count`). This mirrors exactly how `derive_sect_affiliation` is already re-derived at those same three sites, so a value one producer writes can never go stale when a later merge changes `mention_count` (e.g. a bio-only record folded into an attested one → survivor re-derives to `isnad_attested`). Every other stage that rewrites `narrators_canonical.parquet` (date_reconcile / geography / …) preserves the column automatically via the schema-driven read-all/write-all idiom.

The loader maps it onto the Narrator node with an explicit `SET` (a legacy parquet lacking the column defaults to `unknown`). `orphan_narrators.cypher` now excludes `attestation = 'biographical_only'`, so its `// Expected: 0` contract is meaningful again — it catches a real orphan regression without flagging the kept bio-only nodes.

## No deploy change (confirmed)

The narrator prune (`prune-narrators` / deploy `graph-prune-narrators.yml`) keeps by **canonical-set membership**, not degree — its docstring explicitly rejects a degree predicate *because* it would wrongly delete these legitimate zero-degree canonical narrators. Bio-only narrators are canonical rows, so they already survive the prune regardless of degree. This PR does not touch that path, and no deploy edit is needed. (Scoping detail from the #370 investigation.)

## Files

- `src/models/enums.py` — `Attestation` StrEnum (`isnad_attested` / `biographical_only` / `unknown`)
- `src/resolve/attestation.py` — **new** `derive_attestation` helper
- `src/resolve/schemas.py` — `attestation` column on `NARRATORS_CANONICAL_SCHEMA`
- `src/resolve/{disambiguate,bio_promote,fuzzy_cluster}.py` — derive at each build site
- `src/graph/load_nodes.py` — map onto the Narrator node
- `queries/validation/orphan_narrators.cypher` — exclude bio-only

## Tests

`test_attestation.py` (helper), plus assertions that a bio-promoted narrator is tagged `biographical_only`, a disambiguated one `isnad_attested`, a fuzzy merge **re-tags** a folded bio-only survivor as `isnad_attested`, the property lands on the Narrator node (with the `unknown` legacy default), and the orphan query excludes bio-only. Full non-integration suite green (2256 passed); ruff + mypy-strict green.

## Notes

- **Count is not asserted in code.** The tag derives per-row from `mention_count`, so it's population-size-agnostic. The #370 analysis quoted ~42,770 P2 narrators from an earlier artifact; those counts move under da#356 re-key and should be verified at reload — nothing here depends on the number.
- **Not blocked on da#369.** The `mention_count == 0` signal cleanly excludes the P1 matn-fragment population (they carry `mention_count >= 1`), so the tag is correct with or without #369.
- The option-(3) **UI presentation** of bio-only entries is an isnad-graph concern, tracked separately.

Closes #370

https://claude.ai/code/session_01CrsyDcQb1oGgFEVBacZzvs

### DIFF
--- a/queries/validation/orphan_narrators.cypher
+++ b/queries/validation/orphan_narrators.cypher
@@ -1,6 +1,12 @@
-// Find narrator nodes with no relationships
+// Find narrator nodes with no relationships, EXCLUDING biography-only narrators.
 // Expected: 0 orphans in a healthy graph
+// A `biographical_only` narrator (da#370) is a real person carried by a rijal /
+// biographical source with no surviving isnad mention. Such narrators are kept on
+// purpose ("tag & keep") and are zero-degree BY DESIGN, so they are not orphans —
+// excluding them keeps this contract meaningful for a genuine orphan regression.
+// coalesce guards a legacy node loaded before the attestation property existed.
 MATCH (n:Narrator)
 WHERE NOT (n)--()
+  AND coalesce(n.attestation, '') <> 'biographical_only'
 RETURN n.id AS narrator_id, n.name_en AS name
 ORDER BY n.id
--- a/src/graph/load_nodes.py
+++ b/src/graph/load_nodes.py
@@ -34,6 +34,7 @@ from typing import Any
 import pyarrow.parquet as pq
 import yaml
 
+from src.models.enums import Attestation
 from src.parse.composition import (
     canonical_matn_identity,
     is_canonical_hadith,
@@ -226,6 +227,7 @@ SET n.name_ar           = row.name_ar,
     n.external_id       = row.external_id,
     n.source_ids        = row.source_ids,
     n.mention_count     = row.mention_count,
+    n.attestation       = row.attestation,
     n.source_corpus     = row.source_corpus,
     n.source_corpora    = row.source_corpora,
     n.sect_affiliation  = row.sect_affiliation
@@ -287,6 +289,11 @@ def _load_narrators(
                 # ``MATCH (n:Narrator) WHERE any(s IN n.source_ids WHERE s STARTS WITH 'itqan:')``.
                 "source_ids": _val(row, "source_ids", []),
                 "mention_count": _val(row, "mention_count"),
+                # Attestation tag (da#370). Defaults to "unknown" for a legacy
+                # canonical file written before this column existed, so the property
+                # is always present on the node (mirrors source_corpus/sect_affiliation).
+                # Every current producer populates it as isnad_attested/biographical_only.
+                "attestation": _val(row, "attestation", Attestation.UNKNOWN.value),
                 # Sect/corpus provenance (da#103). ``source_corpus`` defaults to ""
                 # so the property is always present even for legacy canonical files
                 # written before these columns existed; ``sect_affiliation`` defaults
--- a/src/models/enums.py
+++ b/src/models/enums.py
@@ -83,6 +83,30 @@ class SectAffiliation(StrEnum):
     UNKNOWN = "unknown"
 
 
+class Attestation(StrEnum):
+    """How a canonical narrator is attested in the corpus (da#370).
+
+    Distinguishes a narrator we have an isnad transmission for from a
+    ``biographical_only`` record — a real historical person carried by a rijal /
+    biographical source (itqan / kaggle / muhaddithat) with **no** surviving isnad
+    mention. Bio-only narrators are deliberately kept ("tag & keep", owner ruling on
+    da#370): they are real, searchable people whose absence of an isnad is our
+    corpus's incompleteness, not theirs. The tag lets the UI present them as
+    biography-only entries and lets ``orphan_narrators.cypher`` exclude them so the
+    zero-degree contract stays meaningful for genuine orphans.
+
+    Derived from ``mention_count`` (see :func:`src.resolve.attestation.derive_attestation`):
+    a bio is provenance, not a chain hit, so a bio-promoted record carries
+    ``mention_count == 0``. ``UNKNOWN`` is not a derived value — every canonical
+    narrator has a known ``mention_count`` — it exists only as the loader's default
+    for a legacy canonical parquet written before this column existed.
+    """
+
+    ISNAD_ATTESTED = "isnad_attested"
+    BIOGRAPHICAL_ONLY = "biographical_only"
+    UNKNOWN = "unknown"
+
+
 class TrustworthinessGrade(StrEnum):
     """Narrator trustworthiness grade from classical rijal criticism."""
 
--- a/src/resolve/attestation.py
+++ b/src/resolve/attestation.py
@@ -1,0 +1,50 @@
+"""Derive a narrator-level :class:`Attestation` tag from its isnad mention count (da#370).
+
+A canonical narrator is ``biographical_only`` when it carries **no** isnad
+transmission mention — a record promoted from a rijal / biographical source
+(itqan / kaggle / muhaddithat) rather than seen transmitting in a chain. The signal
+is already present on the canonical record: ``bio_promote`` stamps ``mention_count = 0``
+on a bio-promoted narrator ("a bio is provenance, not a chain hit"), while a narrator
+disambiguated from real chain mentions carries ``mention_count >= 1``.
+
+Why this is derived at every canonical-table build site, not once
+----------------------------------------------------------------
+``narrators_canonical.parquet`` has more than one producer that (re)computes
+``mention_count``: ``disambiguate`` mints rows from mentions, ``bio_promote`` MERGEs
+bio rows in (``mention_count = 0``), and ``fuzzy_cluster`` collapses ids and **sums**
+the survivors' ``mention_count``. A tag computed once and carried across those steps
+would go stale exactly when a merge changes ``mention_count`` (a bio-only record
+folded into an attested one). So — mirroring :func:`derive_sect_affiliation`, which
+is likewise re-derived from the merged ``source_corpora`` at each of those sites — the
+attestation is re-derived from the row's **final** ``mention_count`` wherever the
+canonical table is built. It is a pure function of that one field, so re-deriving is
+always consistent and never depends on producer ordering.
+
+Note the tag is about *isnad* attestation, not graph degree: a ``biographical_only``
+narrator whose name later resolves to a ``STUDIED_UNDER`` network endpoint at load
+time still has no isnad transmission, so the tag stands. It is not the same predicate
+as "zero-degree" — that is why ``orphan_narrators.cypher`` excludes bio-only nodes
+explicitly rather than inferring bio-only from degree.
+"""
+
+from __future__ import annotations
+
+from src.models.enums import Attestation
+
+
+def derive_attestation(mention_count: object) -> str:
+    """Return the :class:`Attestation` value for a canonical narrator.
+
+    ``isnad_attested`` when the narrator carries a positive isnad ``mention_count``,
+    else ``biographical_only`` (a bio-promoted record with ``mention_count`` 0 or
+    ``None``). Never returns ``unknown``: a canonical narrator always has a known
+    ``mention_count``. Returns the enum *value* (a ``str``) so it drops straight into
+    the ``NARRATORS_CANONICAL_SCHEMA`` string column.
+
+    Accepts ``object`` because the canonical-row dicts in the resolve layer type this
+    field loosely; only a positive ``int`` counts as attested (mirrors the
+    ``mc if isinstance(mc, int)`` guard the fuzzy-cluster merge already uses).
+    """
+    if isinstance(mention_count, int) and mention_count > 0:
+        return Attestation.ISNAD_ATTESTED.value
+    return Attestation.BIOGRAPHICAL_ONLY.value
--- a/src/resolve/attestation.py
+++ b/src/resolve/attestation.py
@@ -11,14 +11,22 @@ Why this is derived at every canonical-table build site, not once
 ----------------------------------------------------------------
 ``narrators_canonical.parquet`` has more than one producer that (re)computes
 ``mention_count``: ``disambiguate`` mints rows from mentions, ``bio_promote`` MERGEs
-bio rows in (``mention_count = 0``), and ``fuzzy_cluster`` collapses ids and **sums**
-the survivors' ``mention_count``. A tag computed once and carried across those steps
-would go stale exactly when a merge changes ``mention_count`` (a bio-only record
-folded into an attested one). So — mirroring :func:`derive_sect_affiliation`, which
-is likewise re-derived from the merged ``source_corpora`` at each of those sites — the
-attestation is re-derived from the row's **final** ``mention_count`` wherever the
-canonical table is built. It is a pure function of that one field, so re-deriving is
-always consistent and never depends on producer ordering.
+bio rows in (``mention_count = 0``), ``fuzzy_cluster`` collapses ids and **sums** the
+survivors' ``mention_count``, and ``narrator_split`` (da#337) reduces a generic
+primary's count and mints peeled band rows. A tag computed once and carried across
+those steps would go stale exactly when one of them changes ``mention_count`` (a
+bio-only record folded into an attested one; a primary peeled down to zero). So —
+mirroring :func:`derive_sect_affiliation`, which is likewise re-derived from the
+merged ``source_corpora`` at each of those sites — the attestation is re-derived from
+the row's **final** ``mention_count`` at every one of those four sites. It is a pure
+function of that one field, so re-deriving is always consistent and never depends on
+producer ordering.
+
+The remaining canonical writers in ``RESOLVE_STEP_ORDER`` — ``date_reconcile`` and
+``tabaqa_dates`` — carry the tag through untouched, and that is safe **only because
+they never recompute ``mention_count``**: they round-trip the schema and edit date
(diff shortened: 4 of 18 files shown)


parametrization:
Requestor: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

Reviewed at head 1edccfe against the code, on the plumbing/contract axis. All four points hold.

Loader maps the property correctly. The narrator MERGE adds an explicit n.attestation = row.attestation to its SET list, which is required because the loader builds an explicit per-field batch dict rather than doing n += row. Legacy resolves to unknown without a crash: the batch sets attestation to _val(row, 'attestation', Attestation.UNKNOWN.value), and _val is v = row.get(key); return default if v is None else v — so both an absent column (row.get -> None) and a present-but-null value fold to the default 'unknown'. _read_parquet_rows yields dicts, so a missing column is a missing key, not a read error. The two loader tests cover the set-through and the legacy-default paths.

orphan_narrators.cypher scoping is correct and complete. The predicate is WHERE NOT (n)--() AND coalesce(n.attestation,'') <> 'biographical_only'; coalesce guards a legacy node loaded before the property existed (it stays eligible as a genuine orphan, the conservative call). This is a full fix, not a half one, because _classify_orphan_narrators does only count = len(rows); passed = count == 0 — it counts whatever the query returns and does NOT re-implement the zero-degree predicate independently, so scoping the cypher is sufficient.

The Attestation StrEnum and schema column are wired consistently. isnad_attested / biographical_only / unknown are spelled identically across enums.py, derive_attestation (which returns the enum .value), the load_nodes default (Attestation.UNKNOWN.value), the schema field comment, and the cypher literal 'biographical_only' (Cypher cannot import the enum, so the literal must match, and it does). The three writer sites all route through the single derive_attestation helper on the row's FINAL mention_count — bio_promote (0 -> biographical_only), disambiguate (chain mentions -> isnad_attested), and fuzzy_cluster (re-derived from the SUMMED count so a folded bio-only survivor correctly flips to isnad_attested). Column is nullable, matching the loader default contract. Tests assert value equality at each site including the merge-retag case.

No deploy change is needed. Verified against src/graph/prune.py (da#413): the only orphan predicate is canonical-set membership (orphans = graph - canonical), and a degree/zero-degree predicate is explicitly rejected there because it would wrongly delete the 57,993 zero-degree-yet-canonical narrators. Bio-only narrators are written into narrators_canonical.parquet by bio_promote, so they are in the keep-set and already survive the prune regardless of zero degree. The claim checks out; no phantom deploy edit.

parametrization:
TechDebt: none

Reviewed at head 1edccfeb, reading each file at that sha (no parent checkout). All four points in the brief hold — but the "re-derive at every canonical build site" invariant has a gap that lets the tag go stale, so I'm requesting a change before merge.

**What checks out**
- The three tagged sites (disambiguate, bio_promote, cluster) each re-derive through the shared derive_attestation, sitting right next to the derive_sect_affiliation call they mirror — verified at head, not from diff framing. Derive-once-and-carry is correctly avoided at these three.
- The fuzzy_cluster case is genuinely load-bearing and its test is non-vacuous. In _merge_cluster the survivor dict is built fresh and never copies any member's attestation (it is not in the scalar back-fill tuple); the sole assignment is derive_attestation over the SUMMED count. So re-derivation is the only path to the tag, and the test is guarded by merged_records == 1 plus a summed mention_count of 5 (5+0) and len == 1 — it drives the collapse-and-sum and cannot pass unless the fold fires and re-tags.
- load_nodes SETs the attestation property from an explicit column with a legacy "unknown" default; the Attestation StrEnum and the nullable schema column are right.
- orphan_narrators.cypher excludes biographical_only (coalesce-guarded), and _classify_orphan_narrators is just count = len(rows) with no independent predicate, so scoping the query is sufficient.

**The blocking gap — narrator_split rewrites the canonical table and changes mention_count without re-deriving**
narrators_canonical.parquet has six writers, not three. RESOLVE_STEP_ORDER runs disambiguate, bio_promote, cluster, then narrator_split, reconcile, tabaqa_dates before the graph load. reconcile and tabaqa_dates are safe — they round-trip the full schema and never touch mention_count, so a carried tag stays valid. narrator_split is not:
- It reduces a primary's count with primary_rec["mention_count"] = max(0, primary_count - plan.peeled_mention_count) (the max(0, ...) clamp shows 0 is reachable) but leaves the primary's attestation untouched, so a primary peeled down to 0 mentions stays isnad_attested when it should now read biographical_only.
- _peeled_record builds each new band row from a fixed column tuple that omits attestation, so every peeled row is written with a null tag and loads as "unknown" — even though it carries band.mention_count isnad mentions and is genuinely isnad_attested.

Both writes land after the last derivation site (cluster), and reconcile/tabaqa_dates carry them verbatim to the loader, so the stale values reach the graph. This is the exact staleness the design exists to prevent, and it contradicts attestation.py's own docstring ("re-derived from the row's final mention_count wherever the canonical table is built") — narrator_split builds the table and does not. narrator_split (da#337) is a live stage, so this fires on any run that peels a generic node.

**Requested change**
Make narrator_split the fourth derivation site, mirroring the established pattern — set the attestation inside _peeled_record from derive_attestation(band.mention_count), and set it on the primary from derive_attestation of its reduced count right after the subtraction. Add one non-vacuous test: a primary whose mentions are fully peeled (count to 0) re-tags to biographical_only, and a peeled band row is isnad_attested. reconcile/tabaqa_dates need no code change; a one-line note in the attestation docstring that carry-through there is safe only because they do not recompute mention_count would keep the next writer honest.

Everything else is solid — enum, schema, loader, cypher, and the fuzzy re-tag are all correct and well-documented. Happy to re-review the moment narrator_split re-derives.

parametrization:
Fixup pushed at ce474a1 addressing Nikolaos's ChangesRequested (the head moved from 1edccfe, so both reviews need to re-confirm here).

**The gap Nikolaos caught (correct + blocking):** `narrators_canonical.parquet` has a **fourth** writer that mutates `mention_count` after the last derivation site. `RESOLVE_STEP_ORDER` is disambiguate → bio_promote → cluster → **narrator_split** → reconcile → tabaqa_dates. `narrator_split` (da#337) both reduces the primary's count (`max(0, count - peeled)`) and mints peeled band rows from a fixed column tuple that omits `attestation` — so a fully-peeled primary kept a stale `isnad_attested`, and every peeled band loaded as the loader's `unknown` default despite carrying real isnad mentions.

**Fix** (same pattern — `derive_attestation` on the FINAL count, as at the other three sites):
- `_peeled_record`: `attestation = derive_attestation(band.mention_count)`
- apply loop: re-derive right after the `max(0, …)` subtraction

`reconcile` / `tabaqa_dates` are unchanged and safe — they round-trip the schema and never recompute `mention_count`; I documented that invariant in `attestation.py` so the next writer stays honest.

**Test** `test_stage_peel_re_derives_attestation` drives a real peel (two equal 50-mention bands; recorded count 50 both screens the name in as generic and equals the peel, so the primary clamps to 0) and asserts the peeled band is `isnad_attested` and the zeroed primary is `biographical_only`. It bites: the input rows carry `attestation=None`, so removing either derivation fails the exact-value assertion.

Green: ruff + mypy-strict (src/, 105 files) + full non-integration suite (2257 passed) via pre-commit/pre-push hooks.

parametrization:
Requestor: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

Re-review at head c1ce2f2 (my prior Approved was at 1edccfe). My four plumbing/contract points are unchanged at this head and still hold — the cypher scoping, the explicit loader SET n.attestation with the _val 'unknown' legacy default, the enum/schema value consistency, and the no-deploy-change conclusion against src/graph/prune.py (canonical-set membership keeps bio-only rows). The only new surface is narrator_split as the 4th derive_attestation site, and it is consistent with all of them.

The narrator_split addition is correct and matches the re-derive-on-final-count contract. _peeled_record derives attestation from band.mention_count: a peeled band carries real isnad mentions so it resolves to isnad_attested, and because the record starts as an all-None schema dict and the inherited-column tuple deliberately omits attestation, this line is what stops an attested band from loading as the loader's 'unknown' default — a real gap without it. In the apply loop the primary is reduced by primary_rec['mention_count'] = max(0, primary_count - plan.peeled_mention_count) and attestation is re-derived on that mutated value on the next line, in the correct order (after the mutation), so a primary whose whole recorded count peels off clamps to 0 and correctly re-tags biographical_only rather than carrying a stale isnad_attested. _peeled_record inherits only name/provenance columns from the primary, never mention_count/attestation, so the primary's mutation does not leak into the band. Both sites write the enum .value string into the nullable NARRATORS_CANONICAL_SCHEMA attestation column via the single helper, identical to the other three producers.

The new test bites both sites end-to-end: it sets the recorded count equal to the single peeled band so the max(0, ...) reduction lands the primary at 0, then asserts peeled -> isnad_attested (count 50) and primary -> biographical_only (count 0). Drop either re-derivation and it fails (the peeled row would be the None/unknown default; the primary would keep a stale isnad_attested). The attestation.py docstring now lists narrator_split (da#337) alongside the other build sites, so the single-source-of-truth invariant note is accurate. Still good to merge on my axis.

parametrization:
RequestOrReplied: Approved
TechDebt: none

Re-reviewed at head c1ce2f25 (post update-branch), reading each file at that sha. The narrator_split staleness gap is fully closed — approving.

**Fix verified — narrator_split is now the fourth derivation site**
- _peeled_record sets `rec["attestation"] = derive_attestation(band.mention_count)` after the fixed column tuple (which still omits attestation), so a peeled band with real isnad mentions is isnad_attested rather than the loader's "unknown" default.
- The apply loop sets `primary_rec["attestation"] = derive_attestation(primary_rec["mention_count"])` immediately after the `max(0, primary_count - peeled)` reduction, so a primary peeled to 0 re-tags to biographical_only. primary_rec is the in-place record_by_id reference that flows into `records`, so the mutation reaches the written table.
- All four canonical writers that (re)compute mention_count now derive: disambiguate (929), bio_promote (414), fuzzy_cluster (1177), narrator_split (628 peeled + 736 primary). reconcile/tabaqa_dates remain schema-round-trip only and never touch mention_count, so carry-through there stays valid — and that is now documented.

**The new test is non-vacuous**
test_stage_peel_re_derives_attestation constructs the recorded primary count equal to the peeled total, so `max(0, 50-50)` lands the primary at mention_count 0 — the exact branch I flagged. Crucially, `_canonical_row` seeds every field from `{f.name: None for f in SCHEMA}` and the test never overrides attestation, so the input rows carry attestation=None. Remove either derivation and the value stays None, failing the exact-value asserts (`peeled[0]["attestation"] == "isnad_attested"`, `by_id[cand]["attestation"] == "biographical_only"`). It drives the collapse-and-reduce path, not a stand-in.

**No regression**
The three original derivation sites, the Attestation enum, the nullable schema column, the loader SET + "unknown" default, and the coalesce-guarded cypher exclusion are all unchanged and correct at head. bio_promote's derivation line moved only because the merged da#385 counters now sit in the same finalize loop — the two coexist cleanly.

Clean, complete fix. Approved.

Issue #444, opened 2026-07-12
resolve: date-axis narrator_split Gate 3 (SPLIT_MIN_SEPARATION) is unreachable dead code


parametrization:
## Summary

The date-axis narrator splitter's third over-split guard (Gate 3, the `SPLIT_MIN_SEPARATION` "adjacent bands too close ⇒ abstain" check in `src/resolve/narrator_split.py::plan_split`) is **unreachable dead code**. It can never fire given the current constant relationship, so the abstain branch and its `gate3_unseparated` reason are never taken.

## Root cause

Two thresholds are in tension:

- `SPLIT_BAND_GAP = 80` — `_cut_bands` starts a new death-band whenever a consecutive estimate gap exceeds 80 AH. So any two distinct bands are separated by a >80 AH gap.
- `SPLIT_MIN_SEPARATION = 50` — Gate 3 abstains when two adjacent qualifying-band **midpoints** are `<= 50` AH apart.

Because two bands only exist when separated by a >80 gap, `min(next_band) > max(prev_band) + 80`, and each midpoint lies within its band's range, so adjacent midpoints are always `> 80` apart — never `<= 50`. Gate 3's condition is therefore always False. Since `SPLIT_BAND_GAP (80) > SPLIT_MIN_SEPARATION (50)`, the guard is structurally unreachable.

## Evidence

Empirically swept a full two-band grid through `plan_split` (band pairs across a wide midpoint range): the only outcomes observed were a clean split or `gate1_single`. `gate3_unseparated` was never returned. Two contemporary bands collapse into a single band (one `_cut_bands` band ⇒ Gate 1), and two separated bands split cleanly — neither path reaches Gate 3.

## Impact

Latent only — no incorrect output. The two live guards (Gate 1 single-band, Gate 2 too-many-bands) carry the real over-split protection, so nothing splits that shouldn't. The dead branch is misleading to a maintainer (it reads as an active contemporaries guard) and any code keying on the `gate3_unseparated` reason will never see it.

Found while reading `plan_split` for the #337/#346 context-discriminator work; independent of that split direction (which the owner has flagged rather than shipped).

## Options (pick at fix time; no fix inline now)

1. **Remove Gate 3** and its `gate3_unseparated` reason as dead code, documenting that Gate 1 + Gate 2 are the complete guard set.
[...]

PR #449, opened 2026-07-18
resolve: remove unreachable Gate 3 (SPLIT_MIN_SEPARATION) dead code (da#444)

parametrization:
## Summary

Removes the date-axis `narrator_split` **Gate 3** (`SPLIT_MIN_SEPARATION = 50`) as
**unreachable dead code** (da#444). The guard — "adjacent qualifying-band midpoints too
close ⇒ abstain" in `plan_split` — can never fire given the current constant
relationship, so its abstain branch is never taken.

## Root cause

Two thresholds are in tension:

- `SPLIT_BAND_GAP = 80` — `_cut_bands` starts a new death-band whenever a consecutive
  estimate gap exceeds 80 AH, so any two distinct bands are separated by a > 80 AH gap.
- `SPLIT_MIN_SEPARATION = 50` — Gate 3 abstained when two adjacent qualifying-band
  **midpoints** were `<= 50` AH apart.

Because two bands only exist when separated by a > 80 gap, `min(next) > max(prev) + 80`
and each midpoint lies inside its band's range, so adjacent midpoints are always `> 80`
apart — never `<= 50`. Since `SPLIT_BAND_GAP (80) > SPLIT_MIN_SEPARATION (50)`, the
guard is structurally unreachable.

## Evidence

Swept two well-supported bands through the real `plan_split` across every separation
0..300. Outcomes were **only** a single-band Gate 1 abstain (separation <= 80) or a
clean split; the minimum adjacent qualifying-band midpoint separation ever observed was
**81**. `gate3` never determined an outcome. Verified the instrument separates the
classes (Gate 1 vs split) before reading the null result.

## Change (Option 1 — remove)

- Remove the Gate 3 block and the `SPLIT_MIN_SEPARATION` constant (+ its `__all__`
  entry). Documented that Gate 1 (single-band) + Gate 2 (too-many-bands) are the
  **complete** guard set, and why a separation gate is structurally unnecessary given
  `SPLIT_BAND_GAP > 50`.
- Chose Option 1 over Option 2 (re-enable by raising the threshold above
  `SPLIT_BAND_GAP`): that would change split behavior on adjacent-but-separated bands
  and needs its own evidence, which does not exist. No product reason to abstain there.

## Impact

Latent only — **no behavior change**. Gate 1 + Gate 2 carry the real over-split
protection; nothing that split before splits differently now.

## Tests

- New `test_removed_gate3_separation_is_dead_code`: sweeps 0..300 and asserts two
  qualifying bands are always `> SPLIT_BAND_GAP` apart and the only outcomes are Gate 1
  abstain or split — no separation-based abstain.
- `pytest tests/test_resolve/test_narrator_split.py` → 19 passed.
- ruff format + lint clean; mypy strict clean; full pre-push green.

## Reviewers

- Peer: **Nikolaos Papadopoulos**

Closes #444
Part of main#928

### DIFF
--- a/src/resolve/narrator_split.py
+++ b/src/resolve/narrator_split.py
@@ -47,8 +47,15 @@ Algorithm (per eligible canonical id ``C``, name ``N``)
 4. **ABSTAIN — leave ``C`` as ONE node — UNLESS ALL hold** (the over-split guard):
    * ≥2 clusters each with ≥ :data:`SPLIT_MIN_SUPPORT` mentions (*qualifying*);
    * ≤ :data:`SPLIT_MAX_CLUSTERS` qualifying clusters (more ⇒ the name is a generic
-     bucket we cannot cleanly resolve → abstain, logged for audit);
-   * adjacent qualifying-cluster midpoints separated by > :data:`SPLIT_MIN_SEPARATION`.
+     bucket we cannot cleanly resolve → abstain, logged for audit).
+
+   A separate "adjacent bands too close ⇒ abstain" separation guard is **not** needed:
+   :func:`_cut_bands` only ever starts a new band on a consecutive gap exceeding
+   :data:`SPLIT_BAND_GAP` (80 AH), so any two distinct bands — and therefore any two
+   qualifying bands — are already > 80 AH apart by construction, well beyond any
+   contemporaries-scatter threshold. (A ``SPLIT_MIN_SEPARATION = 50`` gate once sat here;
+   because ``SPLIT_BAND_GAP > 50`` it was structurally unreachable and was removed as
+   dead code — da#444.)
 5. **Peel-not-partition** — undatable mentions, and any mention in a below-support
    band, STAY on the primary node. Only confidently-dated distinct clusters peel off.
 6. **Id assignment** (deterministic — a pure function of the input, reproducible
@@ -147,7 +154,6 @@ logger = get_logger(__name__)
 __all__ = [
     "SPLIT_BAND_GAP",
     "SPLIT_MIN_SUPPORT",
-    "SPLIT_MIN_SEPARATION",
     "SPLIT_MAX_CLUSTERS",
     "MID_GAP",
     "NARRATOR_SPLITS_SCHEMA",
@@ -176,10 +182,6 @@ SPLIT_BAND_GAP = 80
 # Below it, a band is left on the primary (peel-not-partition) — a handful of
 # adjacency estimates is not enough to mint a distinct historical person.
 SPLIT_MIN_SUPPORT = 10
-
-# Two qualifying bands whose midpoints are closer than this are treated as one
-# referent's scatter, not two people — abstain. Secondary to SPLIT_BAND_GAP.
-SPLIT_MIN_SEPARATION = 50
 
 # More than this many *qualifying* bands means the name is a generic bucket we cannot
 # cleanly resolve into a few people — abstain and log for audit rather than shatter it.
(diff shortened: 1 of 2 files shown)


parametrization:
RequestOrReplied: Approved

**Review: Approved**

Verified the "provably dead" claim against the actual code at head, not just the PR narrative:

- `_cut_bands` (narrator_split.py:278) cuts a new band only when `m.estimate - current[-1].estimate > SPLIT_BAND_GAP` (80), so between any two adjacent bands `min(B) - max(A) > 80`.
- `_band_midpoint` (L293) is the arithmetic mean of a band's estimates, which is always within `[min, max]` of that band. So `midpoint(B) - midpoint(A) >= min(B) - max(A) > 80`.
- Adjacent *qualifying* bands are a subset of all bands, still distinct intervals, so their midpoints are still `> 80` apart. The removed Gate 3 tested `<= SPLIT_MIN_SEPARATION (50)` — with `SPLIT_BAND_GAP (80) > 50` that branch is unreachable for all inputs. Claim holds.

- `git grep SPLIT_MIN_SEPARATION` on the PR branch shows only prose/comment references remain — no live code references the removed constant, and the `__all__` entry is dropped. Nothing else imports it.
- The sweep test (`test_removed_gate3_separation_is_dead_code`, 0..300) is a real instrument: it asserts two bands form only when the requested separation `> 80`, that the observed adjacent-midpoint separation is then always `> 80` (never in the `<= 50` range), and that the only outcomes are Gate 1 abstain or a clean split — never a separation-based abstain.
- Option 1 (remove) over Option 2 (raise threshold) is the right call: re-enabling would change split behaviour on adjacent-but-separated bands with no evidence to justify it.

Latent-only, no behaviour change; Gate 1 + Gate 2 carry the over-split protection. CI 6/6 green.

Must-fix: None
TechDebt: None

Note (informational, not blocking): this PR and da#439 (PR#450) both edit `narrator_split.py`'s module docstring and `__all__`, so a textual merge conflict is likely — whichever merges second will need a trivial rebase. Flagging for wrap-up merge sequencing, not a review finding.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved

**Review: Approved**

QA / data-quality lens — dead-code removal (da#444).

The "provably dead" evidence is sound. Verified at wave-25 HEAD:
- `SPLIT_BAND_GAP = 80`, removed `SPLIT_MIN_SEPARATION = 50`.
- `_cut_bands` cuts a new band only on a consecutive gap `> SPLIT_BAND_GAP` (80 AH).
- `_band_midpoint` is the mean of a band's estimates, which lies within `[min, max]` of that band.

Midpoint argument holds generally, not just for the fixture shape: for consecutive bands, `mean(later) - mean(earlier) >= min(later) - max(earlier) = inter-band gap > 80 > 50`. So the removed `b - a <= 50` check could never fire — its removal cannot change behavior. The 0..300 separation sweep in `test_removed_gate3_separation_is_dead_code` empirically corroborates (asserts every two-band outcome has adjacent midpoints `> SPLIT_BAND_GAP`, and the only outcomes are Gate-1 single-band abstain or a clean split — never a separation abstain). No fixture can produce the removed branch, consistent with the "unrealistic fixture makes a guard assertion inert" discipline — here the guard genuinely has no reachable input.

Must-fix: None
TechDebt: The dead-code proof rests on the invariant `SPLIT_BAND_GAP (80) > the old 50 contemporaries-scatter floor`. With `SPLIT_MIN_SEPARATION` now gone, a future lowering of `SPLIT_BAND_GAP` below ~50 would silently re-open the contemporaries-scatter case the removed guard once covered, with no test asserting that relationship. Optional: a one-line comment or assert near `SPLIT_BAND_GAP` recording that bands are separation-safe only while the gap stays well above a generation. Non-blocking.

Issue #439, opened 2026-07-12
narrator_split ±1 window vs loader TRANSMITTED_TO adjacency diverge on gappy chains (PR-B input)


parametrization:
## Observed during review of PR #437 (issue #411)

PR #437 correctly aligns `narrator_split._build_chain_index` / `_collect_datable` to the loaders' `(hadith_id, chain_index)` composite chain key, removing the cross-chain neighbour fabrication. This is a separate, **pre-existing** divergence surfaced during that review — not introduced by #437 and not blocking it.

### The divergence

The splitter's ±1 neighbour window and the graph loader's TRANSMITTED_TO adjacency use two different notions of "adjacent", and they diverge on chains with position gaps:

- **Loader** (`src/graph/load_edges.py::_build_chain_pairs`, ~L185-190): sorts a chain by `position_in_chain`, filters to **resolved** mentions (non-null `canonical_narrator_id`), then pairs **consecutive resolved** mentions `resolved[i] -> resolved[i+1]` — regardless of any position gap between them.
- **Splitter** (`src/resolve/narrator_split.py::_collect_datable`): looks up neighbours at **exact** `position ± 1` (`npos == position - 1` / `position + 1`) within the same chain; its `chains` map also contains only resolved mentions (pass 2 skips null canonical).

On a contiguous chain `[0,1,2,...]` with every position resolved, the two agree. But when a chain has a **position gap** — e.g. the mention at an intermediate position was left unresolved (null `canonical_narrator_id`) and dropped, or positions are non-contiguous — the loader still creates an edge across the gap (consecutive-in-resolved-list), while the splitter's exact-±1 window finds **no** neighbour across the gap. The splitter therefore **under-counts** real teacher/student adjacencies that exist in the loaded graph.

### Why this matters for PR-B (#337/#346)

PR-B's isnad-neighbour context discriminator is designed to cluster on the same teacher/student adjacency the graph carries. If PR-B inherits the splitter's exact-±1 window, its adjacency signal will diverge from the loaded TRANSMITTED_TO topology on gappy chains — a systematic under-count (not fabrication). Per the drop-gate bidirectional lesson (main#928 / da#423), a narrator that appears only in gappy chains could be systematically erased from PR-B's context clusters, not merely under-weighted.

### Suggested resolution

[...]

PR #450, opened 2026-07-18
fix(resolve): reconcile narrator_split adjacency with loader TRANSMITTED_TO (da#439)

parametrization:
## What & why

Closes #439.

`narrator_split`'s death-band evidence and the graph loader used two different notions of chain adjacency, diverging on **gappy chains**:

- **Loader** (`load_edges._build_chain_pairs`): sorts a chain by `position_in_chain`, filters to **resolved** mentions, and pairs **consecutive resolved** mentions — bridging any position gap.
- **Splitter** (`narrator_split._collect_datable`): matched neighbours at an **exact `position ± 1`**.

When an intermediate position is unresolved (null `canonical_narrator_id`, dropped from the split's chain index), the loader still emits a `TRANSMITTED_TO` edge across the gap, but the exact-±1 window found nothing and left the narrator **undatable**. A narrator appearing only in gappy chains was therefore systematically **erased** from the split's evidence relative to the loaded graph — the drop-gate erasure class (main#928 / da#423), and a PR-B (#337/#346) input concern since PR-B's isnad-neighbour discriminator clusters on that same adjacency.

## Change

- `_collect_datable` now reads a mention's **immediately-preceding / -following resolved** neighbour in its position-sorted `(hadith_id, chain_index)` chain — the loader's consecutive-resolved notion — instead of exact ±1. It mirrors the loader's **self-loop drop** (a neighbour resolving to the mention's own id forms no edge and is not bridged past).
- `_build_chain_index` position-sorts each chain once (deterministic `(position, cid)` order) so the lookup is by list index.
- The composite-key cross-chain confinement (da#411) and the sign convention (teacher = earlier position `+MID_GAP`, student = later position `−MID_GAP`) are unchanged.

## Tests (agreement, not just a fixture)

- `test_gappy_chain_datable_across_dropped_position` — chain with positions 0/2/4 resolved and 1/3 dropped; **RED** under exact-±1 (candidate undatable, verified), **GREEN** now (datable across both gaps).
- `test_split_adjacency_matches_loader_transmitted_to` — runs **both** `_build_chain_pairs` (loader) and the split path on one gappy fixture and asserts the candidate's split neighbours equal the loader's `TRANSMITTED_TO` edges incident to it. This exercises the reconciled adjacency against the real loader function, not a splitter-internal fixture — the da#439 invariant.

## Gates
- Full resolve + graph suites green (`857 passed, 6 skipped`); narrator_split file `20 passed`.
- ruff lint + format, mypy strict: clean. Pre-commit + pre-push hooks passed (cspell skipped — `.py` is out of cspell's `docs/**/*.md` scope; no workflow/secret changes).

Reviewers: @ requested Kavitha Sundaramurthy, Ivana Horvat.

### DIFF
--- a/src/resolve/narrator_split.py
+++ b/src/resolve/narrator_split.py
@@ -31,15 +31,22 @@ not by the name shape.
 Algorithm (per eligible canonical id ``C``, name ``N``)
 -------------------------------------------------------
 1. Gather every mention with ``canonical_narrator_id == C``.
-2. **Isnad-adjacency evidence** — for each such mention, look at the chain neighbours
-   at ``position_in_chain`` ±1 in the same ``(hadith_id, chain_index)`` chain (da#411 —
-   the composite key the graph loaders use; NOT ``hadith_id`` alone, which would
-   fabricate cross-chain neighbours in a multi-isnad hadith) and map them to their
-   *attested* ``death_year_ah`` (precision ≠ ``tabaqa_estimate`` — an *estimated*
-   window is never used as evidence, which also keeps the pass cascade-free on
-   re-run). A teacher (neighbour at position−1) dies a transmission gap *earlier*;
-   a student (position+1) a gap *later* — so ``C``'s per-mention death estimate is
-   ``neighbour_death ± MID_GAP`` (``MID_GAP`` the midpoint of
+2. **Isnad-adjacency evidence** — for each such mention, look at its immediately
+   preceding / following **resolved** chain neighbours in the same
+   ``(hadith_id, chain_index)`` chain (da#411 — the composite key the graph loaders
+   use; NOT ``hadith_id`` alone, which would fabricate cross-chain neighbours in a
+   multi-isnad hadith) and map them to their *attested* ``death_year_ah``
+   (precision ≠ ``tabaqa_estimate`` — an *estimated* window is never used as
+   evidence, which also keeps the pass cascade-free on re-run). "Adjacent" is the
+   **consecutive-resolved** notion the loader pairs into ``TRANSMITTED_TO`` edges
+   (:func:`src.graph.load_edges._build_chain_pairs`, da#439), NOT an exact
+   ``position ± 1`` test: an intermediate position left unresolved (null
+   ``canonical_narrator_id``) opens a position gap that the loader bridges and that
+   an exact-±1 window would silently skip — erasing the adjacency of any narrator
+   seen only in gappy chains (the drop-gate erasure class, main#928 / da#423). A
+   teacher (the earlier-position neighbour) dies a transmission gap *earlier*; a
+   student (the later-position neighbour) a gap *later* — so ``C``'s per-mention
+   death estimate is ``neighbour_death ± MID_GAP`` (``MID_GAP`` the midpoint of
    :data:`~src.resolve.mononym_split._MIN_GAP`/``_MAX_GAP``), averaged over the
    mention's attested neighbours. A mention with no attested neighbour is *undatable*.
 3. **Cluster** the per-mention estimates 1-D, gap-based: sort and cut wherever a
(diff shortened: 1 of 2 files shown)


parametrization:
RequestOrReplied: Request

Review requested for da#439 — reconciling `narrator_split`'s adjacency with the loader's `TRANSMITTED_TO` consecutive-resolved pairing so the split's evidence and the loaded graph agree on gappy chains. Peer reviewer: Kavitha Sundaramurthy; secondary: Ivana Horvat. Base is `deployments/phase-9/wave-25`.

parametrization:
RequestOrReplied: Approved

**Review: Approved**

Reviewed at head `9051199`. This is directly my `narrator_split` domain, and it's a correct, principled reconciliation.

Verified the three things that matter:
- **(a) The invariant is real, not a tautology.** `test_split_adjacency_matches_loader_transmitted_to` imports and runs the *production* `load_edges._build_chain_pairs` and, independently, the splitter path (`_build_chain_index` + `_collect_datable`), then asserts the candidate's split-neighbour set equals the loader's `TRANSMITTED_TO` incident edges. Two independent implementations compared on one input — genuine cross-function agreement. Confirmed `_build_chain_pairs` (load_edges.py:185-219) actually does position-sort → filter-resolved → pair consecutive → drop `from_id==to_id`, so the "mirror the loader" claim is faithful, including the self-loop-not-bridged behavior (for `[A,C,C,Y]` the loader emits A→C and C→Y; the splitter's two C-mentions yield {A} then {Y}, union {A,Y} — they agree at the per-candidate level).
- **(b) The gappy-chain fixture is a real RED→GREEN.** Positions 0/2/4 resolved, 1/3 null-dropped; the old exact-±1 window looked at the (dropped) positions 1 and 3 and left the candidate undatable — the drop-gate erasure class. The consecutive-resolved read recovers both adjacencies.
- **(c) No regression to the split engine.** Sign convention (teacher +MID_GAP / student −MID_GAP) preserved; da#411 composite-key confinement unchanged; full resolve+graph suite 857 green; narrator_split 20 green; CI 6/6 SUCCESS at head.

Must-fix: None

TechDebt:
- The invariant test asserts agreement on a **single** `(hadith_id, chain_index)` chain (it calls `_build_chain_pairs` directly, which is the per-chain pairing unit; production applies it after grouping by composite key in `_load_transmitted_to`). Faithful for the gappy case under test, but a multi-isnad agreement fixture would strengthen the invariant to cover the grouping boundary too. Optional.
- `_collect_datable` locates a mention's slot with `next(i for i,(p,_) in enumerate(chain) if p == position)` — by position **value**. If two resolved mentions ever share a `position_in_chain` in one chain (a data-quality defect, not expected), this returns the first slot and can mislocate `own_id`/neighbours. Marginal (requires a position-uniqueness violation) and the chain tuples carry no `mention_id` to disambiguate on, so I'm not gating on it — noting for the record.

Nice work — the instrument-separation discipline (running the real loader, not a splitter-internal echo) is exactly right for this reconciliation.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved

**Review: Approved**

QA / data-quality lens — narrator_split adjacency reconciled to the loader's TRANSMITTED_TO (da#439).

The cross-function invariant test is genuine, not a tautology. `test_split_adjacency_matches_loader_transmitted_to` runs BOTH independent code paths on one gappy fixture — the loader's `_build_chain_pairs` (position-sort → filter resolved → consecutive pairs) and the splitter's `_build_chain_index` + `_collect_datable` (drops unresolved, position-sort, immediate list neighbours) — and asserts the candidate's neighbour sets are set-equal. Two separate implementations, one asserted agreement. Good.

Gappy-chain RED→GREEN is real. `test_gappy_chain_datable_across_dropped_position`: positions 1 and 3 are unresolved (null `canonical_narrator_id`), so `_build_chain_index` drops them from `chains` (verified: `if cid is None: continue`). The old exact-`position ± 1` window looked only at the dropped positions → `datable` empty (the da#423 erasure class). The new consecutive-resolved walk recovers teacher(pos0)+student(pos4) across the gap, and the estimate assertion `round(((100+MID_GAP)+(180-MID_GAP))/2)` also pins the teacher(+)/student(-) sign convention. Solid.

Self-loop mirroring is correct — verified edge-for-edge against the loader source (not just the PR comment). Loader `_build_chain_pairs` drops a `from_id == to_id` pair with `continue` and does NOT bridge past it to the next resolved mention; the splitter's `if nid == own_id: continue` on the immediate ±1 neighbour behaves identically (skip, no bridge). The candidate's other incident edge is unaffected in both. They agree.

Must-fix: None
TechDebt:
- The cross-function invariant test covers only the gappy (dropped-unresolved) case. The self-loop agreement — a distinct behavior claim in this PR — is proven by reasoning + code comment, not asserted by the invariant test. Add a self-loop fixture to `test_split_adjacency_matches_loader_transmitted_to` so loader/splitter agreement on `from_id == to_id` is a green assertion, given this domain's self-loop history (da#148).
- Minor ordering edge: on a position tie within one chain the loader sorts by `position_in_chain` only (stable) while the splitter sorts by `(position, canonical_id)`; the two could order tied positions differently and diverge. Degenerate for real isnads (unique position per chain), but a tie fixture would close it.

Issue #346, opened 2026-07-09
data-quality: bare 'عبد الله' canonical node holds 19,412 mentions — over-merge narrator_split did not peel


parametrization:
## Context

Companion to the `عائذة` anomaly. From the same 2026-07-09 narrator-resolve run (parquet ref `staged/narrator-resolve/2026-07-09-dcb6205`, post da#337 `narrator_split`).

## Anomaly

Rank 8 by mention_count is the **bare** name `عبد الله` with **19,412** mentions, carrying no nasab, no kunya, no nisba.

`Abd Allah` is one of the most common names in the corpus. A single canonical node holding 19,412 mentions of the bare form is the classic over-merge aggregate: many distinct narrators (Ibn Umar, Ibn Abbas, Ibn Mas'ud, Ibn Amr, Ibn al-Zubayr, and dozens of lesser figures) whose isnad occurrence was recorded without a discriminating element, all collapsed onto one `uuid5(name)` node.

This is precisely the failure class da#337 was built to address, so the interesting question is **why the split did not fire here**.

## Investigation

- Did `narrator_split` consider this node at all? Check `narrator_splits.parquet` for `name_ar_normalized == 'عبد الله'`.
- The splitter discriminates on date-band (`band_midpoint_ah`, `band_earliest_ah`, `band_latest_ah`) plus exemplar anchors. A bare name with no biographical linkage likely has **no date band at all**, so there is nothing to discriminate on and the splitter correctly declines. If so, the fix is not in the splitter — it is upstream, in giving these mentions a discriminating signal.
- Evaluate **contextual disambiguation**: the isnad position (who transmits to/from this node) is a strong signal even when the name string is bare. An `Abd Allah` who narrates from the Prophet and to Nafi' is Ibn Umar.
- Quantify the blast radius: what fraction of the 19,412 mentions have a resolvable neighbor pair?

## Related

Same root class as the pre-da#337 generic-kunya blobs that gated the prod choke-points enrich. Betweenness/pagerank on an over-merged `Abd Allah` node produces a phantom hub.

## Do not start

Deferred by owner pending quota reset. Investigation only when picked up — no fix in flight.

parametrization:
## Update 2026-07-12 — persists in the corrected from-parse artifact; now BLOCKS the choke-point deliverable

[...]

PR #451, opened 2026-07-18
resolve: contextual (isnad-neighbour) disambiguation of over-merged bare names (da#346)

parametrization:
## Summary

Adds a **contextual (isnad-neighbour) disambiguation** resolve stage (da#346) that peels
the confidently-identifiable mentions off an over-merged bare name onto distinct nodes,
reducing the phantom-hub betweenness — while leaving everything it cannot confidently
identify on the flagged primary.

## Why the date-axis splitter could not do this (acceptance (a))

`narrator_split` **correctly abstains** on bare `عبد الله`: a bare mention carries no
*attested* death-band to cut (Gate 1, single band ⇒ abstain), and the #337 spike proved
forcing a date-band split is infeasible (the chimera does not separate) and unsafe
(false-splits genuine hubs — al-Zuhrī into five impossible-date nodes). That negative
result is why #337 was flag-now and node-identity split deferred to external rijāl
(da#443). Documented in the module docstring.

## The corpus-internal lever (acceptance (b))

The isnad **position** identifies a referent even when the name string is bare: an ʿAbd
Allāh who narrates ⟵the Prophet and ⟶Nāfiʿ is ʿAbd Allāh ibn ʿUmar. For each
hand-verified neighbour-pair signature, the stage remaps the matching mentions to a
distinct discriminated node.

**Not a threshold, not an inference** (the da#423 silent-zero discipline). Driven by a
curated signature seed under a **bidirectional acceptance fixture**: a signature MUST fire
on the multi-referent class it names AND MUST NOT fire on the control class or a
partial (one-role) match. A mention matched by **zero** or **≥2** signatures is never
peeled — it stays on the bare primary. Empty seed ⇒ pure no-op (the safe default).

## Instrument separation (verified)

The end-to-end test builds one bare `عبد الله` node carrying three classes and asserts the
detector **separates** them: 3 golden-chain (Prophet→Nāfiʿ) mentions peel to Ibn ʿUmar; a
partial-match (Prophet→Mālik) and a control (Mālik→Zuhrī) both **stay** on the primary.
This is the "prove the instrument separates the classes before trusting the number" lesson
applied directly.

## Residual (acceptance (c))

The bare `عبد الله` node stays `over_merged`-flagged (seed entry 1 of
`over_merged_narrators.yaml` already flags it); the contextual peel only reduces its count
and does not touch the flag. Bare mentions with no corpus-internal discriminating pair are
left to da#443's external-rijāl split — explicitly out of scope here.

## Blast radius

The number (fraction of the 19,412 with a resolvable neighbour pair) is a **runtime
measurement** produced when the stage runs against the staged corpus — `data/` is empty in
the repo, so it cannot be computed in the PR. `neighbour_pair_coverage` computes it
deterministically (unit-tested), and the stage **always** writes
`contextual_coverage.parquet` (total / with-neighbour / with-teacher / with-student /
matched / ambiguous per bare name) so the re-run discloses it for review.

## Wiring

New `contextual_disambiguation` step in `run_all` (3.57), after `narrator_split`, before
reconcile/tabaqa/`over_merged_flag`, so peeled ids flow through the date stages and the
residual flag counts the reduced primary. Added to `RESOLVE_STEP_ORDER` (CLI `--from-step`
auto-includes it); no checkpoint (not in `RESUMABLE_STEPS`). Reuses narrator_split's
`_build_chain_index` / `_remap_split_mentions`.

## Tests / checks

- `tests/test_resolve/test_contextual_disambiguation.py` — 11 tests (see above).
- Full resolve suite: 576 passed, 6 skipped (orchestrator/from-step wiring unbroken).
- ruff format + lint clean; mypy strict clean; full pre-push green.

## Note for reviewers

The seed ships **one** textbook, hand-verifiable entry (Ibn ʿUmar). Grow-later under
da#443. If the team prefers to defer ALL node-minting to da#443, emptying the seed leaves
the mechanism + coverage harness intact as a pure no-op — flagged in my status to the
team lead.

## Reviewers

- Peer: **Ivana Horvat**

Closes #346
Part of main#928

### DIFF
--- a/src/resolve/__init__.py
+++ b/src/resolve/__init__.py
@@ -170,6 +170,7 @@ RESOLVE_STEP_ORDER = (
     "bio_promote",
     "cluster",
     "narrator_split",
+    "contextual_disambiguation",
     "reconcile",
     "tabaqa_dates",
     "over_merged_flag",
@@ -466,6 +467,7 @@ def run_all(
 
     from src.resolve import (
         bio_promote,
+        contextual_disambiguation,
         date_reconcile,
         dedup,
         disambiguate,
@@ -657,6 +659,47 @@ def run_all(
     else:
         outcomes["narrator_split"] = StageSkipped("narrator_split", "precomputed")
         logger.info("resolve_step_skipped_precomputed", step="narrator_split")
+
+    # Step 3.57: Contextual (isnad-neighbour) disambiguation (da#346). The date-axis
+    # narrator_split abstains on a BARE name (no attested death-band to cut), but the
+    # isnad POSITION still identifies a referent: a bare ʿAbd Allāh narrating ⟵Prophet
+    # ⟶Nāfiʿ is Ibn ʿUmar. For each curated, hand-verified neighbour-pair signature,
+    # peel the matching mentions onto a distinct discriminated node; leave every unknown
+    # or ambiguous mention on the bare primary (which over_merged_flag keeps flagged) and
+    # to da#443's external-rijāl split. Runs AFTER narrator_split (operates on the settled
+    # canonical set) and BEFORE reconcile/tabaqa_dates/over_merged_flag so the peeled ids
+    # + remapped mentions flow through the date stages and the residual flag counts the
+    # reduced primary. Empty seed ⇒ pure no-op. Idempotent (peeled ids no longer carry the
+    # bare name; the residual still matches no signature). Emits contextual_splits.parquet
+    # (audit) + contextual_coverage.parquet (the da#346 blast-radius report).
+    if _do("contextual_disambiguation"):
+        try:
+            logger.info("resolve_step", step="contextual_disambiguation", status="running")
+            ctx_path = contextual_disambiguation.apply_contextual_disambiguation(
+                output_dir, staging_dir=staging_dir
+            )
+            ctx_files = [ctx_path] if ctx_path is not None else []
+            outcomes["contextual_disambiguation"] = StageRan("contextual_disambiguation", ctx_files)
+            logger.info(
+                "resolve_step",
+                step="contextual_disambiguation",
+                status="complete",
+                files=len(ctx_files),
+            )
+        except Exception as exc:  # noqa: BLE001
+            outcomes["contextual_disambiguation"] = StageErrored(
+                "contextual_disambiguation", type(exc).__name__, traceback.format_exc()
+            )
+            logger.error(
+                "resolve_step_failed",
+                step="contextual_disambiguation",
+                traceback=traceback.format_exc(),
+            )
+    else:
+        outcomes["contextual_disambiguation"] = StageSkipped(
+            "contextual_disambiguation", "precomputed"
+        )
+        logger.info("resolve_step_skipped_precomputed", step="contextual_disambiguation")
 
     # Step 3.6: Multi-source date reconciliation (da#165). After bio_promote and
     # cluster have built the final canonical set, fold each narrator's per-source
--- a/src/resolve/contextual_disambiguation.py
+++ b/src/resolve/contextual_disambiguation.py
@@ -51,10 +51,12 @@ Algorithm (per curated signature ``S`` on bare name ``N``)
 ----------------------------------------------------------
 1. Resolve ``N`` to its canonical id ``C = make_canonical_id(N)`` and gather every
    mention with ``canonical_narrator_id == C``.
-2. For each such mention, read its ±1 chain neighbours **within the same
-   ``(hadith_id, chain_index)`` chain** (:func:`src.resolve.narrator_split._build_chain_index`,
-   da#411 — never a cross-chain neighbour) and map them to their normalized names: the
-   set of teacher names (position −1) and student names (position +1).
+2. For each such mention, read its **consecutive-resolved** teacher/student neighbours
+   within the same ``(hadith_id, chain_index)`` chain — the SHARED adjacency helper
+   :func:`src.resolve.narrator_split.resolved_chain_neighbours` (da#439), identical to
+   the loader's ``TRANSMITTED_TO`` adjacency and the date-band splitter's, so a gappy
+   isnad cannot silently erase a discriminating pair (da#411 — never a cross-chain
+   neighbour) — and map them to their normalized names.
 3. A signature *matches* a mention iff every neighbour role it constrains is satisfied —
    its ``teacher_ar`` (if given) is among the mention's teacher names AND its
    ``student_ar`` (if given) among the student names. A mention matched by exactly one
@@ -100,7 +102,11 @@ import yaml
 from src.parse.identity import make_canonical_id, make_discriminated_canonical_id
 from src.resolve._run_record import write_canonical
 from src.resolve.attestation import derive_attestation
-from src.resolve.narrator_split import _build_chain_index, _remap_split_mentions
+from src.resolve.narrator_split import (
+    _build_chain_index,
+    _remap_split_mentions,
+    resolved_chain_neighbours,
+)
 from src.resolve.schemas import NARRATORS_CANONICAL_SCHEMA
 from src.utils.arabic import normalize_arabic
 from src.utils.logging import get_logger
@@ -280,26 +286,28 @@ def _neighbour_names_by_mention(
     chains: dict[tuple[str, int], list[tuple[int, str]]],
     name_by_id: dict[str, str],
 ) -> dict[str, tuple[frozenset[str], frozenset[str]]]:
-    """``mention_id -> (teacher_names, student_names)`` from ±1 in-chain neighbours.
+    """``mention_id -> (teacher_names, student_names)`` from the consecutive-resolved neighbours.
 
-    A neighbour is only ever the ±1 position **within the same ``(hadith_id, chain_index)``
-    chain** (da#411), mirroring the graph loaders' composite chain key. Names are the
-    normalized ``name_ar_normalized`` of the neighbour canonical ids.
+    Uses the **shared** isnad-adjacency helper
+    (:func:`src.resolve.narrator_split.resolved_chain_neighbours`, da#439) so this stage's
+    notion of teacher/student is byte-identical to the date-band splitter's and to the
+    loader's ``TRANSMITTED_TO`` adjacency — the immediately-preceding / -following
+    **resolved** mention in the position-sorted ``(hadith_id, chain_index)`` chain (a
+    dropped-position gap bridged as the loader bridges it; self-loop dropped, not bridged
+    past; never a cross-isnad neighbour, da#411). Reading exact ``position ± 1`` here (the
+    pre-da#439 form) would have silently erased a golden-chain teacher/student pair on any
+    gappy isnad — under-peeling a confident identity. Names are the neighbour ids'
+    ``name_ar_normalized``; each role is a singleton or empty set.
     """
     out: dict[str, tuple[frozenset[str], frozenset[str]]] = {}
     for hadith_id, chain_index, position, mention_id in mentions:
-        teachers: set[str] = set()
-        students: set[str] = set()
-        for npos, nid in chains.get((hadith_id, chain_index), ()):
-            if npos == position - 1:
-                name = name_by_id.get(nid)
-                if name:
-                    teachers.add(name)
-            elif npos == position + 1:
-                name = name_by_id.get(nid)
-                if name:
-                    students.add(name)
-        out[mention_id] = (frozenset(teachers), frozenset(students))
+        teacher_id, student_id = resolved_chain_neighbours(chains, hadith_id, chain_index, position)
+        teacher_name = name_by_id.get(teacher_id) if teacher_id else None
+        student_name = name_by_id.get(student_id) if student_id else None
+        out[mention_id] = (
+            frozenset({teacher_name} if teacher_name else ()),
+            frozenset({student_name} if student_name else ()),
+        )
     return out
 
 
--- a/src/resolve/contextual_disambiguation.py
+++ b/src/resolve/contextual_disambiguation.py
@@ -1,0 +1,548 @@
+"""Peel confidently-identifiable mentions off an over-merged bare name by isnad context (da#346).
+
+The problem this addresses (why ``narrator_split`` could not)
+-------------------------------------------------------------
+The bare ism ``عبد الله`` collapses ~19k mentions of *many* historically-distinct ʿAbd
+Allāhs (Ibn ʿUmar, Ibn ʿAbbās, Ibn Masʿūd, …) onto one ``nar:`` node — the classic
+over-merge that inflates betweenness (da#346). The date-axis splitter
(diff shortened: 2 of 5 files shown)


parametrization:
RequestOrReplied: Request

**Review: Changes requested**

Strong stage overall — the curated-seed-not-threshold discipline, the zero/≥2-match residual, the empty-seed and absent-primary no-ops, and the always-emitted coverage report are exactly right, and the three acceptance axes I was asked to verify all pass (details below). But there is one must-fix that I am uniquely placed to catch, because this stage IS the "PR-B" consumer da#439 was filed to protect, and it currently re-commits the very divergence da#439 exists to remove — invisibly, because the fixtures mask it.

**Must-fix**

1. **`_neighbour_names_by_mention` uses exact `position ± 1`; it must use the consecutive-resolved adjacency (da#439).** At contextual_disambiguation.py `_neighbour_names_by_mention`, the teacher/student sets are built with `npos == position - 1` / `npos == position + 1`. That is the exact-±1 window da#439 identifies as diverging from the loader's `TRANSMITTED_TO` adjacency (`load_edges._build_chain_pairs`, which pairs *consecutive resolved* mentions and bridges position gaps). da#439's issue body names #337/#346 — this PR — as the reason for the reconciliation: "If PR-B inherits the splitter's exact-±1 window, its adjacency signal will diverge from the loaded topology on gappy chains." So this stage should read a mention's *immediately-preceding / -following resolved* neighbour in the position-sorted `(hadith_id, chain_index)` chain, not the exact ±1 slot.
   - Direction of the error is **safe** (it under-peels, never false-peels: exact-±1 is a strict subset of the consecutive-resolved neighbours, so it can only *miss* a genuine teacher/student across a gap, leaving that mention on the flagged primary — it never manufactures a wrong match). So this is not data corruption and does not threaten the default-safe gate. But it defeats the stage's da#346 purpose on gappy chains: a genuine Ibn ʿUmar mention whose Prophet-teacher sits across a dropped/unresolved intermediate position is silently not peeled — the da#423 / main#928 erasure class, in the exact place Phase 9 is trying to fix it. And it means da#439's fix never actually reaches its named consumer.
   - Please reconcile to the consecutive-resolved definition. Best done via a shared helper so PR-B, `narrator_split._collect_datable`, and the loader share one adjacency notion (I'm the da#439/PR#450 author — happy to extract `_immediate_resolved_neighbours(chain, position)` there and have this stage consume it, to avoid duplicating the logic a third time; let's coordinate merge order so we don't leave a window where the two disagree).

2. **Add a gappy-chain fixture — the divergence is currently fixture-masked.** Every chain in the test (`_chain`) is contiguous (positions 0/1/2), so no test can distinguish exact-±1 from consecutive-resolved (silent-zero / instrument-doesn't-separate, per feedback_silent_zero_is_not_a_measurement + feedback_fixture_makes_guard_assertion_inert). Add a fixture with a dropped/unresolved interior position (e.g. Prophet@0, unresolved@1, ʿAbd Allāh@2, Nāfiʿ@3) and assert the Ibn ʿUmar signature still fires across the gap. That test should be RED on the current exact-±1 reader and GREEN after fix (1).

**The three axes I was asked to verify (all pass):**

- (a) Bidirectional separation is genuine: `test_stage_peels_matched_class_leaves_control_and_partial` peels the golden Prophet→ʿAbd Allāh→Nāfiʿ chain to Ibn ʿUmar, leaves the Mālik→Zuhrī control whole, and leaves the partial-match (Prophet→…→Mālik) on the primary; `test_match_signature_zero_and_ambiguous_return_none` confirms a ≥2-signature ambiguous pair returns None. The instrument separates the peel-class from the control-class on contiguous chains. (The only gap is the missing gappy-chain case in must-fix 2.)
- (b) Gate is default-safe: zero/≥2-match → primary; empty seed → no-op; absent primary → no-op minting nothing; load-time validation rejects an all-`None` (match-everything) signature; peeled targets carry hand-verified `name_en`/`note` and `over_merged=False`, while the reduced residual stays for `over_merged_flag`. Nothing mints without a hand-verified signature.
- (c) Blast-radius is deferred, not faked: `neighbour_pair_coverage` computes the tally deterministically and is unit-tested; `contextual_coverage.parquet` is always written (even a zero-peel run); no corpus number is hard-coded — it is produced when the stage runs against the staged artifact at the owner-gated re-run. (Note: fixing must-fix 1 will *change* those coverage numbers upward on gappy chains, which is the point — the report should reflect the loader-consistent adjacency.)

**Other observations (non-blocking):**
- Idempotence holds via the target *id* differing from the primary id (peeled mentions are no longer gathered under `C`), which `test_stage_idempotent` confirms. The docstring says "peeled target ids no longer carry the bare name N" — minor imprecision, the target row does inherit the bare `name_ar_normalized`; idempotence is by id, not name. Consider a one-line docstring tweak.
- Run-order (after `narrator_split`, before `reconcile`/`tabaqa_dates`/`over_merged_flag`) and the `__init__` provenance order are correct so peeled ids flow through the date stages and the residual flag counts the reduced primary.

TechDebt: None (must-fix 1/2 are correctness, not debt).

CI is 11/11 green, but that green is exactly why must-fix 2 matters — the contiguous fixtures can't fail on the adjacency divergence.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Request

**Review: Must-fix (1)**

QA / data-quality lens — gated `contextual_disambiguation` stage (da#346). Most of this is strong; one finding blocks.

What is genuinely proven (verified against tests + source):
- **Bidirectional instrument separation** — `test_stage_peels_matched_class_leaves_control_and_partial` peels only the golden class (Prophet→Nāfiʿ ⇒ Ibn ʿUmar, 3 mentions), and leaves BOTH a partial-match (Prophet→Mālik) AND a control (Mālik→Zuhrī, a different ʿAbd Allāh) on the bare primary; primary count reduced 5→2. The signature fires on the class it names and does NOT fire on the control or the partial. This is real class separation, not a silent zero.
- **Default-safe gate** — nothing mints without a hand-verified constrained signature: unconstrained signature (no teacher AND no student) is a hard `ValueError` at load; empty seed → pure no-op (no coverage file written); absent primary → no-op/no fabrication; zero-match and ≥2-match (ambiguous) both leave the mention on the flagged primary.
- **Blast-radius honestly deferred** — `neighbour_pair_coverage` computes coverage deterministically and always emits `contextual_coverage.parquet` on a real run; the corpus number is left to the re-run, not faked.
- No false-peel risk: exact-±1 reads a subset of true neighbours, and matching requires every constrained role present, so a missing neighbour can only make matching HARDER — it can never introduce a spurious name or cause a wrong mint. The peel direction is strictly safe (under-recall, never over-mint).

Must-fix:
1. **The neighbour walk uses the exact-`position ± 1` adjacency that sibling PR#450 (da#439) is removing THIS wave as the gappy-chain erasure class (main#928 / da#423).** `_neighbour_names_by_mention` matches `npos == position - 1` / `+ 1`, but `_build_chain_index` drops unresolved mentions from `chains` (verified: `if cid is None: continue`). So a real Ibn ʿUmar mention in a gappy chain — `Prophet → [unresolved] → ʿAbd Allāh → Nāfiʿ` — finds no teacher at position−1, fails the signature, is NOT peeled, and is under-counted in `contextual_coverage.parquet` (`with_teacher` / `signature_matched`). The peel itself is safe (it just stays flagged for da#443), but `contextual_coverage.parquet` is the *named da#346 acceptance / blast-radius measurement*, and this instrument structurally erases the gappy-chain class — the exact "verify the instrument before the reading" failure the drop-gate history warns about, and it does so right as the sibling function drops this very adjacency. Align `_neighbour_names_by_mention` to the consecutive-resolved (da#439) adjacency — ideally a shared helper with `_collect_datable` — so the peel recall and the acceptance metric count the same neighbours the loader connects. Coordinate with PR#450 (they touch the same adjacency notion in `narrator_split`).

TechDebt: None

parametrization:
## Rebased on merged wave-25 + shared-adjacency-helper fix (re-requesting review)

Force-pushed `d5896ae`. Base wave-25 now contains da#439 (PR#450) and da#444 (PR#449).

**What changed since the last review:**
- Rebased cleanly on merged wave-25.
- **Closed an adjacency-drift this PR would otherwise have re-introduced.** da#439 moved `narrator_split` to the loader's *consecutive-resolved* `TRANSMITTED_TO` adjacency (gap-bridging, self-loop-not-bridged); this PR's `contextual_disambiguation` still used the old exact-`position ± 1` neighbour read — the same gappy-chain erasure class, which would silently under-peel a confident identity (a golden chain `Prophet→[dropped]→ʿAbd Allāh→Nāfiʿ` would miss the Ibn ʿUmar signature).
- Extracted `narrator_split.resolved_chain_neighbours(...)` as the **single** shared adjacency definition; refactored `_collect_datable` to call it (behaviour identical — narrator_split suite green) and switched `_neighbour_names_by_mention` to it. So the date-band splitter, the contextual disambiguator, and (forthcoming) `narrator_unify` da#347 all reason about teacher/student through one function and cannot drift again (the root-class fix; a per-stage copy is how da#439 happened).
- New `test_stage_gappy_chain_still_matches_signature`: RED under exact-±1, GREEN via the helper.

**Checks:** full resolve suite 580 passed / 6 skipped; all CI green (Integration Tests included); mergeable.

Reviewers: **Ivana Horvat** (peer), **Oyunbileg Batbayar** (secondary) — please re-review. @Ivana: `resolved_chain_neighbours` is the helper the team lead asked you to import for the da#347 isnad-neighbour gate rather than rolling your own.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved

**Review: Approved**

Re-review at head `d5896ae` (rebased on merged wave-25). The must-fix from my prior verdict is resolved.

Verified:
1. **Exact-`position ± 1` walk is gone.** `_neighbour_names_by_mention` now imports and calls the shared `narrator_split.resolved_chain_neighbours` helper — the single da#439 consecutive-resolved adjacency definition, byte-identical to `_collect_datable`'s and to the loader's `TRANSMITTED_TO` (`_build_chain_pairs`). No `position - 1` / `position + 1` neighbour test remains (only doc references to the retired form). So `contextual_coverage.parquet` no longer structurally under-counts the gappy-chain class — my key point.
2. **Gappy fixture is a genuine RED→GREEN.** `test_stage_gappy_chain_still_matches_signature`: `Prophet(0) → null(1) → ʿAbd Allāh(2) → null(3) → Nāfiʿ(4)` with positions 1 and 3 unresolved (dropped from `chains`). The candidate at pos2 peels to the Ibn ʿUmar target across BOTH dropped positions — the teacher/student bridge the gaps exactly as the loader does. Under the old exact-±1 read the Ibn ʿUmar signature would not have fired. Coverage is computed from the identical `neighbours` dict used for the peel, so the matched mention is counted in `signature_matched` by construction — peeled AND counted.
3. **All three separation axes still hold post-rebase.** Golden peels / partial-match + control both stay on the flagged primary (`test_stage_peels_matched_class_leaves_control_and_partial`); zero/ambiguous return None; unconstrained signature is a hard `ValueError`; empty seed is a pure no-op. CI Test + Integration Tests green. The helper collapses each role to a single consecutive-resolved neighbour (matching the loader's one-predecessor/one-successor adjacency) and mirrors the self-loop drop-without-bridge — the earlier multi-name set semantics were correctly replaced, not a regression.

Blast-radius still honestly deferred to the re-run (computed deterministically, emitted, no fake zero), now on the same adjacency the loader connects.

Must-fix: None
TechDebt: The gappy fixture asserts the peel (`canonical_narrator_id == target`) but not the `contextual_coverage.parquet` row for that run; a one-line `signature_matched == 1` assertion on the gappy case would make "counted, not just peeled" a direct green rather than an inferred one. Non-blocking.

parametrization:
RequestOrReplied: Approved

**Review: Approved** (re-review of the rework at a5b3b73, rebased on merged-#450)

Both must-fixes from my prior review are resolved, verified by reading the code at head AND executing it — not just trusting green CI.

1. **Shared adjacency helper — ONE definition, no third copy.** `narrator_split.resolved_chain_neighbours(chains, hadith_id, chain_index, position)` is now the single consecutive-resolved definition (exported in `__all__`), returning `(teacher_id, student_id)` as the immediately-preceding / -following resolved neighbour in the position-sorted chain, with the loader's self-loop drop (no bridging past) and the da#411 cross-chain confinement. Confirmed BOTH consumers call it:
   - `narrator_split._collect_datable` (the date-band splitter) — refactored to `teacher_id, student_id = resolved_chain_neighbours(...)`, behaviour identical to merged-#450.
   - `contextual_disambiguation._neighbour_names_by_mention` — now `teacher_id, student_id = resolved_chain_neighbours(...)`; `grep` confirms **no** lingering `position ± 1` / `npos ==` in the stage. So PR-B, the splitter, and the loader's TRANSMITTED_TO now share one adjacency notion — the da#439 reconciliation reaches its named consumer.

2. **Gappy-chain fixture is a genuine RED→GREEN instrument (fixture-masking gone).** `test_stage_gappy_chain_still_matches_signature` builds Prophet@0 / null@1 / ʿAbd Allāh@2 / null@3 / Nāfiʿ@4 and asserts the mention peels to `ctx:ibn-umar` across both dropped positions. I proved it is a real instrument by monkeypatching the helper back to exact-±1 and re-running: the stage then produces **NO PEEL** (`contextual_disambiguation_no_peel`) — i.e. the test genuinely separates the two adjacency notions, it is not a tautology. It also asserts the coverage report counts the gappy mention (`with_teacher=1`, `with_student=1`, `signature_matched=1`), un-masking the divergence at the measurement level too (Oyunbileg's point) — good addition.

**Three original axes re-verified after the rebase (all still hold):**
- (a) Bidirectional separation intact: `test_stage_peels_matched_class_leaves_control_and_partial` still peels the golden Prophet→Nāfiʿ class, leaves the Mālik→Zuhrī control whole, leaves the partial-match on the primary; ambiguous ≥2-match still returns None.
- (b) Gate still default-safe: zero/≥2→primary, empty-seed/absent-primary no-op, all-`None` signature rejected at load, residual stays flagged; the shared helper only *narrows or widens which real neighbour is seen*, never fabricates one (still under-peels-not-over-peels on a pruned slot → `(None, None)`).
- (c) Blast-radius still deferred + deterministic, now additionally exercised on the gappy case.

Executed locally on the head: `tests/test_resolve/test_contextual_disambiguation.py` + `test_narrator_split.py` → **33 passed**. CI 11/11 green, mergeable.

Must-fix: None
TechDebt: None

Nice work — the helper extraction is the right shape and the coverage-level gappy assertion is a stronger instrument than I asked for.

parametrization:
Follow-up (`a5b3b73`): strengthened the gappy-chain fixture to also assert the across-gap mention is **counted** in `contextual_coverage.parquet` (`with_teacher=1`, `with_student=1`, `signature_matched=1`, `with_any_neighbour=1`) — @Oyunbileg's point that the old exact-±1 walk under-counted the named blast-radius measurement, not just under-peeled. The fixture now proves both halves: the peel AND the corrected measurement. All CI green, mergeable.

Issue #317, opened 2026-07-05
Narrator set: matn-sentence / Qur'anic-verse pollution in zero-degree tail (~26%)


parametrization:
## Summary

The canonical narrator set contains a large class of **matn-sentence / Qur'anic-verse pollution** — full sentences mis-extracted as narrator names. On the staging graph reload (#723 clean data), **~26% of the 150,187 narrators are zero-degree orphans whose `name_ar` is a matn sentence**, not a person. This is the first thing a user sees on the `/narrators` list (id-ordered), e.g. row 1 is Qur'an 22:32 (`الله ومن يعظم شعاءر الله فانها من تقوى القلوب`).

## Scope — this is NOT introduced by the #723 reload

Pre-existing. The da#311 scrub removed single-token junk + isnad residue (165,939 → 150,187) but **never addressed matn *sentences***. The old graph carried more of this class. It is orthogonal to da#314 (trailing-isnad leakage) and da#315 (ما/وبه residual) — this is a distinct, larger class: whole narrative/verse sentences as standalone "narrators".

## Evidence (staging Neo4j, post-reload)

| Metric | Count | % of 150,187 |
|---|---|---|
| Zero-degree orphans (no TRANSMITTED_TO/NARRATED edge) | 45,617 | 30.4% |
| …that are matn-sentence-shaped (≥5 words) | 38,763 | **25.8%** |
| Names starting with و/ف, ≥4 words (matn narrative) | 10,958 | 7.3% |
| Orphans containing قال/كان/الله | 7,009 | 4.7% |

Examples (all zero-degree, `in_degree/out_degree/pagerank = null`):
- `الله ومن يعظم شعاءر الله فانها من تقوى القلوب` (Qur'an 22:32)
- `وتهيات للحج وودعت الناس وكنت على الخروج فورد نحن لذلك كارهون والامر اليك`
- `فقلت لو كنت هناكا ما قتلتهم`
- `وما` ("and not")

## Why the earlier da#311 validation passed (blind spot)

Validation checked **top narrators by degree** (clean — matn orphans have zero degree and never rank) and probed `شعائر` with hamza `ئ` while the stored normalized form is `شعاءر` (hamza `ء`) → **false 0**. The pollution lives entirely in the zero-degree tail. Independent inspection of the UI surface caught it.

## Proposed remediation (two levers)

[...]

PR #456, opened 2026-07-18
feat(parse): recover matn-embedded isnads for sanadset "No SANAD" rows (#366)

parametrization:
## What & why

`sanadset`'s 159,558 rows tagged `Sanad = "No SANAD"` are **not** chainless — ~122,000 carry an isnad embedded in the running matn (da#366). This recovers them with a two-convention splitter, **without** the NER `full_text` fallback (#369) that mints matn sentences as narrators (measured on `lk`: 1.44× mention inflation, 53.7% of the extra names are matn substrings vs 0.1% of the isnad).

Disjoint from #368 (the 6,143 `Sanad`-column rows, already deterministic on-branch). Premise re-verified at HEAD: the `else` branch in `sanadset.py` left these rows `isnad_raw_ar = None` — that is exactly where the splitter slots in.

## Approach (precision-first)

- **Detector** `D = head_opener OR (≥2 standalone عن within the first 25 tokens)` — covers both the receipt-verb opener convention (`حدثنا … عن …`) and the bare-name convention (`علي بن إبراهيم عن أبيه عن …`); an opener-only or `عن`-only probe is blind to one of them.
- **Conservative boundary**: `قال` is treated as a chain *link* only when an opener/`عن` follows closely, otherwise it marks the matn; plus standalone `أن`/`أنه` and matn punctuation. Only the isolated isnad **head** is fed to the existing fail-loud `extract_narrator_mentions` — the whole body is never segmented (that IS the #317/#369 pollution).
- **Fails closed** (recovers nothing) on any ambiguity — detector miss, unlocatable boundary, unsegmentable head. Recovery is an accepted lower bound; recovering a matn sentence as a chain is the failure mode we refuse.

## Scope

- `src/parse/isnad_matn_split.py` — splitter (detector + boundary + segmentation).
- `src/parse/sanadset.py` — wiring into the `else` branch; recovered isnad moves out of `matn_ar` into `isnad_raw_ar`; `recovered_isnad_from_matn`/`embedded_mentions` observability (excluded from the B3 firehose metric like the #368 column path).
- `scripts/da366_measure_recovery.py` — **data-gated** measurement harness.

## The data constraint (flagged to the lead before implementation)

The repo ships **no real corpus** (`data/raw` is empty). So the recovery magnitude (~122,000) and the both-direction boundary validation on the 485,285 real tagged rows can only be measured at the **owner-gated re-run** — they are not CI assertions. What CI proves here is splitter **correctness** on fixtures reproducing both conventions + the negative/bare-opener/no-boundary classes, including a direct **pollution-avoidance** assertion vs the naive full-text segmentation and **both-direction boundary fidelity** (recovered isnad narrator-set == isnad-only segmentation; residual matn == original matn). The harness runs the magnitude/precision numbers when the corpus is present and prints a loud SKIP otherwise (no silent zero). Per the analysis this work is **out of the launch gate / additive**.

## Tests

`tests/test_parse/test_isnad_matn_split.py` (14), a parser integration test (recovers an embedded row, leaves a pure-matn row chainless), and `tests/test_scripts/test_da366_measure.py`. Full local⇄CI parity run green: `ruff check`, `ruff format --check`, `mypy src/`, and the full non-integration suite (2299 passed).

## Review

Reviewers: **Ivana Horvat** (peer), **Jean-Claude Habimana** (secondary — prior assignee, authored the da#366 analysis). Charter-format verdict comments, please (no `gh pr review`).

Resolves #366 (auto-close fires on the wave→main merge, not this wave-branch merge).

### DIFF
--- a/scripts/da366_measure_recovery.py
+++ b/scripts/da366_measure_recovery.py
@@ -1,0 +1,131 @@
+"""Measure da#366 matn-embedded isnad recovery against the real sanadset corpus.
+
+This is the instrument the issue asks for: run the two-convention splitter over
+the real ``No SANAD`` rows and report how many carry a recoverable isnad
+(expected ~122,000, a lower bound), broken down by convention; and check the
+boundary does not pull matn into the isnad.
+
+**Why it is data-gated.** No real corpus ships in the repo (``data/raw`` is
+empty), so this cannot run in CI. The recovery magnitude is a re-run
+measurement, not a unit assertion — the splitter's CORRECTNESS is unit-proven in
+``tests/test_parse/test_isnad_matn_split.py``; this script produces the MAGNITUDE
+once the raw CSVs are present. Run with no data and it prints a loud SKIP and
+exits 0 (non-gating), rather than emit a zero that looks like a measurement
+("silent zero is not a measurement").
+
+The pure summarisers (:func:`summarize_recovery`, :func:`boundary_precision`)
+are unit-tested in ``tests/test_scripts/test_da366_measure.py``.
+
+Usage:
+    PYTHONPATH=. python3 scripts/da366_measure_recovery.py [RAW_DIR]
+"""
+
+from __future__ import annotations
+
+import sys
+from collections import Counter
+from collections.abc import Iterable
+from dataclasses import dataclass, field
+from pathlib import Path
+
+from src.parse.isnad_matn_split import split_isnad_matn
+from src.parse.narrator_extraction import extract_narrator_mentions
+
+
+@dataclass(frozen=True)
+class RecoveryStats:
+    """Recovery tally over a set of candidate matn texts."""
+
+    total: int
+    recovered: int
+    by_convention: dict[str, int] = field(default_factory=dict)
+
+    @property
+    def rate(self) -> float:
+        return self.recovered / self.total if self.total else 0.0
+
+
+def summarize_recovery(texts: Iterable[str]) -> RecoveryStats:
+    """Run the splitter over *texts* and tally recoveries by convention."""
+    total = 0
+    conv: Counter[str] = Counter()
+    for text in texts:
+        total += 1
+        result = split_isnad_matn(text)
+        if result is not None:
+            conv[result.convention] += 1
+    return RecoveryStats(total=total, recovered=sum(conv.values()), by_convention=dict(conv))
+
+
+def boundary_precision(pairs: Iterable[tuple[str, str]]) -> tuple[int, int]:
+    """Boundary-precision check over known ``(isnad, matn)`` pairs.
+
+    Reconstructs the untagged shape ``isnad + " " + matn`` and asserts the
+    splitter does not pull a matn token into the recovered isnad. Returns
+    ``(clean, recovered)`` — ``clean`` is the number of recoveries whose isnad
+    narrator set is a subset of the isnad-only segmentation (no matn leaked in).
+
+    Recall is deliberately NOT reported: the tagged rows strip their opener, so
+    they are a different textual representation and cannot calibrate the
+    splitter's sensitivity (da#366). This measures precision only.
+    """
+    clean = 0
+    recovered = 0
+    for isnad, matn in pairs:
+        result = split_isnad_matn(f"{isnad} {matn}")
+        if result is None:
+            continue
+        recovered += 1
+        truth = {s.name for s in extract_narrator_mentions(isnad, "ar")}
+        got = {s.name for s in result.spans}
+        if got <= truth:
+            clean += 1
+    return clean, recovered
+
+
+def _iter_no_sanad_matns(raw_dir: Path) -> list[str]:  # pragma: no cover - needs real data
+    """Yield the matn text of every ``No SANAD`` row across the raw CSVs.
+
+    Mirrors the parser's tag extraction (``src/parse/sanadset.py``) so the
+    population matches what the loader would see.
+    """
+    from src.parse.base import read_csv_robust
+    from src.parse.sanadset import _MATN_RE, _SANAD_RE, _extract_tag, _strip_structural_tags
+
+    texts: list[str] = []
+    for csv_file in sorted(raw_dir.glob("*.csv")):
+        table, _enc = read_csv_robust(csv_file)
+        for i in range(table.num_rows):
+            full = table.column("Hadith")[i].as_py() if "Hadith" in table.column_names else None
+            if not full:
+                continue
+            sanad = _extract_tag(_SANAD_RE, full)
+            if sanad and sanad.lower() != "no sanad":
+                continue
+            matn = _strip_structural_tags(_extract_tag(_MATN_RE, full))
+            if matn:
+                texts.append(matn)
+    return texts
+
+
+def main(argv: list[str]) -> int:  # pragma: no cover - CLI wrapper
+    raw_dir = Path(argv[1]) if len(argv) > 1 else Path("data/raw/sanadset")
+    if not raw_dir.exists() or not any(raw_dir.glob("*.csv")):
+        print(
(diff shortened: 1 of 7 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved

**Review: Approved**

QA / data-quality lens (2nd reviewer, subbing for Ivana) — da#366 matn-embedded isnad recovery. Reviewed diff + tests at head `a727f87`; CI green. The pollution-avoidance crux (#317/#369) is handled well.

(1) **The pollution-avoidance test is genuine and WOULD fail on a boundary leak.**
- `test_opener_and_an_chain_is_both` and `test_qala_is_a_link_not_a_boundary_when_opener_follows` assert the recovered `isnad_ar` by EXACT string and pin exact span names (`["مالك","نافع","ابن عمر"]`); the matn in these fixtures contains its own `عن` (`نهى عن بيع الغرر`), so any boundary leak would lengthen `isnad_ar` / add spans and break the equality.
- `test_boundary_is_faithful_both_directions` reconstructs `isnad + " " + matn` and requires BOTH `result.isnad_ar == isnad` AND `result.matn_ar == matn` — bidirectional fidelity, catching a leak in either direction (matn pulled in, or isnad dropped).
- `test_splitter_does_not_pollute_like_the_ner_fallback` is a proper A/B: it first asserts the naive whole-body segmentation `any("رسول" in s.name)` (the control genuinely produces the #369 pollution) and then `all("رسول" not in s.name)` for the splitter with `len(spans) < len(naive)`. The instrument is shown to separate the classes, not just read a favorable number.

(2) **Fail-closed on ambiguity — verified.** `split_isnad_matn` returns `None` (recovers nothing) on: empty/blank, detector miss, unlocatable boundary (`_matn_boundary` → None rather than guess), empty head, `IsnadSegmentationError` (partial chain blob, da#158), and zero spans. Tested by `test_no_matn_boundary_marker_fails_closed`, `test_bare_opener_introducing_speech_recovers_nothing`, `test_non_isnad_matn_recovers_nothing`, and end-to-end by `test_matn_embedded_isnad_recovered` (the pure-matn "No SANAD" row stays `isnad_raw_ar is None` — never minted into a fake chain). Recovered mentions still pass the SAME `_is_narrator_like` gate as the other two paths, and recovery is non-destructive (`full_text_ar` preserved verbatim; integration test asserts it). Under-recover, never mint — correct direction.

(3) **Magnitude honestly deferred, not a silent zero.** `scripts/da366_measure_recovery.py` is data-gated: on no corpus it prints a loud SKIP and exits 0, explicitly refusing to emit a zero that "looks like a measurement." The ~122,000 is documented as a re-run lower bound and surfaced via the `recovered_isnad_from_matn` parser metric. The pure summarizers (`summarize_recovery`, `boundary_precision`) are unit-tested so the measurement logic itself is covered.

Must-fix: None
TechDebt:
1. The module acknowledges an ~3.1% false-positive cost for the `an_chain` probe (≥2 `عن`) on genuine matn, but there is no negative fixture in that weak class — every negative (`pure matn`, `qala-head matn`, `single-an prose`) fails at the detector, not at the downstream defenses. Add one matn fixture that TRIPS the `an_chain` detector (≥2 windowed `عن`) yet must recover nothing / only clean narrators, to demonstrate the boundary + fail-loud segmenter + `_is_narrator_like` filter actually contain that acknowledged precision cost (this domain's da#423/#369 history warrants proving the second line of defense, not just the detector).
2. `test_da366_measure` covers `boundary_precision` only on the all-clean case (`clean == recovered`). Add a leaking-input case so the precision instrument is shown to report `clean < recovered` when matn leaks — verify the instrument separates before the re-run reads it.
Both are test-completeness; neither blocks merge given the fail-closed design and the reused narrator-likeness gate.

parametrization:
RequestOrReplied: Approved

**Review: Approved**

Reviewed as the da#366 analysis author, with a precision-first lens. Read at head `a727f87`; the detector, boundary, fail-closed paths, parser integration, and both test files all hold up.

**1. The boundary genuinely does not leak matn into the isnad — and the assertion is real, not vacuous.** Two independent, live controls prove it:
- `test_splitter_does_not_pollute_like_the_ner_fallback` is a genuine A/B against the naive da#369 full-body segmentation: it asserts the naive path DOES absorb a matn token (`any("رسول" in s.name for s in naive)`) AND that the splitter does NOT (`all("رسول" not in ...)`), with `len(result.spans) < len(naive)`. The positive-control half means the instrument actually separates the two classes — this is not the inert-fixture trap (main#927): CI is green, so the naive path really does mint the Prophet-formula token as a narrator.
- `test_boundary_is_faithful_both_directions` reconstructs `isnad + " " + matn` and requires BOTH `isnad_ar == isnad` exactly (nothing of matn pulled in) AND `matn_ar == matn` exactly, with spans equal to the isnad-only segmentation (nothing dropped). That is the bidirectional fidelity the domain demands.
The `boundary_precision` harness independently formalizes precision as `got <= truth` (recovered narrator set ⊆ isnad-only set) and deliberately declines to report recall (tagged rows strip their opener → different representation), which is the honest framing.

**2. Fail-closed is correct — it under-recovers, never mints a matn sentence.** Every ambiguity returns `None`: detector miss, unlocatable boundary (`_matn_boundary` returns None when no reliable قال/أن/punct introducer follows the chain), empty head, `IsnadSegmentationError` on an unsegmentable head (da#158 chain blob), and a head yielding zero spans. Covered by `test_no_matn_boundary_marker_fails_closed`, `test_bare_opener_introducing_speech_recovers_nothing`, the pure/qala-head/single-an negatives, and the integration test's pure-matn row (`by_num[2]["isnad_raw_ar"] is None`). The boundary is `min(valid)` — the earliest reliable introducer — so the directional bias is toward cutting early (fewer narrators = under-recovery), never toward absorbing matn. Linking `قال` (opener/عن within 12 chars) is correctly skipped so a `حدثنا … قال حدثنا …` chain is not severed. This is the #317/#369 discipline honored end to end.

**3. The ~122k is correctly a re-run measurement, not a faked CI number.** No test asserts a magnitude; the unit suite proves CORRECTNESS only. The 122k appears solely as a documented expectation/lower-bound in the module docstring and the `recovered_isnad_from_matn` metric comment. `da366_measure_recovery.py` prints a loud SKIP and exits 0 (non-gating) when no corpus is present rather than emit a zero ("silent zero is not a measurement"), and its pure summarizers are unit-tested on synthetic rows. This is exactly the measurement-vs-assertion separation the domain requires.

**Reuse & integration.** Reuses the shared fail-loud `extract_narrator_mentions`, `extract_transmission_phrases`, `normalize_arabic`, and the SAME `_is_narrator_like` gate the other two mention paths apply (identical quality bar) — no segmentation logic reimplemented. The `_process_chunk`/`parse_sanadset` tuple widens 6→8 with every call site and the empty-rows early-return updated (no arity mismatch), `full_text_ar` preserved verbatim while the isnad is moved out of `matn_ar`, the mention branch correctly ordered before the `isnad_raw_ar`/`<NAR>` branch (so a recovered head is not re-parsed for absent tags), position_in_chain gap-free over accepted mentions, and both tag-free recovery counts excluded from the B3 firehose metric. `test_provenance_conformance` correctly registers the module as pure-text (no schema).

Must-fix: None
TechDebt: `boundary_precision` ships as a unit-tested precision instrument but is not wired into `main()` — the CLI report prints only recovery magnitude by convention (`summarize_recovery`), so the owner-gated re-run will surface the ~122k magnitude but not a precision figure. The recovery population is untagged by definition (no in-population ground truth), so precision can only be measured as a proxy over the tagged (has-SANAD) rows; wiring that proxy into the re-run report would strengthen the precision evidence for a domain where precision is the whole point. Non-blocking — the correctness the number depends on is already unit-proven.

Issue #452, opened 2026-07-18
narrator_split peels mentions onto phantom late-dated twin nodes (gen=unknown, impossible death years 546–693 AH)


parametrization:
## Summary

Found during Wave 25 da#347 investigation (Ivana): **206 canonical narrator nodes share a normalized name form with a split-off "twin" that carries `gen=unknown` and impossible death years (546–693 AH)** — yet those phantom late-medieval-dated twins still hold real mentions (up to 207 on a single node).

This looks like `narrator_split` peeling mentions onto phantom late-dated nodes: a genuine early narrator's mentions are being split onto a fabricated node with an impossible (centuries-too-late) date band. It is the inverse failure of the over-merge class — here the splitter is *over*-splitting onto date-incoherent nodes.

## Why it matters
- **Leaderboard-relevant:** phantom nodes with real mentions distort betweenness/pagerank and choke-point centrality (the exact metric the #928/#958 program cleans).
- Impossible death years (546–693 AH for companion-era narrators) are a hard data-quality signal that the split produced a node no rijāl authority could accept.

## Investigation pointers
- `src/resolve/mononym_split.py` / `narrator_split` plan — check the date-band assignment on peeled mentions; a peel that inherits/derives a late band is the suspect.
- Relates to da#439 (split↔loader adjacency, fixed PR#450) and da#346 (contextual peel, PR#451) — confirm this is not already addressed by those, and whether it needs a date-band coherence gate on peeled nodes.
- Quantify: the 206 nodes + total phantom-held mentions; are all 206 the same date-derivation bug or multiple classes?

## Provenance
Surfaced by Ivana Horvat during da#347 (Anas b. Mālik) Wave-25 investigation against the #928 re-run base (git_sha bd133e6, 107,951 canonical). Routed to Kavitha Sundaramurthy (narrator_split owner) for triage / overlap-check with da#439/#346.

PR #458, opened 2026-07-18
resolve: date-coherence abstain gate on narrator_split peels (da#452)

parametrization:
## Summary

Adds a **date-coherence abstain gate** (Gate 0) to `narrator_split.plan_split` so the splitter can no longer peel mentions onto phantom "twin" nodes with `gen=unknown` and impossible death years (546–693 AH) — the 206-node defect Ivana found on the re-run base (da#452).

## Root cause

`plan_split` trusts a chain neighbour's attested `death_year_ah` with **no plausibility bound**. An impossibly-late neighbour date (a chimeric/corrupt bio date — tracked separately as **da#454**) propagates into the band midpoint, and `_peeled_record` stamps a twin at that midpoint with `generation = _generation_for_year(midpoint)` = `UNKNOWN` (the ṭabaqa windows top out at 400 AH). The twin still holds the real mentions it peeled, distorting betweenness/choke-point centrality (the #928/#958 target metric). `mononym_split` is NOT implicated — it dates from a curated registry gated by `_temporally_plausible`.

## Fix — Gate 0 (date coherence)

A qualifying band whose midpoint maps to `NarratorGeneration.UNKNOWN` (outside every ṭabaqa window — impossibly late >400 AH or early <11 AH) does **not** qualify to peel; its mentions stay on the primary (peel-not-partition), exactly like a below-support band. The gate is tied to `_generation_for_year` — the **same** function that stamps a peeled node's generation — so the check and the stamp can never disagree and no peeled node can carry `gen=unknown`. Covers both the impossible-late AND impossible-early tails.

### Gate ordering (kept Gate 2 reachable)

Gate 2 (too-many-bands ⇒ generic bucket) now runs on the **raw** qualifying count *before* the coherence prune: a name throwing off >`SPLIT_MAX_CLUSTERS` date bands is an unresolvable bucket regardless of coherence, and evaluating it first keeps it reachable — a coherent-only count can never exceed the ~5 bands that fit the ṭabaqa window >`SPLIT_BAND_GAP` apart, which would have made Gate 2 dead code (the exact thing da#444 just removed). Order is now Gate 2 (raw) → Gate 0 (prune) → Gate 1 (≥2 coherent).

## Instrument-separation fixture (the merge-blocker)

Per the da#452 / da#423 discipline — prove the gate SEPARATES the classes, both sides:
- `test_gate0_out_of_window_band_abstains_no_phantom_twin`: d.150 + impossibly-late d.600 (UNKNOWN) → node stays whole, **no** phantom twin minted (RED without the gate — it would split and mint a 600 AH gen=unknown twin).
- `test_gate0_in_window_later_band_still_peels`: d.150 + in-window LATER d.340 (280–400) → **still peels** with a real generation — proving the gate does NOT false-abstain a genuine late-but-plausible narrator (the inverse erasure risk).

## Scope / magnitude

Ships the gate + fixture (unit-mechanic correctness). Magnitude — 206 nodes, total phantom-held mentions, single-class-or-not — is an owner-gated **re-run measurement**. The upstream impossibly-late attested neighbour dates are a **separate track (da#454)**, re-run prep; this output gate defends the split regardless of da#454.

## Checks

- New instrument-separation pair (above); existing `test_too_many_bands` / gate3-dead-code tests updated for the reorder + in-window scoping.
- Full resolve suite: 582 passed, 6 skipped. ruff + mypy clean; full pre-push green.

## Reviewers
- Peer: **Nikolaos Papadopoulos**

Closes #452
Part of main#928

### DIFF
--- a/src/resolve/narrator_split.py
+++ b/src/resolve/narrator_split.py
@@ -323,6 +323,21 @@ def _band_label(midpoint: int) -> str:
     return f"d:{(midpoint // 25) * 25}"
 
 
+def _band_is_date_coherent(band: list[DatableMention]) -> bool:
+    """True when the band's midpoint falls inside a known ṭabaqa generation window.
+
+    A midpoint outside every :data:`~src.resolve.tabaqa_dates._GENERATION_DEATH_WINDOW_AH`
+    window maps to :attr:`~src.models.enums.NarratorGeneration.UNKNOWN` — an impossibly
+    LATE (> ~400 AH) or impossibly EARLY (< 11 AH) death that no rijāl authority would
+    accept for a hadith-isnad narrator (da#452). Peeling such a band would mint a phantom
+    twin carrying ``gen=unknown`` + that impossible death year (the 546–693 AH twins Ivana
+    found) while still holding real mentions. Tied to :func:`_generation_for_year`, the
+    SAME function :func:`_peeled_record` uses to stamp a peeled node's generation, so the
+    coherence check and the stamp can never disagree.
+    """
+    return _generation_for_year(_band_midpoint(band)) is not NarratorGeneration.UNKNOWN
+
+
 def plan_split(name_ar_normalized: str, datable: list[DatableMention]) -> SplitPlan:
     """Decide whether/how to split candidate ``name_ar_normalized``; pure + deterministic.
 
@@ -338,11 +353,12 @@ def plan_split(name_ar_normalized: str, datable: list[DatableMention]) -> SplitP
     bands = _cut_bands(datable)
     qualifying = [b for b in bands if len(b) >= SPLIT_MIN_SUPPORT]
 
-    # Gate 1: need ≥2 well-supported bands. A single-band name (a genuinely single
-    # person like Sufyān al-Thawrī / al-Zuhrī) abstains here — the load-bearing guard.
-    if len(qualifying) < 2:
-        return abstain
-    # Gate 2: too many well-supported bands ⇒ generic bucket, not a few people.
+    # Gate 2: too many well-supported bands ⇒ generic bucket, not a few people. Evaluated
+    # on the RAW qualifying count (before the da#452 coherence prune below): a name that
+    # throws off more than SPLIT_MAX_CLUSTERS distinct date bands is an unresolvable
+    # bucket regardless of whether some of those bands are date-incoherent — and running
+    # it first keeps it reachable (a coherent-only count can never exceed the ~5 bands
+    # that fit the ṭabaqa window >SPLIT_BAND_GAP apart, which would make this dead code).
     if len(qualifying) > SPLIT_MAX_CLUSTERS:
         logger.info(
             "narrator_split_abstain_noise",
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved

**Review: Approved**

QA / data-quality lens (2nd reviewer) — da#452 Gate 0 date-coherence abstain. Reviewed diff + full `plan_split` at head `69fa051`; CI green 6/6. The erasure-risk direction is handled correctly.

1. **Both-sides instrument separation — genuine and non-vacuous (merge-blocker satisfied).**
   - Out-of-window side (`test_gate0_out_of_window_band_abstains_no_phantom_twin`): `d.150` (30 mentions) + `d.600` (15). `assert _generation_for_year(600) is UNKNOWN` proves the fixture is real, not a fabricated window. This is genuinely RED without the gate — pre-#458 `plan_split` had only Gate 1 + Gate 2, so two well-supported bands 450 AH apart would split and mint the 600 AH phantom twin; with Gate 0 the incoherent band drops → one coherent band → Gate 1 abstains. The instrument flips.
   - In-window LATER side (`test_gate0_in_window_later_band_still_peels`) — the critical inverse-erasure guard: `d.150` + `d.340` (∈ 280–400 LATER window). `assert _generation_for_year(340) is LATER` pins the fixture; the split STILL fires and peels the d.340 band. This is non-vacuous: a too-aggressive gate (e.g. dropping any late midpoint) would drop d.340 → abstain → this test fails. So the pair proves the gate separates the coherent-late from the incoherent-late, not just "abstains on late."

2. **Non-destructive, abstain at the peel decision — verified in the full function.** Gate 0 sets `qualifying = coherent`; `peel_bands` is derived from that coherent set, so a dropped incoherent band is never in `plan.peeled`. Its mentions are simply not on any peel list, so they stay on the primary (peel-not-partition, identical to below-support/undatable). No node is deleted — the out-of-window case returns `not is_split` (node whole). And because `peel_bands ⊆ coherent`, every peeled `midpoint_ah` is structurally in-window: a phantom-twin midpoint can never be minted (the out-of-window test's `all(... not UNKNOWN)` invariant holds by construction).

3. **Magnitude (206 nodes) correctly deferred.** No corpus number is asserted in code or tests — the gate's correctness is unit-proven on synthetic fixtures; the 206-node effect is a re-run measurement, not a faked CI value. No silent zero.

Also checked: the `test_removed_gate3_separation_is_dead_code` sweep range was tightened 0..301 → 0..281 so both bands stay in-window (120..400) and Gate 0 does not perturb that test — a principled adjustment (out-of-window is now Gate 0's separately-tested domain), not a weakening; 400 is the inclusive coherent boundary.

Must-fix: None
TechDebt: Gate 2 (too-many-bands) is intentionally evaluated on the RAW qualifying count before the Gate 0 coherence prune (documented, to keep Gate 2 reachable). The accepted cost: a name with ≤ SPLIT_MAX_CLUSTERS coherent bands but > SPLIT_MAX_CLUSTERS *raw* bands (i.e. several well-supported impossible-date bands pushing the raw count over 6) would abstain rather than split its coherent pair. Safe direction (abstain keeps the node whole + over_merged-flagged, no data loss) and effectively unreachable in a real corpus (needs 5+ well-supported bands at impossible death years), so non-blocking — worth a one-line note in the issue as an accepted limitation.

parametrization:
RequestOrReplied: Approved

**Review: Approved** (da#452 UNKNOWN-generation coherence gate, head 69fa051)

Verified all three focus areas by reading the code AND executing it, not just trusting green CI.

**1. Gate ordering (Gate 2 on RAW count, before the Gate 0 coherence prune) — reasoning holds.**
The union of the ṭabaqa windows is `[11, 400]` (SAHABI 11–110 … LATER 280–400, overlapping-contiguous). Bands must be `> SPLIT_BAND_GAP` (80 AH) apart, so the max number of *coherent* bands that can coexist is `floor((400-11)/81)+1 = 5` — strictly below `SPLIT_MAX_CLUSTERS` (6). So if Gate 0 pruned first, a coherent-only Gate 2 (`> 6`) could never fire → dead code, recreating exactly what da#444 removed. Running Gate 2 on the raw qualifying count (which incoherent bands *can* push past 6) keeps it reachable. Confirmed this is load-bearing for `test_too_many_bands_is_noise_abstain`: its 7 bands (120,220,…,720) include 4 out-of-window ones; under the shipped order it abstains at Gate 2 on the raw count of 7, but under a Gate-0-first order it would drop the 4 incoherent bands and split the 3 coherent ones — the test would flip. The shipped ordering is the correct one, and the abstain-the-whole-bucket choice for a 7-band name is the safe direction (leaves it flagged).
The `test_removed_gate3_separation_is_dead_code` sweep was correctly narrowed `0..301 → 0..281`: the late midpoint `120+sep` must stay `≤ 400` to remain coherent, else Gate 0 (not the gate3 logic) would change the outcome — sep 281+ is Gate 0's domain, covered by the new out-of-window test. No loss of gate3 coverage (the >80-vs-removed-50 separation claim is fully exercised across 81..280). Both reordered assertions still hold; 23 passed locally.

**2. Bidirectional instrument — both sides real, proven non-vacuous.**
I ran the A/B directly: the two fixtures are identical (`_dm(150,30) + _dm(late,15)`) differing ONLY in the late band (600 vs 340).
- WITH Gate 0: out-of-window d.600 → `is_split=False` (node stays whole, no phantom twin, `narrator_split_drop_incoherent_band dropped_midpoints=[600]` logged); in-window d.340 → `is_split=True`, peels one band at midpoint 340 with `gen=LATER`.
- WITHOUT Gate 0 (monkeypatched `_band_is_date_coherent → True`): out-of-window d.600 → `is_split=True` — it WOULD mint the d.600 phantom twin. So Gate 0 is genuinely load-bearing (RED-without-gate), and the in-window case proves no false-abstain of a genuine late narrator. Each test also asserts its fixture is real (`_generation_for_year(600) is UNKNOWN` / `(340) is LATER`), so neither is vacuous. This is a clean single-variable A/B — the drop-gate bidirectional discipline done right.

**3. Tie to `_generation_for_year` is sound.**
`_band_is_date_coherent` = `_generation_for_year(_band_midpoint(band)) is not UNKNOWN`, and `_peeled_record` stamps `generation = _generation_for_year(band.midpoint_ah)` using the SAME function; `band.midpoint_ah == _band_midpoint(band)` (confirmed empirically — the in-window test asserts `band.midpoint_ah == 340`). So a band that would stamp `gen=unknown` is dropped before Gate 1 — no peeled node can ever carry `gen=unknown`, and `test_gate0_out_of_window...` asserts exactly that over `plan.peeled`. The check and the stamp cannot disagree.

CI 6/6 green, mergeable. Magnitude (how many phantom twins this kills) is correctly deferred to the re-run; da#454 upstream track is out of scope here.

Must-fix: None
TechDebt: None

Clean, well-reasoned gate — the raw-count ordering comment is exactly the non-obvious thing a future reader needs, and the bidirectional fixture is a real instrument.

Issue #380, opened 2026-07-09
death_year_provenance is written and schema-declared but read by no production code


parametrization:
Found while reviewing PR #363 (da#356 / da#376).

`death_year_provenance` is written by `_upsert_canonical` (`src/resolve/disambiguate.py:1032`, `:1408`), declared in `NARRATORS_CANONICAL_SCHEMA` (`src/resolve/schemas.py:82`), and asserted in two tests. No production code reads it. The only `.get("death_year_provenance")` occurrences in `src/` are inside its own writer (the do-not-overwrite guard) and the Arrow array build.

Consequence: `fuzzy_cluster._death_years_conflict` weighs an `uncorroborated` death year with exactly the same authority as a `corroborated` one. An OCR-corrupt bio year attached by a weak Stage-2 fuzzy match can veto a merge against a corroborated year.

The direction is precision-safe (it blocks merges rather than causing them), and against `main` the column did not exist to be read, so this is not a regression introduced by PR #363. Tracking it so the column either grows a consumer or is removed.

Suggested: have `_death_years_conflict` (or `_can_merge`) discount an `uncorroborated` year, and add a guard test that goes red if the provenance tag is ignored.

parametrization:
Folding in a second, related item from the re-review of PR #363 at `333c4f9`, so it has a home rather than living in a review thread.

`tests/test_resolve/test_canonical_identity_invariant.py::TestInflectionFolding::test_every_id_producer_agrees` is misnamed. Its docstring says "the fold lives inside `make_canonical_id`, so all eight minting sites agree," but the body only asserts that `make_canonical_id` and `make_discriminated_canonical_id` agree with each other. It does not exercise any producer.

The property it claims is nonetheless true at `333c4f9` — an AST walk over `src/` finds zero `uuid5` narrator-id constructions outside `identity.py` and no `f"nar:..."` string building, so every producer does route through the folded mint. The test just is not what checks that.

Two options, either fine: rename it to what it tests (`test_both_mint_functions_agree`), or make it an AST assertion over `src/` that no module outside `identity.py` constructs a `nar:` id directly. The second is what would actually catch a new producer bypassing the mint.

Left as-is in PR #363 deliberately, to avoid churning the head under review.

parametrization:
Two debt items from the da#363 review (PR#363, head `333c4f9`), recorded here so they survive the merge. Neither blocks; neither costs a re-run.

[...]

PR #463, opened 2026-07-21
fix(#380): weight fuzzy_cluster death-year veto by death_year_provenance

parametrization:
## Summary

`death_year_provenance` was written by `_upsert_canonical` and schema-declared, but **no production code read it**. As a result `fuzzy_cluster._death_years_conflict` (`src/resolve/fuzzy_cluster.py:396`) weighed an `uncorroborated` death year with exactly the same full authority as a `corroborated` one — an OCR-corrupt year attached by a weak Stage-2 fuzzy match could veto a genuine merge.

This adds the missing **consumer**: the death-year veto is now weighted by provenance.

## What I verified at wave-branch HEAD (`f3edd06`)

- The premise **still holds**: `_death_years_conflict` read only `death_year_ah`, provenance-blind. `_can_merge` / `_can_merge_cached` / `_stateless_merge_ok` all route through it, so it is the single fix point.
- `disambiguate.py:1441-1444` already documents the design intent I'm honoring: an uncorroborated bio year "is still PERSISTED ... tagged `uncorroborated` — it just may not steer the ... identities" of other narrators (da#356).
- A **sibling consumer already exists** — `narrator_unify.py` (da#431, added after the issue's PR #363 review) uses a two-band model: `_corroborated_death_conflict` (tight, corroborated-only) + `_gross_death_spread` (loose band, any provenance). The issue's "no production code reads it" is now narrowly out of date, but the reported defect in `fuzzy_cluster` is real and unaddressed. I mirrored narrator_unify's model rather than invent new semantics.

## The fix

`_death_years_conflict` is now two-band:
- **Both years trusted** (`corroborated` or untagged `None`) → conflict at the tight `_DEATH_YEAR_TOLERANCE` (=2), i.e. **unchanged** from today.
- **Either year `uncorroborated`** → only a gross, generations-apart spread beyond the new `_UNCORROBORATED_DEATH_SPREAD` (=50, mirrors `narrator_unify._gross_death_spread`) blocks.

**Precision-safe:** this only *relaxes* a veto (fewer false blocks), never causes a merge. Corroborated-year behavior is preserved. Untagged (`None`) years keep today's full-authority veto, so nothing regresses for records the provenance column doesn't cover.

## Consumer vs removal

The issue offered removal (dropping the column) as an alternative. Removal is **not** viable: `narrator_unify` (da#431) already reads the column as a load-bearing safety-net input, so dropping it would regress that stage. Adding the `fuzzy_cluster` consumer is the correct direction.

## Instrument + mutation check

- New tests: `test_uncorroborated_death_year_does_not_veto_small_spread` (the RED-on-mutation guard), `test_uncorroborated_gross_death_spread_still_blocks`, `test_corroborated_death_years_keep_tight_tolerance`. Both predicate-level and end-to-end `cluster_records`.
- **Mutation check:** reverted the discount (`tolerance = _DEATH_YEAR_TOLERANCE` unconditionally) → the small-spread guard **failed** (`assert True is False`); restored → all green.
- Full parity green: `tests/test_resolve/` 611 passed / 6 skipped; ruff-format + ruff-lint + mypy(strict) clean; commit- and push-stage pre-commit hooks passed; sync-drift gate OK.

Closes #380

### DIFF
--- a/src/resolve/fuzzy_cluster.py
+++ b/src/resolve/fuzzy_cluster.py
@@ -115,6 +115,16 @@ _CLUSTER_RATIO_THRESHOLD = 90.0
 # are not the same person). One source rounding a death year by a year is fine.
 _DEATH_YEAR_TOLERANCE = 2
 
+# Discounted band for a death year tagged ``uncorroborated`` (da#380). A weak
+# Stage-2 fuzzy bio match persists its year on the record but marks it
+# ``uncorroborated`` because it "may not steer the ... identities" of other
+# narrators (disambiguate.py, da#356). An OCR-corrupt uncorroborated year off by
+# a handful of AH years must therefore NOT veto a merge at the tight
+# ``_DEATH_YEAR_TOLERANCE``; only a gross, generations-apart spread still blocks.
+# Mirrors ``narrator_unify._gross_death_spread``'s da#431 sanity band (=50) so the
+# resolve stage weighs an uncorroborated year the same way everywhere.
+_UNCORROBORATED_DEATH_SPREAD = 50
+
 # Minimum number of *shared* significant (non-connector) name tokens required to
 # merge. token_set_ratio scores a bare given name as a perfect (100) subset of
 # every fuller name carrying it (``محمد`` ⊂ ``محمد بن اسماعيل البخاري``), and a
@@ -394,11 +404,27 @@ def _name_match(keys_a: list[str], keys_b: list[str], *, threshold: float) -> bo
 
 
 def _death_years_conflict(a: dict[str, Any], b: dict[str, Any]) -> bool:
-    """True when both carry a death year that disagree beyond the tolerance."""
+    """True when both carry a death year that disagree beyond the tolerance.
+
+    da#380 — the veto is provenance-weighted. Two *trusted* years (``corroborated``
+    or untagged ``None``) conflict at the tight :data:`_DEATH_YEAR_TOLERANCE`, as
+    before. But when EITHER year is tagged ``uncorroborated`` — a weak Stage-2 fuzzy
+    bio year that disambiguate persists yet excludes from steering identity
+    (da#356) — a small disagreement (e.g. an OCR-corrupt year off by a few) must NOT
+    veto a real merge; only a gross, generations-apart spread beyond
+    :data:`_UNCORROBORATED_DEATH_SPREAD` still blocks. This keeps corroborated-year
+    precision intact while stopping an uncorroborated year from vetoing at full
+    authority. Mirrors ``narrator_unify``'s corroborated-vs-gross-spread split.
+    """
     da, db = a.get("death_year_ah"), b.get("death_year_ah")
-    if isinstance(da, int) and isinstance(db, int):
-        return abs(da - db) > _DEATH_YEAR_TOLERANCE
-    return False
+    if not (isinstance(da, int) and isinstance(db, int)):
+        return False
+    uncorroborated = (
+        a.get("death_year_provenance") == "uncorroborated"
+        or b.get("death_year_provenance") == "uncorroborated"
+    )
+    tolerance = _UNCORROBORATED_DEATH_SPREAD if uncorroborated else _DEATH_YEAR_TOLERANCE
+    return abs(da - db) > tolerance
 
 
 def _genders_conflict(a: dict[str, Any], b: dict[str, Any]) -> bool:
--- a/tests/test_resolve/test_fuzzy_cluster.py
+++ b/tests/test_resolve/test_fuzzy_cluster.py
@@ -130,6 +130,72 @@ def test_conflicting_death_years_block_merge() -> None:
     assert sorted(len(g) for g in clusters) == [1, 1]  # stayed apart
 
 
+# ---------------------------------------------------------------------------
+# da#380 — the death-year veto is weighted by `death_year_provenance`. An
+# `uncorroborated` year (a weak Stage-2 fuzzy bio year, persisted but excluded
+# from steering identity — disambiguate.py da#356) must NOT veto a real merge at
+# the tight `_DEATH_YEAR_TOLERANCE`; only a gross spread still blocks it. These
+# guards go RED if the provenance tag is ignored (the pre-da#380 behaviour).
+# ---------------------------------------------------------------------------
+def test_uncorroborated_death_year_does_not_veto_small_spread() -> None:
+    """A small disagreement with one `uncorroborated` year still merges.
+
+    Spread = 5 AH — beyond `_DEATH_YEAR_TOLERANCE` (2) but far under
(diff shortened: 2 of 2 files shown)


parametrization:
## Merge-gate review — PR #463 (fix da#380)

Reviewed at head `8ae8771c232ff051af350f0553d6a52ede19701a` via an isolated detached worktree (no branch switch in any shared checkout). Findings below cite symbols.

**The defect is real and was unfixed pre-PR.** `fuzzy_cluster._death_years_conflict` previously applied `_DEATH_YEAR_TOLERANCE` (=2) to *any* two int death years, so an OCR-corrupt `uncorroborated` Stage-2 year off by a few AH could veto a genuine merge at full authority. Confirmed against the base blob in the diff.

**The "add a consumer, not remove the column" direction is correct — verified independently.** `death_year_provenance` is read in production by `narrator_unify._corroborated_death_conflict` (`m.get("death_year_provenance") == "corroborated"`), written by `disambiguate.py`, and carried in `schemas.py`. So the column is load-bearing; removal was correctly ruled out.

**Band consistency vs `narrator_unify` — genuinely coherent, with one drift caveat.**
- Tight band: `narrator_unify` *imports* `_DEATH_YEAR_TOLERANCE` from `fuzzy_cluster` — cannot drift. Good.
- Gross band: this PR adds `_UNCORROBORATED_DEATH_SPREAD = 50`; `narrator_unify` has its own `_UNIFY_MAX_DEATH_SPREAD = 50`. These are two independent literals the comment asserts "mirror," with nothing enforcing equality. Non-blocking (the fix is correct today); filed as tech debt — see below.
- Note the two stages intentionally differ on *untagged* `None` years (fuzzy_cluster keeps the tight band, narrator_unify's corroborated-conflict requires both tagged `corroborated`). That asymmetry is pre-existing and correct — Stage-2 clustering stays conservative — and this PR preserves it.

**Single fix point, no bypass.** All three merge entry points route the death-year veto through `_death_years_conflict`: `_can_merge`, `_can_merge_cached`, and the stateless `_stateless_merge_ok`. No provenance-blind path remains.

**Precision — RELAX-only, corroborated + `None` byte-for-byte preserved.** The discount fires *only* when either side's provenance `== "uncorroborated"`; otherwise `tolerance = _DEATH_YEAR_TOLERANCE`, identical to pre-PR for corroborated, `None`, and any other tag. Widening the tolerance can only make `_death_years_conflict` return `False` in strictly more cases — it removes a veto, never creates one, and cannot by itself cause a merge the name/gender guards don't already accept.

**Mutation-tested the guard (not trusted).** Reverting the discount to an unconditional `_DEATH_YEAR_TOLERANCE` in my isolated checkout turns `test_uncorroborated_death_year_does_not_veto_small_spread` RED (`assert True is False`), while `test_uncorroborated_gross_death_spread_still_blocks` and `test_corroborated_death_years_keep_tight_tolerance` stay green — the discount guard is live, and the tight/gross guards pin the other two bands. Restored (worktree discarded).

**CI:** all six checks SUCCESS/COMPLETED at head — Lint, Type Check, Test, Integration Tests, Base-image pin + fixture realism, Pre-commit ⇄ CI sync-drift gate.

Tech debt filed: noorinalabs/noorinalabs-data-acquisition#464 (hoist the shared gross death-year spread band so the cross-stage "50" cannot drift). Non-blocking.

---
Requestee: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: #464

parametrization:
Correctness review of `_death_years_conflict` (`src/resolve/fuzzy_cluster.py` @ 8ae8771, head SHA).

**Verified against origin at head SHA (`gh api contents/... ?ref=8ae8771...`), not a shared checkout:**

1. **Defect premise is real, "no consumer" is stale as the PR body itself flags.** Pre-PR, `_death_years_conflict` read only `death_year_ah`, provenance-blind. `narrator_unify.py` (da#431) DOES already read `death_year_provenance` (`_corroborated_death_conflict` / `_gross_death_spread`), so removing the column (the issue's alternative) would have regressed that stage — confirmed via `cross-repo-graph.json` and direct read of `narrator_unify.py`. Keeping the column and adding the `fuzzy_cluster` consumer is correct.
2. **Change only relaxes, never causes, a veto.** `tolerance` can only grow (2 → 50) when either side is `uncorroborated`; `abs(diff) > tolerance` can only flip True→False, never the reverse. Corroborated/corroborated and untagged-`None`/untagged-`None` pairs keep the unchanged tight `_DEATH_YEAR_TOLERANCE=2` path — confirmed by reading the diff logic directly.
3. **Single fix point.** `cross-repo-graph.json` confirms all three merge entry points (`_can_merge`, `_can_merge_cached`, `_stateless_merge_ok`) route through the one patched predicate — no parallel/duplicated veto logic to miss.
4. **Constant coherence with `narrator_unify`.** `_UNCORROBORATED_DEATH_SPREAD=50` matches `narrator_unify._UNIFY_MAX_DEATH_SPREAD=50` exactly (not a drifting lookalike). One nuance worth naming (non-blocking): `narrator_unify`'s tight band requires BOTH years explicitly `"corroborated"` — an untagged/`None` year there falls through to the loose 50-band backstop regardless of provenance (its docstring says so explicitly: "regardless of provenance"). This PR's tight band instead treats `None` as trusted alongside `"corroborated"`. That's not a bug — it's the PR body's explicitly stated, more conservative scoping decision ("nothing regresses for records the provenance column doesn't cover"), and it's correctly documented in the docstring rather than silently claimed as an exact mirror. Flagging only so it doesn't read as a literal 1:1 port if someone diffs the two later.

**Instrument + mutation check (run myself, isolated worktree at head SHA — not the implementer's or a shared checkout):**
- Clean HEAD: `tests/test_resolve/test_fuzzy_cluster.py` → 25 passed.
- Mutated `tolerance = _UNCORROBORATED_DEATH_SPREAD if uncorroborated else _DEATH_YEAR_TOLERANCE` down to unconditional `_DEATH_YEAR_TOLERANCE` → `test_uncorroborated_death_year_does_not_veto_small_spread` went RED (`assert True is False`) as expected; all other 24 tests still green (confirms the other two new tests aren't accidentally exercising the same code path as false-positive padding).
- Restored, reran → 25 passed again. Worktree removed after use.
- Note: found and used a stale worktree at this exact head SHA that already had an un-reverted mutation applied mid-check (the #1056/#1058 shared-worktree hazard) — did not trust it, built and used a fresh isolated worktree from `pull/463/head` for the actual verification above.

**CI at head (8ae8771):** all 6 checks green — Lint, Type Check, Test, Integration Tests, Base-image pin + fixture realism, Pre-commit ⇄ CI sync-drift gate (`gh pr checks 463`).

No must-fix findings. No tech-debt to file.

---
Requestor: Oyunbileg Batbayar
Requestee: Kwesi Boateng
RequestOrReplied: Approved
TechDebt: none

Issue #464, opened 2026-07-21
tech-debt: share the resolve-stage gross death-year spread band (50) between fuzzy_cluster and narrator_unify


parametrization:
## Problem

The resolve stage now encodes the "gross death-year spread" sanity band as **two independent literals** that a code comment asserts must stay equal, but nothing enforces:

- `src/resolve/fuzzy_cluster.py` — `_UNCORROBORATED_DEATH_SPREAD = 50` (added in PR #463 / da#380)
- `src/resolve/narrator_unify.py` — `_UNIFY_MAX_DEATH_SPREAD = 50` (da#431)

The fuzzy_cluster comment reads: "Mirrors `narrator_unify._gross_death_spread`'s da#431 sanity band (=50) so the resolve stage weighs an uncorroborated year the same way everywhere." That cross-stage invariant is real, but it is a **manual** mirror — retuning one literal (e.g. narrator_unify's 50 → 40) silently breaks "the same way everywhere" with no test or import to catch it.

By contrast the *tight* tolerance is already shared correctly: `narrator_unify` imports `_DEATH_YEAR_TOLERANCE` from `fuzzy_cluster`, so that half cannot drift. Only the gross-spread band (50) is duplicated.

## Why not fixed in-place

`narrator_unify` already imports from `fuzzy_cluster`, so `fuzzy_cluster` cannot import `_UNIFY_MAX_DEATH_SPREAD` back without a circular import. The clean fix is to hoist the shared resolve-stage death-year constants (tolerance + gross spread) into a single small module both stages import.

## Scope

- Extract `_DEATH_YEAR_TOLERANCE` and the gross-spread band (=50) into a shared constants location.
- Point both `fuzzy_cluster` and `narrator_unify` at it.
- Non-behavioral refactor; existing da#380 / da#431 guard tests must stay green.

Surfaced during the merge-gate review of PR #463 (da#380).

parametrization:
Requestor: Fatima Okonkwo
Requestee: Ivana Horvat
RequestOrReplied: Request

**Wave 27 Kickoff — Phase 9**

This issue is assigned to you for p9-wave-27.
- Peer reviewer: Alejandra Reyes-Fuentes
- Secondary reviewer: Nikolaos Papadopoulos
- Branch from: `deployments/phase-9/wave-27`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
[...]

PR #481, opened 2026-07-22
tech-debt: share the resolve-stage gross death-year spread band

parametrization:
## Summary

`fuzzy_cluster._UNCORROBORATED_DEATH_SPREAD` (da#380 / PR #463) and `narrator_unify._UNIFY_MAX_DEATH_SPREAD` (da#431) were two independent `= 50` literals. A code comment on each asserted they must stay equal ("mirrors ... so the resolve stage weighs an uncorroborated year the same way everywhere"), but nothing enforced it — retuning one silently breaks the cross-stage invariant.

This hoists both the tight tolerance (`_DEATH_YEAR_TOLERANCE`) and the gross-spread band (`= 50`) into a new shared module, `src/resolve/_death_year_bands.py` (`DEATH_YEAR_TOLERANCE`, `GROSS_DEATH_SPREAD`), and points both `fuzzy_cluster` and `narrator_unify` at it directly (rather than one importing transitively through the other).

- `narrator_unify` already imported `_DEATH_YEAR_TOLERANCE` from `fuzzy_cluster` (that half couldn't drift); it now imports it from the shared module instead.
- The gross-spread band, previously duplicated as two `= 50` literals, is now a single value both stages alias locally (`_UNCORROBORATED_DEATH_SPREAD` / `_UNIFY_MAX_DEATH_SPREAD` keep their existing local names — and their existing worked-example doc comments — but both now read from the one shared constant).

Non-behavioral refactor: no logic, thresholds, or public names changed. da#380 and da#431 guard tests are unmodified and stay green.

Closes #464

## Test plan
- [x] `ENVIRONMENT=test uv run pytest tests/test_resolve/test_fuzzy_cluster.py tests/test_resolve/test_narrator_unify.py -q` — 45 passed
- [x] `pre-commit run --all-files` (commit-stage) — all green
- [x] `pre-commit run --all-files --hook-stage pre-push` (mypy-strict + full pytest suite) — all green

## Reviewers
- Jean-Claude Habimana (merge-gate)
- Oyunbileg Batbayar

### DIFF
--- a/src/resolve/_death_year_bands.py
+++ b/src/resolve/_death_year_bands.py
@@ -1,0 +1,33 @@
+"""Shared resolve-stage death-year sanity bands (da#464).
+
+``fuzzy_cluster`` (da#380) and ``narrator_unify`` (da#431) each veto a
+merge/unify decision on a death-year conflict, weighed at two bands:
+
+* :data:`DEATH_YEAR_TOLERANCE` — the tight band. Two *trusted* years
+  (corroborated, or an untagged/legacy record) disagreeing by more than this
+  many AH years block the merge outright; one source rounding a year by a
+  year or two is fine, a whole generation apart is not.
+* :data:`GROSS_DEATH_SPREAD` — the loose band. A much wider sanity check that
+  still catches a distinct namesake even when neither year carries full
+  trust — an ``uncorroborated`` fuzzy-bio year in ``fuzzy_cluster``, or a
+  curated unify-group member in ``narrator_unify``.
+
+Both stages must weigh an uncorroborated/noisy year the same way — that
+cross-stage invariant used to be two independent literals a code comment
+merely *promised* stayed equal (one retune away from silent drift). Hoisting
+both constants here makes the invariant structural: there is only one place
+to change either band, and both stages import it.
+"""
+
+from __future__ import annotations
+
+# Two *trusted* death years (corroborated, or untagged/legacy) disagreeing by
+# more than this many AH years block a merge/unify outright.
+DEATH_YEAR_TOLERANCE = 2
+
+# Loose sanity band (in AH years) applied when at least one death year is not
+# fully trusted (fuzzy_cluster's `uncorroborated` tag; narrator_unify's
+# curated-but-noisy group members). Wide enough to pass a genuine
+# generation-default noise gap, tight enough to still refuse a
+# cross-generation namesake.
+GROSS_DEATH_SPREAD = 50
--- a/src/resolve/fuzzy_cluster.py
+++ b/src/resolve/fuzzy_cluster.py
@@ -92,6 +92,7 @@ from src.resolve._checkpoint import (
     resolve_cadence,
     save_checkpoint,
 )
+from src.resolve._death_year_bands import DEATH_YEAR_TOLERANCE, GROSS_DEATH_SPREAD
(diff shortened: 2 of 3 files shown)


parametrization:
## Merge-gate review — APPROVE

Reviewed the diff at head `97c7ec3` against `deployments/phase-9/wave-27`. Pure non-behavioral refactor: hoists the two resolve-stage death-year bands into a new shared `src/resolve/_death_year_bands.py` (`DEATH_YEAR_TOLERANCE = 2`, `GROSS_DEATH_SPREAD = 50`) and points `fuzzy_cluster` and `narrator_unify` at it.

**Value-equivalence verified by inspection:** the diff shows `_DEATH_YEAR_TOLERANCE` (was `2`), `_UNCORROBORATED_DEATH_SPREAD` (was `50`), and `_UNIFY_MAX_DEATH_SPREAD` (was `50`) each now alias the shared constant carrying the identical literal — no threshold, logic, or public name changed. The local names both stages export are preserved, so any external consumer of `fuzzy_cluster._DEATH_YEAR_TOLERANCE` / `_UNCORROBORATED_DEATH_SPREAD` still resolves. `narrator_unify` now imports `_DEATH_YEAR_TOLERANCE` from the shared module directly instead of transitively through `fuzzy_cluster` — that half already couldn't drift; this change makes the gross-spread band structurally single-sourced too, replacing a comment that merely *promised* two `= 50` literals stayed equal.

This is the correct fix for da#464: the cross-stage invariant is now enforced by construction (one place to retune either band) rather than by a comment. The da#380 / da#431 guard tests are unmodified — there is no new assertion to mutation-test here; safety rests on the provable value-identity plus the full suite (2516 passed) and mypy staying green at head across all modules, which would catch any import breakage. Confirmed green.

No must-fix items.

---
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

parametrization:
Reviewed at head `97c7ec3` against base `deployments/phase-9/wave-27`.

Verified this is a genuine non-behavioral hoist, not a value change:

- The new leaf module `src/resolve/_death_year_bands.py` keeps both bands at their existing values (`DEATH_YEAR_TOLERANCE = 2`, `GROSS_DEATH_SPREAD = 50`), and imports only `from __future__` — no cycle risk with either consumer.
- Both stages still consume the constants in real veto logic, not just define them: `fuzzy_cluster` line 492 (`tolerance = _UNCORROBORATED_DEATH_SPREAD if uncorroborated else _DEATH_YEAR_TOLERANCE`), `narrator_unify` line 236 (`abs(yi - yj) > _DEATH_YEAR_TOLERANCE`) and line 248 (`hi - lo > _UNIFY_MAX_DEATH_SPREAD`).
- `narrator_unify` previously pulled `_DEATH_YEAR_TOLERANCE` transitively through `fuzzy_cluster`; it now imports both bands directly from the shared module, so the invariant is structural (one place to retune) instead of a comment promising two `= 50` literals stay equal.
- Grep of both modules confirms no stray death-spread literal was missed — the only remaining `= 50` is the unrelated `top_k` parameter.

CI is green on all six checks at head (Lint, Type Check, Test, Integration Tests, base-image/fixture-realism, pre-commit⇄CI sync gate). The da#380/da#431 guard tests are untouched.

Approving.

---
Requestee: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

Issue #453, opened 2026-07-18
tech-debt: strengthen da#439 adjacency invariant (multi-isnad coverage + same-position slot location)


parametrization:
## Summary

Two minor robustness/coverage gaps surfaced during review of PR#450 (da#439, narrator_split ↔ loader `TRANSMITTED_TO` adjacency reconciliation). Neither is a correctness bug on real data; filing so they are tracked, not lost (per reviewer verdict on #450).

## 1. Cross-function adjacency invariant is asserted on a single chain only

`tests/test_resolve/test_narrator_split.py::test_split_adjacency_matches_loader_transmitted_to` runs the production `load_edges._build_chain_pairs` and the splitter path (`_build_chain_index` + `_collect_datable`) on one `(hadith_id, chain_index)` chain and asserts their neighbour sets agree. It calls `_build_chain_pairs` **directly** — which is the per-chain pairing unit — but production applies it *after* grouping by the composite `(hadith_id, chain_index)` key in `_load_transmitted_to`. So the invariant is proven for the gappy single-chain case but not across the multi-isnad grouping boundary.

**Fix:** add a multi-isnad agreement fixture (one `hadith_id`, ≥2 `chain_index` values, a candidate in each) asserting the splitter's per-chain confinement and the loader's grouped pairing still agree — extending the da#439 invariant to the grouping boundary. The splitter side is already covered in isolation by `test_multi_isnad_no_cross_chain_neighbour`; this joins the two.

## 2. `_collect_datable` locates a mention's slot by position value

`_collect_datable` finds a mention's slot with `next(i for i,(p,_) in enumerate(chain) if p == position)` — by `position_in_chain` **value**. If two resolved mentions ever share a position in one chain (a data-quality defect; `position_in_chain` should be unique per isnad), this returns the first matching slot and can mislocate `own_id` and the ±1 neighbours. The chain tuples carry only `(position, canonical_id)`, no `mention_id`, so there is no in-band way to disambiguate same-position rows today.

**Fix (if ever needed):** either assert position-uniqueness per `(hadith_id, chain_index)` when building `chains` (fail loud on the data defect), or carry `mention_id` into the chain tuple so the candidate's own slot is located exactly. Marginal — gated on a position-uniqueness violation that should not occur.

## Context

[...]

PR #483, opened 2026-07-22
test(#453): extend da#439 adjacency invariant across the multi-isnad grouping boundary

parametrization:
Closes #453.

## Summary

Tier 3 resolve closeout, non-output-changing per scope disposition. `test_split_adjacency_matches_loader_transmitted_to` (da#439) proves the splitter/loader adjacency agreement by calling `_build_chain_pairs` directly on a single chain's mentions — the per-chain pairing unit. Production never calls it that way: `_load_transmitted_to` first buckets mentions by the composite `(hadith_id, chain_index)` key (`by_chain`, keyed via `_chain_index`) and calls `_build_chain_pairs` once **per group**. The single-chain test cannot reach that grouping step, so a regression that flattened two isnads together before pairing would slip through it while still corrupting a real multi-isnad hadith (da#411's failure class).

Adds `test_split_adjacency_matches_loader_transmitted_to_multi_isnad`: one `hadith_id`, two `chain_index` chains (one gappy, one contiguous, overlapping position values), the candidate present in both. Builds the loader side by mirroring `_load_transmitted_to`'s actual grouping (bucket by `(hadith_id, _chain_index(row))`, pair each bucket), and the splitter side by running `_build_chain_index`/`_collect_datable` over both chains at once — then asserts each of the candidate's two mentions agrees with its own chain's grouped-loader neighbours and never leaks into the other chain's.

Manually verified the fixture is not vacuous: pairing the same mentions *without* grouping first fabricates a cross-chain edge and silently drops the gappy chain's candidate mention from the neighbour set — confirming this fixture exercises a real hazard class the single-chain test doesn't reach.

Item 2 from #453 (`_collect_datable`'s position-value slot lookup) is explicitly marginal/"if ever needed" in the issue and gated on a data defect that shouldn't occur — left untouched per the non-output-changing disposition.

## Test plan

- [x] `pytest tests/test_resolve/test_narrator_split.py -v` — 31/31 passed (new test included)
- [x] Full non-integration suite (`pre-push` command): 2516 passed, 10 skipped, 0 failed
- [x] `pre-commit run --all-files`: all hooks passed
- [x] `mypy src/`: no issues
- [x] Manual sanity check confirming the fixture distinguishes grouped-correct from ungrouped-wrong pairing (not vacuous)

Reviewers: Nikolaos Papadopoulos (merge-gate) + Alejandra Reyes-Fuentes.

### DIFF
--- a/tests/test_resolve/test_narrator_split.py
+++ b/tests/test_resolve/test_narrator_split.py
@@ -906,3 +906,93 @@ def test_split_adjacency_matches_loader_transmitted_to(tmp_path: Path) -> None:
 
     # The two notions of adjacency AGREE (the da#439 invariant).
     assert split_neighbours == loader_neighbours
+
+
+def test_split_adjacency_matches_loader_transmitted_to_multi_isnad(tmp_path: Path) -> None:
+    """da#453: extend the da#439 cross-function invariant across the multi-isnad
+    grouping boundary.
+
+    `test_split_adjacency_matches_loader_transmitted_to` proves splitter/loader
+    agreement by calling `_build_chain_pairs` directly on one chain's mentions — the
+    per-chain pairing unit. Production never calls it that way: `_load_transmitted_to`
+    first buckets mentions by the composite ``(hadith_id, chain_index)`` key
+    (``by_chain``, keyed via `_chain_index`) and calls `_build_chain_pairs` once PER
+    GROUP. The single-chain test can't reach that grouping step at all, so a defect
+    that flattened two isnads together before pairing would pass it and still corrupt
+    a multi-isnad hadith (da#411's failure class).
+
+    One ``hadith_id``, two ``chain_index`` chains, the candidate present in both —
+    mirrors `test_multi_isnad_no_cross_chain_neighbour` but cross-checks the
+    splitter against the loader's actual grouped-pairing path (built the same way
+    `_load_transmitted_to` builds it) instead of the splitter alone.
+    """
+    from src.graph.load_edges import _build_chain_pairs, _chain_index
+    from src.resolve.narrator_split import _build_chain_index, _collect_datable
+
+    cand = make_canonical_id(_ABU_ABDALLAH)
+    teacher_a = _anchor_row("nar:teacher_a", 100)
+    student_a = _anchor_row("nar:student_a", 180)
+    teacher_b = _anchor_row("nar:teacher_b", 250)
+    student_b = _anchor_row("nar:student_b", 320)
+    canonical = [
+        _canonical_row(cand, _ABU_ABDALLAH, 2),
+        teacher_a,
+        student_a,
+        teacher_b,
+        student_b,
+    ]
+    hid = "hdt:bukhari:9"
+    mention_rows = [
+        # Chain 0 — gappy (mirrors da#439): candidate at position 2, two dropped
+        # unresolved mentions opening a gap to each attested neighbour.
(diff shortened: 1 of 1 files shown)


parametrization:
## Merge-gate review — APPROVE

Reviewed the diff at head `bfc2a20` against `deployments/phase-9/wave-27`. Test-only change: adds `test_split_adjacency_matches_loader_transmitted_to_multi_isnad`. CI is green at head (Lint / Type Check / Test / Integration / fixture-realism / sync-drift all SUCCESS).

**Guard is non-vacuous — verified by construction, not just by the author's claim.** The splitter side runs the real production functions (`_build_chain_index` / `_collect_datable`) over both `chain_index` chains mixed in one parquet, exactly as the stage does. Because the two isnads reuse overlapping position values (0/1/2), a regression in the da#411 class — flattening the two chains before pairing instead of bucketing by `(hadith_id, _chain_index)` first — would fold chain-1 neighbours into `by_mention["m-cand-0"]`, and the equality assertion `by_mention["m-cand-0"] == loader_neighbours_by_chain[0]` (== `{teacher_a, student_a}`) would fail. So the assertion is genuinely trippable; the disjointness assertions (`not (by_mention[...] & loader_neighbours_by_chain[other])`) add a second independent failure surface for the same defect. This is a real extension of the da#439 invariant across the grouping boundary the single-chain test cannot reach.

The loader side is built by mirroring `_load_transmitted_to`'s actual grouping (`by_chain.setdefault((hid, _chain_index(row)), []).append(row)`, pair per group) rather than re-implementing it, so the two sides are cross-checked against production code on both ends. Item 2 of da#453 correctly left untouched per the non-output-changing disposition.

No must-fix items.

---
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

parametrization:
Reviewed at head `bfc2a20` against base `deployments/phase-9/wave-27`.

Test-only addition. The value of this PR is entirely in whether the new guard is non-vacuous, so I checked it against the real production code rather than the description:

- `src/resolve/narrator_split.py` `_build_chain_index` keys chains on the composite `(hadith_id, chain_index)` (via `_coalesce_chain_index`, mirroring `load_edges._chain_index`), and `_collect_datable` → `resolved_chain_neighbours` confines each mention's neighbours to its own `(hadith_id, chain_index)` chain. The new test drives those real functions over both chains at once, so it exercises exactly the grouping step the single-chain `test_split_adjacency_matches_loader_transmitted_to` (da#439) can't reach.
- The fixture is non-vacuous against the stated regression class (flatten two isnads before pairing): the two chains carry overlapping position values (0/1/2), so a splitter that keyed on `hadith_id` alone — or paired the flattened mention list — would collide the isnads and cross-contaminate the candidate's neighbour sets. The four assertions then fail: `m-cand-0`/`m-cand-1` would no longer equal their own chain's grouped-loader neighbours, and the two disjointness asserts (`not (by_mention[...] & other_chain)`) would trip. Under the correct grouped code they pass, which the green Test check confirms.
- The loader reference side is built by mirroring `_load_transmitted_to`'s actual `by_chain.setdefault((hid, _chain_index(row)), [])` bucketing, so the splitter is cross-checked against a correctly-grouped loader, not the splitter alone.

Item 2 of #453 is correctly left untouched per the non-output-changing disposition. CI green on all six checks at head.

Approving.

---
Requestee: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

Issue #454, opened 2026-07-18
Narrator nodes carry impossibly-late attested death_year_ah (~500–740 AH) — unbounded bio dates propagate into narrator_split (da#452 upstream)


parametrization:
## Summary

Layer-2 finding from the da#452 triage (Kavitha): some narrator nodes carry an **impossibly-late attested `death_year_ah`** — companion/successor-era narrators stamped with death years ~500–740 AH. `_attested_death` only checks `precision != estimate`; it does **not** bound the year against a plausible horizon, so these corrupt values are treated as authoritative attested dates.

## Impact
- These late attested dates flow into `narrator_split` band midpoints (a neighbour's attested death ± MID_GAP), producing the phantom late-dated twin nodes tracked in da#452. da#452 adds an **output** coherence gate on the split side (defends regardless); THIS issue is the **input/source** bug — the late dates should not exist on the nodes in the first place.
- Likely the same corrupt-bio source as the (closed) da#356 chimera class — a mention fuzzy-matched to an OCR/variant-corrupt bio candidate that also carried a wrong/implausible death date — but da#356 fixed the name-keying, not the date. Distinct manifestation, separate fix.

## Investigation pointers
- `src/resolve/` attested-date handling (`_attested_death` and where `death_year_ah` + `precision` are stamped from bio candidates).
- Quantify: how many nodes carry `death_year_ah` outside a plausible horizon (say < 10 AH or > ~400 AH for the corpus era); are they concentrated in the crossref/fuzzy-matched bio class (da#356 residue)?
- Fix direction: bound attested dates to a plausible horizon at stamp time (reject/flag out-of-horizon bio dates), OR trace to the corrupt bio records and exclude them.

## Provenance / relationships
- Surfaced by Kavitha Sundaramurthy in the da#452 triage (narrator_split phantom late-dated twins). da#452 (Wave 25) is the split-side gate; THIS is the upstream data-quality fix. Related closed issue: da#356 (chimera name-keying). Corpus not local (base bd133e6) — magnitude is a re-run measurement.
- Not gating the Phase 9 re-run on its own (da#452's gate makes the split robust to it); scope for re-run prep / a later wave.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Alejandra Reyes-Fuentes
RequestOrReplied: Request

**Wave 27 Kickoff — Phase 9**

This issue is assigned to you for p9-wave-27.
[...]

PR #484, opened 2026-07-22
fix(da#454): bound attested death_year_ah to the isnad-plausibility envelope at stamp time

parametrization:
## Summary

Closes #454.

Narrator nodes were carrying impossibly-late attested `death_year_ah`
(~500-740 AH) — companion/successor-era narrators stamped with a late-
collector/commentator's death year. da#452 (Wave 25) added the **output**-side
gate (`narrator_split._attested_death`) that keeps the split's neighbour-death
evidence clean regardless, but the corrupt values were still landing on
`narrators_canonical.parquet` in the first place, from three independent
producers that each wrote a bio's raw `death_year_ah` with no plausibility
check:

- `disambiguate._load_candidates` — feeds `_temporal_filter` /
  `_bio_corroborated`, so a bad year could sway a match/corroboration decision
  and (via `death_year_index`) narrator_split's own neighbour-death evidence.
- `bio_promote`'s new-node and backfill paths.
- `date_reconcile.reconcile_canonical_dates` — the actual authoritative
  producer of the final canonical `death_year_ah`/bounds/precision; it
  re-derives these straight from the raw per-source bio rows independent of
  what `bio_promote`/`disambiguate` wrote, and would silently re-introduce a
  value the earlier stages had scrubbed.

Also: `fuzzy_cluster`'s ±2-year death-year merge guard runs *before*
`date_reconcile` in the pipeline (`RESOLVE_STEP_ORDER`) and has no plausibility
gate of its own, so an unscrubbed value from `bio_promote`/`disambiguate` could
wrongly block or allow a merge there too.

## Fix

Added `tabaqa_dates.plausible_attested_death_year` — a thin, reused wrapper
around the existing `is_plausible_narrator_death_ah` + `generation_from_value`
(the da#446 envelope) — and call it at every stamp site above:

- `disambiguate._load_candidates`: the `Candidate.death_year_ah` loaded from a
  bio row is bound-checked against the row's own `generation`.
- `bio_promote`: both the new-canonical-node path and the existing-record
  backfill path use the bound-checked value.
- `date_reconcile.reconcile_canonical_dates`: after per-source reconciliation,
  the reconciled death *point* is bound-checked against the record's
  `generation`; when implausible, the **whole** death dating (point + both
  bounds + precision) is scrubbed back to undated — not just the point — so
  the narrator reads as genuinely undated for the ṭabaqa fallback stage that
  runs immediately after, instead of carrying a null point next to stale
  corrupt bounds/precision.

`narrator_split._attested_death` (da#446) is left unchanged — it stays as
deliberate defense-in-depth for a pre-existing curated corpus that predates
this fix, per the issue's own framing (this closes the *input*-side gap; the
magnitude is a re-run measurement, not a live-corpus hotfix).

## Test plan

- [x] New unit tests for `plausible_attested_death_year` (valid year, `None`
      passthrough, absolute-envelope violation, generation-mismatch scrub) in
      `tests/test_resolve/test_tabaqa_dates.py`.
- [x] New tests for `bio_promote`'s new-node and backfill scrub paths (plus a
      plausible-year regression) in `tests/test_resolve/test_bio_promote.py`.
- [x] New tests for `disambiguate._load_candidates`' scrub in
      `tests/test_resolve/test_disambiguate.py`.
- [x] New end-to-end tests for `date_reconcile.reconcile_canonical_dates`'s
      scrub — generation-aware, absolute-envelope, and a plausible-year
      regression — in `tests/test_resolve/test_date_reconcile.py`.
- [x] Full existing suite green: `uv run pytest tests/ -m "not integration"`
      (2527 passed, 10 skipped, 58 deselected — integration tests need Docker).
- [x] `pre-commit run --all-files` and `--hook-stage push` (ruff-format,
      ruff-lint, gitleaks, actionlint, cspell, mypy-strict, pytest) all green.

Reviewers: Jean-Claude Habimana (merge-gate) + Ivana Horvat

### DIFF
--- a/src/resolve/bio_promote.py
+++ b/src/resolve/bio_promote.py
@@ -44,6 +44,7 @@ from src.resolve.sect_affiliation import (
     normalize_corpus,
     primary_corpus,
 )
+from src.resolve.tabaqa_dates import plausible_attested_death_year
 from src.utils.arabic import normalize_arabic
 from src.utils.logging import get_logger
 
@@ -52,10 +53,11 @@ logger = get_logger(__name__)
 __all__ = ["promote_bios_to_canonical"]
 
 # Scalar bio fields back-filled onto an existing canonical record when missing.
+# ``death_year_ah`` is handled separately (da#454, plausibility-bound at stamp
+# time) rather than through this generic pass-through.
 _BACKFILL_FIELDS = (
     "name_en",
     "birth_year_ah",
-    "death_year_ah",
     "generation",
     "gender",
     "trustworthiness",
@@ -308,6 +310,14 @@ def promote_bios_to_canonical(
 
             bio_id = safe_str(row.get("bio_id"))
             cid = make_canonical_id(norm)
+            # da#454: bound the bio's attested death year against the isnad-narrator
+            # plausibility envelope *before* it can be written anywhere — a raw
+            # death_year_ah outside the envelope (the d. ~700-780 AH late-collector
+            # pollution) becomes None rather than landing on a fresh node or
+            # back-filling an existing one.
+            death_year = plausible_attested_death_year(
+                row.get("death_year_ah"), row.get("generation")
+            )
 
             # When `cleaner_removed_content` is True the cleaner removed name-span
             # material beyond its affix normalizations. That covers TWO classes the
@@ -409,7 +419,7 @@ def promote_bios_to_canonical(
                     "name_ar_normalized": norm,
                     "aliases": [],
                     "birth_year_ah": row.get("birth_year_ah"),
-                    "death_year_ah": row.get("death_year_ah"),
+                    "death_year_ah": death_year,
                     "generation": safe_str(row.get("generation")),
                     "gender": safe_str(row.get("gender")),
                     "trustworthiness": safe_str(row.get("trustworthiness")),
@@ -441,6 +451,11 @@ def promote_bios_to_canonical(
                 for field_name in _BACKFILL_FIELDS:
                     if rec.get(field_name) in (None, "") and row.get(field_name) not in (None, ""):
                         rec[field_name] = row.get(field_name)
+                # death_year_ah backfills from the plausibility-bound value (da#454),
+                # never the raw row — an implausible attested year must not sneak
+                # onto an existing (even zero-mention) record either.
+                if rec.get("death_year_ah") in (None, "") and death_year is not None:
+                    rec["death_year_ah"] = death_year
             promoted += 1
             promoted_cids.add(cid)
 
--- a/src/resolve/date_reconcile.py
+++ b/src/resolve/date_reconcile.py
@@ -86,6 +86,7 @@ from src.parse.narrator_dates import (
 )
 from src.resolve._run_record import write_canonical
 from src.resolve.schemas import NARRATORS_CANONICAL_SCHEMA
+from src.resolve.tabaqa_dates import generation_from_value, is_plausible_narrator_death_ah
 from src.utils.arabic import normalize_arabic
 from src.utils.hijri import ah_year_to_ce_range
 from src.utils.logging import get_logger
@@ -510,6 +511,25 @@ def reconcile_canonical_dates(
         obs_slot = obs_by_cid.get(cid) if cid else None
         if obs_slot is not None:
             merged = reconcile_narrator(obs_slot["birth"], obs_slot["death"])
+            # da#454: a reconciled DEATH point outside the isnad-narrator plausibility
+            # envelope (the d. ~700-780 AH late-collector/commentator pollution welded
+            # onto an early node — see tabaqa_dates.is_plausible_narrator_death_ah) must
+            # never be written onto the canonical record. Reconciliation folds per-
+            # source OBSERVATIONS (agreement/conflict/noise-gap) but has no concept of
+            # "impossible for THIS narrator's generation" — a single uncontested corrupt
+            # source sails straight through as a "conflict-free" EXACT point. Scrub the
+            # WHOLE death dating (point + both bounds + precision), not just the point:
+            # leaving corrupt bounds/EXACT precision behind a null point would read as
+            # "dated" to tabaqa_dates' undated-check and skip the ṭabaqa fallback that
+            # runs next, stranding the narrator on a half-scrubbed, still-wrong dating.
+            death_point = merged.get("death_year_ah")
+            if death_point is not None and not is_plausible_narrator_death_ah(
+                death_point, generation_from_value(rec.get("generation"))
+            ):
+                merged["death_year_ah"] = None
+                merged["death_year_ah_earliest"] = None
+                merged["death_year_ah_latest"] = None
+                merged["death_date_precision"] = DatePrecision.UNKNOWN.value
             for key, value in merged.items():
                 # Never null out a point a prior producer back-filled (defensive —
                 # a contributing bio would itself have surfaced that point here).
--- a/src/resolve/date_reconcile.py
+++ b/src/resolve/date_reconcile.py
@@ -523,6 +523,10 @@ def reconcile_canonical_dates(
             # "dated" to tabaqa_dates' undated-check and skip the ṭabaqa fallback that
             # runs next, stranding the narrator on a half-scrubbed, still-wrong dating.
             death_point = merged.get("death_year_ah")
+            # Death columns the da#454 scrub cleared with EXPLICIT intent, so the
+            # back-fill-protection loop below honours the clear instead of swallowing
+            # it (empty on the normal path — populated only when the scrub fires).
+            scrub_cleared: set[str] = set()
             if death_point is not None and not is_plausible_narrator_death_ah(
                 death_point, generation_from_value(rec.get("generation"))
             ):
@@ -530,10 +534,29 @@ def reconcile_canonical_dates(
                 merged["death_year_ah_earliest"] = None
                 merged["death_year_ah_latest"] = None
                 merged["death_date_precision"] = DatePrecision.UNKNOWN.value
+                scrub_cleared = {
+                    "death_year_ah",
+                    "death_year_ah_earliest",
+                    "death_year_ah_latest",
+                }
             for key, value in merged.items():
                 # Never null out a point a prior producer back-filled (defensive —
-                # a contributing bio would itself have surfaced that point here).
-                if key.endswith("_year_ah") and value is None and rec.get(key) is not None:
+                # a contributing bio would itself have surfaced that point here) —
+                # UNLESS the da#454 scrub above cleared it with explicit intent. When
+                # the canonical record ALREADY holds a death point (a persisted
+                # pre-fix corrupt point on an idempotent `resolve --from-step
+                # reconcile` re-run, or a bio_promote-stamped point reconciliation
+                # recomputes as implausible from the fuller unscrubbed bio set), the
+                # guard would otherwise keep that point while its bounds + precision
+                # are nulled — the self-inconsistent half-scrub this scrub exists to
+                # prevent, which then reads as "dated" to tabaqa_dates' undated-check
+                # and skips the ṭabaqa fallback (da#454).
+                if (
+                    key.endswith("_year_ah")
+                    and value is None
+                    and rec.get(key) is not None
+                    and key not in scrub_cleared
+                ):
                     continue
                 rec[key] = value
             reconciled += 1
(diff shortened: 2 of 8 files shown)


parametrization:
## Review — Ivana Horvat

Reviewed at head `d4b53cf` (`gh pr diff 484` + full `tabaqa_dates.py`, `bio_promote.py`, `date_reconcile.py`, `disambiguate.py`, and `narrator_split.py` at that SHA). All six CI checks green at this SHA.

**The envelope itself is unchanged.** `is_plausible_narrator_death_ah` / `generation_from_value` / `NARRATOR_DEATH_MIN_AH` (11) / `NARRATOR_DEATH_MAX_AH` (450 = LATER-window hi 400 + 50 margin) all predate this PR (da#446) and are correctly derived from `_GENERATION_DEATH_WINDOW_AH`, so the envelope this PR binds against is the same one already validated for the output-side gate. `plausible_attested_death_year` is a thin, correctly-implemented wrapper: `None` passes through, a valid year passes through, an envelope/generation violation returns `None`. Verified the arithmetic on all four new unit-test cases by hand (absolute-envelope 805/754, generation-mismatch 300 vs sahabi, 300 vs later).

**Stamp sites all check out:**
- `bio_promote._load_candidates`... sorry, `bio_promote.promote_bios_to_canonical` — `death_year` is computed once per row and reused for both the new-node mint and the backfill branch; `_BACKFILL_FIELDS` correctly drops `death_year_ah` in favor of the explicit bound-checked backfill, so a corrupt value can no longer sneak onto a zero-mention existing record either.
- `disambiguate._load_candidates` — `generation` is captured once and reused for both `Candidate.generation` and the death-year scrub; since the fix lives on the `Candidate` dataclass itself, `_upsert_canonical`'s `candidate.death_year_ah` write (line ~1060) inherits the scrub for free without needing its own edit — nice placement.
- `date_reconcile.reconcile_canonical_dates` — scrubs the full four-column dating (point + both bounds + precision → `UNKNOWN`) when the reconciled point is implausible, with the documented reasoning for why a null point alone isn't enough (tabaqa-fallback would skip a half-scrubbed row).

**One real but non-blocking gap found (filed as #486):** `date_reconcile`'s fold-back loop has a *pre-existing*, unrelated-to-this-PR defensive guard ("never null out a point a prior producer back-filled") that only matches keys ending in `_year_ah` — which covers `death_year_ah` but *not* `death_year_ah_earliest`/`_latest` (different suffix) or `death_date_precision`. For a canonical record whose `death_year_ah` was **already** implausible before this stage runs (only reachable for pre-#484 corrupt data reloaded via `bio_promote`'s `_load_existing_canonical`, since backfill never re-validates an already-populated field) and whose bio observations re-derive that same implausible point, the guard blocks the point from being nulled while the bounds/precision are unconditionally nulled anyway — leaving a self-inconsistent (stale point, null bounds, `unknown` precision) record instead of a fully-scrubbed one.

Traced this isn't a live data-integrity issue: `_death_is_undated` already treated a non-null point as "dated" pre-#484 too (no fallback-skip regression), and `narrator_split._attested_death` (da#446, deliberately kept as defense-in-depth) independently re-validates `death_year_ah` against the same plausibility check using the record's own `generation`, regardless of precision — so the corrupt point still can't reach the neighbour-death evidence pool. It's an untested internal-consistency wart on data that predates this fix, not a defect in the fix's stated (input-side, forward-looking) scope. Filed #486 with a suggested one-line fix (bypass the guard for this specific scrub, or add the missing regression test) for a follow-up wave.

**Test coverage** is genuinely thorough — valid/None/absolute-envelope/generation-mismatch unit tests, new-node + backfill + regression tests for `bio_promote`, candidate-pool scrub + keep tests for `disambiguate`, and generation-aware + absolute-envelope + regression tests for `date_reconcile`'s reconciliation stage. Full suite + pre-commit + push-stage hooks reported green in the PR description, and CI is green at head.

No must-fix items. Approving.

---
Requestor: Ivana Horvat
RequestOrReplied: Approved
TechDebt: #486

parametrization:
## Merge-gate review — da#454 (Tier-1 cutover gater)

Reviewed at head `d4b53cf78dd1d91f76bc475b2bf2aaf5eb0cbfa1`, diffing against `deployments/phase-9/wave-27` via `gh api .../contents?ref=<sha>` (no local checkout of the parent). CI is green at head across all six checks (Lint, Type Check, Test, Integration Tests, base-image/fixture-realism, sync-drift gate). I mutation-tested the guards against the PR code checked out at that exact SHA rather than trusting the assertions.

The envelope core is sound. `plausible_attested_death_year` is a faithful thin wrapper over `is_plausible_narrator_death_ah` + `generation_from_value`; `NARRATOR_DEATH_MIN/MAX_AH` are derived from `_GENERATION_DEATH_WINDOW_AH` so the envelope and the ṭabaqa periodization cannot drift; the `None` passthrough, absolute-envelope, and generation-tightening branches are all real and all tested. The `bio_promote` new-node and backfill sites and the `disambiguate._load_candidates` site each scrub before the value can be written, and their tests exercise the scrub and the plausible-year regressions. No issue there.

### Must-fix — the `date_reconcile` scrub is silently defeated by the back-fill-protection guard (new code, contract-violating, untested path)

The da#454 block sets the reconciled death point + both bounds + precision to `None`/`unknown`, then the pre-existing loop runs:

```python
for key, value in merged.items():
    if key.endswith("_year_ah") and value is None and rec.get(key) is not None:
        continue          # never null out a point a prior producer back-filled
    rec[key] = value
```

`death_year_ah` ends with `_year_ah`; `death_year_ah_earliest`/`_latest` do **not**. So when the canonical record already holds a death point, the scrub's `None` for the *point* is swallowed by this `continue`, while the bound/precision nulls (which don't match the suffix) go through. I ran this against the PR at head:

- Canonical rec with **no** pre-existing point (the case all four new tests use): point → `None`, bounds → `None`, precision → `unknown`. Correct.
- Canonical rec that **already carries** a corrupt point `720` (bios reconcile to `720`): result is `death_year_ah=720, earliest=None, latest=None, precision='unknown'`. The corrupt point **survives** — behind nulled bounds and `unknown` precision. That is exactly the half-scrubbed, internally-inconsistent dating the scrub's own comment says it exists to prevent ("Scrub the WHOLE death dating ... not just the point"), just mirrored. And because `death_year_ah` is non-`None`, `_death_is_undated` reads the row as *dated*, so the ṭabaqa fallback that runs next will not repair it — the corrupt `720` is stuck permanently.

Reachability: a fresh from-scratch resolve run is clean, because `bio_promote`/`disambiguate` scrub the corrupt value first, so `rec`'s point is `None` when the scrub fires and it applies. The hole is the **idempotent re-run over a persisted `narrators_canonical.parquet` that already carries legacy pre-fix corrupt points** — which is the pipeline's documented mode (`bio_promote` has an existing-record backfill path; `date_reconcile` reads the existing canonical table). On that re-run neither `bio_promote` backfill (fills only when empty) nor `disambiguate` upsert (sets only when `None`) nor this scrub removes a legacy corrupt point, so it ships to cutover as a d. 720 AH Companion. This PR is precisely the gate over a corpus that predates the fix, so this is the live path, not a hypothetical.

Test gap that hid it: every `date_reconcile` fixture (`test_stage_scrubs_*`, `test_stage_keeps_*`) uses a canonical rec with no pre-existing death point, so `rec.get("death_year_ah")` is always `None` and the guard branch never fires. The scrub's whole-dating contract is asserted only where the guard cannot contradict it — an assertion no fixture can trip.

Resolution — either is acceptable:

1. Force the clear through the guard for the da#454 case (e.g. null the death columns on `rec` after the loop, or mark death as explicitly-cleared so the back-fill guard does not re-preserve it), and add a test with a pre-existing corrupt point asserting all four death columns end scrubbed. This makes the code do what its comment already claims.
2. If preserving a pre-existing point is intentional and cleaning the legacy corpus is genuinely out of scope (relying on `narrator_split._attested_death` for output protection), then correct the scrub comment — it currently overclaims — add a test pinning the surviving-point behavior so the residual is explicit, and confirm the cutover regenerates `narrators_canonical` from scratch rather than re-running over the persisted table. Otherwise the gate leaks corrupt attested deaths into the cutover dataset.

I recommend option 1: it is the smaller change and it honors both the scrub's stated intent and the "this gates the cutover" mandate.

Confidence: high that the guard collision is real (reproduced against head), and high that it is reachable on the re-run/idempotency path that a pre-fix corpus cutover exercises. Holding the gate until the point-scrub is made effective or the residual is made explicit and shown safe for the cutover mechanism.

---
TechDebt: none

parametrization:
## Re-review — Ivana Horvat (head `67e67fc`)

My prior approval was against `d4b53cf` (stale — content advanced). Re-reviewed the incremental diff, which is scoped exactly to `src/resolve/date_reconcile.py` (+25/-2) and its tests (+71).

**This closes #486.** The fix introduces a per-record `scrub_cleared: set[str]` populated with `{"death_year_ah", "death_year_ah_earliest", "death_year_ah_latest"}` only when the da#454 plausibility scrub fires, and the pre-existing back-fill-protection guard now reads `... and key not in scrub_cleared: continue` — so the guard still protects an ordinary reconciliation-produced `None` from clobbering a prior producer's stamped point (unchanged behavior), but no longer protects a point the scrub itself just decided to null. `death_date_precision` was never subject to the guard (doesn't end in `_year_ah`), so no double-handling there. Traced the `scrub_cleared` reset: it's declared inside the per-record `for rec in rows:` / `if obs_slot is not None:` block, so no cross-record state leak.

The new regression test (`test_stage_scrubs_implausible_death_over_preexisting_canonical_point`) seeds exactly the scenario #486 described — a canonical record already carrying a legacy corrupt point (720/720/720/"exact") before this stage runs, matching an idempotent `resolve --from-step reconcile` re-run over a pre-fix `narrators_canonical.parquet` — and asserts all four columns (point + both bounds + precision) end up fully null/`unknown`, not half-scrubbed. This is the RED-before/GREEN-after case I couldn't get coverage on in the prior round.

CI is 6/6 green at `67e67fc` and the PR is mergeable. No further concerns — approving at the new head.

**#486 disposition:** closing as resolved-by-#484 (comment posted with the closing rationale).

---
Requestor: Ivana Horvat
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge-gate re-review — da#454 (Tier-1 cutover gater) — fix verified

Re-reviewed at new head `67e67fcfe84576eeb64c04bd815759df410be5da`, the only changes since my prior head being `src/resolve/date_reconcile.py` (+25/−2) and its test (+71). CI is 6/6 green at head, mergeable. I re-ran every guard against the PR code checked out at that exact SHA rather than trusting the assertions.

The must-fix is resolved, and cleanly (Option 1). The scrub branch now records an explicit clear-intent sentinel `scrub_cleared = {death_year_ah, death_year_ah_earliest, death_year_ah_latest}`, populated only when the da#454 scrub fires, and the back-fill-protection loop's `continue` gains `and key not in scrub_cleared`. So the intentional clear pierces the guard, while every accidental producer-`None` outside the sentinel is still swallowed — the legitimate protection is not weakened.

Verified against head, not just read:

- Reproduced my original hole at head: a canonical record already carrying corrupt `720` now comes out `death_year_ah=None, earliest=None, latest=None, precision=unknown`. The corrupt point no longer survives behind nulled bounds.
- The `else`/no-scrub path is unchanged: a record with no pre-existing point still scrubs correctly (sentinel empty).
- Alejandra's from-scratch reachability point confirmed: `reconcile_narrator` recomputes the death point from the raw, unscrubbed `narrators_bio_*.parquet`, so a plausibly-stamped point (100 / 150) whose bio set recomputes to an implausible `720` also trips the guard on a clean run — not only on the `--from-step reconcile` resume. In both cases the sentinel now pierces the guard and all three death columns + precision clear. So the fix covers the clean-run path too, not just idempotent re-runs.
- Non-regression on the legitimate case: when the scrub does NOT fire (bio set carries no death observation, reconciled point is `None`) over a plausibly back-filled point, the point is still preserved (`death_year_ah` stays `150`). The back-fill guard the sentinel modifies still does its original job. Birth is untouched — the scrub only ever populates `scrub_cleared` with death keys, so `birth_year_ah` protection is unchanged.
- The regression test `test_stage_scrubs_implausible_death_over_preexisting_canonical_point` is not inert: I ran it against the OLD source (`d4b53cf`) and it fails with `assert 720 is None` — it genuinely exercises the guard branch (canonical rec pre-loaded with `death_year_ah=720`) and asserts point AND both bounds AND precision cleared. RED on old head, GREEN now. Full `test_date_reconcile.py` is green (30 passed).

Confidence: high. The specific hole I flagged is closed on both the idempotent-re-run and the from-scratch-divergence paths, the fix is surgical and does not regress the back-fill or birth protection, and the pinning test would catch any reintroduction. Clearing the gate.

---
RequestOrReplied: Approved
TechDebt: none

PR #502, opened 2026-07-26
perf(arabic): memoize normalize_arabic with lru_cache (#495)

### LINKED ISSUE #495: audit(F3): lru_cache normalize_arabic in the per-token hot path
Filed by: Nadia Khoury (Program Director) — code-audit program noorinalabs-main#1089. Track epic: noorinalabs-main#1096.
Source: CODE_AUDIT.md §9 (noorinalabs-main @ 3bdd620) — item F3.

**F3 — Cache `normalize_arabic`** · Type: perf · Effort: XS
Pure `str→str`, 8 regex/translate passes, called per **token** (~10⁸ calls over a ~10⁵-form vocabulary) in `name_quality` — ~8–10 min of a full resolve for one `@lru_cache(maxsize=1<<18)` line.
*Verify:* existing 21-assertion suite + 110 name_quality tests (caching a pure function can't change them).

Index row (§13): P | ~10⁸ calls ≈8–10 min → cache hits | XS | —

parametrization:
## Summary
Closes #495 (F3, perf, XS).

`normalize_arabic` is a pure single-`str`-arg function called per-token during `name_quality` resolution (~8-10 min of a resolve). This wraps it in `functools.lru_cache(maxsize=1<<18)` so repeated tokens — and the import-time `_NAME_LEXICON` / `_NORM_MARKER_PATTERNS` builds — hit the cache instead of re-running the full normalization pipeline.

## Changes
- `src/utils/arabic.py`: `from functools import lru_cache`; `@lru_cache(maxsize=1 << 18)` on `normalize_arabic`.

## Correctness
Pure function → output is unchanged; the cache only elides recomputation. Import-time callers are safe to cache.

## Testing
- `tests/test_utils/test_arabic.py`, `tests/test_utils/test_arabic_fuzz.py`, `tests/test_parse/test_name_quality.py`: **623 passed**.
- `ruff check` clean; pre-commit (ruff-format/lint, gitleaks) + pre-push pytest passed.

### DIFF
--- a/src/parse/composition.py
+++ b/src/parse/composition.py
@@ -62,7 +62,7 @@ from __future__ import annotations
 import hashlib
 
 from src.parse.identity import HADITH_ID_PREFIX, ID_DELIMITER
-from src.utils.arabic import is_arabic, normalize_arabic, strip_to_letters
+from src.utils.arabic import is_arabic, normalize_arabic_uncached, strip_to_letters
 
 # source_corpus -> allowed collection slugs.
 #   None           = load all collections for this source. Equivalent to the
@@ -189,8 +189,8 @@ def canonical_matn_identity(matn_ar: str | None, matn_en: str | None) -> str | N
     tokens) — a ``None`` identity is never deduped, so the hadith is kept.
 
     The Arabic matn is preferred and normalized via
-    :func:`src.utils.arabic.normalize_arabic` (diacritics / alif / hamza / taa
-    marbuta / tatweel folded, whitespace collapsed), then reduced to letters and
+    :func:`src.utils.arabic.normalize_arabic_uncached` (diacritics / alif / hamza /
+    taa marbuta / tatweel folded, whitespace collapsed), then reduced to letters and
     single spaces by :func:`src.utils.arabic.strip_to_letters`; the English matn
     (lowercased) is stripped the same way and is the fallback when no Arabic matn
     is present.
@@ -209,7 +209,12 @@ def canonical_matn_identity(matn_ar: str | None, matn_en: str | None) -> str | N
     index stays compact for the full ~650k-row corpus.
     """
     if matn_ar and is_arabic(matn_ar):
-        normalized = strip_to_letters(normalize_arabic(matn_ar))
+        # Full matn bodies (matn_ar / full_text_ar) are large and mostly unique,
+        # and the loader runs this over the full ~650k-row corpus, so bypass the
+        # lru_cache on normalize_arabic (da#495) — caching these would retain
+        # every distinct body for the process lifetime and risk OOM (memory-
+        # bounding history da#337/#723). Same reasoning as src/resolve/parallels.py.
+        normalized = strip_to_letters(normalize_arabic_uncached(matn_ar))
     elif matn_en and matn_en.strip():
         normalized = strip_to_letters(matn_en.lower())
     else:
--- a/src/parse/isnad_matn_split.py
+++ b/src/parse/isnad_matn_split.py
@@ -40,7 +40,11 @@ from src.parse.narrator_extraction import (
     NarratorSpan,
     extract_narrator_mentions,
 )
-from src.utils.arabic import extract_transmission_phrases, normalize_arabic
+from src.utils.arabic import (
+    extract_transmission_phrases,
+    normalize_arabic,
+    normalize_arabic_uncached,
+)
 
 __all__ = [
     "DETECTOR_TOKEN_WINDOW",
@@ -148,7 +152,11 @@ def _head_opener(text: str) -> bool:
     editorial mark still counts; the comparison is on the *normalized* first
     Arabic word, so it is exact-token, never a substring of a longer word.
     """
-    match = _FIRST_AR_WORD_RE.match(normalize_arabic(text))
+    # `text` here is a full matn body (split_isnad_matn runs per corpus row), so
+    # use the uncached core: caching whole bodies would retain every distinct one
+    # for the process lifetime and risk OOM (da#495 review; da#337/#723). We only
+    # read the first word, but normalize_arabic would still cache the full string.
+    match = _FIRST_AR_WORD_RE.match(normalize_arabic_uncached(text))
     if match is None:
         return False
     return match.group(1) in _HEAD_OPENERS
--- a/src/resolve/parallels.py
+++ b/src/resolve/parallels.py
@@ -23,8 +23,9 @@ contract are unchanged.
 Matching signal
 ---------------
 Token-set Jaccard over the normalized matn — Arabic via
-:func:`src.utils.arabic.normalize_arabic` when an Arabic matn is present, else the
-English matn lowercased. Scores are tiered into :class:`VariantType` by the same
+:func:`src.utils.arabic.normalize_arabic_uncached` (the uncached core; matn bodies
+are large and mostly unique) when an Arabic matn is present, else the English matn
+lowercased. Scores are tiered into :class:`VariantType` by the same
 thresholds ``dedup.py`` uses. ``cross_sect`` is read from the hadith ``sect``
 column (authoritative), not inferred from the corpus name.
 
@@ -71,7 +72,7 @@ from src.resolve._provenance import (
     write_parallel_links,
 )
 from src.resolve.schemas import PARALLEL_LINKS_SCHEMA
-from src.utils.arabic import is_arabic, normalize_arabic
+from src.utils.arabic import is_arabic, normalize_arabic_uncached
 from src.utils.logging import get_logger
 
 if TYPE_CHECKING:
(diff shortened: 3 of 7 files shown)


parametrization:
Requestor: Oyunbileg Batbayar
TechDebt: none

Verified the claims: checked out the branch and ran the tests myself. `tests/test_utils/test_arabic.py` + `test_arabic_fuzz.py` + `test_parse/test_name_quality.py` = 623 passed; full non-integration suite = 2531 passed / 0 failed; ruff clean. The pure-function/output-unchanged claim checks out.

One must-fix before merge: `@lru_cache` is applied unconditionally to `normalize_arabic`, but its call sites are not uniform in shape. Most (narrator names/tokens, the structural-particle sets, the name lexicon keys) really are the bounded-vocabulary, high-repetition case this PR targets. But `src/resolve/parallels.py::_tokenize` (line 136) calls `normalize_arabic(matn_ar)` on the FULL hadith body text, once per hadith, across the whole corpus — `_load_hadith_rows` loads "every staged hadith," and the anchor-scan comments above it (parallels.py:106-108) confirm this is the long, full-corpus pass.

That call site is the wrong shape for this cache: `matn_ar` values are large (full hadith text, not a short token) and mostly *unique* per hadith — genuine repeats only happen for the true verbatim-parallel pairs parallels.py exists to find. Caching them means every distinct hadith text this pass touches gets BOTH its raw string and its normalized string permanently retained for the rest of the pipeline's process lifetime (`maxsize=1<<18` = 262,144 entries), instead of the prior compute-and-discard. The resolve pipeline has documented OOM sensitivity — `narrator_split`'s da#337 PR-3a explicitly bounds its chain index to avoid keeping full-corpus structures resident, citing #723 — and a blanket cache on a function now also fed full hadith text at corpus scale reintroduces exactly that risk class. It also isn't exercised by this PR's testing, which is unit-level Arabic/name-quality tests only, not a full-corpus memory profile.

Suggest one of:
1. Scope the cache to a narrower name/token-normalization wrapper used by the narrator-name call sites, leaving `normalize_arabic` (or at least the `parallels._tokenize` call site) uncached, or
2. If the blanket cache is kept, back it with an actual corpus-scale measurement (peak RSS during a `parallels` resolve step on a representative staging snapshot) rather than unit-test-only verification, and note the deliberate memory/CPU trade in the docstring.

Otherwise a nice, minimal diff — this is a real scale concern surfaced by tracing every call site, not a style nit.

parametrization:
Thanks for the trace — you're right, the blanket cache is the wrong shape for `parallels.py::_tokenize`, which feeds full `matn_ar` bodies across the whole staged corpus. Fixed in `1118bf3` (additive, no force-push).

**Fix (your option 1 — scope the cache):**
- Split the pipeline into an uncached core `normalize_arabic_uncached` and kept `normalize_arabic` as the memoized wrapper that delegates to it (both in `__all__`).
- All ~23 bounded-vocabulary callers — name_quality tokens, particle/marker sets, grade needles, `canonical_surface`, the import-time `_NAME_LEXICON`/`_NORM_MARKER_PATTERNS` builds — keep the cache win with zero call-site churn.
- The one full-corpus caller, `parallels._tokenize` (line 136), now calls `normalize_arabic_uncached` and computes-and-discards. Call-site comment cites the OOM history (da#337/#723).

Chose the named public-core split over `normalize_arabic.__wrapped__` / a private cross-module import so the bypass is self-documenting.

**Verification:**
- Cached and uncached produce identical output on a fold/taṣliya/empty/latin sample set; `normalize_arabic` has `cache_info`, `normalize_arabic_uncached` does not.
- 643 assertions green: `test_arabic`, `test_arabic_fuzz`, `test_name_quality`, `test_parallels`.
- ruff clean; pre-commit + pre-push (mypy-strict, pytest) passed.

The unbounded-retention path is gone; the memoization now only ever holds the bounded name/token vocabulary. Ready for re-review.

parametrization:
Requestor: Oyunbileg Batbayar

Thanks — the cached/uncached split is the right shape and `parallels._tokenize` is fixed correctly. Re-scanned every `normalize_arabic` call site (not just the ones I'd already flagged) to make sure the fix was complete, and found one more of the same class that's still on the cached path.

`src/parse/composition.py::canonical_matn_identity` (line 212) still calls the memoized `normalize_arabic(matn_ar)`, same as parallels.py did. It's called from `src/graph/load_nodes.py::_build_curated_identity_index` (line 180) and `_hadith_load_outcome` (line 227) — both scan **every hadith row** for the cross-edition dedup identity; `composition.py`'s own docstring cites "the full ~650k-row corpus." `_effective_matn_ar` feeds it `matn_ar`, falling back to `full_text_ar` (an even larger combined isnad+matn body for halimbahae/open_hadith/bihar sources) — so this is the identical large/mostly-unique-per-hadith shape that made `parallels.py` unsafe to cache, just in the graph-load stage instead of resolve.

Suggest the same treatment: switch `composition.py:212` to `normalize_arabic_uncached`, with a comment pointing at this thread the way you did in parallels.py.

TechDebt: none

parametrization:
Added the regression guard the follow-up asked for (`06dbba5`).

`tests/test_resolve/test_parallels.py::TestTokenize::test_matn_body_bypasses_the_normalize_cache` clears the `normalize_arabic` cache, tokenizes a body-sized Arabic matn via `_tokenize`, and asserts the cache `currsize` stayed `0` — i.e. the full-corpus body path never enters the memoized cache. It also asserts `normalize_arabic_uncached` exposes no `cache_info` (structural guard).

I confirmed the test actually catches the regression: pointing the body path back at the cached function makes `currsize` go to `1` and the assertion fails. So this can't silently regress.

644 assertions green (arabic + arabic_fuzz + name_quality + parallels); ruff + mypy-strict + pytest gates passed.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Reply

Nice regression guard — good structural test, and I confirmed it flags the right thing.

Looks like our comments crossed: this commit (06dbba5) only adds the guard for the already-fixed `parallels.py` path. The `composition.py::canonical_matn_identity` finding from my 17:00:30 comment (line 212, called from `load_nodes.py` at lines 180/227 across the full ~650k-row corpus) is still outstanding — `gh api .../pulls/502/files` shows `composition.py` isn't in the changed-files list yet. Still `ChangesRequested` pending that one.

parametrization:
Good catch — fixed `composition.py::canonical_matn_identity` in `bac39af`, and while auditing the remaining callers for the same shape I found **one more of the same class you'd want off the cache**, so I fixed it in the same commit:

**`src/parse/isnad_matn_split.py::_head_opener` (line 151).** It normalizes the FULL `text` just to read its first word, and it runs per corpus row via `split_isnad_matn(matn_text)` (called from `sanadset.py`). Same large/mostly-unique-body shape — the grep for direct `normalize_arabic(matn_ar)` calls wouldn't surface it because the body is one call frame up.

Both now use `normalize_arabic_uncached`:
- `canonical_matn_identity` — uncached core (matn_ar / full_text_ar over the ~650k-row corpus).
- `_head_opener` — uncached core. The module's bounded `_HEAD_OPENERS` build keeps the cached `normalize_arabic`, and `split_isnad_matn` still caches the bounded narrator **names** it extracts (intended).

**Regression guards (both assert cache `currsize == 0` on the body path):**
- `test_composition::test_matn_body_bypasses_the_normalize_cache` — via `canonical_matn_identity`.
- `test_isnad_matn_split::test_full_body_bypasses_the_normalize_cache` — via `_head_opener` directly (guarding the specific body path, since `split_isnad_matn` legitimately caches extracted names).

I re-swept every remaining `normalize_arabic(` call site: all others pass narrator names / particle sets / grade needles (bounded, low-cardinality — the intended cache case). These three (parallels, composition, isnad `_head_opener`) were the only full-corpus body paths.

All directly-affected suites green (755); the pre-push gate ran the full non-integration suite green too. ruff + mypy-strict passed. Ready for re-review.

parametrization:
Requestor: Oyunbileg Batbayar
RequestOrReplied: Approved
TechDebt: none

Nice catch on the third site — isnad_matn_split._head_opener is exactly the kind of thing my grep-by-argument-name approach would've missed (it normalizes the whole body just to read one word), and finding it by re-auditing every remaining call site rather than trusting the fix was already "done" is the right instinct.

Independently re-verified before approving, not just re-reading the diff:
- Checked out bac39af myself and confirmed composition.py:212 and isnad_matn_split.py's _head_opener both now route through normalize_arabic_uncached, with regression guards in test_composition.py and test_isnad_matn_split.py that assert cache currsize stays 0 on the body path.
- Re-swept every remaining `normalize_arabic(` call site in src/ myself (not just yours) — sanadset.py's two text/stripped call sites looked worth a second look given the "full corpus" pattern, but both operate on individual narrator-span text (`_is_narrator_like`, chain-segment extraction), not matn bodies. Everything else is names/tokens/particles. No further full-body sites found.
- Ran the targeted suites (689 passed), the full non-integration suite (2534 passed, 0 failed, matching your "unrelated" characterization — no segfault this run), mypy clean, ruff check + format clean.
- Confirmed cached/uncached output equivalence directly on a small Arabic + Latin + empty sample set.

Good persistence through three rounds — approved.

parametrization:
Reviewed the #495 diff directly (`gh pr diff 502`), not just the resolved thread history.

**Purity/hashability**: `normalize_arabic_uncached` (the pipeline `normalize_arabic` now wraps) does only regex `.sub`/`.translate` on the input `str` — no I/O, no reliance on mutable module state beyond precompiled patterns/lookup dicts that are never mutated at runtime. Deterministic same-input-same-output, and `str` is hashable — safe for `lru_cache`.

**Cache bound**: `@lru_cache(maxsize=1 << 18)` (262,144 entries) is a real bound, not unbounded — good, this directly addresses the memory-leak risk on high-cardinality input the task flagged.

**Call-site audit** (independently re-swept, not just re-reading the prior thread): confirmed the three full-corpus/full-matn-body call sites — `src/resolve/parallels.py::_tokenize`, `src/parse/composition.py::canonical_matn_identity`, `src/parse/isnad_matn_split.py::_head_opener` — all route through the uncached core at head SHA `bac39af`, each with a `cache_info().currsize == 0` regression guard (`test_parallels.py`, `test_composition.py`, `test_isnad_matn_split.py`). Checked `sanadset.py`'s two remaining `normalize_arabic(text)` call sites (`_is_narrator_like`, `_segment_nar_content`) myself — both operate on individual `<NAR>`-tag narrator-span text, not matn bodies, so they are correctly on the cached (bounded-vocabulary) path. No other full-body site found.

**Correctness**: no normalization *output* changes — `normalize_arabic` is a thin memoized wrapper delegating to the same pipeline steps that were inline before; cached vs. uncached outputs are asserted identical on a fold/tasliya/empty/Latin sample.

**CI**: all 6 checks green at head SHA `bac39afe45894528ba6403e26fdaaf4d4c555b26` (Lint, Type Check, Test, Integration Tests, Base-image pin + fixture realism, Pre-commit/CI sync-drift gate).

No further issues. Approving as the second distinct reviewer for the merge gate (alongside Oyunbileg's Opus review).

---
RequestOrReplied: Approved
TechDebt: none

PR #504, opened 2026-07-26
da#490: edge_load_conformance false-positive + GRADED_BY orphan gap

parametrization:
Closes #490.

### Finding A — `APPEARS_IN created=0` is benign idempotency, not a gap
`execute_write_batch` returns only `summary.counters.relationships_created`. On a
re-load onto a graph that already carries the edges, the re-`MERGE` creates **0
new** relationships — the correct idempotent outcome. `edge_load_conformance`
(and PARALLEL_OF's own da#160 zero-warning) keyed on `created == 0`, so they
false-alarmed on every clean reload and needlessly drove the loader toward rc=5.

**Fix:** `EdgeLoadResult` now carries `already_present` (endpoints resolved but
not newly created) and a `resolved` property (`created + already_present`). Every
edge loader computes `already_present = len(valid_batch) - created`. Both
zero-count checks now key on `resolved == 0` — *no endpoint pair present, whether
newly or already* — which is the only genuine gap. `already_present` is also
surfaced in the CLI load summary.

### Finding B — GRADED_BY 0/21,185: the failing endpoint is the **Hadith**, not the Grading
The issue hypothesised a `grading_node_id` scheme mismatch. **It is not.**
`_load_gradings` (node) and `_load_graded_by` (edge) both derive the grading id
from **one** `grading_node_id(sid)` on the **same** `source_id`, so the Grading
side cannot disagree — proven by code inspection and a bite-test.

The real cause: all 21,185 grades are **fawaz six-books editions** (bukhari,
muslim, …), which `_load_hadiths` drops as **non-canonical duplicates of the lk
spine** (`HADITH_COMPOSITION["fawaz"] = {nawawi, dehlawi, qudsi}`, da#191/da#333).
Their **Hadith node never loads**, so:
- `_load_graded_by` counted the deliberate composition drop as `missing_endpoints`
  (21,185, every run), and
- `_load_gradings` still minted an **orphan Grading node** whose `hadith_id`
  points at a Hadith that never loads.

Reconciling the issue's own numbers confirms it: `gradings_loaded merged=69,575`
with `GRADED_BY missing=21,185` means ~48k **lk** grades already resolve
(created=0 on reload = Finding-A idempotency) while the 21,185 fawaz orphans fail
every run.

**Fix:** gate `_load_gradings` and `_load_graded_by` on the node loader's **real
kept set** — the same `build_loaded_hadith_ids` / `_hadith_load_outcome` decision
[...]

PR #505, opened 2026-07-26
fix(da#490): edge_load_conformance false-positive + GRADED_BY composition gap

### LINKED ISSUE #490: edge-load: APPEARS_IN conformance false-positive on idempotent reload + GRADED_BY never loads (0/21,185 endpoints resolve)
## Summary
The reload's `edge_load_conformance` "loaded 0 edges" warning (which drives the loader's rc=5 VALIDATION_FINDINGS on every clean reload) conflates **two distinct situations** for `APPEARS_IN` and `GRADED_BY`. Surfaced during the Phase-9 re-cutover stg reload (deploy#696). One is benign; one is a genuine standing data gap.

## Evidence — prior vs current cutover
| Edge | 2026-07-17 reload ([29598703834](https://github.com/noorinalabs/noorinalabs-deploy/actions/runs/29598703834)) | 2026-07-25 reload ([30143725019](https://github.com/noorinalabs/noorinalabs-deploy/actions/runs/30143725019)) |
|------|------|------|
| `APPEARS_IN` | created=**650,983**, missing_endpoints=75,380 | created=**0**, missing_endpoints=75,380 |
| `GRADED_BY`  | created=**0**, missing_endpoints=21,185 | created=**0**, missing_endpoints=21,185 |
| `zero_count_edge_types` | `['GRADED_BY']` | `['APPEARS_IN', 'GRADED_BY']` |

## Finding A — `APPEARS_IN created=0` is benign idempotency (conformance false-positive)
`created` (from `execute_write_batch`) counts **newly-created** relationships. On a re-load onto a graph that already carries the edges (07-17 created 650,983; the narrator prune doesn't touch hadith→collection edges), the re-MERGE creates **0 new** edges — the correct idempotent outcome, not a load failure. The `edge_load_conformance` check flags any edge type with `created == 0` as "loaded 0 edges", so it false-alarms on every idempotent reload and needlessly pushes the loader to rc=5.

**What needs to be done:** teach the conformance check to distinguish "0 created because already present" from "0 created because 0 valid endpoints" — e.g. report `created + already_present`, or skip the zero-count flag when `valid_batch > 0 && created == 0`. (`src/graph/load_edges.py` result reporting + `src/graph/__init__.py` conformance rollup.) Precedent: da#160 (PARALLEL_OF "0 edges loaded").

## Finding B — `GRADED_BY` has never loaded a single edge (genuine gap)
`GRADED_BY` is `created=0` in **both** runs with an identical `missing_endpoints=21,185` — i.e. no `(:Hadith)-[:GRADED_BY]->(:Grading)` edge has ever been created. Both endpoints derive from the same `source_id` (`src/graph/load_edges.py:1022-1024`): `hid = hadith_node_id(sid)`, `gid = grading_node_id(sid)`; the check (`_GRADED_BY_CHECK`) requires both nodes present, and all 21,185 fail.

Grading **nodes** do load (`gradings_loaded merged=69,575`), and `APPEARS_IN` resolves most of the same hadiths, so the failing endpoint is almost certainly the **Grading** side: `grading_node_id(sid)` likely produces ids that don't match the ids the Grading nodes were actually loaded under.

**What needs to be done:** compare `grading_node_id(sid)` output against a sample of loaded `Grading.id` values on stg; reconcile the id construction (loader node vs edge path) so the endpoint check resolves. Then confirm `GRADED_BY` creates edges on the next reload.

## Impact
- **No effect** on the narrator prune / enrich / choke-point leaderboard (those depend on Narrator + TRANSMITTED_TO only).
- Finding A is monitoring noise (spurious rc=5 findings on clean reloads).
- Finding B means grading-linked queries (`GRADED_BY`) return nothing — a real data-completeness gap, but pre-existing, not introduced by this artifact.

## Context
Phase-9 re-cutover: parent deploy#696 / main#978. Not a cutover blocker; tracked for follow-up.

parametrization:
## Summary

Closes #490. Tier 3 tech-debt, non-blocking, surfaced during the Phase-9 re-cutover (#978).

### Finding A — `edge_load_conformance` false-positive on idempotent reload

`load_all_edges`'s conformance check flagged *any* edge type with `created == 0` as a load failure. That conflates two distinct situations:

- a genuine "0 valid endpoints" (the da#160 PARALLEL_OF regression this check exists to catch), and
- a correct idempotent re-MERGE onto a graph that already carries every edge of that type — Neo4j's own `relationships_created` counter is legitimately 0 on a re-run, by design.

This false-alarmed on `APPEARS_IN` in the 07-25 cutover reload (`created=0`, but 650,983 edges from the prior run were still there, untouched).

**Fix:** added `EdgeLoadResult.attempted` — edges whose both endpoints resolved and were submitted to the write query, regardless of whether the MERGE created something new — and keyed the conformance check (and PARALLEL_OF's own sibling da#160 zero-check) on `attempted` instead of `created`. `attempted` is now threaded through all 6 edge loaders and surfaced in the CLI's per-edge-type load summary.

### Finding B — GRADED_BY never loads a single edge (0/21,185 endpoints resolve)

`_load_graded_by` and the Grading node loader (`_load_gradings`) mint identical ids (`grading_node_id(sid)`/`hadith_node_id(sid)`) from the exact same `hadiths_*.parquet` `source_id` column — I traced this by hand and ruled out a code-level id-scheme drift between the two producers.

What both *do* skip is the da#373 canonical-hadith gate the chain edges (TRANSMITTED_TO / NARRATED) already apply: a graded hadith the node loader composed away (the per-source/collection `HADITH_COMPOSITION` allowlist, or the cross-edition matn dedup) has no Hadith node to attach to. Without the gate, that case was silently inflating `missing_endpoints` indistinguishably from a genuine id mismatch.

**Fix:**
- Route `_load_graded_by` through the same `_staging_loaded_hadith_ids` / `_mention_hadith_survives` gate already shared by TRANSMITTED_TO/NARRATED, so a composed-away graded hadith is now an accounted-for `dropped_composed` drop, not a mystery miss.
- Split the *remaining* endpoint-check failures into `missing_hadith_only` / `missing_grading_only` / `missing_both`, logged on `graded_by_loaded`. This turns the issue's own "likely grading_node_id != loaded Grading.id" into something the next real reload's structured logs will *confirm*, not guess at — I don't have access to the live stg graph to do that comparison myself.

### Incidental: test-mock gap

`MockNeo4jClient.execute_write_batch`'s `self._nodes_created if self._nodes_created else len(batch)` could never express "0 created" (0 is falsy in Python) — so no existing test in the suite could simulate an idempotent re-load at all, which is exactly the scenario Finding A needed to cover. Switched the default to `None` with an `is not None` check; no existing caller passes the kwarg, so this is behavior-preserving for every current test.

## Test plan

- [x] New tests in `tests/test_graph/test_load_edges.py`:
  - `TestEdgeLoadConformanceAttempted`: idempotent reload not flagged, genuine zero-endpoint case still flagged, `EdgeLoadResult.attempted` defaults to 0.
  - `TestLoadGradedByCompositionGate`: composed-away graded hadith dropped (not missing), canonical graded hadith loads, missing-grading-only vs missing-hadith-only distinguished in logs, idempotent GRADED_BY reload reports `attempted` correctly.
- [x] Full suite green: `uv run pytest tests/ -m "not integration"` (2539 passed, 10 skipped, 58 deselected — integration needs Docker).
- [x] `pre-commit run --all-files` and `--hook-stage push` (ruff-format, ruff-lint, gitleaks, actionlint, cspell, mypy-strict, pytest) all green.

Merge-gate: Nikolaos Papadopoulos.

### DIFF
--- a/src/cli.py
+++ b/src/cli.py
@@ -388,8 +388,9 @@ def _cmd_load(
         )
     for er in summary.edge_results:
         print(
-            f"    {er.edge_type}: created={er.created} skipped={er.skipped}"
-            f" missing_endpoints={er.missing_endpoints} malformed_ids={er.malformed_ids}"
+            f"    {er.edge_type}: created={er.created} attempted={er.attempted}"
+            f" skipped={er.skipped} missing_endpoints={er.missing_endpoints}"
+            f" malformed_ids={er.malformed_ids}"
         )
 
     if summary.validation_results:
--- a/src/graph/load_edges.py
+++ b/src/graph/load_edges.py
@@ -48,6 +48,18 @@ class EdgeLoadResult:
     missing_endpoints: int
     # Edges refused because an endpoint id violates the id grammar (da#359).
     malformed_ids: int = 0
+    # da#490: edges whose BOTH endpoints resolved and were submitted to the
+    # write query, regardless of whether Neo4j actually created a new
+    # relationship for them (``created``) or the MERGE found it already present
+    # (an idempotent re-load). This is the field ``created == 0`` alone cannot
+    # distinguish: an edge type that loaded 650,983 edges last run and 0 THIS
+    # run because every one of them already exists (``attempted`` > 0,
+    # ``created`` == 0) is a correct idempotent no-op, not a load failure —
+    # unlike an edge type with ``attempted`` == 0, which never had a single
+    # valid endpoint pair to submit in the first place. See
+    # :func:`load_all_edges`'s conformance check, which keys on this rather
+    # than on ``created``.
+    attempted: int = 0
 
     @property
     def refused(self) -> int:
@@ -415,7 +427,14 @@ def _load_transmitted_to(
         total_pairs=len(all_pairs),
         dropped_noncanonical=dropped_noncanonical,
     )
-    return EdgeLoadResult("TRANSMITTED_TO", created, 0, missing, malformed_ids=malformed_ids)
+    return EdgeLoadResult(
+        "TRANSMITTED_TO",
+        created,
+        0,
+        missing,
+        malformed_ids=malformed_ids,
+        attempted=len(valid_batch),
+    )
 
 
 # ---------------------------------------------------------------------------
@@ -552,7 +571,9 @@ def _load_narrated(
         missing_endpoints=missing,
         dropped_noncanonical=dropped_noncanonical,
     )
-    return EdgeLoadResult("NARRATED", created, 0, missing, malformed_ids=malformed_ids)
+    return EdgeLoadResult(
+        "NARRATED", created, 0, missing, malformed_ids=malformed_ids, attempted=len(valid_batch)
+    )
 
 
 # ---------------------------------------------------------------------------
@@ -668,7 +689,14 @@ def _load_appears_in(
         else 0
     )
     logger.info("appears_in_loaded", created=created, skipped=skipped, missing_endpoints=missing)
-    return EdgeLoadResult("APPEARS_IN", created, skipped, missing, malformed_ids=malformed_ids)
+    return EdgeLoadResult(
+        "APPEARS_IN",
+        created,
+        skipped,
+        missing,
+        malformed_ids=malformed_ids,
+        attempted=len(valid_batch),
+    )
 
 
 # ---------------------------------------------------------------------------
@@ -730,6 +758,7 @@ def _load_parallel_of(
     missing = 0
     created = 0
     malformed_ids = 0
+    attempted = 0
     for rows in _iter_parquet_row_batches(path):
         batch: list[dict[str, Any]] = []
         for row in rows:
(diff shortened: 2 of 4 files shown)


parametrization:
Superseded by #504 — a wave-28 spawn under the same implementer identity (persona-collision on our side, not a code-quality issue) was independently working da#490 in parallel and converged on the same approach (conformance keyed on valid-endpoint-count, GRADED_BY routed through the da#373 canonical-hadith gate).

#504 is further along: it's already merge-gate approved (Nikolaos Papadopoulos, Opus) and its CI ran the Neo4j Integration Tests green, with `test_graded_by_edges` asserting a real GRADED_BY count on a live engine — empirically confirming the Finding B root cause that this PR could only diagnose (missing_hadith_only/missing_grading_only split) without live-DB access.

Closing this one in favor of #504. See https://github.com/noorinalabs/noorinalabs-data-acquisition/pull/504.

PR #487, opened 2026-07-22
Phase 9 Wave 27 → main (data-acquisition)

parametrization:
Phase 9 Wave 27 → main (data-acquisition). All 4 wave-27 PRs reviewed (2 reviewers each, including an Opus merge-gate reviewer) and merged into the wave branch.

- #484 (da#454) — **sole Tier-1 cutover gater.** Bounds attested `death_year_ah` to the isnad-plausibility envelope at stamp time. Merge-gate reviewer held it on a genuine must-fix (the back-fill guard swallowed the scrub's point-null over a pre-existing corrupt point → half-scrubbed record survives; reachable on both the idempotent-resume AND from-scratch cutover paths). Fixed via an explicit `scrub_cleared` clear-intent sentinel + regression test; re-approved fresh. Tech-debt #486 filed then closed as resolved-in-PR.
- #481 (da#464) — share the resolve-stage death-year spread band constant.
- #482 (da#477) — record the ى→ي fold full-corpus A/B artifact.
- #483 (da#453) — extend the da#439 adjacency invariant across the multi-isnad grouping boundary.

Merge with `--merge` (never squash). Wave branch retained.

PR #507, opened 2026-08-09
fix(da#506): correct PROD_RELOAD_RUNBOOK_723 — two checks that fail on a healthy graph, stale figures, false premise

### LINKED ISSUE #506: PROD_RELOAD_RUNBOOK_723 is unexecutable: 2 verification queries pass-on-broken/fail-on-healthy, all figures stale, and prod is already reloaded
## Summary

`docs/PROD_RELOAD_RUNBOOK_723.md` cannot be safely executed as written. Every figure in it describes an **earlier** stage reload, and two of its **verification queries return empty/wrong on a healthy system** — the failure shape where the broken state and the passing state are observationally identical from outside.

All findings below were re-measured live on **2026-08-09** against both hosts.

---

## Defect 1 — §1 precondition 2: curated row counts are stale (would abort a correct reload)

Runbook demands `narrators_canonical` = **232,766** and `narrator_mentions_resolved` = **3,126,954**.

Measured locally:

| artifact | runbook | actual |
|---|---|---|
| `data/curated/narrators_canonical.parquet` | 232,766 | **102,350** |
| `data/curated/narrator_mentions_resolved.parquet` | 3,126,954 | **2,975,629** |

The local artifacts are **correct** — stage Neo4j holds exactly 102,350 Narrators, and provenance proves the same run (`parquet_ref = staged/narrator-resolve/2026-07-25-c5a95be`). The narrator count fell legitimately via merge fixes (da#347/#356/#376) and chains rose via corpus recovery (da#364–369). It is the *runbook* that is stale, so an operator following §1 literally aborts a **correct** reload.

## Defect 2 — §5 pass criteria are stale (and §4 / the header carry the same numbers)

| criterion | runbook | stage actual | prod actual |
|---|---|---|---|
| Narrators | ≈232,766 | **102,350** | 102,350 |
| Chains | ≈585,129 | **1,122,521** | 1,122,477 |
| hollow | 0 | 0 ✅ | 0 ✅ |
| Linkage | ≈69% | **59.87%** | 59.81% |
| total nodes | ≈1,665,760 | **2,722,570** | 2,722,526 |

Same stale figures also appear at: the header (L3–5, "1.67M nodes, ~69% linked"), §1 precondition 3 (`top narrator = أبو هريرة/سفيان`), §4 (`total_nodes≈1,665,760`), and §4's `orphan_narrators: ~72,852` (zero-degree count is now **41,072**).

## Defect 3 — §5 top-narrator query has the arrow backwards (returns empty on a HEALTHY graph)

```cypher
MATCH (n:Narrator)<-[:NARRATED]-() ...   -- returns ZERO ROWS
```

The real direction is `Narrator-[:NARRATED]->Hadith`. Measured on stage: **839,238** `NARRATED` edges, and the as-written query returns **no rows at all**. Corrected direction returns the expected real narrators (ابن عباس، سفيان، وكيع، علي; **أبيه absent**).

An operator running §5 against a healthy prod graph sees an empty top-narrator list and reads it as catastrophic failure — on a graph that is fine.

## Defect 4 — §0 target verification prints an IPv6 address on BOTH hosts, so the safety check guarding the destructive wipe is unusable

The runbook's own §0 command, run verbatim:

| host | §0 as-written (`curl -s ifconfig.me`) | `curl -4 -s ifconfig.me` |
|---|---|---|
| `noorinalabs-prod` | `2a01:4ff:f4:a690::1` | **178.156.214.225** ✅ |
| `noorinalabs-stg` | `2a01:4ff:f4:5e0d::1` | **87.99.137.225** ✅ |

§0 says *"MUST print the prod hostname and 178.156.214.225. If it prints 87.99.137.225, STOP."* As written it prints **neither**, on either host. The operator either stops on the correct host, or — far likelier under a check that never matches — learns to ignore the check. This is the **most severe** of the five: it is the guard in front of an irreversible `DETACH DELETE` of production, and it does not discriminate prod from stage at all. Fix is `curl -4`.

## Defect 5 (premise) — prod has ALREADY been reloaded; the runbook's stated purpose is false

The runbook opens: *"Replaces the polluted/broken prod isnad graph (768k nodes, 8.98% linked, hollow chains, أبيه pollution)"*. Measured on prod today:

- **2,722,526** nodes; 102,350 Narrators; 1,122,477 Chains, **hollow = 0**
- Linkage **59.81%** (not 8.98%)
- Relational-pronoun residue (ابيه/جده/امه/عنه): **0 rows**
- Top narrators are real (ابن عباس، ابن عبس، …); **أبيه absent**
- `LoadProvenance.parquet_ref` = `staged/narrator-resolve/2026-07-25-c5a95be` — **the same artifact ref as stage**

Prod is the corrected graph. Executing §3 as written would `DETACH DELETE` a **healthy** production graph and take a ~1 hr outage to rebuild what is already there.

Prod is, however, slightly **short of stage** on the edge phases — a partial tail, not pollution:

| | stage | prod | delta |
|---|---|---|---|
| Chain | 1,122,521 | 1,122,477 | −44 |
| TRANSMITTED_TO | 3,387,179 | 3,342,894 | −44,285 |
| NARRATED | 839,238 | 826,376 | −12,862 |
| linked Narrators | 61,278 | 61,212 | −66 |

The loaders are MERGE-idempotent, so this is healed by **re-running §4 alone** (~the runbook's own documented "roll forward" recovery) with **no wipe at all**. The wipe is not merely unnecessary — it is strictly worse than the no-op.

---

## Proposed fix

1. Correct all stale figures (§1.2, §1.3, §4, §5, header) to the measured values, dated and sourced.
2. Fix the §5 arrow to `-[:NARRATED]->`.
3. Fix §0 to `curl -4` so the guard actually discriminates prod from stage.
4. **Restructure the premise**: demote the destructive wipe-and-reload to a contingency, and make the default path a §4-only idempotent top-up, since prod is already loaded. Add a "measure first" step so the runbook can never again assert a graph state it has not just checked.

## Owner decision required

The prod-reload sign-off that was pending is **moot as framed** — there is nothing polluted to replace. The remaining real question is whether to run the §4 idempotent top-up to close the ~44k-edge gap between prod and stage. That is non-destructive but does touch prod, so it stays owner-gated.

parametrization:
Closes #506.

Corrects `docs/PROD_RELOAD_RUNBOOK_723.md`, which could not be safely executed as written. All figures below were **re-measured live on 2026-08-09** against both hosts — nothing here is carried from the previous revision's prose.

## The headline: prod is already reloaded

The runbook's stated purpose — *"Replaces the polluted/broken prod isnad graph (768k nodes, 8.98% linked, hollow chains, أبيه pollution)"* — is **false**. Measured on prod:

- **2,722,526** nodes · 102,350 Narrators · 1,122,477 Chains · **hollow = 0**
- Linkage **59.81%** (not 8.98%) · relational-pronoun residue: **0 rows** · أبيه absent
- `LoadProvenance.parquet_ref` = `staged/narrator-resolve/2026-07-25-c5a95be` — **the same ref as stage**

Running the old §3 would have `DETACH DELETE`d a **healthy production graph** and taken a ~1 hr outage to rebuild what was already there.

## Two checks that fail on a healthy system

The more insidious class — where broken and passing are observationally identical from outside:

| § | as written | behaviour on a HEALTHY system |
|---|---|---|
| §5 top-narrator | `(n:Narrator)<-[:NARRATED]-()` | **zero rows** — reads as catastrophic data loss. Real direction is `Narrator-[:NARRATED]->Hadith`, 839,238 edges |
| §0 target check | `curl -s ifconfig.me` | returns **IPv6** on both hosts (`2a01:4ff:f4:a690::1` prod, `2a01:4ff:f4:5e0d::1` stage) — matches *neither* IPv4 literal, so the guard in front of an irreversible wipe **cannot tell prod from stage**. Fixed with `curl -4` |

The §0 one is the most severe in the document: it is the safety check on a destructive operation, and it fails identically on the right host and the wrong one.

## Stale figures

| | old | measured |
|---|---|---|
| `narrators_canonical` rows | 232,766 | **102,350** |
| `narrator_mentions_resolved` rows | 3,126,954 | **2,975,629** |
| Chains | ≈585,129 | **1,122,521** |
| total nodes | ≈1,665,760 | **2,722,570** |
| linkage | ≈69% | **59.87%** |
| `orphan_narrators` | ~72,852 | **≈41,072** |

The artifacts are correct; the runbook was stale. The counts moved legitimately — narrators *fell* via merge fixes da#347/#356/#376, chains *rose* via corpus recovery da#364–369. An operator following §1 literally would have aborted a **correct** reload.

## Restructure

Rather than only patching numbers (which would go stale again), the flow now measures before it decides:

- **New §1** measures the live graph and classifies it HEALTHY / BROKEN / UNCLASSIFIED. Nothing downstream may be chosen from this document's prose.
- **§3** is now an idempotent **no-wipe top-up** — the default path, and the runbook's own documented "roll forward" recovery promoted to primary.
- **§4** (the wipe) is demoted to **contingency only**, reachable solely when §1 says BROKEN.
- **§2** records the real prod↔stage gap: −44 Chain, −44,285 `TRANSMITTED_TO`, −12,862 `NARRATED`, confined to the chain-derived phases. A truncated tail, not pollution — closed by the §3 top-up with no deletion.
- Preconditions now assert **equality against stage** rather than against literals typed into the file.

## Verification

- cspell ✅ · gitleaks ✅ · mypy-strict ✅ · pytest ✅ (pre-push)
- markdownlint-cli2 v0.18.1: **0 errors** across 30 files
- The single relative link (`testing-on-subsets.md`) resolves

## Owner decision still open

The pending prod-reload sign-off is **moot as framed** — there is nothing polluted to replace. The remaining question is whether to run the §3 idempotent top-up to close the ~44k-edge gap. Non-destructive, but it writes to prod, so it stays owner-gated. **This PR changes documentation only; prod was not touched.**

### DIFF
--- a/docs/PROD_RELOAD_RUNBOOK_723.md
+++ b/docs/PROD_RELOAD_RUNBOOK_723.md
@@ -1,11 +1,20 @@
 # Prod graph reload runbook — #723 data-quality fix
 
-Replaces the polluted/broken prod isnad graph (768k nodes, 8.98% linked, hollow
-chains, أبيه pollution) with the corrected graph already validated on stage
-(1.67M nodes, ~69% linked, 100%-populated chains, zero name pollution).
+> **STATUS (measured 2026-08-09, da#506): the #723 reload has ALREADY been applied to prod.**
+> Prod carries the corrected graph — 2,722,526 nodes, 102,350 Narrators, **hollow = 0**, zero
+> relational-pronoun pollution, and the same `LoadProvenance.parquet_ref` as stage
+> (`staged/narrator-resolve/2026-07-25-c5a95be`). **The destructive wipe in §4 is NOT the
+> default path and must not be run to "apply #723" — that work is done.** The only outstanding
+> gap is a small edge-phase shortfall vs stage (§2), healed by the **idempotent top-up in §3**
+> with no wipe. §4 exists solely as a contingency for a graph that is genuinely broken, and it
+> is reachable only after §1 says so.
 
-**This is a destructive, owner-sign-off-only operation.** Do not run any step in
-§3+ without explicit owner approval.
+Historical purpose: replace the polluted/broken prod isnad graph (768k nodes, 8.98% linked,
+hollow chains, أبيه pollution) with the corrected graph validated on stage. That replacement
+has happened. What remains is parity maintenance, not replacement.
+
+**§4 is a destructive, owner-sign-off-only operation.** Do not run any step in §4 without
+explicit owner approval, and not at all unless §1 classifies prod as BROKEN.
 
 ---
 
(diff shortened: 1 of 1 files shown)


parametrization:
Reviewed for operational usability and artifact provenance (complementary to the correctness/Cypher-direction review). Verified against repo-local evidence, not the PR body's own claims.

**Walking the runbook as an operator would (head SHA b3176e6d)**

Every numbered step has a concrete command and a falsifiable pass/fail: §0's target check has a literal-IP match, §1's classification table has three mutually exclusive concrete bands (HEALTHY/BROKEN/UNCLASSIFIED) with numeric thresholds, §1a's preconditions are equality checks against a live comparator (stage) rather than literals, §4's wipe loop prints `WIPE COMPLETE`, §6's reload has an explicit rc-code table, and §7's pass criteria are concrete numbers with a stated re-measurement caveat. This is a real improvement over the "looks roughly right" failure mode the PR is fixing — I didn't find a step whose success/failure is unfalsifiable. §3's "confirm the deltas have closed" is the softest of the bunch (no explicit numeric threshold) but it's checked against §2's own table, so it's minor.

**Artifact provenance — verified from repo-local evidence, not taken on the PR's word**

- `data/curated/narrators_canonical.parquet`: measured 102,350 rows via `pyarrow.parquet.read_metadata` — matches the PR's claim exactly, and matches `data/curated/_manifest.txt`'s MD5 (`8758b3f710e8d82ee6dbb0bb039da9c3`) and a direct `md5sum` of the file, so the artifact isn't just right-sized, it's checksum-verified against its own manifest.
- `narrator_mentions_resolved.parquet`: measured 2,975,629 rows — also matches exactly.
- `data/curated/_resolve_run.txt` shows `run_status=complete` and every `RESOLVE_STEP_ORDER` stage (`ner` through `parallels`) with `outcome=ran` — this is a genuine complete run, not a bounded/partial probe (`src/resolve/_run_record.py` confirms `run_status` is only ever stamped `complete` by the process that mints the final tally).
- The cited ref `staged/narrator-resolve/2026-07-25-c5a95be` matches `scripts/publish_parquet.py::default_parquet_ref`'s format exactly (`staged/narrator-resolve/<date>-<short-sha>`), and `c5a95be` resolves to a real commit in this repo's history (`c5a95bef`, merge of PR #487, 2026-07-22) — not a fabricated-looking hash.
- I could **not** independently verify the live prod/stage Neo4j figures (node counts, linkage %, the `LoadProvenance.parquet_ref` equality between hosts) — that requires SSH access to both hosts, which is outside a read-only repo review. The local-artifact evidence above is fully consistent with those claims but doesn't prove them.
- da#506 (the issue this closes) documents the same 5 defects with the same measured figures; the runbook's rewrite tracks it point-for-point.

**One real ordering defect, non-blocking**

§4 (the wipe) is now physically ordered *before* §5 (backup), even though §5 explicitly says "run BEFORE §4" and the base revision had backup precede wipe. The routing is stated correctly elsewhere (§1's classification table and the sign-off checklist both say `§5 → §4 → §6`), so an operator following the document's own decision-tree usage (which §1 mandates — "nothing may be chosen from this document's prose" alone) lands on the right order. But nothing inside §4 itself points back to §5, so a top-to-bottom skim hits the wipe before the backup instruction. Given this doc exists specifically to close "looks fine, isn't" gaps on a destructive prod operation, I'd rather this get fixed, but since the routing text is already correct and the fix is a pure reorder, I'm not gating merge on it — filed as tech-debt: da#508.

**Rollback path**: not orphaned — §8 correctly references the renumbered §3/§6 (roll-forward) and §5 (dump), both consistent with the restructure.

All 11 CI checks green at the head SHA. Strong piece of work: the "measure before you decide" restructure and the checksum-clean artifact provenance are exactly what this class of defect needs.

---
Requestee: Sofia Cardoso
RequestOrReplied: Approved
TechDebt: #508

parametrization:
Merge-gate review at head SHA `b3176e6d431359f921ba4098922366f09d1fe14f`. I treated every figure in the issue and the PR body as an unverified claim and re-derived each one independently — from repo-local artifacts, from the loader source, from the ontology, and by executing the runbook queries read-only against both live hosts. A doc fix that swaps one wrong number for another passes CI identically, so nothing below is taken from the PR description.

## Verdict

Approved. All four defects are real, the replacement content is correct, and I could not find a fifth query that fails on a healthy graph.

## Relationship direction — the defect with real operational cost

Three independent sources agree the old arrow was genuinely inverted:

- `ontology/domain.yaml` relationship `NARRATED`: `from: narrator`, `to: hadith`
- `src/graph/load_edges.py` `_NARRATED_QUERY`: `MATCH (n:Narrator …) MATCH (h:Hadith …) MERGE (n)-[r:NARRATED]->(h)`
- Live stage: `MATCH (a)-[:NARRATED]->(b) RETURN labels(a), labels(b), count(*)` returns exactly one row — `["Narrator"] -> ["Hadith"], 839238`. There are zero edges in the other direction.

Mutation test, both forms executed verbatim against the healthy stage graph:

```
--- OLD (inverted) arrow
MATCH (n:Narrator)<-[:NARRATED]-() RETURN n.name_ar AS name, count(*) AS deg ORDER BY deg DESC LIMIT 10
   -> zero rows

--- NEW arrow (PR head)
MATCH (n:Narrator)-[:NARRATED]->() RETURN n.name_ar AS name, count(*) AS deg ORDER BY deg DESC LIMIT 10
   -> ابْنِ عَبَّاسٍ 14338 · ابْنُ عَبْسٍ 13118 · أَبُو عَبْدِ اللَّهِ الْحَافِظُ 12415 · … · وَكِيعٌ 8275 · سُفْيَانُ 6845 · عَلِيٌّ 6054
```

Confirmed: the old query returns empty on a graph that is fine, and the §7 note describing the operator hazard is accurate. The example narrators listed in the new §7 pass criteria all appear in the real top-10, and أبيه is absent.

## Figures — verified, not accepted

Repo-local evidence:

- `data/curated/narrators_canonical.parquet` = **102,350** rows, `narrator_mentions_resolved.parquet` = **2,975,629** rows, read via `pyarrow.parquet.read_metadata`. Exactly the §1a values; the old 232,766 / 3,126,954 are confirmed stale.
- `data/curated/_resolve_run.txt`: `canonical_ids=102350 run_status=complete git_sha=c5a95be` — corroborates both the count and the `c5a95be` in the artifact ref.

Live measurement (read-only) against both hosts. Every one of the 15 rows in the new §2 table matched:

| | §2 claims stage / prod | measured stage / prod |
|---|---|---|
| total nodes | 2,722,570 / 2,722,526 | 2,722,570 / 2,722,526 |
| Hadith | 1,427,222 / — | 1,427,222 / 1,427,222 |
| Chain | 1,122,521 / 1,122,477 | 1,122,521 / 1,122,477 |
| Narrator | 102,350 / — | 102,350 / 102,350 |
| Grading · Collection | 69,465 · 1,011 | 69,465 · 1,011 both hosts |
| hollow chains | 0 / 0 | 0 / 0 |
| linked Narrators | 61,278 / 61,212 | 61,278 / 61,212 |
| PARALLEL_OF | 7,549,479 / — | 7,549,479 / 7,549,479 |
| TRANSMITTED_TO | 3,387,179 / 3,342,894 | 3,387,179 / 3,342,894 |
| APPEARS_IN | 1,426,899 / — | 1,426,899 / 1,426,899 |
| NARRATED | 839,238 / 826,376 | 839,238 / 826,376 |
| STUDIED_UNDER · GRADED_BY | 55,248 · 48,283 | 55,248 · 48,283 both hosts |

## Internal consistency — recomputed, not trusted

- **Linkage.** Recomputed by the doc's own query rather than from the stated percentage: stage `t=102350, linked=61278, pct=59.87`; prod `t=102350, linked=61212, pct=59.81`. The §7 "≈59.9%" and the §1 "≈59–60%" band both follow.
- **Node sum.** The five labels in §2 sum to 2,722,569, one short of the stated 2,722,570. The missing node is the singleton `LoadProvenance` — `MATCH (n) UNWIND labels(n) AS l …` returns `"LoadProvenance", 1` on both hosts, and `noorinalabs-deploy/scripts/graph_load_provenance.sh` MERGEs it on `scope`, one per graph, deliberately not on `parquet_ref`. §1 and §2 reconcile exactly, and prod's 2,722,526 is the same sum less the 44 missing Chains.
- **orphan_narrators.** §6's ≈41,072 = 102,350 − 61,278 exactly. The superseded ~72,852 is correctly labelled the pre-merge-fix figure.
- **Attribution.** The claim that the counts moved legitimately holds: da#347/#356/#376 are all narrator over-merge/split fixes (they reduce the canonical count), and da#364 recovers 63,642 discarded `TRANSMITTED_TO` rows while da#369 adds 380,771 matn-derived mentions (they raise chains). The direction of each cited issue matches the direction of its delta.

## Premise rewrite — verified, and the runbook is still usable forward

Prod, measured read-only: `parquet_ref = staged/narrator-resolve/2026-07-25-c5a95be`, `load_status = complete`, `stamped_at = 2026-07-25T18:38:33.669Z` — the same ref stage carries. hollow = 0, linkage 59.81%, relational-pronoun query returns 0 rows. Prod is the corrected graph. Executing the base revision's §3 would have `DETACH DELETE`d it.

On whether correcting the premise left the document describing history: **no — it remains executable for a future reload.** §1, §3, §4, §6 and §7 are all generic procedures with no dependency on the current state; §2 is a dated snapshot, but its instruction ("run the §1 block against `noorinalabs-stg` as well and diff") is the executable part and stays valid. The one place where the rewrite does pin itself to now is §1's HEALTHY criterion — filed as #510 below.

I also verified the §3 idempotency claim on which the new default path rests, rather than taking "the loaders are MERGE-idempotent" on faith: `_CHAIN_MERGE` is `MERGE (n:Chain {id: row.id})`, and all six relationship writers in `load_edges.py` are `MERGE` forms. A top-up will create the missing 44 Chains and the missing edges without deleting anything. The `SSH_HOST` / `LOAD_ARGS` contract in `scripts/load_staging.sh` is real, and since the script defaults to `load --nodes-only`, the explicit `LOAD_ARGS="load"` in §3/§6 is load-bearing and correct.

## The other queries — checked the same way, not assumed

Defect 3 was found by inspection, so I executed every remaining command in the document rather than assuming it was the only one.

- **§0 guard.** Reproduced live on both hosts. Unflagged `curl -s ifconfig.me` returns `2a01:4ff:f4:a690::1` on prod and `2a01:4ff:f4:5e0d::1` on stage; `curl -4 -s ifconfig.me` returns `178.156.214.225` and `87.99.137.225`. The as-written guard matched neither IPv4 literal on either host — it could not discriminate prod from stage at all. The `-4` fix is necessary and correct, and the "load-bearing, do not drop it" note is justified.
- **§1 linkage query.** `sum(CASE WHEN COUNT{(n)--()}>0 …)` nests a COUNT subquery inside an aggregate; I expected this to be the second silent-failure candidate. It executes correctly on Neo4j 5 and returns the right numbers on both hosts. No defect.
- **§1 / §7 pronoun query.** Returns 0 rows on both hosts. I also ran the control the doc claims exists — same query shape with `["سفيان","وكيع"]` returns `سفيان 27249` and `وكيع 10120`. The "controlled check, so an empty result is a real zero" note is itself true, not just asserted.
- **§1 provenance, §7 label and chain/hollow queries.** All execute and return correct results. Every property they read is real: `name_ar`, `name_ar_normalized`, `mention_count` are SET in `load_nodes.py`, `narrator_ids` is SET on Chain, and `LoadProvenance.parquet_ref` is written by the deploy-side provenance script.
- **§4 wipe loop.** Not executed (destructive), but its one external dependency is satisfied — `SHOW PROCEDURES` confirms `apoc.periodic.iterate` is present on prod.

Nothing else in the document fails on a healthy graph.

## Non-blocking

- #509 — §1a scopes the Neo4j memory/swap precondition to §4/§6, but §3 (the new default path) runs the identical loader and the same `PARALLEL_OF` phase that precondition exists to protect. Currently satisfied on prod (verified: 8 GiB swap active, container memory limit 10737418240), so no live exposure; the gating is still wrong.
- #510 — §1's HEALTHY gate pins the `parquet_ref` literal, so after the next artifact refresh a healthy graph classifies UNCLASSIFIED. Fail-safe direction, but it is §1a's own "equality against stage, not a literal in this file" rule not applied to §1.
- #508 (already filed) — §4 physically precedes §5, and §4's own prerequisite line names the §0 re-check but not the backup.

CI: 11/11 checks SUCCESS at `b3176e6d431359f921ba4098922366f09d1fe14f`, confirmed against the check-runs API at that SHA, and the head has not moved since.

Not merging — the owner holds that decision.

---
Requestee: Sofia Cardoso
RequestOrReplied: Approved
TechDebt: #509, #510

